PR #10, opened 2015-08-13
Base project

lmorchard:
Okay, this has Django / Sugardough for the server and a gulp workflow for static assets. Still working on integrating the two, but figured this start would be worth taking a look at until then


idea_town/settings.py
```diff
@@ -0,0 +1,167 @@
...
+    },
+]
+
+# Django-CSP
+CSP_DEFAULT_SRC = (
+    "'self'",
+)
+CSP_FONT_SRC = (
+    "'self'",
+    'http://*.mozilla.net',
```
jaredhirsch, line 141: we'll probably want to also whitelist the `*.firefox.com` and `.mozaws.net` domains for all these CSP rules. I'm not sure if the `*` lets us go two subdirs deep or not (for `ideatown.dev.mozaws.net` and `ideatown.stage.mozaws.net`)

package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/mozilla/idea-town.git"
+  },
+  "author": "",
+  "license": "ISC",
```
jaredhirsch, line 17: MPL-2.0

Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM python:3-slim
```
pdehaan, line 1: Should this be Python 2 (same as the Travis config above)?

jaredhirsch:
I'm actually super excited to land this right away, so I might just file followup bugs for the two things I mentioned above. Testing locally now

jaredhirsch:
Gah, splitting out all the repos and email distracted me. FOR REALS trying it out now

jaredhirsch:
@lmorchard I'm having trouble getting this to work locally: seems like the VMs are loading ok, but I don't quite see how to access them from the OS. Docker [just released](http://blog.docker.com/2015/08/docker-toolbox/) some kind of brand-new thing that replaces boot2docker--maybe I should use the old thing instead?

lmorchard:
Hmm, you probably want to use the newest thing. I didn't have Docker on this machine at all since a reinstall, so that's what I ended up using. It comes with this handy Kitematic GUI that has a handy button which opens up the site in your browser. 

It was vaguely tricky because at first I expected to find the server at like 0.0.0.0:8000, but it's at 192.168.99.100:8000 on my machine

Dockerfile
lmorchard, line 1: Yeah, probably should be. Looks like Sugardough is partway through a Py2 to 3 transition, but I don't think we really want to jump on the Py3 train

lmorchard, line 1: Huh, though apparently the overwhelming advice from #webdev is that it's super Python 3 time if you're starting a new project, which we are.

pdehaan, line 1: In that case, feel free to ignore my nonsense here and maybe we bump from Python 2.7 to Python 3.x in the .travis.yml file (so they're both in sync).

jaredhirsch:
@lmorchard Aha, that works! I had been doing `docker-machine ls`, which gave me a random other port (which I [now see](http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=docker) is a Docker port), though it did list the right local IP:

```
| ~ @ 🚐   docker-machine ls
NAME      ACTIVE   DRIVER       STATE     URL                         SWARM
default            virtualbox   Running   tcp://192.168.99.100:2376   
```

Well, this should be fun! merging like

PR #22, opened 2015-08-17
chore(lint): Add ESLint and AirBnB styleguide

### LINKED ISSUE #19: enforce airbnb style
shoutout to @pdehaan 

We need to land the Giant Blob before we can use https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb, but let's do this soon!

pdehaan:
Fixes #19

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,7 @@
+extends:
+  - eslint:recommended
+  - airbnb
+
+rules:
+  comma-dangle: [2, never]
+  no-console: 1
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,67 +1,66 @@
-var debug = true;
+const autoprefixer = require('gulp-autoprefixer');
+const babelify = require('babelify');
+const browserify = require('browserify');
+const buffer = require('vinyl-buffer');
+const cache = require('gulp-cache');
+const del = require('del');
+const eslint = require('gulp-eslint');
+const gulp = require('gulp');
+const gutil = require('gulp-util');
+const imagemin = require('gulp-imagemin');
+const minifycss = require('gulp-minify-css');
+const notify = require('gulp-notify');
+const rename = require('gulp-rename');
+const runSequence = require('run-sequence');
+const sass = require('gulp-sass');
+const source = require('vinyl-source-stream');
+const sourcemaps = require('gulp-sourcemaps');
+const uglify = require('gulp-uglify');
 
-var gulp = require('gulp');
-var gutil = require('gulp-util');
-var sass = require('gulp-sass');
-var imagemin = require('gulp-imagemin');
-var sass = require('gulp-sass');
-var del = require('del');
-var runSequence = require('run-sequence');
-var rename = require('gulp-rename');
-var browserify = require('browserify');
-var babelify = require('babelify');
-var source = require('vinyl-source-stream');
-var buffer = require('vinyl-buffer');
-var sourcemaps = require('gulp-sourcemaps');
-var notify = require('gulp-notify');
-var eslint = require('gulp-eslint');
-var autoprefixer = require('gulp-autoprefixer');
-var minifycss = require('gulp-minify-css');
-var uglify = require('gulp-uglify');
-var cache = require('gulp-cache');
+const IS_DEBUG = true;
 
 // Lint the gulpfile
-gulp.task('selfie', function(){
+gulp.task('selfie', function selfieTask() {
   return gulp.src('gulpfile.js')
     .pipe(eslint())
     .pipe(eslint.format());
 });
 
 // Lint the *.js files
-gulp.task('lint', function() {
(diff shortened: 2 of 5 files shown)


.eslintrc
```diff
@@ -0,0 +1,7 @@
+extends:
+  - eslint:recommended
+  - airbnb
+
+rules:
+  no-console: 1
+  no-var: 0
```
pdehaan: Silencing all the errors about use of `var` usage from the AirBnB style ruels, but I can convert all the `var`s to `let`s and `const`s if we want.

gulpfile.js
```diff
@@ -1,67 +1,66 @@
...
-var sass = require('gulp-sass');
-var del = require('del');
-var runSequence = require('run-sequence');
+var minifycss = require('gulp-minify-css');
+var notify = require('gulp-notify');
 var rename = require('gulp-rename');
-var browserify = require('browserify');
-var babelify = require('babelify');
+var runSequence = require('run-sequence');
+var sass = require('gulp-sass');
```
pdehaan: Ignore all the diff churn. I just A-Zed the `require()` statements and removed the dupe `sass` var require.

gulpfile.js
```diff
@@ -1,67 +1,66 @@
...
-var autoprefixer = require('gulp-autoprefixer');
-var minifycss = require('gulp-minify-css');
 var uglify = require('gulp-uglify');
-var cache = require('gulp-cache');
+
+var IS_DEBUG = true;
 
 // Lint the gulpfile
-gulp.task('selfie', function(){
+gulp.task('selfie', function selfieTask() {
```
pdehaan, line 23: The AirBnB style guide also dislikes anonymous functions, so I gave all these tasks function names. If that seems like absurdity, I can just disable the anonymous function rule for this file and remove these names.

package.json
```diff
@@ -42,17 +45,17 @@
...
-    "ampersand-view-switcher": "^2.0.0",
-    "bluebird": "^2.9.34",
-    "domready": "^1.0.8",
-    "mozilla-tabzilla": "^0.4.0",
-    "normalize.css": "^3.0.3"
+  "directories": {
+    "doc": "docs"
+  },
+  "homepage": "https://github.com/mozilla/idea-town#readme",
+  "license": "MPL-2.0",
```
pdehaan, line 52: Changed this from `"ISC"` to `"MPL-2.0"`.

package.json
```diff
@@ -42,17 +45,17 @@
...
-    "ampersand-view": "^8.0.1",
-    "ampersand-view-switcher": "^2.0.0",
-    "bluebird": "^2.9.34",
-    "domready": "^1.0.8",
-    "mozilla-tabzilla": "^0.4.0",
-    "normalize.css": "^3.0.3"
+  "directories": {
+    "doc": "docs"
+  },
+  "homepage": "https://github.com/mozilla/idea-town#readme",
```
pdehaan, line 51: I left the `homepage` as-is, but we should change to `https://ideatown.firefox.com/` once we get the splash deployed and site launched.

package.json
```diff
@@ -42,17 +45,17 @@
...
+  "license": "MPL-2.0",
+  "private": "true",
```
pdehaan, line 53: Removed `main` attribute and added `private: true` since we aren't planning on deploying this to npm (as far as I know).

static-src/scripts/App.js
```diff
@@ -30,7 +27,5 @@ module.exports = App.extend({
       hashChange: true
     });
     */
-
-  }
-
+  },
```
pdehaan: Oh, and AirBnB style guide prefers trailing commas everywhere (for prettier git diffs). I'm not a huge fan, but what can you do...

pdehaan:
r? @6a68 

Looks like Travis is barfing all over the place, but I'm guessing that is just Python/lint churn.

.eslintrc
jaredhirsch: we're using babel on the front-end, so yeah, let's bury `var`

package.json
jaredhirsch, line 53: +1

static-src/scripts/App.js
jaredhirsch: That might be a good rule to not use. Those trailing commas are going to drive me crazy ;-)

jaredhirsch:
@pdehaan Thanks for this! Looks good overall.

I'm not a fan of trailing commas, I guess we should remove that and add a comment somewhere (README?) that we use airbnb's style rules with a few modifications. Would you mind making that change, then this should be good to land?

jaredhirsch:
yay merging!!!!!

gulpfile.js
lmorchard, line 23: How does this affect anonymous arrow functions in ES6 / babel? (eg. `(foo) => { foo++; }`) We'll probably want to use a lot of those.

Issue #42, opened 2015-08-18
missing css error on docker static VM startup


jaredhirsch:
STR:
- clone repo
- `docker-compose up`
- see this in stdout, perhaps a gulp race condition or some other thing happening

```
client_build_1 | [18:24:29] Using gulpfile /app/gulpfile.js
client_build_1 | [18:24:29] Starting 'build'...
client_build_1 | [18:24:29] Starting 'clean'...
client_build_1 | [18:24:29] Starting 'connect'...
client_build_1 | [18:24:29] Finished 'connect' after 9.72 ms
client_build_1 | [18:24:29] Server started http://localhost:9988
client_build_1 | [18:24:29] Finished 'clean' after 66 ms
client_build_1 | [18:24:29] Starting 'vendor'...
client_build_1 | [18:24:29] Starting 'npm:tabzilla:img'...
client_build_1 | [18:24:29] Starting 'npm:tabzilla:css'...
client_build_1 | [18:24:29] Starting 'npm:normalize'...
client_build_1 | [18:24:29] Finished 'npm:tabzilla:img' after 22 ms
client_build_1 | [18:24:29] Finished 'npm:tabzilla:css' after 24 ms
client_build_1 | [18:24:29] Finished 'npm:normalize' after 20 ms
client_build_1 | [18:24:29] Finished 'vendor' after 35 ms
client_build_1 | [18:24:29] Starting 'scripts'...
client_build_1 | [18:24:30] Finished 'scripts' after 1.71 s
client_build_1 | [18:24:30] Starting 'styles'...
client_build_1 | Error in plugin 'sass'
client_build_1 | Message:
client_build_1 |     static-src/styles/main.scss
client_build_1 |   4:9  file to import not found or unreadable: ../vendor/normalize.css/normalize
client_build_1 | Current dir: 
client_build_1 | [18:24:30] Finished 'styles' after 17 ms
client_build_1 | [18:24:30] Starting 'images'...
client_build_1 | [18:24:30] Finished 'images' after 17 ms
client_build_1 | [18:24:30] Finished 'build' after 1.86 s
client_build_1 | [18:24:30] Starting 'watch'...
client_build_1 | [18:24:30] Finished 'watch' after 43 ms
client_build_1 | [18:24:30] Starting 'server'...
client_build_1 | [18:24:30] Finished 'server' after 2.97 μs
client_build_1 | [18:24:30] Starting 'default'...
[...]

PR #43, opened 2015-08-18
Issue #42: Look for assets from node_modules found in NODE_PATH

lmorchard:
On a local install of gulp, these CSS dependencies are found in `./node_modules/`.

In Docker, these have been moved to `/home/webdev/node_modules` inside the client build container.

This directory shows up in `process.env.NODE_PATH`, so let's use that when it's specified.


lmorchard:
Closing, because idea-town-app is a thing now and the client-build stuff should be excised from this here server thing

PR #45, opened 2015-08-25
Experiment models

lmorchard:
This PR now includes both FxA login and a first stab at Experiment models & a simple REST API. The two are not tied together yet, but that should hopefully be my next step.

If you can get this running in a Docker server, some things to try:
- Login as admin/admin at http://ideatown.dev:8000/admin
- Add some experiments and detail cards for experiments
- Visit http://ideatown.dev:8000/api/ to explore the API


jaredhirsch:
Cool, I'll take a look in a bit

idea_town/experiments/models.py
```diff
@@ -0,0 +1,52 @@
...
+    users = models.ManyToManyField(User, through='UserInstallation')
+
+    created = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        return self.title
+
+
+class ExperimentDetail(models.Model):
```
meandavejustice, line 27: Should we have a property for the xpi url? or are we going to be hosting all of our experiments?

idea_town/experiments/views.py
```diff
@@ -0,0 +1,45 @@
...
+from .serializers import (ExperimentSerializer, ExperimentDeepSerializer,
+                          ExperimentDetailSerializer)
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+class ExperimentViewSet(viewsets.ModelViewSet):
+    """
+    Returns a list of all Experiments in the system.
```
meandavejustice, line 14: Inconsistent with oneline doc block above

idea_town/experiments/models.py
lmorchard, line 27: Actually, that's a question I had and I'd punted for now in the model.

We can (maybe should?) host the experiments, and can treat them as uploads like the images? Or, we might put them on AMO? Not sure what exactly we want to do there yet. 

FWIW, I was thinking images (at least) would end up on S3 in deployment, since it's not usually good to keep a filesystem around on a VM in AWS. We could upload XPIs there, too, and manage them like the rest of the content in admin.

idea_town/accounts/providers/fxa/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+class FirefoxAccountsOAuth2Adapter(OAuth2Adapter):
+    provider_id = FirefoxAccountsProvider.id
+
+    def __init__(self):
+        provider = self.get_provider()
+        provider_settings = provider.get_settings()
+
+        self.access_token_url = provider_settings.get(
+            'ACCESS_TOKEN_URL', 'https://oauth.accounts.firefox.com/v1/token')
```
meandavejustice, line 20: aren't all of these declared in `settings.py`? ie: `SOCIALACCOUNT_PROVIDERS`

lmorchard, line 20: <strike>Yeah, `provider.get_settings` is a helper that looks up `SOCIALACCOUNT_PROVIDERS['fxa']`</strike>

Oh wait, I think I misunderstood. You mean these defaults?

idea_town/experiments/models.py
jaredhirsch, line 27: Well, we'll definitely need the experiments to be signed by AMO. AMO has an auto-signing tool built into the submission process; it was really easy to get the universal-search addon to pass the auto-review, and you get a private AMO URL for free along with the signing. That seems like the simplest starting point for hosting addons?

idea_town/experiments/migrations/0001_initial.py
```diff
@@ -0,0 +1,60 @@
...
+                ('modified', models.DateTimeField(auto_now=True)),
+            ],
+        ),
+        migrations.CreateModel(
+            name='ExperimentDetail',
+            fields=[
+                ('id', models.AutoField(auto_created=True, verbose_name='ID', serialize=False, primary_key=True)),
+                ('order', models.IntegerField(default=0)),
+                ('headline', models.CharField(max_length=256)),
+                ('image', models.ImageField(upload_to=idea_town.utils.HashedUploadTo('image'))),
```
jaredhirsch, line 34: we'll probably need (large) thumbnails of these images - or maybe we can just resize on demand?

idea_town/experiments/migrations/0001_initial.py
```diff
@@ -0,0 +1,60 @@
...
+            ],
+            options={
+                'ordering': ('experiment', 'order', 'modified'),
+            },
+        ),
+        migrations.CreateModel(
+            name='UserInstallation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, verbose_name='ID', serialize=False, primary_key=True)),
+                ('rating', models.FloatField(null=True)),
```
jaredhirsch, line 48: do we want to capture text feedback as well? something like a user testimonial might be cool to use as a pull quote on the experiment detail page

idea_town/experiments/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+        fields = ('url', 'order', 'headline', 'image', 'copy', 'experiment_url')
+
+    def get_experiment_url(self, obj):
+        request = self.context['request']
+        path = reverse('experiment-detail', args=(obj.experiment.pk,))
+        return request.build_absolute_uri(path)
+
+
+class ExperimentDeepSerializer(serializers.HyperlinkedModelSerializer):
+    """Deep Experiment serializer that includes ExperimentDetails"""
```
jaredhirsch, line 21: this is great

idea_town/accounts/providers/fxa/views.py
meandavejustice, line 20: yeah, it looks like we are duplicating since we are passing in the same defaults as are already in SOCIALACCOUNT_PROVIDERS

README.md
```diff
@@ -34,11 +42,8 @@ WIKI: https://wiki.mozilla.org/Idea-Town
     try `docker-compose up` again. ([Ugh, this is a known bug in
     docker-compose][dc-bug]. Maybe we can find a workaround?)
 
-* To see the IP address of the default Docker machine:
-  * `docker-machine ip default` 
-
 * To visit the Django server:
-  * `open http://$(docker-machine ip default):8000/`
+  * `open http://ideatown.dev:8000/`
```
jaredhirsch, line 46: These instructions worked really well for me, nice work

lmorchard:
Quick commit update to make flake8 happier & pass on travis

lmorchard:
@6a68 said it's okay so BOOM

lmorchard:
FWIW, for further changes after this big bang, let's open up smaller issues for quicker reviews

idea_town/experiments/migrations/0001_initial.py
lmorchard, line 48: And that was another of my questions I forgot to write down: What kind of input do we want to capture from users? 

I saw a 1-5 stars thing in wires, so I threw in rating. But there was also mention of a survey. We can expand the model for a quick comment field, but a survey is another app in and of itself (unless we, say, used an external service or something). This also plays into nailing down what metrics we might want to collect

idea_town/accounts/providers/fxa/views.py
lmorchard, line 20: Hmm, I may yank the defaults out of settings. This custom fxa provider for allauth may eventually turn into a Pull Request to the upstream project, where these URLs are hardcoded in the out-of-box fxa provider and not exposed as settings. So, I'd want to keep the defaults with the provider itself since these are integral to the FxA service

idea_town/experiments/migrations/0001_initial.py
lmorchard, line 34: Yup, that should be a TODO. I started looking into a couple Django thumbnail apps to maybe take care of this easily, but hadn't decided on one yet. I think one of them does cached on-demand resizing from URL parameters, so that might be handy. Or, we use another one with sizes decided up front and push pre-generated thumbnails to S3 or something

PR #55, opened 2015-09-02
Re-introduce the frontend asset build process

lmorchard:
- Create new `idea_town/frontend` Django app
- Single catch-all view that handles all URLs not taken by other apps
- Frontend sources live in `idea_town/frontend/static-src`
- `gulpfile.js` build lives in `idea_town/frontend/static`
- Container running Gulp and auto-restarts on `gulpfile.js` changes


lmorchard:
Okay, @6a68, have a look! :) In a nutshell, you should be able to run `docker-compose build && docker-compose up` and get a Gulp build running that handles stuff you edit in `idea_town/frontend/client-src`

Oh, and you can edit these templates for great justice:
- `idea_town/base/templates/idea_town/base.html`
- `idea_town/frontend/templates/idea_town/frontend/index.html`

The directory path naming weirdness is due to how Django handles mashing together template namespaces between apps.

jaredhirsch:
@lmorchard LGTM, works locally, urlpattern looks sane, let's do this

PR #69, opened 2015-09-11
DO NOT MERGE! Docker tweaks for ideatown

mostlygeek:
Not ready to merge yet!

Start of the design on what's required for an idea-town prod deploy docker image.


lmorchard:
Hmm. So, that `Dockerfile-server` currently works to spin up the server part of our dev environment. It came mostly from Sugardough as boilerplate and [it's used in `docker-compose.yml`](https://github.com/mozilla/idea-town/blob/master/docker-compose.yml#L16) alongside [another container that does client-side builds with Gulp](https://github.com/mozilla/idea-town/blob/master/Dockerfile-client_build).

That's mainly for dev convenience, so a single `docker-compose up` starts everything you need. The client build container will never be deployed to a production server, though conceivably it could be run as part of an automated build process. (Then again, just installing node & running gulp directly on a build machine will work too.)

Do we need separate `Dockerfiles` for dev vs prod? Or, will it work stripped down like this with just a rename in `docker-compose.yml`? (I haven't tried it yet)

mostlygeek:
The `Dockerfile` should build a production asset. Having it build development assets breaks conventions everywhere. 

For development, we (cloudops) don't have any recommendations yet. Though, it makes sense to keep things named consistently. Also using the Docker created images, https://hub.docker.com/_/node/ is recommended even for dev. It'll result in fewer surprises and translations when working on the main Dockerfile.

jaredhirsch:
@mostlygeek 

> The Dockerfile should build a production asset. Having it build development assets breaks conventions everywhere.

This statement might be true for ops, but we use docker in development, and we will continue to specialize the dev docker setup to optimize for developer happiness over production hardness.

The [docker-compose docs](https://docs.docker.com/compose/production/) suggest using separate yaml config files for the dev and prod cases, so maybe that's a better direction to take with this PR.

Also related: https://docs.docker.com/articles/dockerfile_best-practices/

mostlygeek:
> This statement might be true for ops, but we use docker in development, and we will continue to specialize the dev docker setup to optimize for developer happiness over production hardness.

Can you use `Dockerfile-dev` instead of `Dockerfile`? 

We'd all ship faster if we can agree on the convention upfront. Having `Dockerfile` build the production asset means ops can tool up automation around that convention and be able to push towards the glorious continuous integration/deployment future.

lmorchard:
> Can you use Dockerfile-dev instead of Dockerfile? We'd all ship faster if we can agree on the convention upfront.

:+1: We have `Dockerfile-server`, we can split that into `Dockerfile` (prod) and `Dockerfile-dev` (local dev). 

Ideally, those two would be identical, except for maybe one or two lines on the -dev side for debugging exceptions. <strike>Trying to remember if it's `Dockerfile` or `docker-compose.yml` that has a kind of inheritance mechanism to share things...</strike> [Looks like it's docker-compose](https://docs.docker.com/compose/extends/)

lmorchard:
Noticed the stripped down Dockerfile is also missing stuff for postgresql and libjpeg, which we need but aren't provided by the python image:

```
RUN apt-get update && \
    apt-get install -y --no-install-recommends build-essential python-dev \
        libpq-dev postgresql-client gettext libjpeg62-turbo-dev \
        vim wget curl git && \
    rm -rf /var/lib/apt/lists/*
```

Do we have another standard image that supplies that stuff? (FWIW vim, wget, curl, and git are there for dev hacks and shouldn't go to prod of course) 

That comes in part from [the Sugardough Dockerfile template](https://github.com/mozilla/sugardough/blob/master/%7B%7B%20cookiecutter.project_name%20%7D%7D/Dockerfile) that I think gets used by a number of Mozilla webdev projects. So, it'll probably be a concern for more projects than ours.

mostlygeek:
If it's needed in the container for the prod deploy let's add it in. Free free to PR into (https://github.com/mostlygeek/idea-town/commits/docker-tweaks).  

Ultimately, the goal is for devs to manage system dependencies for their application. You guys know the requirements better than anybody. Also each system package dependency should be justifiable and necessary. 

For `Dockerfile-dev`, there aren't any restrictions. Install whatever packages makes development easier/faster. 

Also, why would we need postgresql-client? 

Usually we only install dev packages like `libpq-dev`, `libjpeg62-turbo-dev`, etc. In this case, I wasn't sure they were necessary so I didn't blanket include them.

lmorchard:
So, after a period of my head meeting desk repeatedly, I came up with this:

https://github.com/mostlygeek/idea-town/pull/1

That seems to work for dev without changes, and also I was able to fire it up in a pretend-prod setting on my laptop without mounting the local app source.

The thing I'm uncomfortable with is the copypasta'd bit to bring node into the Python container. That's needed to run gulp to build the deployable frontend assets. It's not needed to actually run the site. So, I deleted it at the end. 

But, the result should be a working site with only a dependency on an external postgres DB.

lmorchard:
Thinking about it, I guess an alternative for a shorter Dockerfile would be to start from the official node image (from which I stole the temporary install instructions), never remove node, and install python3 from a plain Debian package. The main issue is that Debian ships the Python version we need, but the packaged node version is v0.10 and too old.

lmorchard:
So, thinking out loud about what we need to deploy this thing:
- [ ] Need the server container, with static frontend assets built. Though hacky, [this PR does that](https://github.com/mostlygeek/idea-town/pull/1). Is there a better option? (e.g. build static assets in a different container or by a different mechanism, push to S3 / CDN?)
- [ ] Need a postgres instance. In dev, we use [the official `postgres:9.3` image](https://github.com/mozilla/idea-town/blob/master/docker-compose.yml#L2). Can we use that in a container in prod, too? Or is there a better option like [Amazon RDS for Postgres](https://aws.amazon.com/rds/postgresql/)?
- [ ] Need a place to store uploaded content - e.g. images for experiments, thumbnails, etc. In dev, we store that on the host filesystem. [In prod, I had thought we could switch to Amazon S3](https://github.com/lmorchard/idea-town/tree/amazon-s3-media-uploads). Is there a better option?
- [ ] At some point soon, we'll want to add [a celery work queue](http://www.celeryproject.org/) to the mix. I assume that will require a rabbitmq instance alongside postgres. Hopefully that's easy, or there's an easier thing I don't know about yet :)

Am I missing anything? Do I have incorrect assumptions? Kind of hoping someone else has done all the above, and there's a path we can follow to make this stuff easy on everyone

mostlygeek:
Some of these are from the meeting we just had: 

Building static assets: 
- build static assets in a nodejs based container, into a mounted directory (to make it easier)
- build app based on Python container and just copy in the static assets

Postgres 
- in prod we'll use RDS 
- for dev, use whatever you want

Storing uploaded content
- for AWS hosted dev/stage/prod environments use S3 
- for localhost development, use the file system, either in the container or a mounted directory

Celery
- use the redis store. For aws we can use elasticache so less stuff for us to manage
- for dev, using redis would be easier than deploying rabbitmq ;)
- for prod, we can use rabbitmq (easy enough to deploy) but prefer the elasticache managed service

lmorchard:
Going to rudely close this and point us to pull #143

PR #76, opened 2015-09-21
chore(package): Tweak the package.json file to use latest repo URL

pdehaan:


package.json
```diff
@@ -31,12 +31,12 @@
   },
   "devDependencies": {
     "babel": "^5.8.21",
-    "babel-eslint": "4.0.10",
+    "babel-eslint": "4.1.3",
     "babelify": "^6.1.3",
     "browserify": "^11.0.1",
-    "del": "^1.2.1",
-    "eslint-config-airbnb": "0.0.7",
-    "eslint-plugin-react": "3.2.3",
```
pdehaan, line 39: Removing the **eslint-plugin-react** dependency since we don't use React.

PR #78, opened 2015-09-23
Add endpoint for metrics

meandavejustice:
endpoint is at `/api/metrics`.

For our first run we will just push events to `datadog`. Going forward we can setup datadog's statsd integration.

### DIFF
--- a/idea_town/metrics/views.py
+++ b/idea_town/metrics/views.py
@@ -1,0 +1,33 @@
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework.permissions import IsAuthenticated
+from django.conf import settings
+import datadog
+import json
+import datetime
+import logging
+
+logger = logging.getLogger(__name__)
+
+datadog.initialize(**settings.DATADOG_KEYS)
+
+
+class MetricsView(APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request):
+        return Response(request.data)
+
+    def post(self, request):
+        d = request.data
+        if 'title' not in d:
+            d['title'] = 'ideatown-generic-event'
+        d['created'] = datetime.datetime.now().isoformat()
+        handle_metric(d)
+        return Response(d)
+
+
+def handle_metric(metric):
+    # will eventually handle multiple types of metrics, including
+    # statsd types.
+    datadog.api.Event.create(title=metric['title'], text=json.dumps(metric))
--- a/idea_town/settings.py
+++ b/idea_town/settings.py
@@ -117,6 +117,10 @@ ROOT_URLCONF = 'idea_town.urls'
 
 WSGI_APPLICATION = 'idea_town.wsgi.application'
 
+DATADOG_KEYS = {
+    'api_key': '4f3b967bbf13c7769ac4fa89efda0fae',
+    'app_key': 'ddd45a7b3a3cb90ff1baf20ae8cfab04d1937038'
+}
 
 # Database
 # https://docs.djangoproject.com/en/1.7/ref/settings/#databases
--- a/idea_town/urls.py
+++ b/idea_town/urls.py
@@ -4,7 +4,7 @@ from django.contrib import admin
 from rest_framework import routers
 
 from .experiments import views as experiment_views
-
+from .metrics import views as metrics_view
 
 # Allow apps to contribute API parts
 router = routers.DefaultRouter()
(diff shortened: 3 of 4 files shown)


lmorchard:
So, I see I'm assigned, but I have some dumb questions: What's datadog & what are we doing with it? Do we have docs & issues & etc on it? I'm kind of light on context about this stuff.

jaredhirsch:
@lmorchard So, cloud ops has been suggesting datadog (https://app.datadoghq.com) as a privacy/legal-approved third party metrics endpoint, mainly because it offers relatively simple dashboarding. They can give you an account if you need one (maybe ping mostlygeek). Firefox Accounts, WebPush, and some other services teams are using it, IIRC. We don't want datadog to receive any personally-identifiable information, which means that any metrics emitters need to sanitize locally before sending to the server (probably a good practice to get into).

Datadog isn't really well-documented anywhere, though there are references in the private mana wiki and on wikimo.

For the purposes of this bug, it's a place to put data. The long-term solution might well be to use S3 as the initial aggregation point, then use AWS lambda to do map-reducey things to it, and only direct things we want to visualize to dashboarding tools like datadog or even google analytics.

Does this help? Happy to attempt to provide more context

jaredhirsch:
@mostlygeek This is a first metrics-to-datadog PR. Feedback welcome :-)

lmorchard:
One top-of-head concern I've got about this API end point is that it looks like any rando can POST any random metric event. Might help to have a vocabulary of what we're collecting?

But then, I guess that just constrains the randos to submitting noise in the defined events. Hmm... maybe overthinking the concern here

lmorchard:
<strike>We might want to add a `@login_required` to this API (however REST framework handles that, I'm not sure) and/or require a CSRF token, if only to toss in a speed bump on random spam bots mating with the metrics endpoint</strike>

Actually, [I think this is already in place](https://github.com/mozilla/idea-town/pull/78/files#diff-867a2641e83e4e2a804f960f368f4449R16) & I just missed it. So, like, you can ignore me and I think this looks fine. Getting hooked up with a datadog account, so hopefully I can give this API endpoint an actual drive around the block and just :+1: it.

jaredhirsch:
> We might want to add a @login_required to this API (however REST framework handles that, I'm not sure) and/or require a CSRF token

Yeah, we've already got `csrftoken` and `sessionid` cookies set by Django, so maybe it'll just work. Looks like docs are here: http://www.django-rest-framework.org/api-guide/authentication/

lmorchard:
Yeah, can confirm - POSTing as a rando to the endpoint gets rejected:

```
➜  idea-town git:(4495f00) ✗ curl -ksi -D'{ "title": "foo", "bar": "baz" }' -XPOST http://ideatown.dev:8000/api/metrics
HTTP/1.0 403 FORBIDDEN
```

meandavejustice:
@lmorchard moved to using `import datadog` as you suggested.

lmorchard:
So, after sorting our my datadog & CSRF token business, it looks like this does a thing for me! My last question is... does datadog have a view somewhere that I can see what I POST'd?

```
➜  idea-town git:(4495f00) ✗ curl -ksi -H 'X-CSRFToken: VVgoSjG9NKSjzLm6VTv1UefE8eRgfLEZ' -H "Cookie: csrftoken=VVgoSjG9NKSjzLm6VTv1UefE8eRgfLEZ; sessionid=w7o7tanryoxjg2034507242e1tgxyzlj" -H "Accept: text/html,application/json" -H "Content-Type: application/json"   -d'{ "title": "idea-town-foo", "bar": "baz" }' -XPOST http://ideatown.dev:8000/api/metrics/ | more
HTTP/1.0 200 OK
Date: Wed, 23 Sep 2015 18:37:41 GMT
Server: WSGIServer/0.2 CPython/3.4.3
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json

{"created":"2015-09-23T18:37:41.444044","title":"idea-town-foo","bar":"baz"}
```

lmorchard:
Eh, going to say this is good, and we can keep refining it with actual usage

meandavejustice:
@lmorchard here is the main event stream I've been using to debug https://app.datadoghq.com/event/stream?tags_execution=and&show_private=true&per_page=30&aggregate_up=true&use_date_happened=false&display_timeline=true&from_ts=1442448000000&is_zoomed=false&to_ts=1443052800000&is_auto=false&incident=true&only_discussed=false&no_user=false&page=0&live=true&bucket_size=10800000

mostlygeek:
Looks ok. I would suggest using tags on your event metrics so they can be filtered/found/correlated with the stastd information into a dashboard later. 

Did you guys commit real api keys into the code?

Issue #63, opened 2015-09-10
add /user endpoint to pass session info to front-end app


jaredhirsch:
- if user's logged in, provide email address, whether the user has the main addon, and a list of any installed experiments
- if user's not logged in, return empty object / null / TBD

lmorchard:
Was thinking of taking some pokes at this, if it doesn't trample on anything @meandavejustice already has in progress. (I'm an issue thief, feel free to smack me)

meandavejustice:
no smacks necessary, happy you picked this up!

jaredhirsch:
This was fixed by #82

PR #82, opened 2015-09-24
Issue #63: Add /api/users/self endpoint

lmorchard:
- Add /api/users/self API endpoint to expose data about the currently
  logged in user. (i.e. user ID, installed experiments)
- Move sadface JS data into template markup
- Add the user and addon props to overall Me model
- Frontend client tweaks to use the API
- Add .editorconfig to idea_town/frontend/static-src so my editor can do
  different indenting in Python & JS
- Server-side tests

### DIFF
--- a/idea_town/frontend/static-src/.editorconfig
+++ b/idea_town/frontend/static-src/.editorconfig
@@ -1,0 +1,12 @@
+# http://editorconfig.org
+
+[*]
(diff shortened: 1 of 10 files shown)


lmorchard:
Hmm, and now my tests are failing. Ugh. Hopefully fixed in a short bit

Issue #83, opened 2015-09-24
Switch from xhr to fetch-polyfill in frontend app?


lmorchard:
Poked at the frontend client code, saw we're using [xhr](https://www.npmjs.com/package/xhr). 

Since I like promises and future things, it occurred to me we might like to use [a fetch() polyfill](https://www.npmjs.com/package/fetch-polyfill).

Main benefits of fetch:
- Built into Firefox 39+
- Promises, so we can easily fire many API GETs in parallel & carry on when they're all done
- Easier JSON handling? (but I probably just haven't figured out xhr)

jaredhirsch:
Closed by #84

PR #84, opened 2015-09-24
Issue #63: Add /api/users/self endpoint

lmorchard:
- Add /api/me API endpoint to expose data about the currently
  logged in user. (i.e. user ID, installed experiments)
- Add API, user, and addon props to overall Me model
- Move sadface JS data and some hardcoded URLs into index.html markup
- Frontend client tweaks to use the API
- Add .editorconfig to idea_town/frontend/static-src so my editor can do
  different indenting in Python & JS
- Tweak tox.ini to use python3.4
- Server-side tests

### DIFF
--- a/idea_town/frontend/static-src/app/main.js
+++ b/idea_town/frontend/static-src/app/main.js
@@ -1,3 +1,7 @@
+import es6Promise from 'es6-promise';
+es6Promise.polyfill();
+import 'isomorphic-fetch';
+
 import app from 'ampersand-app';
 
 import webChannel from './lib/web-channel';
@@ -8,33 +12,46 @@ import PageManager from './lib/page-manager';
 import Router from './lib/router';
 
 app.extend({
+
   initialize() {
-    app.experiments = new ExperimentsCollection();
-    app.experiments.fetch();
     app.webChannel = webChannel;
-    app.me = new Me();
 
-    // session won't change without a hard refresh, but addon state could, so:
-    // if addon state changes, dump user back to '/' and let the router handle
-    // redirecting to the correct landing page
-    app.me.on('change:hasAddon', () => { app.router.reload(); });
+    fetch('/api/me?format=json', {
+      credentials: 'same-origin'
+    }).then((response) => response.json()).then((userData) => {
+      app.me = new Me({
+        user: userData
+      });
 
-    // the header is independent of the page container logic, so it lives
-    // outside the page container element
-    // TODO: seems like the PageManager should know when to refresh / hide
-    //       the header
-    app.headerView = new HeaderView({el: document.querySelector('header') });
-    app.headerView.render();
+      app.experiments = new ExperimentsCollection();
+      app.experiments.fetch();
 
-    app.pageManager = new PageManager({
-      pageContainer: document.querySelector('[data-hook=page-container]')
+      // session won't change without a hard refresh, but addon state could, so:
+      // if addon state changes, dump user back to '/' and let the router handle
+      // redirecting to the correct landing page
+      app.me.on('change:hasAddon', () => { app.router.reload(); });
+
+      // the header is independent of the page container logic, so it lives
+      // outside the page container element
+      // TODO: seems like the PageManager should know when to refresh / hide
+      //       the header
+      app.headerView = new HeaderView({el: document.querySelector('header') });
+      app.headerView.render();
+
+      app.pageManager = new PageManager({
+        pageContainer: document.querySelector('[data-hook=page-container]')
+      });
+
+      app.router = new Router();
+      app.router.history.start();
+    }).catch((err) => {
+      // for now, log the error in the console & do nothing in the UI
+      console && console.error(err); // eslint-disable-line no-console
     });
-    app.router = new Router();
-    app.router.history.start();
   }
+
 });
 app.initialize();
-
 
 // make app accessible from window for debuggin'
 window.app = app;
--- a/idea_town/frontend/static-src/app/models/me.js
+++ b/idea_town/frontend/static-src/app/models/me.js
@@ -8,18 +8,16 @@ import State from 'ampersand-state';
 //       some kind of session-check API that sends over the user model
 //       (email, name, avatar, addon status) if the user's logged in.
 export default State.extend({
+
   props: {
-    hasAddon: ['boolean', true, false]
+    user: 'object',
+    hasAddon: {type: 'boolean', required: true, default: false}
   },
 
   derived: {
     csrfToken: {
       cache: false,
       fn: () => { return cookies.get('csrftoken'); }
-    },
-    session: {
-      cache: false,
-      fn: () => { return window.sadface.userId; }
     }
   },
 
--- a/idea_town/frontend/static-src/app/views/header-view.js
+++ b/idea_town/frontend/static-src/app/views/header-view.js
@@ -1,5 +1,4 @@
 import app from 'ampersand-app';
-import xhr from 'xhr';
 
 import BaseView from './base-view';
 
@@ -15,7 +14,7 @@ export default BaseView.extend({
   initialize() {
     // since we reload the page on every session change, no need to observe
     // the model; just treat it as a static data structure
-    this.session = app.me.session;
+    this.session = app.me.user.id;
   },
 
   render() {
@@ -30,20 +29,15 @@ export default BaseView.extend({
   //       to pick up csrftoken cookie changes.
   logout(evt) {
     evt.preventDefault();
-    xhr({
+    fetch('/accounts/logout/', {
       method: 'POST',
-      uri: '/accounts/logout/',
-      headers: {
-        'X-CSRFTOKEN': app.me.csrfToken
-      }
-    // note, the xhr callback also has a third 'body' argument
-    }, (err, resp) => {
-      if (err || resp.statusCode >= 400) {
-        // for now, log the error in the console & do nothing in the UI
-        console && console.error(err); // eslint-disable-line no-console
-      } else {
-        window.location.reload();
-      }
+      credentials: 'same-origin',
+      headers: { 'X-CSRFTOKEN': app.me.csrfToken }
+    }).then(() => {
+      window.location.reload();
+    }).catch((err) => {
+      // for now, log the error in the console & do nothing in the UI
+      console && console.error(err); // eslint-disable-line no-console
     });
   }
 });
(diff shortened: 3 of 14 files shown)


idea_town/urls.py
```diff
@@ -17,6 +19,7 @@
     url(r'^api-auth/', include('rest_framework.urls', namespace='rest_framework')),
     url(r'^api/', include(router.urls)),
     url(r'^api/metrics/', metrics_view.MetricsView.as_view(), name='metrics'),
+    url(r'^api/users/self', users_views.SelfView.as_view(), name='users_self'),
```
jaredhirsch: Should this have a trailing slash for consistency?

idea_town/frontend/templates/idea_town/frontend/index.html
```diff
@@ -3,6 +3,9 @@
 <head>
     <title>Idea Town</title>
     <link rel="stylesheet" href="{{ static('styles/main.css') }}">
+    <link rel="x-current-user" href="{{ url('users_self') }}">
+    {# TODO: Make this a dev / prod setting? Or, expose in an API endpoint? #}
```
jaredhirsch: Once nice thing about using an API endpoint: the addon can reuse it to look up experiment addon URLs. But this is fine for now, if you just want to land it as-is

idea_town/frontend/static-src/app/main.js
```diff
@@ -7,12 +8,41 @@ import Me from './models/me';
...
+    .querySelector('head link[rel="' + relName + '"]')
+    .getAttribute(attrName);
+}
+
 app.extend({
+
   initialize() {
+
     app.experiments = new ExperimentsCollection();
     app.experiments.fetch();
```
jaredhirsch: Mind moving this into `dataReady`? If the user doesn't have a session, we don't need to fetch the list of experiments.

idea_town/frontend/static-src/.editorconfig
```diff
@@ -0,0 +1,12 @@
+# http://editorconfig.org
```
jaredhirsch: This is rad! I'll add editorconfig to the contributor metabug

idea_town/frontend/static-src/app/models/me.js
```diff
@@ -8,18 +8,18 @@ import State from 'ampersand-state';
 //       some kind of session-check API that sends over the user model
 //       (email, name, avatar, addon status) if the user's logged in.
 export default State.extend({
+
   props: {
-    hasAddon: ['boolean', true, false]
+    user: 'object',
+    addonUrl: 'string',
+    addonName: 'string',
+    hasAddon: {type: 'boolean', required: true, default: false}
```
jaredhirsch, line 14: Nice, much more readable

jaredhirsch:
@lmorchard Looking good! I'm happy to merge this once the few nits above are addressed

idea_town/urls.py
lmorchard: I figured no, since we just removed trailing slashes from other APIs. I usually want to see slashes on things that give collections and no slash for single resources

idea_town/frontend/templates/idea_town/frontend/index.html
lmorchard: Hmm, we do have /api/ that lists links to experiments & details in the hypermedia-ish tradition. DRF doesn't list this particular user resource, though. 

Might be worth landing this as-is and I can follow up to see about adding /api/users/self to the "table of contents". Can circle back with that to make /api/ the main thing linked in the head

idea_town/frontend/static-src/app/main.js
lmorchard: Just pushed a commit to do that... 

I was also thinking with issue #83 we could use fetch & promises to more easily batch up many HTTP requests, if we end up needing to do more than these two that at some point.

idea_town/frontend/static-src/.editorconfig
lmorchard: Yeah, we might want to circle back & refine this too. I mainly added it out of annoyance because I indent 4-spaces in Python and 2-spaces in JS. Could add an .editorconfig in the root that handles both *.py and *.js

lmorchard:
So this will probably generate some new nits and more-than-nits, but: I reworked things a bit and roped in issue #83 to simplify API access.

idea_town/frontend/static-src/.editorconfig
jaredhirsch: Yeah, I've got filetype-based rules set up in vim, but that's awfully neckbeardy to expect of community members.

If you want to keep this PR going even longer, feel free to move this file into the root. Otherwise, might be worth it to file a bug.

lmorchard: Yeah, I'd say let's just file a bug for that

idea_town/frontend/static-src/app/views/header-view.js
```diff
@@ -30,20 +29,15 @@ export default BaseView.extend({
   //       to pick up csrftoken cookie changes.
   logout(evt) {
     evt.preventDefault();
-    xhr({
+    fetch(app.getDocHeadLink('x-logout', 'href'), {
```
jaredhirsch: I'm not totally convinced we need to abstract out the account-related urls, but I'm interested to see where this abstraction takes us, so let's roll with it.

Random questions: Exactly which kinds of data do you want to specify in the DOM? Just enough to get the page connected to the API server? Is one `x-app-startup` tag, containing a serialized object, cleaner than a bunch of individual tags & running to the DOM over and over?

Something to think about for the future: it seems weird that we get the logout URI via `app.getDocHeadLink`, but the login URI (which contains a redirect endpoint, too) is just hard-coded [in the view](https://github.com/lmorchard/idea-town/blob/63-user-api-endpoint/idea_town/frontend/static-src/app/views/landing-page.js#L12). Also seems like we might want to put the csrftoken (`app.me.csrfToken`) in the same place inside the app as the endpoints?

idea_town/frontend/templates/idea_town/frontend/index.html
```diff
@@ -3,6 +3,10 @@
 <head>
     <title>Idea Town</title>
     <link rel="stylesheet" href="{{ static('styles/main.css') }}">
+    <link rel="x-api-root" href="{{ url('api-root') }}?format=json">
+    <link rel="x-logout" href="{{ url('account_logout') }}">
+    {# TODO: Make this a dev / prod setting? Or, expose in an API endpoint? #}
+    <link rel="x-addon" title="Idea Town" href="https://github.com/mozilla/idea-town-addon/blob/master/dist/idea-town-addon-0.0.1.xpi?raw=true">
```
jaredhirsch: Should we push this URL down into settings.py, since we're already in here?

idea_town/frontend/static-src/app/views/header-view.js
jaredhirsch: Piggybacking on my last comment there, we've also got the addon URL specified via `app.me.addon.url`.

Would be nice to unify all that at some point.

jaredhirsch: To be clear, I'm happy to merge this bit as-is, just asking discussion questions

idea_town/frontend/static-src/app/views/header-view.js
```diff
@@ -1,5 +1,4 @@
 import app from 'ampersand-app';
-import xhr from 'xhr';
```
jaredhirsch, line 2: Mind removing xhr from package.json as well?

idea_town/frontend/static-src/app/main.js
```diff
@@ -8,33 +12,66 @@ import PageManager from './lib/page-manager';
...
-    app.pageManager = new PageManager({
-      pageContainer: document.querySelector('[data-hook=page-container]')
+
+    const apiRoot = this.getDocHeadLink('x-api-root', 'href');
+    let apiRootData;
+    let userData;
+
+    fetch(apiRoot).then((response) => response.json()).then((data) => {
+      apiRootData = data;
+      return fetch(apiRootData.me, { credentials: 'same-origin' });
```
jaredhirsch: It's a little concerning that we have to do two round-trips before we can start rendering the page (since we won't know, until the second fetch returns, whether we have a session or not, hence, whether we need to redirect or render).

I'm fine to leave this for now, provided we file a bug to optimize startup time in the not-too-distant future.

idea_town/frontend/static-src/app/main.js
```diff
@@ -8,33 +12,66 @@ import PageManager from './lib/page-manager';
...
+    }).then((response) => response.json()).then((data) => {
```
jaredhirsch: Could the callback argument be renamed from `data` to `userData` here? It seems like we don't need `userData` outside this scope.

idea_town/frontend/static-src/app/views/header-view.js
lmorchard: > I'm not totally convinced we need to abstract out the account-related urls

It starts feeling to me like trouble to hard-code URLs anywhere, and try to use Django's reverse URL lookup. It's bitten me rarely, but it has bitten me. Don't Repeat Yourself, magic numbers, and all that.

> Random questions: Exactly which kinds of data do you want to specify in the DOM? Just enough to get the page connected to the API server? Is one x-app-startup tag, containing a serialized object, cleaner than a bunch of individual tags & running to the DOM over and over?

Basically, I don't really have a plan in mind. That markup pattern in the DOM looked like a less-terrible spot to stick a few shared URLs and strings than the inline script.

Figured we could iterate on that as things come up that the app needs from the server. I can go back to the drawing board and try coming up with more of a startup data API resource. Might be worth a follow up issue though

> Something to think about for the future: it seems weird that we get the logout URI via app.getDocHeadLink, but the login URI (which contains a redirect endpoint, too) is just hard-coded in the view. 

I'd missed the login link in that view. I'd want to change that, too. 

> Also seems like we might want to put the csrftoken (app.me.csrfToken) in the same place inside the app as the endpoints?

We get that from a cookie right now. Do you mean, it should get included in an API resource too?

idea_town/frontend/templates/idea_town/frontend/index.html
lmorchard: Well, that touches settings & env vars & docker setup, so I was being lazy :) But, I can circle back to that now

idea_town/frontend/static-src/app/main.js
lmorchard: Ironically, that's one point toward putting individual URLs in the markup rather than having a single API endpoint with a table of contents :) Basically inlines that stuff

But, there's already at least 2 round trips on startup, given the fetch for user details and the initial load of experiments

PR #96, opened 2015-09-27
Prevents JS error when unauthenticated.

chuckharmston:
`app.me.user` [is empty](https://github.com/mozilla/idea-town/blob/master/idea_town/users/views.py#L18-L19) when unauthenticated, so an attempt to access `app.me.user.addon` throws an error blocking page load when you haven't yet authenticated:

```
TypeError: _ampersandApp2.default.me.user.addon is undefined
```

This prevents the error and allows login by only setting `downloadUrl` when logged in.

### DIFF
--- a/idea_town/frontend/static-src/app/views/landing-page.js
+++ b/idea_town/frontend/static-src/app/views/landing-page.js
@@ -20,8 +20,9 @@ export default BaseView.extend({
               </section>`,
 
   render() {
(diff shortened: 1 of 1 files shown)


lmorchard:
Oh, derp. Good catch!

lmorchard:
Some day soon, we should probably look into getting some frontend tests working on travis

PR #98, opened 2015-09-28
Hide Docker insanity behind a (mediocre) shell script

jaredhirsch:
Docker is a little bit of a rough ride. It might be nice to abstract away some of the most common papercuts behind some kind of a shell script.

I've created a really terrible first version of such a script below, in a classic open source nerdbait move: if you build it (badly), they will come (and make it better).

Thoughts? I'm not at all attached to using bash over python or javascript, just copied the commands that I most often use successfully and pasted them into this file.

### DIFF
--- a/bin/docker-build.sh
+++ b/bin/docker-build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# This is an absolutely terrible shell script which should always
+# successfully rebuild the client image, and probably also always
+# emit confusing semi-meaningless error messages.
+
+docker-machine create -driver virtualbox default
+docker-machine start default
+eval "$(docker-machine env default)"
+docker rmi $(docker images -f dangling=true -q)
+npm install
+docker-compose build
+docker-compose up


lmorchard:
Looks fine, but I kind of go back & forth on stuff like this. Should this be a script? Or should it be README.md advice? Most of the commands are kind of Docker basics, which itself is abstracting a bunch of insanity :neckbeard: 

I still think the `npm install` part on the host is totally strange, though. Hoping it doesn't end up actually being a symptom of something broken with our Docker setup altogether. :fearful:

jaredhirsch:
> Should this be a script? Or should it be README.md advice? 

I know, I'm wrestling with the same question.

Maybe the right answer is to have both a script for beginners (might be nice if it swallowed all those scary-looking error messages), and a markdown doc for folks who are deeper into the project, or more comfortable with docker. Thoughts?

lmorchard:
I'll just merge this for now, so we have it. Maybe part of issue #100 can be to rework this and the expanding hints & tips for Docker into CONTRIBUTORS.md or whatever

PR #107, opened 2015-09-30
Fixes #106: check for docker-machine in shell script

### LINKED ISSUE #106: check for docker-machine in docker-build.sh
`docker-machine` is not a hard dependency of the project. We should check to ensure the user has it installed before running it's commands

meandavejustice:
fixes #106

### DIFF
--- a/bin/docker-build.sh
+++ b/bin/docker-build.sh
@@ -4,9 +4,13 @@
 # successfully rebuild the client image, and probably also always
 # emit confusing semi-meaningless error messages.
 
-docker-machine create -driver virtualbox default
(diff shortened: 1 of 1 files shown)


lmorchard:
Hmm. Giving travis & eslint the side eye :eyes: because that's [what failed for this PR](https://travis-ci.org/mozilla/idea-town/jobs/83001576). Kind of an alarming error - a downloaded package's SHA hash mismatched.

[Works on a re-run though](https://travis-ci.org/mozilla/idea-town/jobs/83001576). So, I'll assume that was a network hiccup

PR #115, opened 2015-10-01
Fix #111: implement hover styles on experiment list

### LINKED ISSUE #111: Style import: implement hover styles on experiments list
This is for the 101-style-home-and-experiment-pages branch

jaredhirsch:
@meandavejustice @lmorchard R? (@johngruen comments welcome, but you're probably asleep right now ^_^)

It seemed like the intent of the wires was to use CSS tricks to render thumbnail images as circular images, but the scratchpad site hacked around this by using round images.

We can't simply use an img here to create that effect, because replaced els don't work with pseudoclasses. Instead, use a div, with the ::before hacks seen on the scratchpad site, but set the div's background via inline style attribute, so that the JS templating code can insert the correct URL.

inactive:

screenshot 2015-10-01 13 07 01

inactive hovered:

screenshot 2015-10-01 13 07 05

active:

screenshot 2015-10-01 13 07 20

active hovered:

screenshot 2015-10-01 13 07 27

### DIFF
--- a/idea_town/frontend/static-src/app/templates/experiment-row-view.js
+++ b/idea_town/frontend/static-src/app/templates/experiment-row-view.js
@@ -1,10 +1,8 @@
 export default `
-  <li class="idea-card" data-hook="show-detail"
-    // toggle a class for styling purposes:
-    class="{{#isInstalled}} active {{/isInstalled}}">
-    <img width="200" height="200" src="{{thumbnail}}">
+  <li data-hook="show-detail" class="idea-card {{#isInstalled}} active {{/isInstalled}}">
+    <div class="idea-preview-image" style="background: url({{thumbnail}}) no-repeat"></div>
     <h2>{{title}}</h2>
-    <p>Status: {{^isInstalled}} Not {{/isInstalled}} Installed</p>
+    <p>{{description}}</p>
   </li>
 `;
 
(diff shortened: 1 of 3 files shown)


idea_town/frontend/static-src/app/templates/experiment-row-view.js
```diff
@@ -1,10 +1,8 @@
 export default `
-  <li class="idea-card" data-hook="show-detail"
-    // toggle a class for styling purposes:
-    class="{{#isInstalled}} active {{/isInstalled}}">
-    <img width="200" height="200" src="{{thumbnail}}">
+  <li data-hook="show-detail" class="idea-card {{#isInstalled}} active {{/isInstalled}}">
+    <div class="idea-preview-image" style="background: url({{thumbnail}}) no-repeat"></div>
     <h2>{{title}}</h2>
-    <p>Status: {{^isInstalled}} Not {{/isInstalled}} Installed</p>
+    <p>{{description}}</p>
```
jaredhirsch, line 5: One thing I'm not doing in this patch is ellipsizing any description overflow. I'm thinking the better answer might be to have separate `description` and `tagline` properties in the API, or something.

meandavejustice:
@6a68 lgtm the Travis failure looks unrelated https://travis-ci.org/mozilla/idea-town/jobs/83189103#L533 Just restarted the build, we'll see if that was just a fluke

jaredhirsch:
@meandavejustice rebased with your travis fix, now it's green, mind hitting the big button?

PR #117, opened 2015-10-01
Fix #113: Default avatar / fix gulpfile of linting errors

### LINKED ISSUE #112: Style import: log out button in "expanded" header view is not clickable
This is for the 101-style-home-and-experiment-pages branch

Looks like the z-index of the header is lower than that of the background image, probably the simplest fix is to bump the z-index of the avatar header.

### LINKED ISSUE #113: Style import: show avatar image in header
This is for the 101-style-home-and-experiment-pages branch

Replace the current "logged in as {{user}} <button>" with the avatar image as spec'd. Click the head to log out.

meandavejustice:
- Got rid of white border around avatar.(not in mockups)
- click avatar to logout
- remove unused deps from `gulpfile` (they've been making the travis build fail)
- bump z-index of avatar so it's clickable fixes #112

### DIFF
--- a/idea_town/frontend/static-src/app/templates/header-view.js
+++ b/idea_town/frontend/static-src/app/templates/header-view.js
@@ -10,7 +10,7 @@ export default `
           </header>
           <div class="town-background"></div>
           <div id="avatar-wrapper">
-            <p>Logged in as {{session}} <button data-hook="logout">Log out</button>
+            <span class="default-avatar" data-hook="logout"></span>
           </div>
         </div>
       </div>
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,5 @@
 const autoprefixer = require('gulp-autoprefixer');
 const babelify = require('babelify');
-const bourbon = require('node-bourbon');
 const browserify = require('browserify');
 const buffer = require('vinyl-buffer');
 const cache = require('gulp-cache');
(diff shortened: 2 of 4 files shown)


idea_town/frontend/static-src/app/templates/header-view.js
```diff
@@ -10,7 +10,7 @@ export default `
           </header>
           <div class="town-background"></div>
           <div id="avatar-wrapper">
-            <p>Logged in as {{session}} <button data-hook="logout">Log out</button>
+            <i class="default-avatar" data-hook="logout"></i>
```
jaredhirsch: Why an i tag?

jaredhirsch:
@meandavejustice Looks like I broke the python linter, too :-(

Would you mind removing the `ExperimentSerializer` import from `experiments/views.py`? That should actually get us back to green

idea_town/frontend/static-src/app/templates/header-view.js
meandavejustice: bootstrap used to use `<i>` for icons, guess that was just stuck in my head. I suppose `span` would be more semantic. What do you think?

jaredhirsch: From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i#Notes): 

> The &lt;i> tag should represent a range of text with a different semantic meaning whose typical typographic representation is italicized.

jaredhirsch:
Awesome! Nice work!

PR #118, opened 2015-10-01
Fixes #103: Make session cookie last effectively forever

### LINKED ISSUE #103: Make the ideatown website session cookie last forever, by default
Long-lasting sessions will make it easy for the addon to send lifecycle events to the server, even if the user only rarely visits the site.

jaredhirsch:
@lmorchard r? Any other settings I need to tweak here?


lmorchard:
Yeah, I think this should work

Issue #110, opened 2015-10-01
Style import: add third header layout + change header mgmt


jaredhirsch:
This is for the 101-style-home-and-experiment-pages branch.

The experiments page header should have a third, thinner header state.

I'm not sure if the cleanest approach is to have the header view expose an API to allow pages to tell it to update itself, or if the header should be managed by the PageManager (i.e., its node inside the page-container node, and the header treated as a subview of pages). The latter seems cleaner, but I'm not sure.

jaredhirsch:
@meandavejustice Are you working on this one?

meandavejustice:
@6a68 Haven't written any code against it yet.

PR #119, opened 2015-10-02
101 style home and experiment pages

jaredhirsch:
@lmorchard r?

This patch brings over styles in the https://github.com/johngruen/idea-town-style-scratchpad repo (as of commit 706ef21, more updates might land overnight). The main focus was styling the experiment detail page and the home page, but it led to some changes in the landing page as well.

@meandavejustice is still working on the skinny header and some related refactoring (issue #110), but it's relatively self-contained and could land later, so, let's get this big review underway ^_^


lmorchard:
Taking a looksee

idea_town/experiments/views.py
```diff
@@ -2,7 +2,7 @@
 from rest_framework.decorators import detail_route
 from rest_framework.response import Response
 from .models import (Experiment, ExperimentDetail)
-from .serializers import (ExperimentSerializer, ExperimentDeepSerializer,
```
lmorchard, line 5: If we're not using the shallow `ExperimentSerializer`, it should probably get removed from `serializers.py` and `ExperimentDeepSerializer` just renamed to `ExperimentSerializer`.

idea_town/experiments/views.py
```diff
@@ -14,7 +14,7 @@ class ExperimentViewSet(viewsets.ModelViewSet):
     Returns a list of all Experiments in the system.
     """
     queryset = Experiment.objects.all()
-    serializer_class = ExperimentSerializer
+    serializer_class = ExperimentDeepSerializer
```
lmorchard: FWIW, the shallow `ExperimentSerializer` is used here to try to keep DB queries & response sizes down in case we have a large number of `Experiments` and don't need **all** the `ExperimentDetail`  for list views. Do we need to tweak the shallow serializer rather than just use the deep one?

I'm still working through the views here, though, so maybe all the details are needed. Maybe pagination is what will be more important if/when we have a lot of experiments.

package.json
```diff
@@ -25,8 +25,9 @@
     "mustache": "^2.1.3",
     "node-bourbon": "^4.2.3",
     "node-neat": "^1.7.2",
+    "node-normalize-scss": "^1.0.3",
     "nodemon": "^1.4.1",
-    "normalize.css": "^3.0.3"
+    "normalize.scss": "^0.1.0"
```
pdehaan, line 30: I'm not sure if we need both of these normalize libs (and I don't think we're using this one anywhere).

idea_town/frontend/static-src/styles/_theme-helpers.scss
```diff
@@ -4,10 +4,15 @@ $default: #0c99d5;
 $primary: #6fbe4a;
 $wash: #e2eaed;
 $dark: #56565a;
+$light: #f3f3f3;
+$transparent-white: rgba(255, 255, 255, .6);
+$transparent-black: rgba(0, 0, 0, .3);
+$trasparent-wash: rgba(226, 234, 237, .96);
+$transparent-stick: rgba(255, 255, 255, .96);
```
pdehaan: IIRC, in Sass we can shorten these to `rgba($white, .6)`

idea_town/experiments/views.py
lmorchard: Okay, I see what's up: The experiments set gets loaded on startup, and this API change includes all the data so that the same set can be used on detail views. 

IMO the better way is to use the shallow serializer as-is, then request the `details_url` resource on demand when an experiment detail view is visited. I might try ginning up a PR to this branch that does that.

The full set will mostly contain data the user will rarely see (i.e. because they rarely navigate to all experiments). So, the API response will be mostly wasted work (i.e. one DB query for the list of experiments, then an additional DB query for each experiment to populate details, then the whole thing bundled up as a JSON response)

idea_town/frontend/static-src/styles/main.scss
```diff
@@ -1,11 +1,8 @@
-@import '../vendor/normalize.css/normalize';
-
-// pull in tabzilla just so the build tools see it
+@import 'normalize';
```
pdehaan: Not sure if we want to move the `normalize` and `tabzilla.scss` in the main.scss or move it into a `_vendor.scss` file, so we can just do an `@import 'vendor';` here.

On a side note, I have a forked branch of Tabzilla which makes it easier to use w/ Sass so we wouldn't need to move into a /vendor/ directory and do a bunch of steps. We'd just need to hack the package.json to use [**pdehaan/tabzilla#sass-alternate**](https://github.com/pdehaan/tabzilla/tree/sass-alternate).
I'll ping the tabzilla people again and see if there is still interest in merging my PR so this would be in the tabzilla official repo.

idea_town/experiments/views.py
lmorchard: This also reminds me I need to finish tests for the API, since this change should have caused a failure :)

gulpfile.js
```diff
@@ -13,7 +12,7 @@ const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
 const imagemin = require('gulp-imagemin');
 const minifycss = require('gulp-minify-css');
-const neat = require('node-neat');
+const normalize = require('node-normalize-scss');
```
pdehaan, line 15: Now that we're using **node-normalize-scss**, I'm not sure if we need the `npm:normalize` Gulp task below.

lmorchard:
Styles & frontend build fine and look good to me. 

Main red flag I have is the small API change that will load all the things. That said, we could just move forward with that for now, because it won't have significant impact until we have more experiments. I can hack on refining to use on-demand experiment detail fetch in the frontend along with some actual tests for that API.

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -1,12 +1,91 @@
...
+              <button data-hook="uninstall" class="button primary">Disable {{title}}</button>
+            {{/isInstalled}}
+            <div class="user-count">7,654,321</div>
+          </div>
+        </div>
+      </div>
+
+      <div class="details-content">
+        <div class="details-overview">
+          <img src="{{thumbnail}}" width="260" height="auto">
```
pdehaan: I don't think `auto` is valid in HTML markup, per the [w3c validator](https://validator.w3.org/nu/#textarea):
showing_results_for_contents_of_text-input_area_-_nu_html_checker

idea_town/experiments/views.py
jaredhirsch, line 5: Sounds good

idea_town/experiments/views.py
jaredhirsch: @lmorchard Yeah, I could have explained this change more clearly: it keeps the front-end code super simple if we pull down the entire experiment when we fetch the collection at app startup. I see what you're saying about DB and API efficiency, but we're going to launch with at most 3 experiments, and the lineup isn't going to change often: there will be an Idea Town committee that signs off on changes.

So, the DB join result is going to be eminently cacheable (either inside postgres or inside django's ORM), and I think deeper performance enhancements are probably premature optimizations at this point.

If you feel strongly about keeping the deep and shallow serializers, I'm happy to back out these python changes, and update the front-end code to separately fetch the model details when the details page is loaded.

package.json
jaredhirsch, line 30: The change removes one of the libs and replaces with the sass version, and it's used [here](https://github.com/mozilla/idea-town/pull/119/files#diff-1fbf9d34e7bbb1cc8e088ac2d7105c4eR1)

idea_town/frontend/static-src/styles/main.scss
jaredhirsch: That's a good idea, I'll move the imports into a `_vendor.scss` file.

idea_town/frontend/static-src/app/templates/experiment-page.js
lmorchard: Hmm, yeah I think you can specify that in CSS, but not in markup. But I think if you leave the height out, I think it will auto scale in correct aspect ratio to width

jaredhirsch: You are [indeed](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-dim-height) [correct](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-height): the `height` attribute should be specified as a non-negative integer.

Since it's just getting discarded anyway by the browser, I'll just remove those height="auto" attributes.

BTW, here's an interesting question: is your w3c validator the same version as the one [linked](https://validator.whatwg.org/) by the whatwg? (Seriously, I can't tell)

idea_town/frontend/static-src/styles/_theme-helpers.scss
jaredhirsch: Nice catch

idea_town/experiments/views.py
lmorchard: I mainly get antsy when I see an API resource potentially return a lot of data with a lot of DB queries without pagination. It's the kind of thing that's easy to forget about later when the site starts looking hammered.

But, like I said in a later comment, I'm fine with going with this for now and queue up a refinement for later because there really won't be much data at first or even ever (relatively speaking)

idea_town/frontend/static-src/styles/_theme-helpers.scss
jaredhirsch: Turns out all the other transparent-\* colors were rgb versions of hex colors, so that code got much more readable :-)

gulpfile.js
jaredhirsch, line 15: I'm not sure either, but I really don't want to dig around in the gulpfile. PRs welcome?

lmorchard:
Oh, there we go: I _did_ have [one test that fails with this.](https://travis-ci.org/mozilla/idea-town/jobs/83361346) The `/api/me` resource is also affected by the serializer swap and a test needs a tweak:

``` diff
diff --git a/idea_town/users/tests.py b/idea_town/users/tests.py
index d79aa75..97d544a 100644
--- a/idea_town/users/tests.py
+++ b/idea_town/users/tests.py
@@ -78,7 +78,7 @@ class MeViewSetTests(TestCase):
                     {
                         'id': 4,
                         'description': 'This is a test',
-                        'details_url': 'http://testserver/api/experiments/4/details',
+                        'details': [],
                         'slug': 'test-1',
                         'thumbnail': None,
                         'title': 'Test 1',
```

jaredhirsch:
Yeah, fixing that now ^_^

lmorchard:
:shipit:

PR #122, opened 2015-10-05
Fix #110: new header for experiment page

### LINKED ISSUE #110: Style import: add third header layout + change header mgmt

meandavejustice:
- Adds new header to experiment detail page
- changes how header is managed

### DIFF
--- a/idea_town/frontend/static-src/app/main.js
+++ b/idea_town/frontend/static-src/app/main.js
@@ -6,7 +6,6 @@ import app from 'ampersand-app';
 
 import webChannel from './lib/web-channel';
 import ExperimentsCollection from './collections/experiments';
-import HeaderView from './views/header-view';
 import Me from './models/me';
 import PageManager from './lib/page-manager';
 import Router from './lib/router';
@@ -30,13 +29,6 @@ app.extend({
       // if addon state changes, dump user back to '/' and let the router handle
       // redirecting to the correct landing page
       app.me.on('change:hasAddon', () => { app.router.reload(); });
-
-      // the header is independent of the page container logic, so it lives
-      // outside the page container element
-      // TODO: seems like the PageManager should know when to refresh / hide
-      //       the header
-      app.headerView = new HeaderView({el: document.querySelector('header') });
-      app.headerView.render();
 
       app.pageManager = new PageManager({
         pageContainer: document.querySelector('[data-hook=page-container]')
--- a/idea_town/frontend/static-src/app/templates/error-page.js
+++ b/idea_town/frontend/static-src/app/templates/error-page.js
@@ -1,5 +1,6 @@
 export default `
   <section class="page" data-hook="error-page">
+    <header data-hook="main-header"></header>
     <h1>Whoops!</h1>
     <p>Looks like we broke something. Maybe try again later :-\ </p>
   </section>
--- a/idea_town/frontend/static-src/app/templates/experiment-page.js
+++ b/idea_town/frontend/static-src/app/templates/experiment-page.js
@@ -1,5 +1,6 @@
 export default `
   <section id="details" class="page" data-hook="experiment-page">
+    <header data-hook="main-header"></header>
     <div id="details">
       <div class="details-sticker"></div>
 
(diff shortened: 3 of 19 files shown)


meandavejustice:
@pdehaan This one should be less wonky :)

idea_town/frontend/static-src/app/lib/page-manager.js
```diff
@@ -37,6 +41,16 @@ export default class PageManager {
       const Ctor = this.pages[page];
       const newPage = new Ctor(opts);
 
+      if (newPage.headerScroll) {
```
lmorchard: Same basic comment I had from the other PR: This might work better as a single subview on the `base-view`, using the [built-in ampersand-view subview handling](https://github.com/AmpersandJS/ampersand-view#subviews-viewsubviews). That handles the DOM lookup, rendering, and appending. Then, let the view handle the logic of when to respond to scrolling based on options from the parent view or some other hint.

lmorchard: Alternatively, we can have the "page" views live in a wrapper "app" view that owns the header directly and swaps main content with the page switcher. [I did something like that with my C25k app this summer that manages visibility of back & help buttons via CSS based on the current view](https://github.com/lmorchard/c25k-web/blob/master/src/js/views/App.js).

meandavejustice: @lmorchard I made a patch here to see what the subviews in the base view would look like. https://gist.github.com/meandavejustice/4106d276de3a9262da0d The other method looks interesting as well

idea_town/frontend/static-src/app/lib/page-manager.js
```diff
@@ -17,6 +19,8 @@ export default class PageManager {
       throw new Error('PageManager constructor must be passed a pageContainer element');
     }
 
+    this.header = document.querySelector('header');
```
pdehaan: This selector seems a bit _generic_ and could possibly break if we had multiple `<header>` elements in the DOM, no?

idea_town/frontend/static-src/app/views/scroll-header-view.js
```diff
@@ -0,0 +1,58 @@
+import app from 'ampersand-app';
+
+import BaseView from './base-view';
+import template from '../templates/scroll-header-view';
+const changeHeaderOn = 120;
```
pdehaan: Not sure if we have any standards/guidelines on `const` variable naming yet and if we want to go with `CHANGE_HEADER_ON` vs `changeHeaderOn`.

Also, may need a brief code comment explaining what the constant/value is for, for future code travelers.

jaredhirsch: The airbnb style guide uses lowerCamelCase for consts.

I think the meaning of the const is clear from how it's used, FWIW.

idea_town/frontend/static-src/app/lib/page-manager.js
jaredhirsch: @meandavejustice About your gist: I think it might be worth splitting out a base page view (has a header, is managed by the page manager) from the base view (other reusable stuff that applies to non-page level modules).

lmorchard: FWIW, I'd still unify `ScrollHeaderView` and `HeaderView` and have that single component handle that scrolling logic internally. The [built-in `prepareView` supplied by default subview registration](https://github.com/AmpersandJS/ampersand-view/blob/master/ampersand-view.js#L290) passes a reference to the parent view, so that component could check for something like `this.parent.headerScroll` and change based on the context where it's used.

meandavejustice:
R for the updates? @pdehaan @6a68 @lmorchard

lmorchard:
Looks cleaner on a skim-through. Giving it a run through now!

Works for me!

PR #135, opened 2015-10-07
Build complete site container in Dockerfile

lmorchard:
- Temporarily install node to build frontend assets with Gulp
- Clean up node and other build tools as last step
- Separate docker-compose files for with and without containerized gulp


lmorchard:
Oops, wrong repo!

PR #136, opened 2015-10-07
Alternative docker-compose manifest that supports host-side Gulp

lmorchard:
Gulp-in-a-container seems to be going not as smoothly as I'd hoped. Here's an alternative that just runs the server and lets you run Gulp on the host without problems. It works like so:

```
docker-compose -f docker-compose-only-server.yml build
docker-compose -f docker-compose-only-server.yml up
```

That builds and runs just the Django server and Postgres, and skips setting up Gulp in a container. Then, you can do this on the host like usual:

```
npm install
gulp
```

If this works a lot better than the everything-in-Docker approach, maybe we back away from that and make this the `docker-compose.yml` replacement altogether.

@meandavejustice, you might like to try this out

### DIFF
--- a/docker-compose-only-server.yml
+++ b/docker-compose-only-server.yml
@@ -1,0 +1,8 @@
+db:
+  image: postgres:9.3
+server:
+  extends:
+    file: docker-compose-base.yml
+    service: server
+  links:
+    - db
--- a/README.md
+++ b/README.md
@@ -102,7 +102,19 @@ Idea Town is not intended to replace trains for most features, nor is it a test 
     docker-compose -f docker-compose-s3.yml build
     docker-compose -f docker-compose-s3.yml up
   ```
- 
+
+* If you'd like to run Gulp on your host computer because there are issues
+  running it in a Docker container, try this:
+  ```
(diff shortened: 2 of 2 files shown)


docker-compose-only-server.yml
```diff
@@ -0,0 +1,24 @@
+db:
+  image: postgres:9.3
+server:
+  build: .
+  dockerfile: Dockerfile-server
```
jaredhirsch: Probably should just extend docker-compose-base, now that #137 has landed

lmorchard:
Just updated with a simplified `docker-compose-only-server.yml` extending from the new base. Also added instructions to the README. I'd say if this works, we can merge it. 

If it works as the best way going forward, we can later make it the default and either offer another docker-compose file to run Gulp in docker or just forget about that notion altogether in development.

We'll still need a way to run node + gulp in building for deployment (probably using Jenkins CI), and it looks like that Docker container works best for that so far. It's just file watching and gulp restarts that really suck. One-shot runs are fine. So, it wasn't wasted work at least :)

meandavejustice:
This is working well for me. Really like having this as an option. Thanks @lmorchard

PR #137, opened 2015-10-07
Enable usage of S3 as a target for media upload & serving

lmorchard:
- Add django-storages and settings to use S3 for media
- Split up docker-compose so local customizations are easier

### DIFF
--- a/docker-compose-s3.yml-dist
+++ b/docker-compose-s3.yml-dist
@@ -1,0 +1,18 @@
+db:
+  image: postgres:9.3
+client_build:
+  extends:
+    file: docker-compose-base.yml
+    service: client_build
+server:
+  extends:
+    file: docker-compose-base.yml
+    service: server
+  links:
+    - db
+  environment:
+    - DEFAULT_FILE_STORAGE=storages.backends.s3boto.S3BotoStorage
+    - AWS_ACCESS_KEY_ID={your access key id goes here}
+    - AWS_SECRET_ACCESS_KEY={your secret access key goes here}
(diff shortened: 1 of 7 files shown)


lmorchard:
Still wrapping my head around our deployment strategy with elasticbeanstalk. But, using Amazon S3 as storage and web hosting for uploaded media (e.g. images for experiments) seems like a good idea.

Even if we don't end up using S3, django-storages seems handy for [all the backends it offers](https://github.com/jschneier/django-storages/tree/master/storages/backends) besides the basic local filesystem.

jaredhirsch:
I can take a look at this

docker-compose-s3.yml-dist
```diff
@@ -0,0 +1,18 @@
+db:
+  image: postgres:9.3
+client_build:
+  extends:
+    file: docker-compose-base.yml
```
jaredhirsch, line 5: Aha, so this 'extends' thing gives us a really simple way to toggle between prod and dev, yes? We just modify the env vars?

jaredhirsch:
Got this working without any problems.

Might make sense to also serve static files (styles, scripts) from the same S3 bucket; that'd make CDNification really simple. But it sounds like scope creep for the current issue, so I'll file a bug to discuss (#141).

Merging.

docker-compose-s3.yml-dist
lmorchard, line 5: Yeah, I'm hoping we can keep the image for the Django server built in a way that the only dev / prod difference is env vars. Don't think we'll be able to use docker-compose in prod, but we should be able to play with prod-ish features in dev like S3 by just juggling env vars

PR #139, opened 2015-10-08
add convenience script

meandavejustice:
This is a little convenience script that replaces the docker ip of the server container in your /etc/hosts file

### DIFF
--- a/bin/update-ip.sh
+++ b/bin/update-ip.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Gets the ip address of the server container and sets it in place in
+# your /etc/hosts file.
+
+IPBASE=`/sbin/ifconfig docker0 | grep 'inet addr:'| awk '{print $2}' | sed -e 's/.*://g' | cut -d '.' -f 1-2`
+NEWIP=`docker inspect ideatown_server_1 | grep '"IPAddress' | awk '{sub(/["|,]/, "", $2); print $2}' | sed 's/["|,]//g'`
+
+sed -ri 's/'$IPBASE'\.[0-9]{1,3}\.[0-9]{1,3}/'$NEWIP'/g' /etc/hosts


lmorchard:
Hmm, yeah, Marketplace found a need for this too!

bin/update-ip.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Gets the ip address of the server container and sets it in place in
+# your /etc/hosts file.
+
+IPBASE=`/sbin/ifconfig docker0 | grep 'inet addr:'| awk '{print $2}' | sed -e 's/.*://g' | cut -d '.' -f 1-2`
+NEWIP=`docker inspect ideatown_server_1 | grep '"IPAddress' | awk '{sub(/["|,]/, "", $2); print $2}' | sed 's/["|,]//g'`
```
pdehaan, line 7: http://linuxconfig.org/how-to-retrieve-docker-container-s-internal-ip-address

I think you may be able to get the IPAddress using a custom formatter:

``` sh
$ docker inspect -f '{{ .NetworkSettings.IPAddress }}' ideatown_server_1
172.17.0.2
```

lmorchard, line 7: <strike>That's actually the internal VM IP address, and not useful in the host-side /etc/hosts</strike>

Actually, I think this is okay for Linux. But, since I'm on OS X, things are different. So on Linux, that IP works - but on OS X, we need the docker machine VM IP.

lmorchard:
Oh wait, I think I merged too soon. This is actually a built-in feature of `docker-machine` on OS X (probably not applicable on Linux):

```
➜  idea-town git:(amazon-s3-media-uploads) docker-machine ip default
192.168.99.100
```

Argh, also this script as-is didn't work on OS X. I'll tweak and submit a PR

PR #140, opened 2015-10-08
Tweaks to IP update script to work on OS X and Linux

lmorchard:
@meandavejustice, does this work for you on Linux still?


meandavejustice:
Tested. works on linux. Thanks for cleaning up, your version is actually readable :sob:

Issue #125, opened 2015-10-06
experiment-detail page: experiment not rendering?


jaredhirsch:
This is what I see when I click an experiment on the /home page:

<img width="1387" alt="screenshot 2015-10-05 17 49 57" src="https://cloud.githubusercontent.com/assets/96396/10297546/823f2d76-6b89-11e5-927c-bd3c9b8951a3.png">

Might be just me, but if I revert the 2nd scratchpad PR, the experiment loads ok.

Need to investigate + fix.

jaredhirsch:
Looks (on my machine) like the page is, indeed, not rendering. The `page-container` element seems to be empty:

<img width="1247" alt="screenshot 2015-10-05 18 00 09" src="https://cloud.githubusercontent.com/assets/96396/10297670/f2fa844c-6b8a-11e5-8ef0-664efb70315d.png">

lmorchard:
fixed with https://github.com/mozilla/idea-town/commit/4c5e24075fdcfd50a18c649d2b3811330f56d04d (I presume)

PR #142, opened 2015-10-12
Fixes #105: router handling/multiple race conditions

### LINKED ISSUE #105: Frontend always redirects to /home on initial load
Whenever I visit or reload `http://ideatown.dev:8000/experiments/foobar`, I get redirected to `http://ideatown.dev:8000/home/`. We should probably navigate to the requested view on load rather than always redirect to `/home/`

meandavejustice:
fixes #105 #125 

There were some race conditions with the addon-status and experiments model.
- Addon status is now set as a variable `ideatownAddon` in the addon on `window.navigator`. This saves us a round trip to the server on hard refreshes [addon pr](https://github.com/mozilla/idea-town-addon/pull/31).
- There was a race condition where we could not navigate to an experiment detail page because the experiments collection was not populated. We now ensure that the collection is populated before calling `blastoff()`

### DIFF
--- a/idea_town/frontend/static-src/app/lib/router.js
+++ b/idea_town/frontend/static-src/app/lib/router.js
@@ -32,7 +32,7 @@ export default Router.extend({
     }
   },
 
-  // 'experiment' is a URL slug: for example, 'universal-search'
+  // 'experiment' is a URL slug: for example, 'universal_search'
   experimentDetail(experiment) {
     if (!app.me.user.id || !app.me.hasAddon) {
       this.redirectTo('');
--- a/idea_town/frontend/static-src/app/main.js
+++ b/idea_town/frontend/static-src/app/main.js
@@ -9,8 +9,10 @@ import ExperimentsCollection from './collections/experiments';
 import Me from './models/me';
 import PageManager from './lib/page-manager';
 import Router from './lib/router';
+import domReady from 'domready';
 
 app.extend({
+  router: new Router(),
 
   initialize() {
     app.webChannel = webChannel;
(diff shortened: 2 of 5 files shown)


idea_town/frontend/static-src/app/lib/router.js
```diff
@@ -32,7 +32,7 @@ export default Router.extend({
     }
   },
 
-  // 'experiment' is a URL slug: for example, 'universal-search'
+  // 'experiment' is a URL slug: for example, 'universal_search'
```
jaredhirsch, line 35: I thought Django hyphenated the URL slugs, is that not the case?

idea_town/frontend/static-src/app/views/base-view.js
```diff
@@ -1,16 +1,36 @@
...
     return mustache.render(this._template, ctx);
   },
 
+  events: {
+    'click a[href]': 'handleLinkClick'
+  },
+
+  // Handles all `<a>` clicks in the app not handled
+  // by another view. This lets us determine if this is
+  // a click that should be handled internally by the app.
```
jaredhirsch: Would you mind explaining the intent here? It seems like we're putting router information into the base view, so that every single view will be doing fallback routing. I'm guessing the intent was more to have a top-level catchall anchor click handler, but I'm not sure why we would need to do that in the first place.

idea_town/frontend/static-src/app/main.js
```diff
@@ -18,32 +20,50 @@ app.extend({
...
-      app.me.on('change:hasAddon', () => { app.router.reload(); });
+  blastOff() {
+    app.pageManager = new PageManager({
+      pageContainer: document.querySelector('[data-hook=page-container]')
+    });
 
-      app.pageManager = new PageManager({
-        pageContainer: document.querySelector('[data-hook=page-container]')
-      });
+    app.router = new Router();
```
lmorchard: Seems like you don't need to create a Router here, [since it's getting done back on line 15 in the `app.extend`](https://github.com/mozilla/idea-town/pull/142/files#diff-9fca6c8d9a32a0b5c86a4223cfac57c3R15)

idea_town/frontend/static-src/app/main.js
```diff
@@ -18,32 +20,50 @@ app.extend({
...
+
+    app.me.on('change:hasAddon', () => {
+      app.router.reload();
     });
+  },
+
+  // This is how you navigate around the app.
+  // this gets called by a global click handler that handles
+  // all the <a> tags in the app.
+  // it expects a url without a leading slash.
```
jaredhirsch: Would you mind explaining this a bit as well? It seems like all you're doing is stripping leading slashes and calling `router.history.navigate`. Could we configure the router to automatically strip leading slashes, or integrate in a more pluggable way (maybe a middleware-ish approach)?

jaredhirsch:
@meandavejustice Mind explaining what the race conditions were, and how this code resolves them? I was banging my head on the funky redirect behavior, but didn't uncover the root of the problem

idea_town/frontend/static-src/app/main.js
```diff
@@ -18,32 +20,50 @@ app.extend({
...
+  },
 
-      app.experiments = new ExperimentsCollection();
-      app.experiments.fetch();
-
-      // session won't change without a hard refresh, but addon state could, so:
-      // if addon state changes, dump user back to '/' and let the router handle
-      // redirecting to the correct landing page
-      app.me.on('change:hasAddon', () => { app.router.reload(); });
+  blastOff() {
```
lmorchard: Hmm, we get here after 1) the addon signals availability and 2) the experiments have loaded.

But, if the addon isn't installed, it never signals availability. And so experiments never load - probably okay, since we don't need them. 

But, if the experiments never load, we never end up in `blastOff()` to initialize the PageManager & Router. That seems like a problem.

idea_town/frontend/static-src/app/views/base-view.js
meandavejustice: Catchall _was_ the intent. As with the handler in `main.js` this was another thing I pulled from the ampersand generated boilerplate project. Both of these things were my way of creating a more solid foundation(trying to follow by example the ampersand way of doing things) in order to debug easier. Will reassess now that I understand the issues a bit better :)

idea_town/frontend/static-src/app/lib/router.js
meandavejustice, line 35: This actually doesn't have to do with django as we pull the slug out and match it from the experiments collection. We could easily convert hyphens to underscores in this process. Seems unnecessary for this.

idea_town/frontend/static-src/app/views/base-view.js
meandavejustice: Looks like this is totally fine to remove. Thanks for catching this!!

meandavejustice:
@6a68 @lmorchard Updates have been made :) r?

lmorchard:
Whoops! missed this update yesterday, will take a look!

lmorchard:
Works for me & cleans some things up nicely

PR #143, opened 2015-10-15
Support for deployment to Elastic Beanstalk

lmorchard:
- bin/build-prod.sh: build frontend & server, push to Docker Hub, build
  EB zip bundle to deploy from Docker Hub.
- .ebextensions/00-environment.config: initial set of EB env vars
- Management commands to help bootstrap the environment
  - createinitialsuperuser: create the first admin user from settings
  - updatefxaprovider: ensure allauth provider for FxA matches settings
- initial_data_dev.json: Remove admin user & FxA provider data, use the
  new management commands like prod
- Dockerfile simplified for minimal server build
- Dockerfile-frontend-build produces a utility container to build
  frontend assets with node + Gulp
- Use Amazon RDS database settings when available
- Move hardcoded Datadog credentials into env vars
- Make AWS credentials mandatory when using S3 storage
- .editorconfig tweak to use 2-space indentation in *.{yml,config}


lmorchard:
Got this working reliably to build a working site on Elastic Beanstalk using Docker Hub and the AWS web UI. Some loose ends:
- This uses node + Gulp in a Docker container to build frontend assets. But, if the machine we use to build the server has node>=v0.12, we can just `npm install && gulp` on the host instead.
- Haven't gotten `eb deploy` from the CLI to work. That seems important for automation.
- Used Docker Hub to get things working, but assume we'll move to an internal Docker repository.
- This doesn't include a Celery queue, because we're not using one yet.

@mostlygeek: Could definitely use some feedback here, particularly on [.ebextensions/00-environment.setup-dist](https://github.com/mozilla/idea-town/pull/143/files#diff-bc50c16765dc9905b5e41c3d2fc5e1c1R1) and [bin/build-prod.sh](https://github.com/mozilla/idea-town/pull/143/files#diff-e76513d3f243835301e3c6248f3877caR1). Feel free to look through the rest of the changes, but those might be the most interesting for deployment. 

@meandavejustice & @6a68 & anyone else interested (@nchapman ?) : Could use some eyes here for general sanity checking and to make sure we all have some familiarity with how this stuff will work.

lmorchard:
Oh, and a couple of tasks this will need before launch:
- [ ] [Contact FxA team for real ID/key for production use](https://github.com/mozilla/idea-town/issues/145)
- [ ] [Set up S3 bucket for production media](https://github.com/mozilla/idea-town/issues/146)

mostlygeek:
This is good progress! Almost exactly what I had in mind. 
- We will have a self hosted, private registry. Using DockerHub is fine for now. I can also give you access to the dev version of our registry. This one is for us to use without worry about affecting the actual registry used for deployments. 

_build-prod.sh_
- We don't need to upload a .zip file to EB. Just the `Dockerrun.aws.json` file pointing to the correct registry/app:tag is preferred. 
- Why are you tagging with timestamps?. For dev deployments we should use `latest` and for stage/prod deploys we want a specific version. The `tag` should be an argument we pass in. 

_Front end asset build process_
- What do you think about having just a node+gulp in the container? Skip the `COPY . /app`. 
- We can reuse the container, no need to npm install for every build
- We would have to volume mount + provide a command during `docker run`. It's scripted so not big deal
- We might be able to reuse / make it more generic since a lot of other projects use gulp to build assets too

lmorchard:
> - We will have a self hosted, private registry. Using DockerHub is fine for now. I can also give you access to the dev version of our registry. This one is for us to use without worry about affecting the actual registry used for deployments. 

Cool! I was hoping the process wouldn't have to change dramatically to switch to the private registry. If you give me access to the dev registry, I can try pushing to that too.

> _build-prod.sh_
> - We don't need to upload a .zip file to EB. Just the `Dockerrun.aws.json` file pointing to the correct registry/app:tag is preferred.

I ended up having to zip a `Dockerrun.aws.json` along with `.ebextensions/00-environment.config` to get env vars settings picked up. It would be great if that's just me being an EB newb, and there's a better way to handle the vars in automation.

> - Why are you tagging with timestamps?. For dev deployments we should use `latest` and for stage/prod deploys we want a specific version. The `tag` should be an argument we pass in. 

No reason this particular timestamp thing needs to stick around. I was playing with fake versioning to see changes happen while I was tinkering. In prod, I would think a git commit hash should be a better tag.

> _Front end asset build process_
> - What do you think about having just a node+gulp in the container? Skip the `COPY . /app`. 
> - We can reuse the container, no need to npm install for every build
> - We would have to volume mount + provide a command during `docker run`. It's scripted so not big deal

Huh, yeah, no reason for the `COPY` to be there! The `build-prod.sh` already mounts a volume in the docker run command. I'll remove that `COPY` and update this PR.

> - We might be able to reuse / make it more generic since a lot of other projects use gulp to build assets too

The thing that might make this less reusable is that we keep project-specific modules in the `package.json`. So, `npm install` downloads both the stuff Gulp needs to run _and_ the JS libs used in the frontend code. Hopefully that stuff changes very infrequently though.

lmorchard:
@mostlygeek Not sure if this is looking more like it, but I tweaked [bin/build-prod.sh](https://github.com/mozilla/idea-town/pull/143/files#diff-e76513d3f243835301e3c6248f3877caR1) to take a server name as a parameter - e.g. `./bin/build-prod.sh hub.stage.mozaws.net/lmorchard/idea-town:latest`

lmorchard:
What's left to push this PR forward? If it looks good, can we merge it and then work on further PRs to refine it?

bin/build-prod.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+SERVER_NAME=${1:-docker.io/lmorchard/idea-town:latest}
```
jaredhirsch, line 3: It'd probably be good to use a different default server name, or just spit out a help message and halt if one isn't set

docker-compose-base.yml
```diff
@@ -23,8 +21,15 @@ server:
     - PYTHONDONTWRITEBYTECODE=1
     - DATABASE_URL=postgres://postgres@db/postgres
     - DEBUG=True
+    - INITIAL_ADMIN_USERNAME=admin
```
jaredhirsch, line 24: Seems weird to hard-code config values here, but not in the EB config file.

Not sure how best to move to a single source of configuration truth, but would be nice to clean up in the future.

jaredhirsch:
@lmorchard This branch works locally for me, though I'm still not clear on the docker hub and elasticbeanstalk parts of this. Filed #165 to add some documentation. Merging.

docker-compose-base.yml
lmorchard, line 24: So, `docker-compose.yml` will only be used for development. Values are hardcoded here to give a consistent dev setup for everyone. 

But, Elastic Beanstalk takes the place of Docker Compose in production & staging. So, something like [00-environment.setup-dist](https://github.com/mozilla/idea-town/pull/143/files#diff-bc50c16765dc9905b5e41c3d2fc5e1c1) will be used there.

We won't want to share any config values between dev/stage/prod, I don't think.

Issue #65, opened 2015-09-10
other endpoint cleanup


jaredhirsch:
Remove templated views other than accounts/, replace (if we need them) with django-rest-framework APIs

jaredhirsch:
We should also make 404s work correctly: if assets are missing, just return a 404 and not the index page (django handles this). if a page doesn't exist, show a 404 page (ampersand handles this).

lmorchard:
FWIW, in dev / debug mode, I think Django will serve up 200 OK with an HTML error message wherever things are not found - even for static resources, since we're serving those up from Django right now. If you switch off `DEBUG` in `idea_town/settings.py` or `docker-compose.yml`, you should see normal 404s.

But yeah, there is a `frontend` app catch-all route for things outside the `static` path. So, that might complicate things too.

PR #148, opened 2015-10-16
Remove unused views

### LINKED ISSUE #64: accounts/ cleanup
- Remove templated APIs / views

lmorchard:
Removing the profile and signup views, since they're unnecessary. Leaving the /accounts/logout view for safety: In normal use, the frontend POSTs to /accounts/logout and skips the logout form. But, if someone somehow visits /accounts/logout with a GET, it would be an internal server error without a Jinja template.


lmorchard:
Also removed a view from the `idea_town/base` app, which I think is the only other unused view hanging around for issue #65

Issue #158, opened 2015-10-19
Need to implement a clearer logout button?


lmorchard:
Currently, there's a placeholder avatar in the upper right after login. Clicking on it logs the user out.

Should this be clearer? Maybe something that drops down on hover with user details & a logout button?

johngruen:
#217

Issue #157, opened 2015-10-19
Need to implement a user profile page?


lmorchard:
Currently, there's a placeholder avatar in the upper right after login. Clicking on it logs the user out.

Should this lead to a user profile page, instead? If so, what should it display?

johngruen:
#217

PR #182, opened 2015-10-21
List user profile info for contributors on experiment detail page

### LINKED ISSUE #150: Implement "brought to you by" section on experiment detail page
The "Brought to you by" section is a placeholder. Let's fill it with real content.

Some TODOs:
- [ ] Assume experiment contributors are Idea Town users. 
- [ ] Ensure users have name, title/bio, and avatar fields - ideally using an associated profile model, if not available on built-in Django User
- [ ] Generate schema migrations as needed
- [ ] Add a `contributors` field listing users to the `/experiment/` API results
- [ ] Use that data in the frontend experiments model to populate the section.

Alternatively (if we can't assume contributors are Idea Town users):
- [ ] Add a new `ExperimentContributor` data model with avatar, name, and title fields.
- [ ] Generate a schema migration for the new model
- [ ] Modify `ExperimentAdmin` in `idea_town/experiments/admin.py` to add a new inline editor for `ExperimentContributor`

lmorchard:
- Add UserProfile model, admin, and serializer
- Add `contributors` field to Experiment model, admin, and serializer
- Display user profile data for contributors on experiment detail page
- Migrations & tests

Closes #150

### DIFF
--- a/idea_town/experiments/admin.py
+++ b/idea_town/experiments/admin.py
@@ -18,6 +18,8 @@ class ExperimentAdmin(admin.ModelAdmin):
 
     prepopulated_fields = {"slug": ("title",)}
 
+    raw_id_fields = ('contributors',)
+
     inlines = (ExperimentDetailInline,)
 
 
--- a/idea_town/experiments/migrations/0003_experiment_contributors.py
+++ b/idea_town/experiments/migrations/0003_experiment_contributors.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+from django.conf import settings
+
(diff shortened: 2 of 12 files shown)


lmorchard:
This adds a UserProfile model for extra info associated with users. That info is used along with a contributors field on experiments to list people on the detail page.

To keep this PR small, the ability for users to edit their own profiles is out of scope. See also #157 & #158 for next steps / questions in that area.

But, you can edit profiles & contributor lists in the Django admin. When you create a new Experiment, there's now a contributors field that you can fill with user IDs from the search/spyglass button next to it. Blank profiles get auto-created for users as they're listed on the page. You can then edit the profiles from admin to change names, upload avatars, etc.

Issue #152, opened 2015-10-19
Implement "measurements" section on experiment detail page


lmorchard:
So, the "measurements" section in the left column is a placeholder. Let's implement it!

A possible blocker is #48 - what kind of questions should we ask / support experiments asking? Should the content in that sidebar actually represent some metrics buckets supported in an API somewhere?

If we want to punt on that issue and just make this section a list of bullet points, we could add a free-form HTML field to the Experiments model & API results for now. Adding a relational model for the list might be overkill until/unless we have it joined up with an actual metrics feature

Also, there's a "Learn More" link. Should this go to a privacy policy? Maybe related to #68

lmorchard:
For Q4, maybe let's just make this a straightforward bullet point list and look for some kind of metrics system integration in the future.

lmorchard:
f9ccfa6d6d6eba5f22c54b5651f185058b9e6c33

Issue #151, opened 2015-10-19
Implement "details" section on experiment detail page


lmorchard:
The "details" section in the left column of the experiment detail page is a placeholder. Let's fill it with real content!

Some proposed TODOs:
- [ ] Add version, updated, and contribute fields to `Experiment` model
- [ ] Generate schema migration for model changes
- [ ] Add new fields to API experiment serializer
- [ ] Use new fields in API JSON response to populate content in frontend

lmorchard:
I know [I just naysayed using free-form JSON fields in models](https://github.com/mozilla/idea-town/issues/150#issuecomment-149276244), but I wonder if this might actually be an appropriate place to use it?

If we could somehow extract add-on metadata to populate this area, that might be better than having to copy/paste it into Django admin fields. And if that extraction just plonks the manifest into a text field on the Experiment, that could be interesting.

lmorchard:
For Q4, let's just make this a simple set of addition DB fields on the Experiment model. That makes it easy to edit in the Django admin. In the future, we can migrate to something more clever like extracting an addon manifest.

lmorchard:
f9ccfa6d6d6eba5f22c54b5651f185058b9e6c33

PR #188, opened 2015-10-21
Experiment detail page sidebar content management

### LINKED ISSUE #150: Implement "brought to you by" section on experiment detail page

lmorchard:
- Add measurements sidebar module as editable content field on experiments
- Display user profile data for contributors on experiment detail page
- Display Experiment model content on detail page sidebar
- Add UserProfile model, admin, and serializer
- Add `contributors` field to Experiment model, admin, and serializer
- Add version, changelog_url, and contribute_url fields to Experiment
  model, admin, and serializer
- Migrations & tests

Closes #150, #151, #152

### DIFF
--- a/idea_town/experiments/tests.py
+++ b/idea_town/experiments/tests.py
@@ -1,7 +1,12 @@
+import json
+
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.test import Client
+from django.contrib.auth.models import User
+from rest_framework import fields
 
+from ..users.models import UserProfile
 from .models import (Experiment)  # , ExperimentDetail, UserInstallation)
 
 import logging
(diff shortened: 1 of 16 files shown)


lmorchard:
Turns out PR #182, #185, and #186 all mash together on the same things. So, it would be easier to merge & review together rather

lmorchard:
Copying these review tips over from #182:

This adds a UserProfile model for extra info associated with users. That info is used along with a contributors field on experiments to list people on the detail page.

To keep this PR small, the ability for users to edit their own profiles is out of scope. See also #157 & #158 for next steps / questions in that area.

But, you can edit profiles & contributor lists in the Django admin. When you create a new Experiment, there's now a contributors field that you can fill with user IDs from the search/spyglass button next to it. Blank profiles get auto-created for users as they're listed on the page. You can then edit the profiles from admin to change names, upload avatars, etc.

jaredhirsch:
Looking now. Thanks for the detailed reviewer notes

docker-compose-base.yml
```diff
@@ -29,7 +29,7 @@ server:
     - FXA_ACCESS_TOKEN_URL=https://oauth-stable.dev.lcip.org/v1/token
     - FXA_AUTHORIZE_URL=https://oauth-stable.dev.lcip.org/v1/authorization
     - FXA_PROFILE_URL=https://stable.dev.lcip.org/profile/v1/profile
-    - FXA_CLIENT_ID=a6cbfd1c48c97307
-    - FXA_SECRET_KEY=57a02718ead7e7186cb22ce329dddc1abf01ad63bdedd85868a4836da4b8204c
+    - FXA_CLIENT_ID=2e13dbd92f77f7df
+    - FXA_SECRET_KEY=a93644d71fa684da464fd2dba4d687c14cb21e1dd740af83895e70ede82f54cc
```
jaredhirsch, line 33: Thanks for adding this, nice to get redirected back to ideatown.dev post-sign in

idea_town/experiments/tests.py
```diff
@@ -10,20 +13,120 @@
...
+                "version": experiment.version,
+                "changelog_url": experiment.changelog_url,
+                "contribute_url": experiment.contribute_url,
+                "xpi_url": "",
+                "details": [],
+                "created": date_field.to_representation(
+                    experiment.created),
+                "modified": date_field.to_representation(
+                    experiment.modified),
+                "contributors": [
```
jaredhirsch: It would probably be cleaner to _just_ have this test verify that the contributor data is displayed--otherwise breaking the experiment API in other ways will produce a false failure here.

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -22,55 +22,59 @@ export default `
...
           <img src="{{thumbnail}}" width="260">
-          <section>
+          {{#model.measurements}}
+          <section class="measurement">
             <h3>Measurements</h3>
-            <p class="disclaimer">All data is collected anonymously and used only to help us improve this test. <a href="">Learn more.</a></p>
+            <p class="disclaimer">All data is collected anonymously and used only to help us improve this test.</p>
+            {{{model.measurements}}}
+            <!--
             <ul class="measurement">
```
jaredhirsch: It looks like the css was modified to apply styles to this via `.measurement ul`, so you probably want to remove the `measurement` class here

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -22,55 +22,59 @@ export default `
...
-                  <p class="title">Senior Strategy Dingus</span>
-                </div>
-              </li>
-              <li>
-                <img src="images/default-avatar@2x.png" width="56" height="56">
+                <img src="{{avatar}}" width="56" height="56">
                 <div class="contributor">
-                  <p class="name">Chris P. Bacon</span>
-                  <p class="title">Senior Strategy Dingus</span>
+                  <p class="name">{{display_name}}</span>
```
jaredhirsch, line 39: the other templates use lowerCamelCase for names, seems like we probably want to keep snake_case on the python side

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -22,55 +22,59 @@ export default `
       <div class="details-content">
         <div class="details-overview">
           <img src="{{thumbnail}}" width="260">
-          <section>
+          {{#model.measurements}}
```
jaredhirsch, line 25: It seems weird to pass a model reference into the templates in this way.

The approach I had been using was to manually copy state from model into view, inside the view's `render` method, then pass the view itself into the `mustache.render(template, context)` call as the context object, for example, [here](https://github.com/mozilla/idea-town/blob/48cd247/idea_town/frontend/static-src/app/views/base-view.js#L11). This approach is verbose, but you always know that the `render` will explain where the template vars come from, as seen in [this view](https://github.com/mozilla/idea-town/blob/48cd247/idea_town/frontend/static-src/app/views/experiment-page.js#L21-L25), for example.

If you'd rather use the model directly as the data object for the view, maybe do that throughout, or at least throughout the experiment-page template. Another option might be to explicitly override the BaseView's `template` method, passing `this.model` into the `mustache.render` call. That would let you avoid writing `model.foo` everywhere for keys that are top-level in the model.

Ultimately, I think it's good for the views to be consistent, but I'll leave it to you to define how that should work :-)

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -24,6 +24,9 @@ export default PageView.extend({
     this.thumbnail = this.model.thumbnail;
     this.description = this.model.description;
 
+    this.modified_date = new Date(this.model.modified);
```
jaredhirsch, line 22: Aha, there's the snake_case I was complaining about in the template ;-)

jaredhirsch:
@lmorchard Looks good overall, and the experiment data is working in my local install :-)

I've picked at the view-model relationship a bit, r+ from me once you make that relationship consistent (in whatever way you prefer). Feel free to land yourself once it's updated, or name-check me and I'll hit the big green button.

idea_town/frontend/static-src/app/templates/experiment-page.js
lmorchard, line 25: I kind of wanted to avoid having to copy every model property to a view property everywhere. Ampersand views using data binding usually reference model props directly, unless it's a view-specific derived property that transforms a model prop for display. 

We're doing things a bit differently using Mustache in place of Ampersand bindings, but figured some of the same principles could help.

idea_town/frontend/static-src/app/templates/experiment-page.js
lmorchard, line 39: Hmm, yeah, we're going to have continuing naming collisions between JS & Python standards where Python automagically names JSON API result keys for the pep8-compliant fields and Ampersand models in JS automagically carry that data into models & views. Could fix that by mapping all the names back & forth, but that might be a bit too much bureaucracy :/

idea_town/frontend/static-src/app/views/experiment-page.js
lmorchard, line 22: Yeah, Python pep8 hates lowerCamelCase variables. Django REST Framework might have some mapping facilities, but I haven't played with them yet. :/

lmorchard:
So, this might sound crazy, but... I switched over to using `model.{field}` everywhere I could in the experiment-page template, and I kind of like the verbosity. Like, in particular, in the contributors looping section there's a `title` attribute - but there's also a `model.title` attribute and the difference is nice to see.

lmorchard:
I think I'm going to merge it to get over the hump, and we can continue to evolve this stuff.

Issue #181, opened 2015-10-20
Restructure install methods to accept experiment objects from client.


meandavejustice:
<a href="https://github.com/meandavejustice"><img src="https://avatars.githubusercontent.com/u/1844554?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [meandavejustice](https://github.com/meandavejustice)**
_Monday Oct 19, 2015 at 18:52 GMT_
_Originally opened as https://github.com/mozilla/idea-town-addon/issues/36_

---

When receiving message to install an experiment (currently `update-approve`), the addon needs to accept and object from the client with all of the data to install the experiment.
will probably look like: 
`{id: "universal-search@mozilla.com", xpi: "https://people.mozilla.com/~666/universal-search-addon-ver-1.xpi"}`

Issue #180, opened 2015-10-20
Initial load: installed experiments state.


meandavejustice:
<a href="https://github.com/meandavejustice"><img src="https://avatars.githubusercontent.com/u/1844554?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [meandavejustice](https://github.com/meandavejustice)**
_Monday Oct 19, 2015 at 18:48 GMT_
_Originally opened as https://github.com/mozilla/idea-town-addon/issues/35_

---

Need to maintain a list of installed experiments to be managed.

Issue #176, opened 2015-10-20
Only send metrics updates for idea-town experiments


meandavejustice:
<a href="https://github.com/meandavejustice"><img src="https://avatars.githubusercontent.com/u/1844554?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [meandavejustice](https://github.com/meandavejustice)**
_Thursday Oct 01, 2015 at 17:20 GMT_
_Originally opened as https://github.com/mozilla/idea-town-addon/issues/30_

---

We need to filter on uninstall and install events to only send the updates we need.

Issue #174, opened 2015-10-20
Clean up URL-related prefs


meandavejustice:
<a href="https://github.com/6a68"><img src="https://avatars.githubusercontent.com/u/96396?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [6a68](https://github.com/6a68)**
_Wednesday Sep 30, 2015 at 00:18 GMT_
_Originally opened as https://github.com/mozilla/idea-town-addon/issues/28_

---

Rather than separately specifying protocol and domain name, it might be nice to just keep the current HOSTNAME pref, and have a BASE_URL pref of the form `http://ideatown.dev:8000`.

While we're at it, it would be good to change the CONTENT_URL to `ideatown.dev:8000`, instead of `localhost:1337/stub-content`.

For the sake of consistency, probably should get rid of ALLOWED_ORIGINS, and instead use BASE_URL.

PR #196, opened 2015-10-24
happy path for install flow, enable/disable addons.

### LINKED ISSUE #162: Experiment Clientside model should have installed state.
Experiment model should have a method that will check against the addon and see if the experiment is installed.

meandavejustice:
Fixes #162 #181 #180 #176 #161 #155 #174


meandavejustice:
Looks like I got some conflicting migrations in my rebase. I'll take care of that either tomorrow morning or Monday.

jaredhirsch:
@meandavejustice It can _definitely_ wait till Monday. Enjoy your weekend!

addon/index.js
```diff
@@ -25,19 +24,22 @@ const mod = new PageMod({
   onAttach: setupApp
 });
 
+if (!store.experiments) store.experiments = {};
+if (!store.availableExperiments) store.availableExperiments = [];
```
jaredhirsch: please use curly braces and put the conditional contents on a newline:

```
if (!store.experiments) {
  store.experiments = {};
}
```

addon/index.js
```diff
@@ -25,19 +24,22 @@ const mod = new PageMod({
...
+
 require('sdk/system/unload').when(function(reason) {
   if (reason === 'uninstall') {
     app.send('addon-self:uninstalled');
     if (store.experiments) {
-      uninstall(store.experiments);
+      Object.keys(store.experiments).forEach(e => uninstall({addon_id: e}));
```
jaredhirsch: Since you don't care about iterating experiments in order, and because the keys are the addon IDs (which is what you're after), you can just get the keys via `for in`:

```
for (e in store.experiments) {
  uninstall({addon_id: e});
}
```

MDN docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

addon/index.js
```diff
@@ -45,13 +47,39 @@ const button = ActionButton({ // eslint-disable-line new-cap
...
+        console.error('No experiments:: ', res);
+      }
+
+      store.availableExperiments = res.json.results.map(exp => {
+        return exp.addon_id;
+      });
+
+      AddonManager.getAllAddons(addons => {
+        addons.forEach(ex => {
+          if (isIdeatownAddon) {
```
jaredhirsch: I think you want `isIdeatownAddon(ex.id)` here

addon/index.js
```diff
@@ -45,13 +47,39 @@ const button = ActionButton({ // eslint-disable-line new-cap
...
-  badge: state.updates.length,
-  badgeColor: prefs.BADGE_COLOR,
   onClick: function() {
     tabs.open({ url: prefs.CONTENT_URL });
   }
 });
 
+function getInstalledAddons(cb) {
+  request({
+    url: 'http://ideatown.dev:8000/api/experiments?format=json',
```
jaredhirsch: Please use the base url defined in package.json

addon/index.js
```diff
@@ -45,13 +47,39 @@ const button = ActionButton({ // eslint-disable-line new-cap
...
+        addons.forEach(ex => {
+          if (isIdeatownAddon) {
+            store.experiments[ex.id] = ex.version;
+          }
+        });
+
+        // send back to client as an array
+        cb(Object.keys(store.experiments).map(k => {
+          return {id: k, version: store.experiments[k]};
+        }));
```
jaredhirsch: OK, so you have an object of the form

```
store.experiments = { id: version, id: version }
```

and you want to convert it into an array of objects of the form 

```
{id: 'some-addon-id', version: '1.0.1'}
```

One nice way to do this might be

```
let output = [];
for (ex in experiments) {
  output.push({
    id: ex,
    version: experiments[ex]
  });
}
return output;
```

Seems like the `for in` version is more readable. Thoughts?

addon/index.js
```diff
@@ -63,45 +91,35 @@ function setupApp() {
...
+    uninstall(msg);
   });
 
   app.on('uninstall-all', function() {
-    uninstall(store.experiments);
-  });
-
-  app.on('uninstall', function(msg) {
-    uninstall(msg);
+    Object.keys(store.experiments).forEach(key => uninstall({addon_id: key}));
```
jaredhirsch: Again, I think `for in` is easier to read:

```
for (id in store.experiments) {
  uninstall({addon_id: id})
}
```

addon/index.js
```diff
@@ -63,45 +91,35 @@ function setupApp() {
...
 }
 
-function uninstall(toUninstall) {
-  toUninstall.forEach(u => {
-    AddonManager.getAddonByID(u.id, function(a) {
-      a.uninstall();
-    });
-  });
+function uninstall(experiment) {
+  AddonManager.getAddonByID(experiment.addon_id, a => a.uninstall());
```
jaredhirsch: This function could be used to uninstall any addon; better wrap it in the `isIdeatownAddon` check

addon/index.js
```diff
@@ -63,45 +91,35 @@ function setupApp() {
...
 }
 
-function triggerInstalls(approvedUpdates) {
-  approvedUpdates.forEach(u => {
-    AddonManager.getInstallForURL(u.url, function(install) {
-      install.install();
-    }, 'application/x-xpinstall');
-  });
+function installExperiment(experiment) {
+  AddonManager.getInstallForURL(experiment.xpi_url, install => {
```
jaredhirsch: Ditto here: this could be used to install any addon. let's make sure it's an ideatown addon

addon/index.js
```diff
@@ -130,28 +149,39 @@ function formatInstallData(install, addon) {
...
+
+      if (store.experiments[addon.id]) {
+        delete store.experiments[addon.id];
+      }
+    }
   }
 });
 
+function isIdeatownAddon(id) {
+  return Boolean(store.availableExperiments.filter(ae => ae === id).length);
```
jaredhirsch: If I understand correctly, this function just checks if a string is in an array of strings. `indexOf` is a pretty widely-used way to do this:

```
function isIdeatownAddon(id) {
  return store.availableExperiments.indexOf(id) > -1;
}
```

idea_town/frontend/static-src/app/models/experiment.js
```diff
@@ -1,9 +1,28 @@
...
   extraProperties: 'allow',
-
+  props: {
+    enabled: {type: 'boolean', default: false}
+  },
   // This shouldn't be necessary; see comments in collections/experiments.js
-  ajaxConfig: { headers: { 'Accept': 'application/json' }}
+  ajaxConfig: { headers: { 'Accept': 'application/json' }},
+
+  // TODO(DJ): this will be removed when we start tracking
```
jaredhirsch, line 13: Please file a bug for this issue (if there isn't one already), and put the issue number in the TODO comment

idea_town/experiments/models.py
```diff
@@ -21,6 +21,7 @@ class Experiment(models.Model):
     version = models.CharField(blank=True, max_length=128)
     changelog_url = models.URLField(blank=True)
     contribute_url = models.URLField(blank=True)
+    addon_id = models.CharField(max_length=500, default='addonid@mozilla.com')
```
jaredhirsch: Every addon published or signed by AMO has an ID, so while it should be required, I don't think we should provide a default.

idea_town/frontend/static-src/app/models/experiment.js
```diff
@@ -1,9 +1,28 @@
...
+  // install state through the User Installation model.
+  initialize() {
+    app.on('addon-install:install-ended', addon => {
+      if (addon.id === this.id) {
+        this.enabled = true;
```
jaredhirsch, line 19: Addons can be enabled and disabled, as well as installed and uninstalled, so the `enabled` state may not actually be reliable. That's a bit of an edge case, so let's file a followup bug for now.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -44,10 +47,19 @@ export default PageView.extend({
...
+
+    if (!isInstall) {
+      eventType = 'uninstall-experiment';
+    }
+
+    app.webChannel.sendMessage(eventType, {
+      addon_id: this.model.addon_id,
+      xpi_url: this.model.xpi_url
+    });
+    // TODO:(DJ) need to setup some databinding and progress ui Since
```
jaredhirsch, line 60: Please file a bug and insert bug number for this todo also

jaredhirsch:
@meandavejustice OK, did the initial read-through of the code, now I'm going to try to actually install some stuff via the happy path ^_^

addon/README.md
```diff
@@ -113,14 +114,9 @@ window.addEventListener("from-addon-to-web", function (event) {
 
 Submit updates:
 ``` javascript
-sendToAddon({type: 'update-approve', detail: approvedUpdates});
+sendToAddon({type: 'install-experiment', detail: {xpi_url: 'https://people.mozilla.com/~jhirsh/universal-search-addon/addon.xpi'}});
```
jaredhirsch: it's hirs_c_h ;-)

idea_town/frontend/static-src/app/models/experiment.js
meandavejustice, line 19: We currently have the addon setup so we uninstall on disable. Should we keep them installed?

addon/README.md
meandavejustice: :blush: whoops

meandavejustice:
@6a68 Thanks for the review!

idea_town/frontend/static-src/app/views/experiment-page.js
meandavejustice, line 60: added link to issue, gh just didn't clear it since it wasn't on this line

idea_town/frontend/static-src/app/models/experiment.js
meandavejustice, line 13: added link to issue, gh just didn't clear it since it wasn't on this line

chuckharmston, line 13: You need a DJ name now.

jaredhirsch:
@meandavejustice For sure! Thanks for writing the code ^_^

I didn't have a chance to go all the way through the happy path (install idea town, install experiment, verify experiment works), but I'm in lots of meetings today, so rather than slow things down, @lmorchard will take it from here

meandavejustice:
@lmorchard Got the tests passing again.

lmorchard:
Finally going to get a look at things here this morning

addon/package.json
```diff
@@ -63,21 +56,7 @@
...
-      "name": "Universal Search",
-      "desc": "Expandable embed search from your address bar",
-      "url": "https://people.mozilla.org/~jhirsch/universal-search-addon/addon.xpi"
-    },
-    {
-      "id": "foxsplitter@piro.sakura.ne.jp",
-      "name": "Fox Splitter",
-      "desc": "Split your browser windows like a tiling wm",
-      "url": "https://addons.mozilla.org/firefox/downloads/latest/324976/addon-324976-latest.xpi?src=search"
+      "value": "http://ideatown.dev:8000/*,https://ideatown.firefox.com/*, https://people.mozilla.org/~jhirsch/universal-search-addon/*"
```
lmorchard, line 59: Does this need to list a people.mozilla.org URL? My impression was this lists the hosts where the addon will listen to messages from content. Hopefully we don't need to list the URLs of every experiment addon. At launch this should just be ideatown.firefox.com and hopefully a solution for issue #175 will let us easily switch to ideatown.dev 

(also, this nitpick need not stop this PR since the URL was already there. Maybe it just carries over into #175)

idea_town/frontend/static-src/app/models/experiment.js
lmorchard, line 13: DJ DJ?

idea_town/experiments/migrations/0003_auto_20151021_2014.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ('experiments', '0002_auto_20150831_1658'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='experiment',
+            name='addon_id',
+            field=models.CharField(max_length=500, default='addonid@mozilla.com'),
```
lmorchard: <strike>Do we maybe instead want to make `addon_id` a mandatory field with no default?</strike>

Oh wait, the next migration does just that.

idea_town/experiments/migrations/0005_auto_20151027_0242.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('experiments', '0004_merge'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
lmorchard: You could probably just lift the params for this `AlterField` and plonk them into the `AddField` for `addon_id` in `0003_auto_20151021_2014.py` and delete these two other migrations as a manual squash. That would muck up your existing dev DB, but should apply cleanly to master

addon/package.json
meandavejustice, line 59: I think you are right about the intent of allowed origins. I'll make sure to address this in #175

idea_town/experiments/migrations/0005_auto_20151027_0242.py
meandavejustice: Sweet, I was wondering if that could be done to make this simpler. Will do.

idea_town/frontend/static-src/app/lib/web-channel.js
```diff
@@ -23,9 +23,9 @@ class WebChannel {
       this.channelId = newChannelId;
     }
 
-    const message = evt.detail;
+    const message = evt.data;
```
lmorchard: Getting this error now [from a few lines above](https://github.com/meandavejustice/idea-town/blob/install-flow-happy-path/idea_town/frontend/static-src/app/lib/web-channel.js#L21): `TypeError: evt.detail is null`

Not sure if the culprit is around here, though.

idea_town/frontend/static-src/app/collections/experiments.js
```diff
@@ -13,6 +15,19 @@ export default Collection.extend({
   // updated yet? TODO: investigate
   ajaxConfig: { headers: { 'Accept': 'application/json' }},
 
+  initialize() {
+    // TODO:(DJ) This is not being run on the experiment-page.
```
lmorchard, line 19: Ah, okay, I think this might be a TODO that corresponds to what I'm seeing... if I load the home page, then click through to an installed experiment, the button says "disable". But, if I reload the experiment page, the button flips back to "enable" even though the experiment is already installed.

<strike>Probably want to file an issue for this TODO, but can probably just merge this and move forward</strike>

Ah, yep, looks like #195 is what I'm looking for in terms of issues.

lmorchard:
Okay, I think this is pretty much :+1: from me. Maybe do the migration squashing, then feel free to merge if I don't get to it.

idea_town/experiments/migrations/0003_auto_20151021_2102.py
```diff
@@ -49,4 +49,14 @@ class Migration(migrations.Migration):
             name='measurements_markup_type',
             field=models.CharField(choices=[('', '--'), ('html', 'HTML'), ('plain', 'Plain'), ('markdown', 'Markdown'), ('restructuredtext', 'Restructured Text')], max_length=30, default='plain', blank=True),
         ),
+        migrations.AddField(
```
lmorchard, line 52: I think this is squashed _too_ far back. This migration already exists with a previously merged commit, which means it would already have been run on prod & etc. We need a new migration for this new set of changes.

lmorchard:
Going ahead and merging this with a couple of tweaks c8397053fd3860e34617a34b42a8b9b1d0c29b2b

Issue #26, opened 2015-08-17
Server metrics


jaredhirsch:
Set up a basic pipeline to ship our server logs and server-side KPI data somewhere

jaredhirsch:
Maybe useful:

https://github.com/mozilla-services/heka
https://github.com/mozilla-services/django-raven-heka
https://wiki.mozilla.org/CloudServices/DataPipeline (and #datapipeline on irc)

meandavejustice:
We have a dev install to play with here: https://pipeline-cep1.dev.mozaws.net/#health
And more info: https://mana.mozilla.org/wiki/display/CLOUDSERVICES/Exploring+with+the+Mozilla+Data+Pipeline+Demo

meandavejustice:
Switching back off of this shortly, need to get a script working to post to the dev instance above, after authenticating with fxa.

Issue #175, opened 2015-10-20
[addon] Cleanly switch between dev and prod prefs


meandavejustice:
<a href="https://github.com/6a68"><img src="https://avatars.githubusercontent.com/u/96396?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [6a68](https://github.com/6a68)**
_Wednesday Sep 30, 2015 at 00:20 GMT_
_Originally opened as https://github.com/mozilla/idea-town-addon/issues/29_

---

The `jpm` package seems to assume that prefs live in package.json. What, then, is a clean way to toggle between dev and prod URLs?

Do we want to build package.json from several other files?

Or, maybe we want to have `package.json.dev` and `package.json.prod`, and some build process copies one or the other to `package.json`, so that jpm can see it?

Or maybe we just always build the addon using the prod settings, then go into FF and manually set prefs in about:config?

¯\_(ツ)_/¯

This bug covers figuring out the most add-on idiomatic way of toggling prefs, and then implementing it.

meandavejustice:
<a href="https://github.com/lmorchard"><img src="https://avatars.githubusercontent.com/u/21687?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [lmorchard](https://github.com/lmorchard)**
_Wednesday Sep 30, 2015 at 00:33 GMT_

---

> Or maybe we just always build the add-on using the prod settings, then go into FF and manually set prefs in about:config?

I think that's it. The prefs in package.json are just initial defaults - and it seems like prod settings would be the best defaults for shipping. Then, you can overlay dev-specific things in local profile via about:config. (Maybe even with a second add-on that configures everything for you. (yo dawg))

meandavejustice:
<a href="https://github.com/lmorchard"><img src="https://avatars.githubusercontent.com/u/21687?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [lmorchard](https://github.com/lmorchard)**
_Monday Oct 19, 2015 at 17:51 GMT_

---

Here's what might be a crazy idea: What if the idea town addon listens for client messages from the known hosts for prod, dev, staging, local dev, etc? One of the messages on login could supply the addon with config details for that server - API endpoint, etc. Then, when you login, the addon can auto-configure itself for that particular server. 

[...]

PR #206, opened 2015-10-29
metrics endpoint

### LINKED ISSUE #178: Expose metrics API to experiment add-ons
<a href="https://github.com/6a68"></img></a> **Issue by [6a68](https://github.com/6a68)**
_Thursday Oct 15, 2015 at 18:40 GMT_
_Originally opened as https://github.com/mozilla/idea-town-addon/issues/33_

---

It would be great if the idea town addon exposed a metrics recording API to other addons, so that metrics data from any experiment could land in a dashboard that's provided by idea town.

meandavejustice:
fixes #178 #26 

cc/ @lmorchard @6a68 

Experiment addons will use [`addon-api`](./addon-api) to send metrics


lmorchard:
Looks like a couple of small code style issues:

```
idea_town/metrics/test.py:1:1: F401 'json' imported but unused
idea_town/metrics/views.py:38:101: E501 line too long (105 > 100 characters)
```

addon/package.json
```diff
@@ -19,7 +19,7 @@
   "dependencies": {
     "array-includes": "^3.0.1",
     "exclude": "^1.0.0",
-    "xtend": "^4.0.0"
+    "idea-town": "0.0.1"
```
lmorchard: Maybe it's not possible, but could this npm package be part of the main idea-town repo in a subdirectory rather than a new github repo? We just spent a bit of gymnastics to stand up a cluster of repos and then merged them. It would be great if we could keep everything still all together.

meandavejustice: for sure

addon-api/ping-server.js
```diff
@@ -4,13 +4,14 @@
  * http://mozilla.org/MPL/2.0/.
 */
 
-const prefs = require('sdk/simple-prefs').prefs;
+const config = require('./package.json').CONFIG;
```
lmorchard: We should have a way for an individual projects to override this config - maybe these end up being methods on an object initialized with the config? (Or could pass config to each function, but that feels messy.) An experiment addon will probably want to change these with prefs, or pull the values from the Idea Town addon prefs.

Should also ship this to NPM with `ideatown.firefox.com` production config by default (even though it doesn't exist yet)

Maybe this plays into #175 for settings in dev / prod switching?

addon-api/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+``` javascript
+
+const addonData = {
+  'id': addon.id,
+  'name': addon.name,
+  'version': addon.version
+};
+
+require('idea-town').metric('universal-search:result-chosen', {
```
lmorchard: Yeah, re: CONFIG - maybe this module should work more like a factory like so:

``` javascript
var ideaTown = require('idea-town')({
    "BASE_URL": "https://ideatown.firefox.com",
    "HOSTNAME": "ideatown.firefox.com"
});
ideaTown.metric('universal-search:result-chosen', {
...
```

Could ship with CONFIG defaults that then get overwritten by merging in whatever the user supplies to the module function.

meandavejustice: This looks much better.

lmorchard:
LGTM :+1:

Issue #209, opened 2015-11-02
Abstract `#main-footer` into a view


johngruen:
Right now the `#main-footer` links are loaded statically in the index template. Since we're using flex box, this causes a gnarly little flash while the ampersand-y parts of the DOM rerender as views are swapped out. Even though the footer persists across all pages, we should load it in dynamically to mitigate this issue.

johngruen:
#209

lmorchard:
Sounds like merging #210 fixed this - reopen if I'm mistaken!

PR #210, opened 2015-11-02
fix(footer): move footer into page views

johngruen:
@lmorchard @meandavejustice so, the main point of this PR is to move the footer into views (#209), but there's some bonus sass fiddling going on as well. 

Not sure why github is showing so much diff on all the template pages, but i suspect it's b/c i modified the root elements.

### DIFF
--- a/idea_town/frontend/static-src/app/templates/home-page.js
+++ b/idea_town/frontend/static-src/app/templates/home-page.js
@@ -1,7 +1,10 @@
 export default `
-  <section class="page" data-hook="home-page">
-    <header data-hook="main-header"></header>
-    <ul id="idea-card-list" class="experiments"></ul>
-  </section>
+  <div class="page" >
+    <section data-hook="home-page">
+      <header data-hook="main-header"></header>
+      <ul id="idea-card-list" class="experiments"></ul>
+    </section>
+    <div data-hook="main-footer"></div>
+  </div>
 `;
 
--- a/idea_town/frontend/static-src/styles/modules/_buttons.scss
+++ b/idea_town/frontend/static-src/styles/modules/_buttons.scss
@@ -1,7 +1,7 @@
 .button {
   border: 0;
   border-radius: 4px;
-  font-weight: 200;
+  font-weight: 500;
   height: 36px;
   padding: 0 16px;
   text-align: center;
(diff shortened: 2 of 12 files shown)


idea_town/frontend/static-src/app/templates/home-page.js
```diff
@@ -1,7 +1,10 @@
 export default `
-  <section class="page" data-hook="home-page">
-    <header data-hook="main-header"></header>
-    <ul id="idea-card-list" class="experiments"></ul>
-  </section>
+  <div lass="page" >
```
pdehaan: Typo: 'lass' -> 'class'

idea_town/frontend/static-src/styles/modules/_buttons.scss
```diff
@@ -9,8 +9,10 @@
   transition: background 50ms;
 
   &.large {
-    font-size: 3.2rem;
-    height: 72px;
+    border-radius: 8px;
+    box-shadow: 0 -4px 0 rgba(0, 0, 0, .2) inset;
```
pdehaan: I believe in Sass-land, you can also do 'rgba($black, 0.2)'

idea_town/frontend/static-src/app/templates/home-page.js
johngruen: yeesh

idea_town/frontend/static-src/styles/modules/_buttons.scss
johngruen: sass-land

lmorchard:
Taking a peek at this, too!

johngruen:
Thank @lmorchard ... I fixed @pdehaan's nits, not sure why github doesn't nuke the comments on that diff

idea_town/frontend/static-src/app/views/footer.js
```diff
@@ -0,0 +1,6 @@
+import BaseView from './base-view';
+import template from '../templates/footer';
```
lmorchard: Was this file maybe included by mistake? There's no `templates/footer.js` and only `footer-view.js` gets used in `page-view.js`.

Gulp build on my machine breaks unless I delete this. (<strike>Also weird that Travis didn't complain.</strike> [Travis didn't complain because we don't do a full build as a check](https://github.com/mozilla/idea-town/pull/211))

lmorchard:
Except for the one file Gulp tripped over on my machine, this works & looks good for me. If it helps, I could just delete that as part of merging unless there's something else that depends on it that I missed.

idea_town/frontend/static-src/app/views/footer.js
johngruen: hey @lmorchard...good catch. i had originally called this file footer.js then changed it to footer-view.js to better match the naming convention for views and pages. i guess i forgot to git remove this one

Issue #163, opened 2015-10-19
Set up some kind of offline work queue


lmorchard:
We've tossed around ideas of needing to send notifications, emails, & etc. We should use an offline work queue for that. Celery & RabbitMQ has been popular, but maybe not the best way going forward. Celery backed by Redis in dev and [Amazon ElastiCache Redis](https://aws.amazon.com/elasticache/) in production could be interesting.

lmorchard:
It occurs to me that we might want to hold off actually implementing something here until we also have a defined feature that wants to use it.

meandavejustice:
closing since we don't have use for a queue yet

Issue #129, opened 2015-10-06
Subscribe new users to testpilot mailing list


nchapman:

meandavejustice:
We may have another issue for tracking this, but I'm not seeing it. @nchapman Where can I look for updates to idea-town mailing list status? I imagine this will come after any tweaks on the name. But would be nice to have this for internal launch for other users that stumble in.

meandavejustice:
reassigning this to @clouserw to continue the discussion.

clouserw:
I think this is talking about a mailing list that doesn't exist right now?  What would be the purpose of the mailing list?

lmorchard:
Currently, we have an auth system that lets you sign up / log into Test Pilot with FxA. If you don't have an *@mozilla.com email address, your account is set to inactive and you're shown a "coming soon" kind of message.

So, in effect, the "mailing list" is the set of email addresses of inactive accounts. Presumably, we'd notify them when they can come back and actually use Test Pilot. 

We don't have a tool to actually email those people yet, though. But, that was the notion for doing an internal launch followed by opening up incrementally to a larger audience. We could set progressively bigger sets of inactive accounts as active and notify them as we go.

(Oh, and there's also code to auto-activate vouched mozillians.org users as active, but I think there was some discussion to keep that turned off for now.)

lmorchard:
We'd also had some discussion about having an actual google-groups-style mailing list for announcements and such, but I think we drifted away from that a bit

clouserw:
Got it, thanks.  Based off of that comment I'm going to close this issue -- it sounds like it has expired.  The addresses with an inactive account aren't a mailing list.  When we want to email them we could write a tool, or we could just run a query and export the addresses to a text file.  Once we figure out how we want to handle the accepting-new-users flow we can sort that out.  Thanks!

PR #212, opened 2015-11-04
Invite-only mode with auto-activation for mozilla.com emails and vouched Mozillians

lmorchard:
TL;DR: This PR adds an invite-only mode to the site with exceptions for pre-defined qualifications. 

Broke this up into some more bite-sized commits that might be easier to review in order. 

Anyone can login with FxA and have an account created. But, if they don't match qualifications to be auto-activated, their account will be flagged inactive and they will see a special page for inactive accounts. 

Users who match the qualifications will be auto-activated and can use the site. Currently, those qualifications are either having an @mozilla.com email address or being a vouched member of mozillians.org.

If we later turn off invitation mode, all previously inactive accounts will become auto-activated on the next login attempt. If a user gains qualification to be activated, they will also be auto-activated on the next login attempt - e.g. If they've been vouched on mozillians.org since the initial signin attempt.

Users can be manually activated in the Django admin tool.

Next steps from this PR:
- # 213 - The inactive account page is terrible. It should be reworked to look like mozilla/idea-town-splash and have some actually decent messaging
- a way to invite users in bulk with email notifications. We can handle that in a future PR that adds such a tool to the Django admin. We'll need a background job queue (#163) to make that happen, too.
- a way add users to an Idea Town mailing list (#129). But, we can come up with addresses for that list eventually with a query on user profiles with `invite_pending=True`


docker-compose-base.yml
```diff
@@ -33,3 +33,5 @@ server:
     - FXA_SECRET_KEY=a93644d71fa684da464fd2dba4d687c14cb21e1dd740af83895e70ede82f54cc
     - DATADOG_API_KEY=4f3b967bbf13c7769ac4fa89efda0fae
     - DATADOG_APP_KEY=ddd45a7b3a3cb90ff1baf20ae8cfab04d1937038
+    - ACCOUNT_INVITE_ONLY_MODE=True
+    - MOZILLIANS_API_KEY=ff33fa1255ce0ad59be17271093ab851c6357936
```
pdehaan, line 37: Sorry, ignorant question here, but are all these KEYS kosher in GitHub? Part of me mildly panics when I see `x_SECRET_KEY` or thinking that somebody could use our `MOZILLIANS_API_KEY` for their apps (although not really an issue if the keys only work for certain pre-approved URLs).

lmorchard, line 37: I'm thinking they're okay, because we'll use different credentials & keys for production. They're just here for consistent dev environments that work out of the box without requiring new devs to dig up all these things on their own. Any data that ends up submitted or used with these APIs is throwaway.

lmorchard:
Rebased against master

lmorchard:
Rebase against master again again

Issue #141, opened 2015-10-08
Serve static files from S3?


jaredhirsch:
PR #137 allows S3 to be used for user-uploaded media, which means experiment thumbnails, in our case.

If we push static files (scripts, styles, fonts) to S3 as well, then we'd have a really nice first version of a build chain: put stuff on S3, use a CDN in prod.

Worth doing?

lmorchard:
Some thinking out loud ahead... TL;DR: Seems like a good idea, will introduce moving parts & gotchas.

In one sense, this could make the Django docker image simpler. I'm playing with [a build script](https://github.com/mozilla/idea-town/compare/docker-tweaks...lmorchard:docker-tweaks#diff-1e5c4882847e3eefa76de60a300139b8R1) that builds a throwaway container with gulp & node that prepares the frontend assets before building the server container image around them.

Instead, we could build the Django server image without frontend assets and change the [STATIC_URL setting](https://docs.djangoproject.com/en/1.8/howto/static-files/) to point at a CDN. 

But, then, we have two sources of static files: We need frontend assets built by node + gulp. But then we also need a few static assets from Django using the `collectstatic` command (e.g. `/admin/` stuff, etc). Not infeasible, just another detail to automate.

We'd also no longer have a single atomic container holding everything that represents the current version of the site. Deployment would be multi-part: Update both the CDN and the server container. 

We'd have to version static assets, maybe stick the SHA hash of the git commit used to build the server container in front of the URL path - e.g. `/2ac76d8b2f49e2eedf842a2bf7e1c536ceeee358/app/app.js`. That way, if we have to roll back the server, the old assets are still there and at the same presumably expires-never cached URLs. I _think_ we can do this via env vars and the `STATIC_URL` setting in Django - deployment would just have to include that SHA hash in the var.

Wastes a little space on S3, but we could eventually have scripts that clean up old versions once we're sure we'll never roll back to them. And the more I think about it, the more versioning assets like this sounds like something we should be doing anyway, CDN or not

So, overall, I think it's worth putting together. Just want to be sure to cover all the gotchas

meandavejustice:
ping @lmorchard is this issue still relevant?

lmorchard:
Going to say no, since we're serving up static assets from the Docker container and that's a happy path for ops so far

PR #215, opened 2015-11-05
Use a Gravatar for user avatar by default

lmorchard:
- Use profile avatar in header view
- Add UserProfile data to /api/me resource
- Add Gravatar URL as fallback for avatar in UserProfile serializer
- Add avatar to admin
- Add gravatar_url() utility function
- Tweaks to tests

### DIFF
--- a/idea_town/frontend/static-src/app/templates/scroll-header-view.js
+++ b/idea_town/frontend/static-src/app/templates/scroll-header-view.js
@@ -8,7 +8,12 @@ export default `
         </header>
         {{^scrolled}}
           <div id="avatar-wrapper">
-            <span class="default-avatar" data-hook="logout"></span>
+            {{#avatar}}
+              <img class="avatar" src="{{avatar}}" width="41" height="41" data-hook="logout">
+            {{/avatar}}
+            {{^avatar}}
+              <span class="default-avatar" data-hook="logout"></span>
+            {{/avatar}}
           </div>
         {{/scrolled}}
         {{#scrolled}}
--- a/idea_town/frontend/static-src/styles/modules/_avatar.scss
+++ b/idea_town/frontend/static-src/styles/modules/_avatar.scss
@@ -5,10 +5,14 @@
   flex: 0 0 48px;
   height: 48px;
   z-index: 2;
-  
+
   .default-avatar {
     @include hidpi-background-image('default-avatar', 41px 41px);
     flex: 0 0 41px;
     height: 41px;
   }
+}
+
+img.avatar {
+  border-radius: 50%;
 }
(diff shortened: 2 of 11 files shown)


idea_town/frontend/static-src/styles/modules/_avatar.scss
```diff
@@ -1,14 +1,18 @@
-#avatar-wrapper {
+#avatar-wrapper, .avatar-wrapper {
```
meandavejustice: Doesn't look like this class is being used anywhere

idea_town/frontend/static-src/app/templates/scroll-header-view.js
```diff
@@ -8,7 +8,12 @@ export default `
         </header>
         {{^scrolled}}
           <div id="avatar-wrapper">
-            <span class="default-avatar" data-hook="logout"></span>
+            {{#avatar}}
+              <img class="avatar" src="{{avatar}}" width="41" height="41" data-hook="logout">
```
meandavejustice, line 12: Should we just set `context.avatar = 'static/default-avatar.png'`to avoid the mustache templating here, and simplify the template? re: our conversation yesterday.

idea_town/utils.py
```diff
@@ -66,6 +66,31 @@ def show_image_display(obj):
...
+            img_url = "%s%s" % (settings.MEDIA_URL, avatar_path)
+        elif hasattr(obj, 'user'):
+            img_url = gravatar_url(obj.user.email)
+        return ('<a href="%s" target="_new"><img src="%s" height="%s" /></a>' %
+                (img_url, img_url, height))
+
+    show_avatar_display.allow_tags = True
+    show_avatar_display.short_description = field_name
+
+    return show_avatar_display
```
meandavejustice, line 91: Nice touch adding these into the admin

idea_town/frontend/static-src/styles/modules/_avatar.scss
lmorchard: Oh, good catch!

idea_town/frontend/static-src/app/templates/scroll-header-view.js
lmorchard, line 12: Hmm, that could get vaguely complicated. We can't hardcode a literal `/static/`, because that's actually a configurable path we might change if/when we go to a CDN / S3 per #141. Sticking the default avatar in as a background image follows the CSS paths around nicely

lmorchard:
Rebased against master & resolved some conflicts

idea_town/frontend/static-src/app/templates/scroll-header-view.js
meandavejustice, line 12: Ahh yes, that makes sense. I'm cool with this for now.

Issue #224, opened 2015-11-10
indicate idea town is still WIP on the landing page


johngruen:
Indicate to users that idea town is not yet shipping.

lmorchard:
https://github.com/mozilla/idea-town/commit/5c2f837ee8a47680f32bef493c5e77ba2b1eacd6

Issue #159, opened 2015-10-19
Implement transitions between page views?


lmorchard:
We're building this shiny single-page-app. Maybe it should have some tasteful transitions between page views?

jaredhirsch:
The ampersand-page-switcher has hooks built in to handle cross-page transitions

Edited: sorry, I mean the [ampersand-view-switcher](https://github.com/AmpersandJS/ampersand-view-switcher). See the readme for animation examples.

johngruen:
@lmorchard yeah, probably should. Preferably we could get away with using CSS to fade into views a bit.

Only issue I see is that RN the `#main-footer` is in that outer index file and isn't injected with the views.

johngruen:
Oh, yeah so PR #210 should make this a lot easier to implement.

lmorchard:
https://github.com/mozilla/idea-town/commit/5c2f837ee8a47680f32bef493c5e77ba2b1eacd6

Issue #91, opened 2015-09-25
Tabzilla tab invisible on 'install addon' landing page


jaredhirsch:
White-on-white. What's needed to make the tab shape blue?

pdehaan:
It looks like we'd probably have to convert the Tabzilla image assets to our custom colors: https://github.com/mozilla/tabzilla/tree/master/media/img (and probably host them ourselves -- or convert to SVG or whatever).

I think @johngruen was actually looking into this.

lmorchard:
https://github.com/mozilla/idea-town/commit/5c2f837ee8a47680f32bef493c5e77ba2b1eacd6

PR #225, opened 2015-11-10
feat(landing-page): let users know idea town is still WIP

johngruen:


johngruen:
#224 #159 and #91

lmorchard:
Experiment detail page is a bit wonky now, but everything else is better. Tests pass, works for me

PR #228, opened 2015-11-10
Also use pageshow event for URL change cue

### LINKED ISSUE #124: experiment-detail page: back button broken
Seems like the history / pushState behavior needs some work.

STR:
1. Load an experiment detail page
2. Hit the 'back' button
3. The 'home' route is shown in the urlbar, but the 'home' page doesn't load

lmorchard:
Closes #124

### DIFF
--- a/idea_town/frontend/static-src/app/main.js
+++ b/idea_town/frontend/static-src/app/main.js
@@ -50,6 +50,9 @@ app.extend({
 
     if (!app.router.history.started()) {
       app.router.history.start();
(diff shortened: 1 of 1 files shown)


lmorchard:
I'm not _entirely_ sure if this is the right fix, since I'm still wrapping my head around the events and issues with them. But, this seems to fix the issue. Might be worth an upstream ampersand patch if it's the right thing to do

jaredhirsch:
Per all the crazy links I dropped in IRC earlier, adding a no-op `unload` listener should resolve any bfcache issues in FF. I kinda think this fix is cleaner, though

PR #236, opened 2015-11-11
Fix header disappearance and clean up templates in experiment view

### LINKED ISSUE #194: Header disappears on rerender (experiment detail page)
You can trigger a rerender by disabling/enabling the experiment.

meandavejustice:
Fixing templates in experiment view

fixes #194 

cc/ @6a68

### DIFF
--- a/idea_town/frontend/static-src/app/templates/experiment-page.js
+++ b/idea_town/frontend/static-src/app/templates/experiment-page.js
@@ -7,83 +7,53 @@ export default `
 
         <div class="details-header-wrapper">
           <div class="details-header">
-            <h1>{{model.title}}</h1>
+            <h1 data-hook="title"></h1>
             <div class="idea-controls">
-              {{#model.enabled}}
-                <button data-hook="uninstall" class="button primary">Disable {{model.title}}</button>
-              {{/model.enabled}}
-              {{^model.enabled}}
-                <button data-hook="install" class="button primary">Enable {{model.title}}</button>
-              {{/model.enabled}}
-              <div class="user-count">7,654,321</div>
+              <button data-hook="uninstall" class="button primary">Disable <span data-hook="title"></span></button>
+              <button data-hook="install" class="button primary">Enable <span data-hook="title"></span></button>
             </div>
           </div>
         </div>
 
-        <div class="details-content">
-          <div class="details-overview">
-            <img src="{{model.thumbnail}}" width="260">
-            {{#model.measurements}}
-            <section class="measurement">
-              <h3>Measurements</h3>
-              <p class="disclaimer">All data is collected anonymously and used only to help us improve this test.</p>
-              {{{model.measurements}}}
-            </section>
-            {{/model.measurements}}
-            <section>
-              <h3>Brought to you by</h3>
-              <ul class="contributors">
-                {{#model.contributors}}
-                <li>
-                  <img class="avatar" src="{{avatar}}" width="56" height="56">
-                  <div class="contributor">
-                    <p class="name">{{display_name}}</span>
-                    <p class="title">{{title}}</span>
-                  </div>
-                </li>
-                {{/model.contributors}}
-                {{^model.contributors}}
-                  <!-- TODO: need a blank slate case for no contributors? -->
-                  <li>
-                    <img src="images/default-avatar@2x.png" width="56" height="56">
-                    <div class="contributor">
-                      <p class="name">To be announced</span>
-                    </div>
-                  </li>
-                {{/model.contributors}}
-              </ul>
-            </section>
-            <section>
-              <h3>Details</h3>
-              <table class="stats">
-                {{#model.version}}
-                <tr>
-                  <td>Version</td>
-                  <td>{{model.version}}{{#model.changelog_url}} <a href="{{model.changelog_url}}">changelog</a>{{/model.changelog_url}}</td>
-                </tr>
-                {{/model.version}}
-                <tr>
-                  <td>Last Update</td>
-                  <td>{{modified_date}}</td>
-                </tr>
-                <tr>
-                  <td>Contribute</td>
-                  <td><a href="{{model.contribute_url}}">{{model.contribute_url}}<a></td>
-                </tr>
-              </table>
-            </section>
-          </div>
+        <div data-hook="details">
+          <div class="details-content">
+            <div class="details-overview">
+              <img data-hook="thumbnail" src="" width="260">
+              <section data-hook="measurements-container" class="measurement">
+                <h3>Measurements</h3>
+                <p class="disclaimer">All data is collected anonymously and used only to help us improve this test.</p>
+                <div data-hook="measurements-html"></div>
+              </section>
+              <section>
+                <h3>Brought to you by</h3>
+                <ul class="contributors"></ul>
+              </section>
+              <section>
+                <h3>Details</h3>
+                <table class="stats">
+                  <tr data-hook="version-container">
+                    <td>Version</td>
+                    <td>
+                      <span data-hook="version"></span>
+                      &nbsp;<a data-hook="changelog-url">changelog</a>
+                    </td>
+                  </tr>
+                  <tr>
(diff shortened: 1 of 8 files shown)


idea_town/frontend/static-src/app/views/details-view.js
```diff
@@ -0,0 +1,83 @@
...
+    extraProperties: 'allow'
+  })
+});
+
+export default BaseView.extend({
+  _template: template,
+
+  bindings: {
+    'model.modified_date': {
+      type: function(el, value) {
```
lmorchard: Getting an eslint warning here:

```
20:13  warning  Missing function expression name  func-names
```

Not sure I actually like this particular rule. Helps sometimes in debugging, but usually feels like boilerplate to me. But, we should either resolve this warning or disable the rule in `.eslintrc`

lmorchard: Actually, you could use an arrow function here, which I think skips the function naming issue too.

idea_town/frontend/static-src/app/views/details-view.js
```diff
@@ -0,0 +1,83 @@
...
+      type: 'text',
+      hook: 'detail-copy'
+    },
+
+    'model.details.headline': {
+      hook: 'detail-headline'
+    }
+  },
+
+  afterRender() {
```
lmorchard: [`renderCollection()` in ampersand-view](https://github.com/AmpersandJS/ampersand-view/blob/master/ampersand-view.js#L363) would be handy here. It takes care of building a generic `CollectionView` and rendering it as a subview. [It works something like this](https://github.com/lmorchard/c25k-web/blob/master/src/js/views/WorkoutList.js#L12).

idea_town/frontend/static-src/app/views/details-view.js
```diff
@@ -0,0 +1,83 @@
+import Model from 'ampersand-model';
+import Collection from 'ampersand-collection';
+import CollectionRenderer from 'ampersand-collection-view';
```
lmorchard: If you use [`renderCollection()`](https://github.com/AmpersandJS/ampersand-view/blob/master/ampersand-view.js#L363) you don't need to import `ampersand-collection-view` here like this to create a generic renderer

idea_town/frontend/static-src/app/views/details-view.js
meandavejustice: thanks!

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -13,35 +14,39 @@ export default PageView.extend({
...
-      document.body.classList.add('active');
-    } else {
-      document.body.classList.add('inactive');
-    }
     document.body._id = document.body.id;
     document.body.id = 'idea-view';
 
     PageView.prototype.render.apply(this, arguments);
+
+    this.renderSubview(new DetailsView({
```
lmorchard: FWIW, you can simplify this a bit by declaring something like this after `bindings`:

```
subviews: {
  details: {
    container: '.details-contain',
    constructor: DetailsView
  }
}
```

Then, `renderSubview` will get automatically called. Also, by default the constructor [will get called with `{ parent: this }`](https://github.com/AmpersandJS/ampersand-view/blob/master/ampersand-view.js#L276) from this view, and so `DetailsView` can access the parent model with `this.parent.model`. You can use that directly or [simplify the reference with an alias in the subview](https://github.com/lmorchard/c25k-web/blob/master/src/js/views/AudioCues.js#L9).

idea_town/frontend/static-src/app/views/header-view.js
```diff
@@ -5,21 +5,66 @@ import template from '../templates/header-view';
 import scrollTemplate from '../templates/scroll-header-view';
 const changeHeaderOn = 100;
 
+import Model from 'ampersand-model';
+const HeaderModel = Model.extend({
```
lmorchard: Another Ampersand fun fact: [ampersand-view](https://github.com/AmpersandJS/ampersand-view/blob/master/ampersand-view.js#L36) is a subclass of [ampersand-state](https://github.com/AmpersandJS/ampersand-state), also the parent class for [ampersand-model](https://github.com/ampersandjs/ampersand-model).

What that means is, you can declare all these props directly on the view itself and make bindings without the `model.` prefixes or the private model class & instance.

[Here's an example Clock view from my c25k app](https://github.com/lmorchard/c25k-web/blob/master/src/js/views/Clock.js) that binds to its own properties & derived properties that get calculated from the parent model.

idea_town/frontend/static-src/app/views/experiment-page.js
meandavejustice: I actually had to place this here since the view that this view extends from uses the subviews hash. 

Unfortunately, the way ampersand handles subviews is that the hash is overridden when extended. I originally started to write a thing where you could set `_subviews` hash and in the `BaseView.initialize` method we would merge them `merge(this.subviews, this._subviews)` but that was still causing issues. :(

lmorchard: Hmm, looking at that again, this seems like a problem with the way we're doing the base -> page view subclasses. Might be better to have page view use `renderSubview` instead of declaring `subviews` so we can keep most views simpler.

You might also be able to write something like this [with lodash](https://lodash.com/docs#assign) (or something else) to selectively inherit & override:

```
subviews: _.assign({
  details: {
    container: '.details-contain',
    constructor: DetailsView
  }
}, PageView.prototype.subviews)
```

idea_town/frontend/static-src/app/views/details-view.js
```diff
@@ -0,0 +1,78 @@
...
+      hook: 'changelog-url'
+    },
+
+    'model.thumbnail': {
+      type: 'attribute',
+      name: 'src',
+      hook: 'thumbnail'
+    },
+
+    'model.details.image': {
```
lmorchard: `details` is not just a single object, but [an array of `ExperimentDetails` entries from the API](https://github.com/mozilla/idea-town/blob/master/idea_town/experiments/serializers.py#L26). This should be a collection like `Contributors` with an associated view & template.

idea_town/frontend/static-src/app/views/header-view.js
meandavejustice: Oh this is much cleaner

idea_town/frontend/static-src/app/templates/contributor-view.js
```diff
@@ -0,0 +1,9 @@
+export default `
```
lmorchard: This might sound backward, but: For simple templates like this - i.e. under 20 lines, no mustache logic - you might just skip the separate .js and just embed the template string directly in the view. [That felt pretty convenient (and React-like) to me in my c25k project](https://github.com/lmorchard/c25k-web/blob/master/src/js/views/Workout.js#L9), keeping the small markup fragments close to the bindings & events declarations.

Also, if you're _not_ using mustache - and instead are using _all_ ampersand DOM bindings - [overriding `template`](https://github.com/mozilla/idea-town/blob/master/idea_town/frontend/static-src/app/views/base-view.js#L11) rather than `_template` in your view will skip the whole mustache rendering step and provide just the plain markup string.

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -7,83 +7,16 @@ export default `
...
-                  <td>{{modified_date}}</td>
-                </tr>
-                <tr>
-                  <td>Contribute</td>
-                  <td><a href="{{model.contribute_url}}">{{model.contribute_url}}<a></td>
-                </tr>
-              </table>
-            </section>
-          </div>
+        <div data-hook="details"></div>
```
lmorchard: Hmm, the more I look at this, the more I think all the stuff in the `details-view` could be merged into `experiment-page` rather than handled as a subview. But, I might be missing something...

lmorchard:
Nice! I think I'm done picking on this PR :)

PR #247, opened 2015-11-13
add settings ui to header

### LINKED ISSUE #217: Settings for Pre-Release Idea Town
@nchapman @lmorchard  This doesn't have to be permanent, but just to get the ball rolling let's implement this UI for the Mozlando alpha. This UI also deprecates Avatar support for now.

alpha settings 2x

meandavejustice:
fixes #217 

settings-menu-ui

### DIFF
--- a/idea_town/frontend/static-src/styles/modules/_avatar.scss
+++ b/idea_town/frontend/static-src/styles/modules/_avatar.scss
@@ -1,18 +1,3 @@
-#avatar-wrapper {
-  @include flex-container(default, center, center, nowrap, false);
-  border-radius: 50%;
-  cursor: pointer;
-  flex: 0 0 48px;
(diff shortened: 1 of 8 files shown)


idea_town/frontend/static-src/styles/modules/_avatar.scss
```diff
@@ -1,18 +0,0 @@
-#avatar-wrapper {
```
lmorchard, line 1: FWIW, I think we might still need this stuff for the contributors list in the experiment details page. (At least the img.avatar part)

meandavejustice, line 1: Added back `img.avatar` styles

lmorchard:
Works for me!

Issue #267, opened 2015-11-18
A few HTML errors in experiment-page template


pdehaan:
I tried running [/idea_town/frontend/static-src/app/templates/experiment-page.js](https://github.com/mozilla/idea-town/blob/4b60baf1f25c9f6e6fb0b462faa977f77248bdaa/idea_town/frontend/static-src/app/templates/experiment-page.js) through an HTML5 validator (https://html5.validator.nu/) and it came back with a few errors.

Notably, I think it is just:
1. unclosed `<a>`, and `<div>` tags
2. duplicate `id` attribute

I'll see if I can run the rest of the templates through the validator and see what else shakes out. Not sure if there is a good Node based HTML5 linter that supports HTML fragments, ES6 strings, and Mustache templating, but we could try adding a HTML linter to the build chain though.

pdehaan:
Also seeing a couple warnings in [/idea_town/frontend/static-src/app/templates/landing-page.js](https://github.com/mozilla/idea-town/blob/4b60baf1f25c9f6e6fb0b462faa977f77248bdaa/idea_town/frontend/static-src/app/templates/landing-page.js)

> Error: The element `button` must not appear as a descendant of the `a` element.
> From line 27, column 55; to line 27, column 85
> 
> ``` html
> xt=/home"><button class="button primary">Sign m
> ```

---

**UPDATE:** Filed separately as https://github.com/mozilla/idea-town/issues/284

pdehaan:
**UPDATE:** duplicate `id` attribute issue filed as https://github.com/mozilla/idea-town/issues/285

pdehaan:
Closing. The unclosed `<a>` and `<div>` tags were fixed as of #277 PR, and the other 2 issues mentioned above were extracted into separate bugs.

PR #276, opened 2015-11-18
feat(inactive-page): style inactive page

lmorchard:
Let's try this PR again

### DIFF
--- a/idea_town/frontend/static-src/app/templates/account-inactive-page.js
+++ b/idea_town/frontend/static-src/app/templates/account-inactive-page.js
@@ -1,11 +1,16 @@
 export default `
-  <div class="page">
-    <section class="page main-content" data-hook="account-inactive-page">
-      <header data-hook="main-header"></header>
-      <h1>Account Inactive</h1>
-      <h2>We're not quite ready for you yet!<br>
-          Please come back in a few weeks.</h2>
+  <div class="page" id="full-page-notice">
+    <section class="main-content" data-hook="account-inactive-page">
+      <div class="vertical-centering-container">
+        <div id="beta-notice-modal" class="delayed-fade-in">
+          <h1>Thanks for signing up!</h1>
+          <h2>We'll let you know as soon as Idea Town is ready.</h2>
+          <div class="modal-controls">
+            <a href="/"><button class="button primary">Home</button></a>
+          </div>
+        </div>
+        <div class="copter fly-up-fade-in"></div>
+      </div>
     </section>
-    <div data-hook="main-footer"></div>
   </div>
 `;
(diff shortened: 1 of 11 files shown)


johngruen:
@pdehaan this is what i'm talking about...I'll merge once it passes!

thanks @lmorchard!

idea_town/frontend/static-src/app/templates/account-inactive-page.js
```diff
@@ -1,11 +1,16 @@
...
-      <header data-hook="main-header"></header>
-      <h1>Account Inactive</h1>
-      <h2>We're not quite ready for you yet!<br>
-          Please come back in a few weeks.</h2>
+  <div class="page" id="full-page-notice">
+    <section class="main-content" data-hook="account-inactive-page">
+      <div class="vertical-centering-container">
+        <div id="beta-notice-modal" class="delayed-fade-in">
+          <h1>Thanks for signing up!</h1>
+          <h2>We'll let you know as soon as Idea Town is ready.</h2>
```
pdehaan, line 7: What does _"We'll let you know"_ mean? Are we collecting email addresses somewhere?
How will we notify these random people?

johngruen:
@pdehaan yerp

idea_town/frontend/static-src/app/templates/account-inactive-page.js
```diff
@@ -1,11 +1,16 @@
...
+          <div class="modal-controls">
+            <a href="/"><button class="button primary">Home</button></a>
```
pdehaan, line 9: I spotted this in #267, but https://html5.validator.nu/ doesn't seem to like nesting `<button>` tags in an `<a>` element for unexplained reasons.

> **Error:** The element `button` must not appear as a descendant of the `a` element.
> From line 15, column 25; to line 15, column 55
> 
> ``` html
> href="/"><button class="button primary">Home</
> ```

I tried shuffling some elements and classes in the dev tools locally, but all experiments looked pretty tragic, so we may need to adjust some CSS styling as well (in a future PR).

idea_town/frontend/static-src/app/templates/account-inactive-page.js
lmorchard, line 7: We collect the email address passed along with the FxA sign-in, as part of an invitation mode system introduced here:

https://github.com/mozilla/idea-town/issues/128#issuecomment-148826004
https://github.com/mozilla/idea-town/commit/f1d09e9cb35d3c1e5b0df39aa31249d13bd61074

We don't actually have the machinery to notify these users via email yet, though:

https://github.com/mozilla/idea-town/issues/129

johngruen:
@pdehaan i'll fix these nits in my 404 page pr which is incoming soon

johngruen:
@pdehaan which is to say, i'll make that string explicitly mention the email

pdehaan:
@johngruen

PR #288, opened 2015-11-19
Ensure static assets are collected on server startup

lmorchard:
- Also, include the current git commit hash at static/revision.txt


lmorchard:
This fixes issues with static assets when `DEBUG = False` in prod

Issue #281, opened 2015-11-18
Hide empty elements on experiment details pages


pdehaan:
via https://github.com/mozilla/idea-town/issues/258#issuecomment-157508739

---

> @lmorchard: _"Hmm, though we should probably hide the elements if they're optional and not filled in (like Contribute, also)"_

lmorchard:
https://github.com/meandavejustice/idea-town/commit/26e3e207f2cac4351a446b2bdab78b4724800c96

Issue #258, opened 2015-11-17
changelog link looks mis-aligned on experiment details page


pdehaan:
### Steps to reproduce:
1. Sign in and go to http://idea-town-dev.elasticbeanstalk.com/experiments/test
2. Look at the "Details" -> "Version" section.

![idea_town](https://cloud.githubusercontent.com/assets/557895/11224307/2e463928-8d29-11e5-8efc-28fe395d3b0f.png)
### Actual results:

The "changelog" link doesn't align with the last updated text.
Looking through the source, I'm guessing it's because the [`<span data-hook="version"></span>`](https://github.com/mozilla/idea-town/blob/6526f8f58b52596c871df8aa644c02e326350508/idea_town/frontend/static-src/app/templates/experiment-page.js#L38-L39) isn't getting set (or has no value in the database). That, combined with the `&nbsp;<a data-hook="changelog-url">changelog</a>` causes the extra space and alignment issue.

I guess this is a non-issue as long as the version number for an add-on is always set.

lmorchard:
Hmm, though we should probably hide the elements if they're optional and not filled in (like Contribute, also)

pdehaan:
Should "Version" and "Last Update" be optional?

pdehaan:
> we should probably hide the elements if they're optional and not filled in

Filed as https://github.com/mozilla/idea-town/issues/281

lmorchard:
https://github.com/meandavejustice/idea-town/commit/26e3e207f2cac4351a446b2bdab78b4724800c96

PR #298, opened 2015-11-19
details version/changelog/last update

### LINKED ISSUE #261: Details page doesn't display version number or last updated date
http://idea-town-dev.elasticbeanstalk.com/experiments/test-title

idea_town
**Figure 1:** Note that:
1. version number is missing, 
2. changelog text is linked (but not specified in admin),
3. an description is being set in the **Last Update** text field.

---

change_experiment___django_site_admin
**Figure 2:** Here's a snippet from my admin entry. Note that version is specified (but "changelog url" isnt)

meandavejustice:
fixes #261, #258, #281

### DIFF
--- a/idea_town/frontend/static-src/app/views/experiment-page.js
+++ b/idea_town/frontend/static-src/app/views/experiment-page.js
@@ -42,13 +42,32 @@ export default PageView.extend({
       hook: 'feedback'
     }],
 
-    'model.modified_date': {
-      type: (el, value) => new Date(value),
+    'model.modified': {
+      type: (el, value) => {
+        const d = new Date(value);
+
+        // check for invalid date
+        if (isNaN(d)) {
+          el.style.display = 'none';
+          return;
+        }
+
+        let formatted = '';
+
+        // safari is the new IE :(
+        try {
+          formatted = d.toLocaleDateString();
+        } catch (e) {
+          formatted = `${d.getMonth() + 1} / ${d.getDate()} / ${d.getFullYear()}`;
+        }
+
+        el.innerHTML = formatted;
+      },
       hook: 'modified-date'
     },
 
-    'model.description': {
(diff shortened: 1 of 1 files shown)


idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -42,13 +42,16 @@ export default PageView.extend({
       hook: 'feedback'
     }],
 
-    'model.modified_date': {
-      type: (el, value) => new Date(value),
+    'model.modified': {
+      type: (el, value) => {
+        const d = new Date(value);
+        el.innerHTML = `${d.getMonth() + 1} / ${d.getDate()} / ${d.getFullYear()}`;
```
pdehaan: Not sure if we want to use [`toLocaleDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) or something similar here to avoid a potential "MM/DD/YYYY" vs "DD/MM/YYYY" fist fight.

pdehaan: Actually, on that note, not sure if you want to verify `d` before trying to use `.getMonth()` and gang.
Otherwise we may get something back like "NaN/NaN/NaN" or "Invalid Date" if you pass in an empty string or unexpected value.

meandavejustice: toLocaleDateString is not supported in safari :( I can look around for a little shim for it

pdehaan:
Not sure if this fixed #281 (or maybe I was expecting something different).

I'm currently seeing this on http://idea-town-dev.elasticbeanstalk.com/experiments/example-experiment

idea_town

I was expecting the **Version** and **Contribute** labels to be hidden since they don't currently have values.

PR #299, opened 2015-11-19
open settings links in new tab

### LINKED ISSUE #293: Open Settings links in new tabs?
idea_town

Currently if you select the ["Idea Town Wiki"](https://wiki.mozilla.org/Idea-Town) or ["File an Issue"](https://github.com/mozilla/idea-town/issues/new) links, it replaces the current tab. We may want to see if we can add `target="_blank"` or some Ampersand magic to get those two links to open in a new tab since it's likely that we'll want to keep the user on the site while they file a bug or whatever.

meandavejustice:
fixes #293


johngruen:
Boom

PR #312, opened 2015-11-23
Track per-client experiment installations

### LINKED ISSUE #195: persist enabled addons per user / experiment detail doesn't detect enabled correctly
experiment detail doesn't detect enabled correctly when visiting the page directly(not navigating from `/home`)
We need to track the enabled state through the User Installation Model on the server and verify with addon after initial render.

cc/ @lmorchard for discussion on User Install model changes.

lmorchard:
- (addon, frontend) Sync installation state of experiments to server on
  initial page load
- (addon) Generate random UUID for Idea Town addon
- (addon) Refactor tracking of available & installed experiments
- (server) Rework UserInstallation model, admin, and tests
- (server) UserInstallation PUT / DELETE API
- (addon, frontend) Misc refactorings & promisifications & cleanup
- (addon-api) Use `Accept: application/json` header to get JSON data
  from metrics API and quiet XML errors in console.

Closes #195

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -5,8 +5,9 @@
  */
 
 let app;
+
 const store = require('sdk/simple-storage').storage;
-const {Cu} = require('chrome');
+const {Cc, Ci, Cu} = require('chrome');
 const self = require('sdk/self');
 const tabs = require('sdk/tabs');
 const {PageMod} = require('sdk/page-mod');
@@ -14,7 +15,10 @@ const {ActionButton} = require('sdk/ui/button/action');
 const request = require('sdk/request').Request;
 const AddonManager = Cu.import('resource://gre/modules/AddonManager.jsm').AddonManager;
 const prefs = require('sdk/simple-prefs').prefs;
+const URL = require('sdk/url').URL;
 const Router = require('./route');
+const cookieManager2 = Cc['@mozilla.org/cookiemanager;1']
+                       .getService(Ci.nsICookieManager2);
 
 const mod = new PageMod({
   include: prefs.ALLOWED_ORIGINS.split(','),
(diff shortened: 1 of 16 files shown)


lmorchard:
Okay, so... I may have gotten a little carried away on this. But, hopefully the refactoring simplifies frontend client start up a bit along with adding the experiment installation state management

addon/index.js
```diff
@@ -24,26 +28,19 @@ const mod = new PageMod({
...
 require('sdk/system/unload').when(function(reason) {
   if (reason === 'uninstall') {
     app.send('addon-self:uninstalled');
-    if (store.experiments) {
-      for (let ex in store.experiments) { // eslint-disable-line prefer-const
-        if (store.experiments.hasOwnProperty(ex)) {
-          uninstall({addon_id: ex});
-        }
+    if (store.installedAddons) {
+      for (let id of store.installedAddons) { // eslint-disable-line prefer-const
```
pdehaan, line 39: Should we just do `for (const id of store.installedAddons) {` (since `id` never changes) and then we can delete that ESLint rule override?

addon/index.js
```diff
@@ -172,9 +171,61 @@ function formatInstallData(install, addon) {
...
+    request({
+      url: opts.url,
+      headers: Object.assign(headers, opts.headers || {}),
+      contentType: 'application/json',
+      onComplete: res => (res.status < 400) ? resolve(res) : reject(res)
+    })[opts.method || 'get']();
+  });
+}
+
+// source: http://jsfiddle.net/briguy37/2mvfd/
```
pdehaan, line 216: Easier to use something like https://www.npmjs.com/package/node-uuid ?

lmorchard, line 216: Figured the full node-uuid module would be overkill, since we only ever need a vaguely random UUID. This is actually what the README links to from "Want something smaller? Check this crazy stuff out!" :)

addon/index.js
lmorchard, line 39: I want to say I used `let` because `const id` didn't work, but that might be a Chrome / V8 / node bug.

meandavejustice, line 39: id is updated in the loop each time so is not valid `const`

lmorchard:
Rebased against latest master

PR #319, opened 2015-11-24
199: Track progress for install/uninstalls

### LINKED ISSUE #199: Track progress for installations on experiment detail view.
We need to show another state to show the user that the experiment is installing.

johngruen:
~~@meandavejustice this shows the UX for enabling/disabling experiments. the uninstall and install methods both use arbitrary setTimeouts to simulate the actual latency of a download install or uninstall. When we implement here, it's probably a good idea to keep a minimum time in place, so that the transition doesn't happen too fast for users to notice.~~

Adds `disabling/enabling` state animations to experiment detail page.

fixes #199 
fixes and issue with the addon loading.


idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -22,7 +22,7 @@ export default PageView.extend({
 
   events: {
     'click [data-hook=install]': 'install',
-    'click [data-hook=uninstall]': 'uninstall',
+    'click [data-hook=uninstall]': 'renderUninstallSurvey',
```
pdehaan: Since you added `id` attributes above, you may be able to convert these 3 things to something like:

``` js
'click #uninstall-button': 'renderUninstallSurvey'
```

meandavejustice: I feel like it's pretty good practive to keep style hooks and js hooks separate.

idea_town/frontend/static-src/app/views/feedback-view.js
```diff
@@ -56,9 +62,9 @@ export default BaseView.extend({
     // Hide the feedback form right away & let the submission happen in the
     // background - user probably doesn't care about progress or confirmation
     this.animateRemove();
+    this.parentUninstall(this.parentUpdateAddon, this.parentModel);
```
lmorchard: This survey view is also feedback view is also used for the surveys while the experiment is installed & in progress. It should _not_ be uninstalling addons.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -184,7 +203,10 @@ export default PageView.extend({
         { value: 'dislike', title: 'I don\'t like this feature.' },
         { value: 'notuseful', title: 'This isn\'t useful for me.' },
         { value: 'other', title: 'Something else.' }
-      ]
+      ],
+      uninstall: this.uninstall,
```
lmorchard: Making the feedback view also uninstall the addon is giving it too many responsibilities. It's also reused for the other feedback case. If we definitely want to hold off the uninstall until an action is taken in the feedback form, it would be better to just add a generic callback hook when done. Then, perform the uninstall in a callback here

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -6,9 +6,9 @@ export default `
         <div class="details-header">
           <h1 data-hook="title"></h1>
           <div class="idea-controls">
-            <button data-hook="feedback" class="button primary">Give Feedback</button>
-            <button data-hook="uninstall" class="button neutral">Disable <span data-hook="title"></span></button>
-            <button data-hook="install" class="button primary">Enable <span data-hook="title"></span></button>
+            <button data-hook="feedback" id="feedback-button" class="button primary">Give Feedback</button>
+            <button data-hook="uninstall" id="uninstall-button" class="button neutral"><span class="state-change-inner"></span><span class="transition-text">Disabling...</span><span class="default-text">Disable <span data-hook="title"></span><span></button>
```
pdehaan: This... doesn't look kosher.

There is something funny around the end bit of `</span><span></button>`.
Should be `</span></span></button>` since we're in a nested `span`.

lmorchard:
Doh. Ended up commenting on the wrong PR, going to copy things over here.

addon/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "license": "MPL-2.0",
   "dependencies": {
-    "idea-town": "^0.3.0"
+    "idea-town": "^0.3.0",
+    "xtend": "^4.0.1"
```
lmorchard, line 22: FWIW, might not need this module - [we the built-in `Object.assign`, which seems to do the same thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -156,24 +156,43 @@ export default PageView.extend({
...
   },
 
-  uninstall(evt) {
+  uninstall(cb, model) {
+    const uninstallButton = document.getElementById('uninstall-button');
+    const feedbackButton = document.getElementById('feedback-button');
+    const width = uninstallButton.offsetWidth;
+    uninstallButton.style.width = width + 'px';
+    uninstallButton.classList.add('state-change');
+    feedbackButton.style.display = 'none';
```
lmorchard, line 183: Hmm, I don't think we need to mess with feedback button styles here. [We have these hooked up with data binding to the `enabled` prop of the experiment, no?](https://github.com/mozilla/idea-town/blob/master/idea_town/frontend/static-src/app/views/experiment-page.js#L36) When `enabled` gets set to false, the feedback button should toggle to invisible. Also, I'm seeing an issue where the "give feedback" button doesn't actually go away on experiment disable - not sure if that's related to this.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -184,7 +203,10 @@ export default PageView.extend({
         { value: 'dislike', title: 'I don\'t like this feature.' },
         { value: 'notuseful', title: 'This isn\'t useful for me.' },
         { value: 'other', title: 'Something else.' }
-      ]
+      ],
+      cb: () => {
```
lmorchard, line 207: Probably worth naming this option something more explicit - e.g. `onSubmit`

idea_town/frontend/static-src/app/views/feedback-view.js
```diff
@@ -56,9 +60,9 @@ export default BaseView.extend({
     // Hide the feedback form right away & let the submission happen in the
     // background - user probably doesn't care about progress or confirmation
     this.animateRemove();
+    this.parentCb();
```
lmorchard, line 63: Need to check if `cb` was actually passed as an option - it's not in the "Give feedback" button case, and so an error gets thrown and submission never happens.

lmorchard:
Should get these commits squashed down to one before merging the PR, to clean up the merge commit noise generated during work in progress

addon/package.json
meandavejustice, line 22: Thanks!

meandavejustice:
closed in favor of https://github.com/mozilla/idea-town/pull/324

PR #191, opened 2015-10-22
happy path WIP DO NOT MERGE.

meandavejustice:
opening for a quick review from @lmorchard 

Mostly changes to addon and client.

Addon:
we now check against a list of availableExperiments from the server and listen for install events from client.

Client:
We check installed experiments by pinging the user's addon. Then we set `enabled` locally.

Model:
added `addon_id` property.

PR #323, opened 2015-12-01
fixes #191: Additional fixes for pr 319

meandavejustice:
Hooks up @johngruen's changes to the addon. Addresses comments from @lmorchard and @pdehaan


idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -171,24 +166,29 @@ export default PageView.extend({
...
     uninstallButton.classList.add('state-change');
     feedbackButton.style.display = 'none';
-    setTimeout(() => {
+
+    cb(false, model);
+
+    app.on('webChannel:addon-uninstall:uninstall-ended', () => {
+      model.enabled = !model.enabled;
       uninstallButton.classList.remove('state-change');
       feedbackButton.style.display = 'initial';
```
lmorchard: Hmm, I don't think we need to mess with button styles & such here. [We have these hooked up with data binding to the `enabled` prop of the experiment, no?](https://github.com/mozilla/idea-town/blob/master/idea_town/frontend/static-src/app/views/experiment-page.js#L36) Also, I'm seeing an issue where the "give feedback" button doesn't actually go away on experiment disable - not sure if that's related to this.

addon/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "license": "MPL-2.0",
   "dependencies": {
-    "idea-town": "^0.3.0"
+    "idea-town": "^0.3.0",
+    "xtend": "^4.0.1"
```
lmorchard: FWIW, might not need this module - [we the built-in `Object.assign`, which seems to do the same thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

idea_town/frontend/static-src/app/views/feedback-view.js
```diff
@@ -62,7 +60,7 @@ export default BaseView.extend({
     // Hide the feedback form right away & let the submission happen in the
     // background - user probably doesn't care about progress or confirmation
     this.animateRemove();
-    this.parentUninstall(this.parentUpdateAddon, this.parentModel);
+    this.parentCb();
```
lmorchard: Need to check if `cb` was actually passed as an option - it's not in the "Give feedback" button case, and so an error gets thrown and submission never happens.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -204,9 +204,9 @@ export default PageView.extend({
         { value: 'notuseful', title: 'This isn\'t useful for me.' },
         { value: 'other', title: 'Something else.' }
       ],
-      uninstall: this.uninstall,
-      updateAddon: this.updateAddon,
-      model: this.model
+      cb: () => {
```
lmorchard: Probably worth naming this something more verbose - e.g. `onSubmit`

PR #334, opened 2015-12-11
Collapse all addon prefs into a single menu selector

lmorchard:
- Add canned configs for local/dev/stage/prod server environments
- Reconfigure & reinitialize addon on pref settings change
- Merge router.js into index.js, because it seemed easier to share the
  same module globals that change on pref update
- Bump addon version to 0.3.1 and include latest build
- Misc code shufflings because I can't help it


lmorchard:
@meandavejustice this might be interesting to you

meandavejustice:
This looks great, thanks @lmorchard !

PR #336, opened 2015-12-14
fixes 320: detect devtools theme and set appropriate icon

### LINKED ISSUE #320: Icon is hard to see with dark theme
The default theme for Firefox Developer Edition is Dark, the idea town icon is hard to see against this theme.

icon-demo

meandavejustice:
fixes #320 

Here we detect the 'devtools.theme' to see if we need to set the `inverted` icon.


lmorchard:
Oh hey, this is hawt. :shipit:

PR #337, opened 2015-12-14
add COC and CONTRIBUTING docs

meandavejustice:
coc from http://contributor-covenant.org/

Thanks for a reminder from @6a68


lmorchard:
:two_hearts: Merging this before some rando decides to spawn a 1000 reply comment thread here :two_hearts:

PR #339, opened 2015-12-16
remove sdk from addon-api

### LINKED ISSUE #311: addon-api should work without SDK
The addon-lib looks great, but the search addon doesn't use the SDK, so we'll need to make a few changes. (More detail to come)

meandavejustice:
fixes #311 

Hey @6a68, is this what you had in mind? Also thanks for the gecko walkthrough last week, it was very helpful when digging though dxr for this.

If this gets the :+1: I'll publish to npm.


meandavejustice:
_cleaned up for merge_

jaredhirsch:
@meandavejustice I actually think #355 might be the simplest way for idea town to support metrics for individual experiments: if idea town sets CORS headers, submitting metrics becomes as easy as firing an xhr, and the addon-api library wouldn't be needed.

This code relies on `require` and `exports` throughout, and these aren't available without the SDK. The Gecko-flavored alternative, [documented here on mdn](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Using), loads JS by passing a funny URL to `Cu.import`. The URLs are of the form `chrome://myfunnypath/foo.js`, where Gecko reads a `chrome.manifest` file to translate `myfunnypath` to a file path relative to the addon. As an example, [here's](https://github.com/mozilla/universal-search-addon/blob/master/src/chrome.manifest) the chrome.manifest for search. I don't think SDK addons ever touch the `chrome.manifest` file, so I don't see how you could use the same library for SDK and non-SDK cases. Or, if you did it, it would probably be way more trouble than it's worth :-P

meandavejustice:
Thanks @6a68 I'm going to close this pr. I've opened another pr at #356 with the documentation updates and experiment.js cleanup from this branch.

PR #340, opened 2015-12-16
Control the order of experiments

### LINKED ISSUE #333: I want to customize the order of experiments in the list view
Right not, the most recently updated experiment goes to the end. Let's not do this.

meandavejustice:
SHOULD BE MERGED AFTER #346 

fixes #333 

You can now set the `order` field in django admin to reorder experiments on the home page.

### DIFF
--- a/idea_town/experiments/models.py
+++ b/idea_town/experiments/models.py
@@ -12,6 +12,9 @@ experimentdetail_image_upload_to = HashedUploadTo('image')
 
 class Experiment(models.Model):
 
+    class Meta:
+        ordering = ['order']
+
     title = models.CharField(max_length=128)
     slug = models.SlugField(max_length=128, unique=True, db_index=True)
     thumbnail = models.ImageField(upload_to=experiment_thumbnail_upload_to)
@@ -20,6 +23,7 @@ class Experiment(models.Model):
                                default_markup_type='plain')
     xpi_url = models.URLField()
     version = models.CharField(blank=True, max_length=128)
+    order = models.IntegerField(default=0)
     changelog_url = models.URLField(blank=True)
     contribute_url = models.URLField(blank=True)
     privacy_notice_url = models.URLField(blank=True)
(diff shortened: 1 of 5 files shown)


idea_town/experiments/models.py
```diff
@@ -18,6 +18,8 @@ class Experiment(models.Model):
     description = models.TextField()
     measurements = MarkupField(blank=True, default='',
                                default_markup_type='plain')
+
+    order = models.IntegerField(default=0)
```
lmorchard: Might also be interesting to add [a model Meta class](https://docs.djangoproject.com/en/1.9/topics/db/models/#meta-options) here [with an `ordering` default for this field](https://docs.djangoproject.com/en/1.9/ref/models/options/#ordering).

meandavejustice: Oh nice, thanks

idea_town/experiments/migrations/0007_experiment_order.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
jaredhirsch: need to bump the migration number

idea_town/experiments/migrations/0007_experiment_order.py
```diff
@@ -0,0 +1,19 @@
...
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('experiments', '0006_auto_20151119_1849'),
```
jaredhirsch: should bump this value, too

idea_town/experiments/migrations/0007_experiment_order.py
meandavejustice: will fix after #346 is merged

lmorchard:
Can we get this disentangled from #346? I think this PR is simple enough, and [I'm not sure we need to maintain a counter field](https://github.com/mozilla/idea-town/pull/346#issuecomment-168753424)

meandavejustice:
Sure thing @lmorchard

meandavejustice:
@lmorchard Got rid of the user installation count stuff, this should be mergable now :)

Issue #130, opened 2015-10-06
Add localization


nchapman:
Tasks so far:
- [x] Implement L10N in frontend app via l20n.js
- [x] Chasing down any remaining strings lurking, e.g. in page titles & [CSS](https://github.com/mozilla/idea-town/issues/380) etc
  - See #382 & #383 
- [x] Localizing database content on the server side (#392)
- [ ] Add locales in addition to en-US (#381)
- [ ] Integrating with Pontoon (or some interface acceptable to localizers) (#414)

lmorchard:
Relevant to these interests:
https://developer.mozilla.org/en-US/docs/Implementing_Pontoon_Mozilla
https://github.com/fabi1cazenave/webL10n
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localizing_with_Verbatim

pdehaan:
http://l20n.org/ also looks interesting (and built by us).

lmorchard:
Gotcha from #342: Look for all places where we're doing plain concatenation of strings like `"Idea Town - " + pageTitle` and turn them into localizable templates. (i.e. because word order and such don't work the same way in all locales, especially for RTL languages)

lmorchard:
Got a start on this in 8b6034b33a373d8b0e2d4d21f09a62621a755b0c

But, some things are left, such as:
- [ ] Integrating with Pontoon (or some interface acceptable to localizers)
- [ ] Localizing database content on the server side (#392)
- [x] Chasing down any remaining strings lurking, e.g. in page titles & [CSS](https://github.com/mozilla/idea-town/issues/380) etc
  - See #382 & #383

lmorchard:
Closing this. Remaining tasks have issues.

PR #342, opened 2015-12-22
update Titles with page change

### LINKED ISSUE #286: Page title never changes as you navigate site
Minor thing, but makes it difficult to navigate back through history using the back button:

idea_town_and_new_issue_ _mozilla_idea-town_and_odd_experiment_layout_with_5_items_ _issue__282_ _mozilla_idea-town

meandavejustice:
fixes #286

### DIFF
--- a/idea_town/frontend/static-src/app/lib/page-manager.js
+++ b/idea_town/frontend/static-src/app/lib/page-manager.js
@@ -27,7 +27,13 @@ export default class PageManager {
       'error': ErrorPage
     };
 
-    this._viewSwitcher = new AmpersandViewSwitcher(opts.pageContainer);
+    this._viewSwitcher = new AmpersandViewSwitcher(opts.pageContainer, {
+      show: (view) => {
+        document.title = 'Idea Town - ' + view.pageTitle || 'Idea Town';
+        document.body.scrollTop = 0;
+      }
+    });
+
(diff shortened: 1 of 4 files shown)


idea_town/frontend/static-src/app/lib/page-manager.js
```diff
@@ -27,7 +27,13 @@ export default class PageManager {
       'error': ErrorPage
     };
 
-    this._viewSwitcher = new AmpersandViewSwitcher(opts.pageContainer);
+    this._viewSwitcher = new AmpersandViewSwitcher(opts.pageContainer, {
+      show: (view) => {
+        document.title = 'Idea Town - ' + view.pageTitle || 'Idea Town';
```
lmorchard, line 32: Whenever we get into localization, this could be a gotcha spot. Simple concatenations don't generally work across languages. Instead you might want to make this a template string that gets filled with `view.pageTitle` - e.g. something like <code>`Idea Town - ${view.pageTitle}`</code> and we can expose that template string itself for localization.

Of course, since we haven't even started on l10n, this is probably putting the cart before the horse :)

lmorchard:
LGTM except for the string concatenation nit

lmorchard:
On second thought, I'm just going to merge this. I'll mention #130 here, so this issue shows up as a link over there and remember it's one of the things that needs doing for L10N :)

PR #343, opened 2015-12-22
cache npm and python deps in travis

### LINKED ISSUE #207: Cache node & python dependencies for faster Travis CI runs
Most of the time in Travis CI runs is spent on installing node & python dependencies because nothing is preserved between runs. But, those dependencies barely ever change. So, we could save a huge amount of time by caching these dependencies in some form - say in an Amazon S3 bucket.

[Travis CI docs have advice on this](http://docs.travis-ci.com/user/caching/)

meandavejustice:
fixes #207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ services:
 install:
   - pip install tox coveralls
   - nvm install 0.12
+cache:
+  - pip
+  directories:
+    - node_modules
 env:
   global:
     - DOCKER_EMAIL=me@lmorchard.com


lmorchard:
I don't think this is going to work, because [Travis CI caching](https://docs.travis-ci.com/user/caching/) requires either a private repo (which we don't use) or [container-based infra](https://docs.travis-ci.com/user/workers/container-based-infrastructure/) (which we currently disable [with `sudo: true` in our `.travis.yml`](https://github.com/meandavejustice/idea-town/blob/207-cache-deps-travis/.travis.yml#L1)). Have you seen this working / offering speedups on your own fork?

If not, we probably have to do something more complex & bespoke - e.g. upload tarballs of dependencies to an S3 bucket on changes to requirements.txt & project.json, download & unpack those tarballs on testing start. 

Alternatively, we could see about removing the need for that `sudo: required`. But, I want to say the current Tox-based testing plus Docker auto-deploy setup prevents that. It would be really cool if I'm wrong

meandavejustice:
@lmorchard Just gunna close this for now

PR #344, opened 2015-12-22
add support for privacy notice urls for experiments

meandavejustice:
fixes #338

PR #346, opened 2015-12-23
Implement install counter

### LINKED ISSUE #153: Implement install counter on experiment detail page
There's a counter of installs on the experiment detail page. Let's implement it! Of course, we have to have this data collected before we can display it. So, #38 is a blocker for now.

meandavejustice:
fixes #153
install_count_uninstalled
install_count_installed

### DIFF
--- a/idea_town/experiments/models.py
+++ b/idea_town/experiments/models.py
@@ -1,5 +1,6 @@
 from django.db import models
 from django.contrib.auth.models import User
+from django.utils.functional import cached_property
 
 from markupfield.fields import MarkupField
 
@@ -34,6 +35,11 @@ class Experiment(models.Model):
 
     created = models.DateTimeField(auto_now_add=True)
     modified = models.DateTimeField(auto_now=True)
+
+    @cached_property
+    def installation_count(self):
+        return UserInstallation.objects.distinct('user').filter(
+            experiment=self).count()
 
     def __str__(self):
         return self.title
--- a/idea_town/experiments/serializers.py
+++ b/idea_town/experiments/serializers.py
@@ -3,7 +3,8 @@ from django.core.urlresolvers import reverse
 
 from ..users.models import UserProfile
 from ..users.serializers import UserProfileSerializer
-from .models import (Experiment, ExperimentDetail, UserFeedback, UserInstallation)
+from .models import (Experiment, ExperimentDetail, UserFeedback,
+                     UserInstallation)
 from ..utils import MarkupField
 
 
@@ -34,7 +35,7 @@ class ExperimentSerializer(serializers.HyperlinkedModelSerializer):
                   'version', 'changelog_url', 'contribute_url',
                   'privacy_notice_url', 'measurements',
                   'xpi_url', 'addon_id', 'details', 'contributors',
-                  'installations_url',
+                  'installations_url', 'installation_count',
                   'created', 'modified', 'order',)
 
     def get_contributors(self, obj):
--- a/idea_town/experiments/tests.py
+++ b/idea_town/experiments/tests.py
@@ -83,6 +83,9 @@ class ExperimentViewTests(BaseTestCase):
                         "xpi_url": "",
                         "details": [],
                         "contributors": [],
+                        "installation_count": UserInstallation.objects
+                        .distinct('user').filter(experiment=experiment)
+                        .count(),
                         "installations_url":
                             "http://testserver/api/experiments/%s/installations/" %
                             experiment.pk,
--- a/idea_town/frontend/static-src/app/views/experiment-page.js
+++ b/idea_town/frontend/static-src/app/views/experiment-page.js
@@ -71,6 +71,10 @@ export default PageView.extend({
       hook: 'version'
     },
 
+    'model.installation_count': {
+      hook: 'install-count'
+    },
+
     'model.contribute_url': [{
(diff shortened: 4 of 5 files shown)


idea_town/experiments/models.py
```diff
@@ -60,6 +61,18 @@ class UserInstallation(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     modified = models.DateTimeField(auto_now=True)
 
+    def save(self, *args, **kwargs):
+        if self.pk is None:
+            self.experiment.installation_count += 1
```
jaredhirsch: I'm not sure this approach will be thread-safe. Maybe use an F expression instead?

 `self.experiment.installation_count = F('installation_count') + 1`

Docs: https://docs.djangoproject.com/en/1.9/ref/models/expressions/#avoiding-race-conditions-using-f

meandavejustice: Thanks @6a68 , updated.

idea_town/experiments/models.py
```diff
@@ -60,6 +61,18 @@ class UserInstallation(models.Model):
...
 
+    def save(self, *args, **kwargs):
+        if self.pk is None:
+            self.experiment.installation_count = models.F('installation_count') + 1
+            self.experiment.save()
+        super(UserInstallation, self).save(*args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        if self.experiment.installation_count > 0:
+            self.experiment.installation_count = models.F('installation_count') - 1
```
jaredhirsch: Multithreading is fun: what's true when line 71 executes may no longer be true when line 72 executes. Without an atomic read-then-write statement, you can't be sure what the value is when line 72 executes.

To ensure you don't clobber any other writes, you should get rid of that positive value check on line 71.

Enforcing the counter > 0 thing is a challenge, and I'm not sure of the best way to handle it. To keep things moving, I'd say you should be fine to remove line 71, file a followup bug to handle the counter < 0 case, and see how @lmorchard wants to handle it when he's back.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -172,6 +176,7 @@ export default PageView.extend({
     app.on('webChannel:addon-install:install-ended', () => {
       this.model.enabled = !this.model.enabled;
       evt.target.classList.remove('state-change');
+      this.model.fetch();
```
jaredhirsch, line 193: It'd be good to manually bump the installation_count in the view before firing the xhr, so that the install button and the counter appear to update at the same time.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -186,6 +191,7 @@ export default PageView.extend({
     app.on('webChannel:addon-uninstall:uninstall-ended', () => {
       model.enabled = !model.enabled;
       uninstallButton.classList.remove('state-change');
+      model.fetch();
```
jaredhirsch, line 211: Ditto here, manually adjust the count downward before you fire the xhr

idea_town/experiments/migrations/0007_experiment_installation_count.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
jaredhirsch: Derp, need to rename this migration now that #344 has landed

idea_town/experiments/models.py
meandavejustice: issue created at #348

lmorchard:
Why maintain a new `installation_count` field when we can get this information with something like `UserInstallation.objects.filter(experiment=some_experiment).count()`? Can define that as a property on the `Experiment` model and cache if needed

lmorchard:
One caveat is that a plain count would catch multiple installations per user. If that's a problem, we can throw a [`distinct('user')`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.query.QuerySet.distinct) in there

jaredhirsch:
Sounds good, @lmorchard. Assigning you to manage which PRs to land, in which order :beers:

meandavejustice:
@lmorchard All fixed up 4 you

idea_town/experiments/serializers.py
```diff
@@ -51,6 +52,9 @@ def get_installations_url(self, obj):
                        args=(obj.pk,))
         return request.build_absolute_uri(path)
 
+    def get_installation_count(self, obj):
```
lmorchard: This would be better implemented as [a property](https://docs.python.org/3.5/library/functions.html#property) on the Experiment model itself, rather than just in the API serializer. When necessary, we can cache it there along with other model instance data.

idea_town/experiments/tests.py
```diff
@@ -82,6 +82,8 @@ def test_index(self):
                         "xpi_url": "",
                         "details": [],
                         "contributors": [],
+                        "installation_count": UserInstallation.objects.distinct('user')
```
lmorchard: To declutter this line a bit, you might want to assign this to a temp variable (e.g. `installation_count`) before the `self.assertJSONEqual`.

idea_town/experiments/serializers.py
meandavejustice: Oh I see, I misunderstood your previous comment

PR #354, opened 2015-12-31
Add tests for frontend.

meandavejustice:
Tests will be written in [tape](https://github.com/substack/tape)
They are run with locally with [testling-cli](https://github.com/substack/testling)
They are piped to [faucet](https://github.com/substack/faucet) for pretty output.

todo:
- [x] Get remaining tests to pass (waiting on a change in #371)
- [x] set up coverage stats (coverify maybe)
- [x] add documentation for testing
- [x] add option to run tests in browser(they are currently running headless only)
- [x] hook up to run on travis.

The list of tests are [here](https://github.com/mozilla/idea-town/issues/147#issuecomment-168237808)

cc/ @lmorchard @chuckharmston @6a68 for discussion.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,13 @@ the Mozilla [wiki][wiki] has instructions.
 
 We use the same process for community members and paid staff.
 
+#### Tests
+Tests should be written for new code contributed.
+* [server tests example](idea_town/experiments/tests.py)
+* [client tests example](idea_town/frontend/static-src/test/)
+
+[Further docs on testing](./TESTING.md).
+
 #### Branches
 
 Fork the project and create one branch per bug on your personal repo.
@@ -69,9 +76,8 @@ Reviewers will run through roughly the following checklist:
 - Are class, function, event, and variable names descriptive? Do they match the existing style?
 - Are boundary cases considered?
 - Are error cases handled?
-
-(Once we have some tests, the reviewer will check that all the tests are passing,
-and that any new tests cover the changed code paths / cases adequately.)
+- All tests and linting pass on travisci?
+- Are there tests for new/changed code?
 
 #### Merging
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,54 @@
+# Testing
+
+We currently have two different ways to write tests. Server tests are fairly
+standard [django tests](https://docs.djangoproject.com/en/1.9/topics/testing/).
+The client tests are written with [tape](https://www.npmjs.com/package/tape)
+to produce [tap](https://en.wikipedia.org/wiki/Test_Anything_Protocol) output,
+and run with [testling](https://www.npmjs.com/package/testling).
+We use [tape-after](https://www.npmjs.com/package/tape-around) to extend tape
+for `before` and `after` methods.
+
+Here are a couple of tips to help out when testing client-side code:
+
+* If you must modify `ampersand-app`, do it per test.
+* Always call `app.reset()` at the end of each test(in the `after` method).
+* If a new view is added, create a sanity test at the very least.
+``` js
+
+const test = around(tape)
+               .after(t => {
+                 app.reset();
+                 t.end();
+               });
+
+test('Experiment row view renders', t => {
+  t.plan(1);
+  const view = new View();
+  view.render();
+  t.ok(view.query('#my-view-id'));
+});
+
+```
+* Each test file should be started with the following snippet so we can have context for errors.
+
+``` js
+test(`Running Tests for ${__filename}`, a => a.end());
+```
+
+#### Running the tests
+
+* `npm test` will run all of the tests in node or headless browser.
+* `npm run test:general` will run general tests in node.
+* `npm run test:client` will run client tests in a headless browser.
+* `npm run test:browser` will start a web server for the tests to be run in a browser window.
+* `npm run test:browser:live` "     " + live reload.
+* `npm run test:cov` stats on code coverage.
+
+#### OSX
+If you are on OSX you will need to have a headless browser installed to run the client-side tests
+in the terminal.
+You can do this via https://github.com/graingert/slimerjs
+
+or
+
+`npm install -g slimerjs`
--- a/idea_town/frontend/static-src/app/views/base-view.js
+++ b/idea_town/frontend/static-src/app/views/base-view.js
@@ -34,6 +34,7 @@ export default AmpersandView.extend({
     // HACK: Slap the same data-l10n-args data on every localized node, because
     // the most common case is they all need the same model data.
     const nodes = this.queryAll('[data-l10n-id]');
+
     for (const node of nodes) {
       node.setAttribute('data-l10n-args', argsJSON);
     }
--- a/idea_town/frontend/static-src/app/views/base-view.js
+++ b/idea_town/frontend/static-src/app/views/base-view.js
@@ -33,7 +33,7 @@ export default AmpersandView.extend({
 
     // HACK: Slap the same data-l10n-args data on every localized node, because
     // the most common case is they all need the same model data.
-    const nodes = this.el.querySelectorAll('[data-l10n-id]');
+    const nodes = this.queryAll('[data-l10n-id]');
     for (const node of nodes) {
       node.setAttribute('data-l10n-args', argsJSON);
     }
--- a/idea_town/frontend/static-src/test/general.js
+++ b/idea_town/frontend/static-src/test/general.js
@@ -1,0 +1,21 @@
+var test = require('tape-catch');
+var path = require('path');
+var fs = require('fs');
+
+test('Running Tests for '+__filename, function(a) {
+  a.end();
+});
+
+test('Each image in images directory has a retina version', function(t) {
+  t.plan(1);
+    t.comment(path.resolve(__dirname, '../images'))
+  fs.readdir(path.resolve(__dirname, '../images'), function(err, files) {
+    if (err) t.fail('fs.readdir call failed, check your paths');
+
+    const retinaFiles = files.filter(function(f) {
+      return !!~f.indexOf('@2x');
+    });
+
+    t.same(retinaFiles.length, files.length / 2);
+  });
+});
--- a/idea_town/frontend/static-src/test/views/contributor-view.js
+++ b/idea_town/frontend/static-src/test/views/contributor-view.js
@@ -1,0 +1,11 @@
+import test from 'tape-catch';
+import View from '../../app/views/contributor-view';
+
+test(`Running Tests for ${__filename}`, a => a.end());
+
+test('Contributor view renders', t => {
+  t.plan(1);
+  const view = new View();
+  view.render();
+  t.ok(view.query('.contributor'));
+});
--- a/idea_town/frontend/static-src/test/browser.js
+++ b/idea_town/frontend/static-src/test/browser.js
@@ -1,0 +1,10 @@
+if (typeof(window) === 'object') {
+  var tapeDom = require('tape-dom');
+  tapeDom.installCSS();
+  tapeDom.stream(require('tape-catch'));
+}
+
+require('./collections/*js', {mode: 'expand'});
+require('./lib/*.js', {mode: 'expand'});
+require('./models/*.js', {mode: 'expand'});
+require('./views/*.js', {mode: 'expand'});
--- a/idea_town/frontend/static-src/test/collections/experiments.js
+++ b/idea_town/frontend/static-src/test/collections/experiments.js
@@ -1,0 +1,16 @@
+import test from 'tape-catch';
+import app from 'ampersand-app';
+import Model from '../../app/models/experiment';
+import Experiments from '../../app/collections/experiments';
+
+test(`Running Tests for ${__filename}`, a => a.end());
+
+test('addon-uninstalled event unsets enabled for each experiment', t => {
+  t.plan(1);
+  const experiments = new Experiments([{title: 'test experiment 1', enabled: true},
+                                       {title: 'test experiment 2', enabled: true}]);
+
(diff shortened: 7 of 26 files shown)


lmorchard:
This is kind of a general process comment / question, but: Would it be better to keep work-in-progress things like this out of the PR queue? Instead we can just work from issues & branches, invite look-sees & discussion there. 

FWIW, this isn't unique to this project - it's come up as a small debate on pretty much every project I've been on :) 

I cruise the PR queue for things I can jump on and shepherd into master branch. Unmergeable WIP PRs like this just stick there as noise. Occasionally I've accidentally merged WIP PRs and some of mine have been merged too.

meandavejustice:
@lmorchard I've never really liked WIP prs either, just gotten in the habit from other teams i've worked on. I'm :+1: for only tracking via branches/issues from now on. Would you be okay with starting _after_ this one though? I should have this one finished up today.

lmorchard:
Per the status meeting, you can ignore my grumblings for sure on this PR - and possibly on future ones too :)

idea_town/frontend/static-src/test/general.js
```diff
@@ -0,0 +1,21 @@
+var test = require('tape');
+var path = require('path');
+var fs = require('fs');
+
+test(`🖳  Running Tests for ${__filename} ␄`, function(a) {
```
lmorchard: Hmm, node v0.12 on Travis seems to dislike some characters in this line.

Node on my machine dislikes them too (v0.12.7)

Oh, actually I think it might be an issue with the ES6 template string... 0.12 doesn't support them. Are you running these tests in one of the Docker containers? If not you may have a different version of node than me & travis.

v0.12 isn't exactly a new version of node, so we can look into upgrading in Docker & on Travis... but that might be a task out of scope for this particular issue

meandavejustice: I'm removing them for this test. Since this test is run outside of the babel transform.

meandavejustice:
@lmorchard @6a68 r?

lmorchard:
Taking a look... 

[Travis seems to generally like the tests](https://travis-ci.org/mozilla/idea-town/jobs/102016897#L702), but not sure what's up with the [unknown unicode character](https://github.com/mozilla/idea-town/pull/354/files#diff-d2dbb9957d3b230f8edea716289c5fc3R6) before each "Running tests for" line.

idea_town/frontend/static-src/app/views/#experiment-page.js#
```diff
@@ -0,0 +1,249 @@
+import app from 'ampersand-app';
```
lmorchard: Is this file meant to be here? `#experiment-page.js#` seems like maybe a backup filename or something.

TESTING.md
```diff
@@ -0,0 +1,24 @@
+# Testing
+
+We currently have two different ways to write tests. Server tests are fairly
+standard [django tests](https://docs.djangoproject.com/en/1.9/topics/testing/).
+The client tests are written with [tape](https://www.npmjs.com/package/tape), and
+run with [testling](https://www.npmjs.com/package/testling).
+
+Here are a couple of tips to helpout when testing clientside code:
```
lmorchard: Nits:
`s/helpout/help out/`
`s/clientside/client-side/`

idea_town/frontend/static-src/app/views/#experiment-page.js#
meandavejustice: whoops, nice catch. Need to add all of these to my global gitignore.

lmorchard:
Some docs on running these tests locally would be handy. I've tried `npm test` as it appears in `tox.ini`, but it errors out with "No headless browser found". Tried `gulp test`, but it just seems to sit there. Not sure what it's for.

lmorchard:
This is a good & massive PR. Would be really worthwhile to amend & expand the commit message. Add some details that aren't obviously explained by "add tests for frontend". For example:
- Added dependencies to package.json 
  - (what do they all do? maybe copy some of the comments from the issue into the commit message?)
- Moved external templates inline with views 
  - (_might_ be worth some git surgery to split those changes off into a separate commit to de-tangle the changes. But, that might also be more trouble than it's worth.)

IMO, more details the better in the commit message - especially for large PRs. It comes in handy later when someone uses `git blame` to figure out when & why something was changed.

lmorchard:
Hmm, looks like if I run `gulp test` and then visit http://192.168.99.1:9966/, I see test results. Is there any way to run these in a console? I want to hook these up to a file watcher so they'll auto-run when I save changes. Not sure if it's a helpful suggestion at this point, but I've had success with [Karma](https://karma-runner.github.io/0.13/index.html) - it uses websockets to drive a browser & reports in the console.

meandavejustice:
@lmorchard `gulp test` runs `npm run test:browser:live` which watches the `test/` dir and reloads the tests in the browser. These cannot be run in the console, but the simple `npm test` or `npm test:client` commands can.

CONTRIBUTING.md
```diff
@@ -70,7 +77,7 @@ Reviewers will run through roughly the following checklist:
 - Are boundary cases considered?
 - Are error cases handled?
 
-(Once we have some tests, the reviewer will check that all the tests are passing,
+(the reviewer will check that all the tests are passing,
```
jaredhirsch: maybe just pull this into the list of bullet points above. speaking of which, does that list of reviewer things to look for actually track with how reviews are done? if not, should we modify the list or just remember to use the list when reviewing? cc @lmorchard

gulpfile.js
```diff
@@ -100,6 +101,8 @@ gulp.task('scripts', ['lint'], function scriptsTask() {
   return bundledStream;
 });
 
+gulp.task('test', shell.task('npm run test:browser:live'));
```
jaredhirsch: It seems funny to use both gulp and npm for running the same set of tasks. Is there any reason we want it in both places, or can we standardize on one?

idea_town/frontend/static-src/test/lib/router.js
```diff
@@ -0,0 +1,100 @@
...
+  app.on('router:new-page', evt => {
+    if (evt.page === 'landing') {
+      t.ok(evt.page, 'Should redirect to landing if missing addon or creds');
+        app.me.hasAddon = true;
+        app.router.landing()
+      }
+
+    if (evt.page === 'home') {
+      t.ok(evt.page, 'Should redirect to home if logged in with creds');
+      app.reset();
```
jaredhirsch: If `app.reset` needs to be called after every test completes, you'll need to make sure it gets called even if the test fails or throws. For this reason, it's a good idea to extract it out to a test runner.

idea_town/frontend/static-src/test/lib/router.js
```diff
@@ -0,0 +1,100 @@
+import test from 'tape';
+import Router from '../../app/lib/router';
+import app from 'ampersand-app';
+import Me from '../../app/models/me';
+import Experiments from '../../app/collections/experiments';
+
+test(`🖳  Running Tests for ${__filename} ␄`, a => a.end());
```
jaredhirsch: Seems like this bit of boilerplate could be extracted out to a runner or parent class, too.

idea_town/frontend/static-src/test/lib/router.js
```diff
@@ -0,0 +1,100 @@
...
+
+test('landing redirects correctly if logged in with addon', t => {
+  t.plan(2);
+
+  app.me = new Me({
+    user: {
+      id: 'gary@busey.biz',
+        addon: {url: '/lol nothing matters'}
+      }
+  });
```
jaredhirsch: Looks like quite a bit of duplicated test setup in this file. Can you extract this to a setup method?

idea_town/frontend/static-src/test/views/contributor-view.js
```diff
@@ -0,0 +1,9 @@
+import test from 'tape';
+import view from '../../app/views/contributor-view';
+
+test(`🖳  Running Tests for ${__filename} ␄`, a => a.end());
+
+test('Contibutor view renders', t => {
```
jaredhirsch: 'Contributor'

idea_town/experiments/models.py
```diff
@@ -42,6 +43,11 @@ class Meta:
     created = models.DateTimeField(auto_now_add=True)
     modified = models.DateTimeField(auto_now=True)
 
+    @cached_property
+    def installation_count(self):
+        return UserInstallation.objects.distinct('user').filter(
+            experiment=self).count()
+
```
jaredhirsch, line 50: Hmm, is this supposed to be here? Seems like a rebase might be needed?

CONTRIBUTING.md
meandavejustice: I guess _travis_ checks to see that the tests are run.

idea_town/experiments/models.py
meandavejustice, line 50: woah nice catch. lemme get rid of this

gulpfile.js
meandavejustice: Thanks for bringing this up. I should probably move over to only using gulp.

CONTRIBUTING.md
jaredhirsch: > I guess travis checks to see that the tests are run.

For sure, but _humans_ need to verify that the tests make sense, are reasonably independent of one another (so that one bug doesn't cause lots of seemingly-unrelated tests to fail), and that the most important or tricky corners have a little extra thorough coverage.

idea_town/frontend/static-src/app/views/base-view.js
```diff
@@ -33,7 +33,7 @@ export default AmpersandView.extend({
 
     // HACK: Slap the same data-l10n-args data on every localized node, because
     // the most common case is they all need the same model data.
-    const nodes = this.el.querySelectorAll('[data-l10n-id]');
+    const nodes = this.queryAll('[data-l10n-id]');
```
jaredhirsch, line 36: It's fine if you want to include this bug fix in this PR, but it needs to be in a separate commit.

There are tons of ways to do this, but here's one way I'd do it:

``` sh
# pull this file out of the main commit
git reset HEAD~1 idea_town/frontend/static-src/app/views/base-view.js
# amend the main commit
git commit --amend --no-edit
# create a new commit
git add idea_town/frontend/static-src/app/views/base-view.js
git commit -m "Use an array, not a NodeList, to allow for-of iteration of l10n nodes"
```

From there, you can `git rebase -i HEAD~2` and swap the order of the two commit lines to put the bug fix earlier in history than the main testing commit.

meandavejustice, line 36: Thanks @6a68 I'll take care of that :)

jaredhirsch:
@meandavejustice OK, this has been open a really long time, so let's get it landed. Things I think are essential and need to be addressed:

Bigger things:
- Because this PR has been dragging on, if the view test bug can't be fixed in an hour (the `can't find variable: Symbol` bug), just pull those tests out of this PR, stash them in a branch somewhere, and file a followup bug to add view tests. Timebox this aggressively: there be dragons in generated code debugging :dragon_face: 
- The test runner has some problems currently: (1) test runner shouldn't bail if a test throws, and (2) tests currently don't call teardown code if a test fails/throws. To fix this, add graceful exception handling, and separate invocation of setup/test/teardown methods (or beforeEach/afterEach, whatever). The [gist](https://gist.github.com/6a68/c262519b2c1fee3fd251) I mentioned in IRC could work, or use whatever solution you like, so long as we solve the two problems.
- Once the test runner has before/after capability, refactor the tests to pull before/after code out into separate methods (figure out whether you're keeping the view tests before you bother refactoring them). Organize the code however you'd like, just be sure that:
  - shared setup isn't duplicated
  - teardown is in a separate method (to guard against test exceptions), and restores fixture state regardless of whether the test passed, failed, or threw.

Smaller things:
- Get rid of the weird unicode bits (`🖳`  and `␄`) at the start and end of the test names. The test runner output would look fine without them, and this could be confusing to less-experienced contributors.
- Test names should be consistently uppercased or downcased, pick any one. Try to make the format of the test name as consistent and predictable as possible, to make it easy for contributors to add or split tests.
- If you do keep the view tests in this PR, rather than test for the existence of innerHTML, it would be better to test for the presence of a specific node that's inserted when the template has rendered correctly. Any attribute is fine, l10n ids or classes or ids, just test that the right template has rendered.
- Tests are easier to trust when they are obvious, so I'd favor obviousness over terseness. Try to do one thing per line (within reason). In particular, please separate the test code from the assert code:

``` js
// this is really terse
t.ok(new View({headerScroll:false}).render().el);

// this version is really simple: easy to scan, separates the test
// from the assertion, and is more friendly to a novice contributor
const view = new View({headerScroll: false});
view.render();
const someEl = view.query('#some-el-in-template');

t.ok(someEl);
```

Please ping me if you have questions or get stuck on any of these items. Once this list is addressed, I'll probably do one quick nitpicky review, aimed at minor line-level syntactic things, then we'll land this. :+1:

jaredhirsch:
Ah, one other thing I missed: when everything else is done, please expand the commit message, as Les mentioned [above](https://github.com/mozilla/idea-town/pull/354#issuecomment-171372474). Happy to offer feedback/advice on this step too. There are no stupid questions :-)

idea_town/frontend/static-src/app/views/base-view.js
meandavejustice, line 36: https://github.com/meandavejustice/idea-town/commit/454e232792a4cf8dbb07bc87e89338ea47be1519

meandavejustice:
@6a68 I believe I have addressed all of your concerns. Thanks for the feedback.

meandavejustice:
@6a68 or @lmorchard r?

lmorchard:
lgtm, let's get this thing merged

PR #356, opened 2016-01-01
Couple of simple cleanups to addon documentation

meandavejustice:


lmorchard:
Looks like this works for me

PR #357, opened 2016-01-06
Canned experiment data for dev & demos

### LINKED ISSUE #149: Create some canned experiments data for dev & demos
Starting from a fresh dev machine, there are no experiments. I start from a fresh dev machine often, since I frequently blow away my whole setup while tinkering with Docker concerns. It would be nice to have some placeholders out of the box.

We have a mechanism for this: [`fixtures/initial_data_dev.json`](https://github.com/mozilla/idea-town/blob/master/fixtures/initial_data_dev.json) gets loaded on every startup by [`bin/run-dev.sh`](https://github.com/mozilla/idea-town/blob/master/bin/run-dev.sh). 

We could add some canned experiment and experiment detail records there. One tricky bit is adding some sample images (e.g. maybe under `idea_town/frontend/static-src` or `idea_town/base/static`) and ensuring the proper paths in the records to point to them.

lmorchard:
- Tweak README-DOCKER.md to mention the command needed to add demo data
- Add fixtures/demo_data.json for use with Django loaddata
- Add natural keys to Experiment, ExperimentDetail, and UserProfile for
  easier fixture import/export without hardcoded primary keys
- Sample images under media/demo
- Ignore media/\* for Docker

Closes #149


pdehaan:
Would it be worth base64 encoding and inlining all the images in the demo_data.json file as data-uris? Or would that just make a ridiculous 3MB JSON file for no real gain?

lmorchard:
Data URLs won't work - these are filenames, pointing at files on disk. Conveniently, the dev web server also serves them up at URLs that can be built by concatenating the filenames with a base URL. But, Django manages the files for uploads & deletion & etc.

lmorchard:
FWIW, there's also [django-fixture-media](https://github.com/leetrout/django-fixturemedia) but that just copies media files from one directory to another mostly :/

meandavejustice:
I needed to rm -rf `media/` in order to successfully checkout this branch locally. This might be an issue after merging to master.

Beyond that, this all looks great to me. Ready to merge after the above comment is discussed.

lmorchard:
> I needed to rm -rf media/ in order to successfully checkout this branch locally. This might be an issue after merging to master.

Oh, ugh. Maybe I really do need to use something like django-fixture-media, that or move the images to `fixtures/` and add a copy command to the README docs. Might not be a good idea to include them directly in `media/` in the git repo, after all. :/

lmorchard:
Okay, let's try this instead: I just moved `media/demos` to `fixtures/media/demos` and updated `README-DOCKER` to include copying them over as a manual step. Bad idea in retrospect to include the `media/` directory in source control, since that's meant to be local data anyway

PR #367, opened 2016-01-13
Use nsIObserver to shuttle metrics from experiments to main addon.

jaredhirsch:
@lmorchard r?

This is pretty much ready, getting it out tonight so you can review before I'm online tomorrow ^_^

I'm not sure about the addon code lifecycle, in particular, whether we need `isInitialized` to enforce a singleton metrics listener, or the placement of the `destroy()` call, which releases the strong reference to the metrics object held by the ObserverService. If we don't release that pointer properly, we'll introduce a memory leak; I haven't yet investigated how 'zombie compartment' detection works.

I can dig into these two concerns before landing, or if you're feeling good about all this, merge away.


jaredhirsch:
Oh, also, I'm not totally sure these events are actually getting to the dev server; the SDK seems to be throwing some kind of error when `request.post()` is invoked. Not sure if this came up with the original addon-api PR or not.

screenshot 2016-01-12 18 22 21

lmorchard:
Going to take a look this morning and see what it's doing. One place where I've for sure seen those XML errors was in ping-server, but [adding an `Accept: application/json` header to the request](https://github.com/mozilla/idea-town/pull/312/files#diff-2d602b131af78a077f02cdfe940bcc39R13) seemed to fix it earlier last month. Weird that it seems to be happening again even with that header in place.

lmorchard:
Looks like some lines [run afoul of eslint](https://travis-ci.org/mozilla/idea-town/jobs/102008105#L279) here, too, and failed in travis testing.

addon/index.js
```diff
@@ -48,6 +48,7 @@ const URL = require('sdk/url').URL;
 const IdeaTown = require('idea-town');
```
pdehaan: Re: ESLint errors,

```
48:7  error    "IdeaTown" is defined but never used  no-unused-vars
```

We can probably remove this import since we removed lines 66-68 below.

addon/index.js
```diff
@@ -36,6 +35,7 @@ const SERVER_ENVIRONMENTS = {
 
 const store = require('sdk/simple-storage').storage;
 const {Cc, Ci, Cu} = require('chrome');
+Cu.import('resource://gre/modules/Services.jsm');
```
pdehaan, line 40: Re: ESLint errors,

```
   88:5  error    "Services" is not defined             no-undef
   91:5  error    "Services" is not defined             no-undef
  210:5  error    "Services" is not defined             no-undef
```

We probably just need to throw something like this at the top of the file:

``` js
/* global Services:false */
```

addon/index.js
```diff
@@ -77,6 +75,51 @@ function updatePrefs() {
...
+const metrics = {
+  isInitialized: false,
+  init: function() {
+    if (this.isInitialized) {
+      return;
+    }
+    this.isInitialized = true;
+    // Note: the observer service holds a strong reference to this observer,
+    // so we must detach it on shutdown / uninstall by calling destroy().
+    Services.obs.addObserver(metrics, 'idea-town::send-metric', false);
```
lmorchard: Nit: Might be handy to stick `idea-town::send-metric` in a constant var (e.g. `EVENT_SEND_METRIC`) near the start of the file

jaredhirsch:
Hmm, what's the right way to run the linter locally?

```
$ gulp lint
[15:14:10] Using gulpfile ~/codez/github/mozilla-idea-town/gulpfile.js
[15:14:10] Starting 'lint'...
[15:14:16] Finished 'lint' after 5.79 s
```

jaredhirsch:
> Hmm, what's the right way to run the linter locally?

Looks like `eslint .` inside the addon directory is how travis-ci does it, but I'm getting an error due to a missing dependency: `Error: Cannot find module 'eslint-config-idea-town'`

pdehaan:
IIRC, It should just be `gulp lint`, but let me grab your repo/branch and try running it locally.

jaredhirsch:
Doh! Got it! From [tox.ini](https://github.com/mozilla/idea-town/blob/master/tox.ini#L49): 

```
cd addon
npm install
npm run lint
```

addon/index.js
```diff
@@ -77,6 +75,51 @@ function updatePrefs() {
...
+    // The nsIObserverService sends non-useful positional arguments; the third
+    // is the only one we need.
+    const data = arguments[2];
+
+    let d;
+    try {
+      d = JSON.parse(data);
+    } catch (ex) {
+      console.error('Idea Town metrics error: cannot process event, ' +
+                    ' JSON.parse failed: ', ex);
```
jaredhirsch: @lmorchard Thoughts on the best way to notify individual addons if their metrics packets are malformed? Right now I've got these `console.error` lines, because it seemed weird to actually throw just because a fire-and-forget packet looked wrong. OTOH the eslint config complains about the use of console.

pdehaan: ESLint should only throw warnings on `console.log()` and `console.error()`. We can disable that rule globally in the .eslintrc if you want, or, you can disable it for a single line using a doc comment:

``` js
console.log('this rule is kind of nonsense'); // eslint-disable-line no-console
```

pdehaan: Ah. Yes, this is very confusing. I didn't see that you were playing in the /addon directory, which is a completely separate .eslintrc and .eslintignore and package.json, so our global `gulp lint` wouldn't even look in that folder.

addon/index.js
pdehaan: On that note, if we remove this `require()`, we can remove it from the [package.json](https://github.com/mozilla/idea-town/blob/master/addon/package.json#L21) as well:

```
➜  addon git:(355-clientside-metrics) ✗ depcheck
Unused Dependencies
* idea-town
```

addon/index.js
lmorchard: > Thoughts on the best way to notify individual addons if their metrics packets are malformed?

`console.error` with an eslint exception is probably best for now.

That doesn't really answer how we could have 2-way communication between addons, though, beyond fire-and-forget events. My Firefox-fu falls down in this area.

jaredhirsch: > That doesn't really answer how we could have 2-way communication between addons, though, beyond fire-and-forget events.

In a hilarious plot twist, I've figured out that it _is_ actually possible to have a single JS file that's compatible with the SDK or the Gecko-centric JSM code loaders. Getting the URI to this file for Cu.import() loading is unforgivably strange, but we can give implementers a snippet to copy-and-paste as they see fit.

Cu.import requires getting the module path relative to the top level of the addon, so a separate addon-api folder still can't work, but we can avoid nsIObserver and just call a callback if an error occurs.

Working on the updated PR now.

lmorchard:
> In a hilarious plot twist, I've figured out that it is actually possible to have a single JS file that's compatible with the SDK or the Gecko-centric JSM code loaders

Hmm, I do still kind of want to keep going in the direction of having the Idea Town addon offering a service API via messaging - rather than try to embed & maintain shared Idea Town client code in all the experiment addons. Feels like that puts a wall around where the complex Idea Town specific stuff lives and who has to deal with it (ie. idea town maintainers rather than experiment maintainers)

jaredhirsch:
@lmorchard Yeah, I hear you. There's really two reasons you would want an errback: if the idea town server is unreachable, for later retries, or to notify sender that the packet was malformed in some way. The first one seems like the idea town addon's responsibility. The second is really a debugging requirement; if you're staring at the console anyway, a callback function doesn't really give you more than a callback message.

`nsIObserver` messaging it is. I'll push the couple of linter fixes and we can call it a day.

addon/index.js
jaredhirsch: Great idea, updated.

addon/index.js
jaredhirsch, line 40: Yup, thanks!

addon/index.js
jaredhirsch: Cool, updated.

jaredhirsch:
@lmorchard OK, all updated. Mind taking another look?

jaredhirsch:
@lmorchard bump :-)

lmorchard:
Looks good to me! I was able to do this in the Browser Console:

```
var observerService = Components.classes["@mozilla.org/observer-service;1"]
    .getService(Components.interfaces.nsIObserverService);
observerService.notifyObservers(null, 'idea-town::send-metric', JSON.stringify({ 
    key: 'foo', value: 'foo', addonName: 'foo' }));
```

And, I saw a request hit the backend server. [That request resulted in a 500 error](https://pastebin.mozilla.org/8857025), so we probably have some work to do there, but we can call that out of scope for this PR.

PR #371, opened 2016-01-15
clean up and consolidate views and templates

### LINKED ISSUE #370: cleanup client templates

meandavejustice:
fixes #370 

Consolidate view and templates when templates are small enough to fit in file reasonably.
Remove some dead code and deps.


jaredhirsch:
@meandavejustice could you add some explanation to this PR? I'm curious why you're inlining the templates, what the goals are here

lmorchard:
FWIW, I think the templates in these views are small enough that it helps clarity to have everything all in the same place.

idea_town/frontend/static-src/app/views/feedback-view.js
```diff
@@ -51,7 +51,6 @@ export default BaseView.extend({
   },
 
   cancel(e) {
-    this.animateRemove();
```
jaredhirsch, line 54: @meandavejustice `this.animateRemove` just calls `this.remove` by default, so it seems like removing this line altogether will not remove the feedback view when the user dismisses it. Is that right, or am I missing something?

PR #372, opened 2016-01-18
Introduce l20n.js for client-side localization

### LINKED ISSUE #130: Add localization

lmorchard:
- Use custom build of l20n to cope with bug 1240192
- Copy locales over in gulp build
- Mark elements for localization in all HTML templates
- Use l20n API to localize feedback dialog strings
- Try localizing strings for en-US

This does not cover server-side localization of DB content, just the client-side app templates. This also doesn't include an actual UI for localizing the strings.

### DIFF
--- a/idea_town/frontend/static-src/app/templates/experiment-page.js
+++ b/idea_town/frontend/static-src/app/templates/experiment-page.js
@@ -6,9 +6,9 @@ export default `
         <div class="details-header">
           <h1 data-hook="title"></h1>
           <div class="idea-controls">
-            <button data-hook="feedback" id="feedback-button" class="button primary">Give Feedback</button>
-            <button data-hook="uninstall" id="uninstall-button" class="button neutral"><span class="state-change-inner"></span><span class="transition-text">Disabling...</span><span class="default-text">Disable <span data-hook="title"></span></span></button>
-            <button data-hook="install" id="install-button" class="button primary"><span class="state-change-inner"></span><span class="transition-text">Enabling...</span><span class="default-text">Enable <span data-hook="title"></span></span></button>
+            <button data-l10n-id="giveFeedback" data-hook="feedback" id="feedback-button" class="button primary">Give Feedback</button>
+            <button data-hook="uninstall" id="uninstall-button" class="button neutral"><span class="state-change-inner"></span><span data-l10n-id="disableExperimentTransition" class="transition-text">Disabling...</span><span data-l10n-id="disableExperiment" class="default-text">Disable <span data-hook="title"></span></span></button>
+            <button data-hook="install" id="install-button" class="button primary"><span class="state-change-inner"></span><span data-l10n-id="enableExperimentTransition" class="transition-text">Enabling...</span><span data-l10n-id="enableExperiment" class="default-text">Enable <span data-hook="title"></span></span></button>
             <span class="user-count"></span>
             <span data-hook="install-count"></span>
           </div>
@@ -20,32 +20,31 @@ export default `
           <div class="details-overview">
             <img data-hook="thumbnail" src="" width="260" height="260">
             <section data-hook="measurements-container" class="measurement">
-              <h3>Measurements</h3>
-              <p class="disclaimer">All data is collected anonymously and used only to help us improve this test.</p>
+              <h3 data-l10n-id="measurements">Measurements</h3>
+              <p data-l10n-id="measurementsDisclaimer" class="disclaimer">All data is collected anonymously and used only to help us improve this test.</p>
               <div data-hook="measurements-html"></div>
-              <a data-hook="privacy-notice-url">
-              <span data-hook="title"></span> Privacy Notice</a>
+              <a data-l10n-id="experimentPrivacyNotice" data-hook="privacy-notice-url"><span data-hook="title"></span> Privacy Notice</a>
             </section>
             <section>
-              <h3>Brought to you by</h3>
+              <h3 data-l10n-id="contributorsHeading">Brought to you by</h3>
               <ul class="contributors"></ul>
             </section>
             <section>
-              <h3>Details</h3>
+              <h3 data-l10n-id="detailsHeading">Details</h3>
               <table class="stats">
                 <tr data-hook="version-container">
-                  <td>Version</td>
+                  <td data-l10n-id="version">Version</td>
                   <td>
                     <span data-hook="version"></span>
-                    &nbsp;<a data-hook="changelog-url">changelog</a>
+                    &nbsp;<a data-l10n-id="changelog" data-hook="changelog-url">changelog</a>
                   </td>
                 </tr>
                 <tr>
-                  <td>Last Update</td>
+                  <td data-l10n-id="lastUpdate">Last Update</td>
                   <td data-hook="modified-date"></td>
                 </tr>
                 <tr>
-                  <td>Contribute</td>
+                  <td data-l10n-id="contribute">Contribute</td>
                   <td><a data-hook="contribute-url"></a></td>
                 </tr>
               </table>
@@ -60,4 +59,3 @@ export default `
     <div data-hook="main-footer"></div>
   </div>
 `;
-
--- a/idea_town/frontend/static-src/app/views/base-view.js
+++ b/idea_town/frontend/static-src/app/views/base-view.js
@@ -14,6 +14,29 @@ export default AmpersandView.extend({
     this.beforeRender();
     AmpersandView.prototype.render.apply(this, arguments);
     this.afterRender();
+
+    this.localizeRendered();
+    if (this.model) {
+      this.model.on('change', () => this.localizeRendered);
+    }
+  },
+
+  getL10nArgs() {
+    // Most common case for l10n args comes from the current model - e.g. on
+    // experiment detail page.
+    return this.model ? this.model.toJSON() : {};
+  },
+
+  localizeRendered() {
+    const args = this.getL10nArgs();
+    const argsJSON = JSON.stringify(args);
+
+    // HACK: Slap the same data-l10n-args data on every localized node, because
+    // the most common case is they all need the same model data.
+    const nodes = this.el.querySelectorAll('[data-l10n-id]');
+    for (const node of nodes) {
+      node.setAttribute('data-l10n-args', argsJSON);
+    }
   },
 
   // implement in subclasses
--- a/idea_town/frontend/static-src/app/views/experiment-page.js
+++ b/idea_town/frontend/static-src/app/views/experiment-page.js
@@ -7,6 +7,7 @@ import DetailView from './detail-view';
 import ContributorView from './contributor-view';
 import template from '../templates/experiment-page';
 import FeedbackView from './feedback-view';
+
 const changeHeaderOn = 120;
 
 const CollectionExtended = Collection.extend({
@@ -214,16 +215,15 @@ export default PageView.extend({
 
   renderUninstallSurvey(evt) {
     evt.preventDefault();
-    // TODO: Hardcoded survey, for now. Populate via API later?
     this.renderSubview(new FeedbackView({
       id: 'disabled-feedback',
       experiment: this.model.url,
-      title: 'Why are you stopping?',
+      title: 'feedbackUninstallTitle',
       questions: [
-        { value: 'broken', title: 'This thing is broken!' },
-        { value: 'dislike', title: 'I don\'t like this feature.' },
-        { value: 'notuseful', title: 'This isn\'t useful for me.' },
-        { value: 'other', title: 'Something else.' }
+        { value: 'broken', title: 'feedbackUninstallAnswerBroken' },
+        { value: 'dislike', title: 'feedbackUninstallAnswerDislike' },
+        { value: 'notuseful', title: 'feedbackUninstallAnswerNotUseful' },
+        { value: 'other', title: 'feedbackUninstallAnswerOther' }
       ],
       onSubmit: () => {
         this.uninstall(this.updateAddon, this.model);
@@ -245,16 +245,15 @@ export default PageView.extend({
 
   feedback(evt) {
     evt.preventDefault();
-    // TODO: Hardcoded survey, for now. Populate via API later?
     this.renderSubview(new FeedbackView({
       id: 'enabled-feedback',
       experiment: this.model.url,
-      title: 'Tell us what you think',
+      title: 'feedbackTitle',
       questions: [
-        { value: 'broken', title: 'Something seems broken.' },
-        { value: 'feature', title: 'Request a feature.' },
-        { value: 'cool', title: 'This is cool!' },
-        { value: 'other', title: 'Something else.' }
+        { value: 'broken', title: 'feedbackAnswerBroken' },
+        { value: 'feature', title: 'feedbackAnswerFeature' },
+        { value: 'cool', title: 'feedbackAnswerCool' },
+        { value: 'other', title: 'feedbackAnswerOther' }
       ]
     }), 'body');
   }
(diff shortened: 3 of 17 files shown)


lmorchard:
Things seem to work here, but could use some more eyes on it. Also, it touches all the templates, so it might run into conflicts with what @meandavejustice was doing with inlining them into view.

This also doesn't do anything for getting us integrated with a localization tool like Pontoon. If this looks like the right approach to l20n, that could be a follow up issue.

Might also help to ping @stasm and/or @zbraniecki to see if I'm using their work properly, here.

idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -6,9 +6,9 @@ export default `
...
           <h1 data-hook="title"></h1>
           <div class="idea-controls">
-            <button data-hook="feedback" id="feedback-button" class="button primary">Give Feedback</button>
-            <button data-hook="uninstall" id="uninstall-button" class="button neutral"><span class="state-change-inner"></span><span class="transition-text">Disabling...</span><span class="default-text">Disable <span data-hook="title"></span></span></button>
-            <button data-hook="install" id="install-button" class="button primary"><span class="state-change-inner"></span><span class="transition-text">Enabling...</span><span class="default-text">Enable <span data-hook="title"></span></span></button>
+            <button data-l10n-id="giveFeedback" data-hook="feedback" id="feedback-button" class="button primary">Give Feedback</button>
+            <button data-hook="uninstall" id="uninstall-button" class="button neutral"><span class="state-change-inner"></span><span data-l10n-id="disableExperimentTransition" class="transition-text">Disabling...</span><span data-l10n-id="disableExperiment" class="default-text">Disable <span data-hook="title"></span></span></button>
```
jaredhirsch, line 10: @lmorchard If we're intending to localize the individual experiment names, will this part (scrolled off the screen) work?

`<span data-l10n-id="disableExperiment" class="default-text">Disable <span data-hook="title"></span></span>`

Back when I did fxa localization on Gaia (l10n.js era), IIRC the flow was like this:
- Replace any nested html with a mustache inside the localized property: `disableExperiment=Disable {{ title }}`
- Manually call the l10n function to translate 'title' and 'Disable'
- Concat localized strings and html by building up a string
- Finally, insert into the DOM

One major reason for not embedding html in the properties files was to keep html invisible to localizers, since they might easily typo the tag when copying it over. Another is that putting html in properties files makes future html updates dependent on updating localizations.

I may be misremembering part of this flow, @stas should know more.

Unfortunately, the l20n.js docs don't mention nested html and localized text, so it's not clear what the recommended approach would be.

jaredhirsch, line 10: OK, I found the [old PR](https://github.com/mozilla-b2g/gaia/pull/18392) I referred to above.

In order to localize this bit of html:

``` html
<p>By proceeding, I agree to the <a href="">terms</a> and 
<a href="">privacy policy</a> of Firefox Cloud Services.</p>
```

I wound up doing this:

properties file:

```
fxa-notice = By proceeding, I agree to the {{tos}} and {{pn}} of Firefox cloud services.
# LOCALIZATION NOTE(fxa-tos): This string corresponds to the {{tos}} placeholder
# in fxa-notice.
fxa-tos = Terms of Service
# LOCALIZATION NOTE(fxa-pn): This string corresponds to the {{pn}} placeholder
# in fxa-notice.
fxa-pn = Privacy Notice
```

in the view rendering method:

``` js
// translate outer text
var localize = navigator.mozL10n.localize;
var noticeText = _('fxa-notice');

// string replace to insert inner nodes
var tosReplaced = noticeText.replace(
  '{{tos}}',
  '<a id="fxa-terms" href="#" class="disabled">Terms of Service</a>'
);
var tosPnReplaced = tosReplaced.replace(
  '{{pn}}',
  '<a id="fxa-privacy" href="#" class="disabled">Privacy Notice</a>'
);

// for each level of nesting, we had to render to the DOM
// to have an element to pass to `localize` :-(
this.fxaNotice.innerHTML = tosPnReplaced;

// grab refs to inner els and localize them
this.fxaPrivacy = document.getElementById('fxa-privacy');
localize(this.fxaPrivacy, 'fxa-pn');
this.fxaTerms = document.getElementById('fxa-terms');
localize(this.fxaTerms, 'fxa-tos');
```

Hopefully the new tool handles this common case in a nicer way, looking forward to hearing from the l20n crew ^_^

lmorchard, line 10: > If we're intending to localize the individual experiment names, will this part (scrolled off the screen) work?

Localizing content from the database - as opposed to app templates - isn't covered in this PR. I should add that as a note to the PR. Not sure how we're going to do that yet, but it will probably happen at the Django server model / API level rather than the Ampersand client level. (i.e. experiment names will come pre-localized by the time they reach the client)

Would be nice if DB content like experiment names could be localized through the same interface that the app itself uses, but I can't remember a Mozilla project where things worked out that way.

Of course, this PR doesn't cover a localization interface, either. Kind of thinking Pontoon, but I don't have any experience with it yet.

> Unfortunately, the l20n.js docs don't mention nested html and localized text, so it's not clear what the recommended approach would be.

[In the localized strings](https://github.com/mozilla/idea-town/pull/372/files#diff-4eb6e773dc567e34f0afea71a400cb5fR37), I didn't include the `<span>` and punted like so:

```
<disableExperiment "Disable {{title}}">
```

The only reason the `<span>` is there, anyway, is for Ampersand to have something as a hook for the experiment title binding. But, L20N takes care of that as part of the localization anyway. Not that that solves the general problem of markup in l10n strings, but I don't think it's an issue in this particular case. I should probably also remove the Ampersand binding, since it becomes redundant now.

lmorchard:
Off the top of my head, another approach to localizing the DB content could be to produce a DB-based dynamically generated `.l20n` file alongside the one for app templates. 

We can't simply roll the DB content localizations into that same static `.l20n` file for the app. To do that, we'd have to re-deploy the site on every content change, which kind of defeats the purpose of having a Django-based CMS. That said, I'm not sure if that would help us take advantage of a common l10n interface for app & DB content.

idea_town/frontend/static-src/app/templates/experiment-page.js
zbraniecki, line 10: There is no need to hack around HTML localization. L20n supports DOM Overlays which allows you to do sth like:

```
<div data-l10n-id="foo"></div>
```

```
<foo "Hello <small>to</small> <strong>L20n</strong>">
```

idea_town/frontend/static-src/app/views/base-view.js
```diff
@@ -16,6 +16,29 @@ export default AmpersandView.extend({
...
+  localizeRendered() {
+    const args = this.getL10nArgs();
+    const argsJSON = JSON.stringify(args);
+
+    // HACK: Slap the same data-l10n-args data on every localized node, because
+    // the most common case is they all need the same model data.
+    const nodes = this.el.querySelectorAll('[data-l10n-id]');
+    for (const node of nodes) {
+      node.setAttribute('data-l10n-args', argsJSON);
+    }
```
zbraniecki, line 39: @Pike, @stas - this use case may be useful when we talk about per-context variables for l20n.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -215,20 +218,28 @@ export default PageView.extend({
...
+      this.renderSubview(new FeedbackView({
+        id: 'disabled-feedback',
+        experiment: this.model.url,
+        title: strings.feedbackUninstallTitle,
+        questions: [
+          { value: 'broken', title: strings.feedbackUninstallAnswerBroken },
+          { value: 'dislike', title: strings.feedbackUninstallAnswerDislike },
+          { value: 'notuseful', title: strings.feedbackUninstallAnswerNotUseful },
+          { value: 'other', title: strings.feedbackUninstallAnswerOther }
+        ],
```
zbraniecki: why do you prefer to mass-l10n using `formatValues` rather than build DOM with `data-l10n-id`?

idea_town/frontend/static-src/app/views/feedback-view.js
```diff
@@ -37,6 +37,14 @@ export default BaseView.extend({
     this.parentOnSubmit = opts.onSubmit;
   },
 
+  render() {
+    BaseView.prototype.render.apply(this, arguments);
+    // HACK: This feels super-dirty. Is there a better way to localize attributes?
+    document.l10n.formatValue('feedbackExtraPlaceholder').then((string) => {
+      this.el.querySelector('.extra textarea').setAttribute('placeholder', string);
+    });
```
zbraniecki:
```
<feedbackExtra
  placeholder: "foo">
```

and then:

```
this.el.querySelector('.extra textarea').setAttribute('data-l10n-id', 'feedbackExtra');
```

idea_town/frontend/static-src/app/templates/experiment-page.js
zbraniecki, line 10: > Unfortunately, the l20n.js docs don't mention nested html and localized text, so it's not clear what the recommended approach would be.

https://blog.mozilla.org/l10n/2012/07/16/l20n-features-explained-dom-overlays/

zbraniecki:
Not sure what are the questions for me here (sorry! There's a lot going on in this issue).

Pontoon supports l20n file format, so it should work OOTB.

idea_town/frontend/static-src/app/templates/experiment-page.js
lmorchard, line 10: Not actually hacking around HTML localization here, FWIW. We don't actually need the embedded HTML in this string, it's just there as a DOM hook for population by Ampersand. But, L20N does that for us now, instead.

idea_town/frontend/static-src/app/views/base-view.js
lmorchard, line 39: Yeah, I was thinking it would be very nice if there was a page-global set of variables. Maybe allow overrides at the DOM node level, but be able to share a common set of vars.

lmorchard:
> Not sure what are the questions for me here

Don't really have any specific questions, but your feedback so far has been great thanks! :)

idea_town/frontend/static-src/app/views/experiment-page.js
lmorchard: Hmm... I hadn't thought of that, and this did seem unnecessarily complex. To clarify that for myself - I think you're saying we should just populate the feedback form widget with empty elements bearing the appropriate `data-l10n-id` values.

idea_town/frontend/static-src/app/templates/experiment-page.js
jaredhirsch, line 10: > https://blog.mozilla.org/l10n/2012/07/16/l20n-features-explained-dom-overlays/

@zbraniecki Thanks for the link. DOM overlays look like _exactly_ what I would want from an l10n library, nice work. (FWIW, I was looking (and not finding) over at http://l20n.org/learn.)

lmorchard:
Updated the commit. Tweaked the feedback view to just use `data-l10n-id` attributes, rather than fetching the localizations itself. Also, got rid of the `<textarea>` placeholder hack.

idea_town/frontend/static-src/locales/en-US.l20n
```diff
@@ -0,0 +1,68 @@
...
+<menuTour "Tour Idea Town">
+<menuWiki "Idea Town Wiki">
+<menuFileIssue "File an Issue">
+<menuLogout "Logout">
+
+<landingWelcome "Welcome to Idea Town">
+<landingIntro "Test new features.<br> Give us feedback.<br> Help build Firefox.">
+<landingFxaGetStartedButton "Get started with your Firefox Account">
+<landingFxaAlternateButton "Don't have an account? Sign up.">
+<landingCtaLegal "By proceeding, you agree to the <a href=\"https://www.mozilla.org/about/legal/terms/services/\">Terms of Service</a> and <a href=\"https://www.mozilla.org/privacy/firefox-cloud/\">Privacy Notice</a> of Idea Town.">
```
jaredhirsch: Per stas' comment about DOM overlays, you can pull the `href`s out of this string

lmorchard:
Just as a note... This and #371 / #370 will collide badly, since I touch every template in this PR and the other moves all the templates into views. So, one will need some massaging if the other lands first.

idea_town/frontend/static-src/locales/en-US.l20n
lmorchard: Oh hey, what do you know, that works great!

idea_town/frontend/static-src/app/views/base-view.js
zbraniecki, line 39: http://l20n.org/learn/using-external-variables-in-translations/ - Context Data is exactly that, but recently @Pike was not sure if we have use cases for it.

lmorchard:
Merged #371 and then rebased this PR against the new master

idea_town/frontend/static-src/app/views/base-view.js
lmorchard, line 39: Hmm, can I set that context data in the current l20n? After getting confused by different versions between l20n.org, MDN, and github - I've been mostly ignoring docs & reading the code to get this far :)

zbraniecki, line 39: Unfortunately, we didn't implement it yet. I'll bring this use case at the next API meeting :)

lmorchard:
Another tweak, slight reorganization of the `locales` directory and moved to the root of the project for simplicity

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -214,16 +215,15 @@ export default PageView.extend({
...
+      title: 'feedbackUninstallTitle',
       questions: [
-        { value: 'broken', title: 'This thing is broken!' },
-        { value: 'dislike', title: 'I don\'t like this feature.' },
-        { value: 'notuseful', title: 'This isn\'t useful for me.' },
-        { value: 'other', title: 'Something else.' }
+        { value: 'broken', title: 'feedbackUninstallAnswerBroken' },
+        { value: 'dislike', title: 'feedbackUninstallAnswerDislike' },
+        { value: 'notuseful', title: 'feedbackUninstallAnswerNotUseful' },
+        { value: 'other', title: 'feedbackUninstallAnswerOther' }
```
jaredhirsch, line 226: these l10n ids are great, nice work

idea_town/frontend/templates/idea_town/frontend/index.html
```diff
@@ -4,11 +4,18 @@
...
+    <meta name="availableLanguages" content="en-US">
+    <link rel="localization" href="{{ static('locales/{locale}/app.l20n') }}">
+
 </head>
 <body>
+
   <div data-hook="page-container"></div>
 
   <script src="{{ static('app/app.js') }}"></script>
+  <script src="https://pontoon.mozilla.org/pontoon.js"></script>
```
jaredhirsch, line 18: Ehh, this isn't awesome, but it's also not a huge deal. I'll file a bug to remove this and either host it locally or not include it till the pontoon part is ready.

jaredhirsch, line 18: Filed as #373, cleared for landing

Issue #92, opened 2015-09-25
handle build and serving of addon


jaredhirsch:
The addon should be auto-signed without problems by the AMO tools, so we should easily be able to replace the raw.github.com URL (which isn't served with the right mime-type) with an AMO link.

Assigning to @meandavejustice because he uploaded the addon originally and should have a pointer to the link.

jaredhirsch:
This depends on https://github.com/mozilla/idea-town-addon/pull/22

meandavejustice:
Still having trouble with this for some reason... after signing and trying to install I still get this error
![](http://i.imgur.com/CWRU6yv.png)

meandavejustice:
Unfortunately we cannot link to AMO since our addon is unlisted. I tried testing to see if it would still work as hidden. But cannot switch visibility https://bugzilla.mozilla.org/show_bug.cgi?id=1122107

jaredhirsch:
Yup, you're exactly right.

There's a bug open to find a way to promote idea town addons from the AMO 'unlisted' state to the 'hidden' state: https://bugzilla.mozilla.org/show_bug.cgi?id=1225536

meandavejustice:
So my plan for this atm is to have a simple signing/packaging `npm script` in the addon called `deploy`. From our `gulpfile` we can build and move it into our `static` dir.

I'll need to do some investigation on hosting our own addon, via updates, access control etc...
@lmorchard how does this sound? Am I leaving anything out here?

lmorchard:
Hmm, so I haven't had a change to play with `jpm sign` myself yet... Does that come back immediately with a signed add-on? Or do you have to wait for an email notification or something?

If it's an immediate thing, it would be cool to roll add-on build & signing into the scripts that build our Docker container for deployment. We're already running gulp in there to build the frontend JS.

If it waits on a notification, we'll probably have to keep manually building the add-on, keep it in a dist directory & copy it over during container build. Either that, or add a file upload field in the admin pages so we can update the add-on separately from site deployment (though that sounds like a bad idea)

jaredhirsch:
@meandavejustice Helpful links for hosting add-ons:

[...]

PR #377, opened 2016-01-20
Fixes to enable / disable flow in addon & frontend

### LINKED ISSUE #325: Enable/Disable state gets messed up after one install cycle.
https://github.com/mozilla/idea-town/pull/324#issuecomment-161717938

lmorchard:
- Use a one-shot listener for install-ended / uninstall-ended messages
  in frontend enable / disable buttons
- Simplify uninstall / renderUninstallSurvey handlers in frontend
- Skip addon uninstall call when frontend calls to uninstall an
  experiment that's not installed
- Ensure messages from addon to frontend happen before metrics ping
- notifyObserver -> notifyObservers typo in addon metrics ping
- Remove AddonManager listeners at unload time to prevent "can't access
  dead object" warnings due to zombie listeners

Fixes #325


meandavejustice:
Would you do a `npm run package` in the addon directory?

lmorchard:
Updated with a new addon package. It occurs to me, though that 1) we should automate that as part of the Docker build and 2) should we bump the version whenever that happens? and 3) is there anywhere we're already using this URL for addon installation? (maybe all concerns for #92)

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -185,32 +185,32 @@ export default PageView.extend({
     evt.target.style.width = width + 'px';
     evt.target.classList.add('state-change');
 
-    this.updateAddon(true, this.model);
-
-    app.on('webChannel:addon-install:install-ended', () => {
-      this.model.enabled = !this.model.enabled;
+    // TODO: Should this maybe be an event listener at a global level?
```
jaredhirsch, line 189: About that TODO: I think it might make sense to abstract these install/uninstall listeners out into an Addon model, and toggle view changes by listening to `change` events on that model's attributes, something like `addonModel.on('change:isInstalled', cb)` or whatever you want to call the attrs. Dunno if you want to fold that into this PR or file a followup bug.

idea_town/frontend/static-src/app/views/experiment-page.js
```diff
@@ -225,9 +225,7 @@ export default PageView.extend({
         { value: 'notuseful', title: 'feedbackUninstallAnswerNotUseful' },
         { value: 'other', title: 'feedbackUninstallAnswerOther' }
       ],
-      onSubmit: () => {
-        this.uninstall(this.updateAddon, this.model);
-      }
+      onSubmit: () => this.uninstall(evt)
```
jaredhirsch, line 229: Should this read `(evt) => this.uninstall(evt)`?

addon/index.js
```diff
@@ -205,6 +205,7 @@ Router.prototype.on = function(name, f) {
 };
 
 Router.prototype.send = function(name, data, addon) {
+  this.mod.port.emit('from-addon-to-web', {type: name, data: data});
```
jaredhirsch, line 210: Is there any particular reason this line was moved to the top of the function? If it's order dependent, would be good to insert a comment above this line to that effect.

addon/index.js
```diff
@@ -426,4 +415,21 @@ AddonManager.addInstallListener({
   onDownloadFailed: function(install) {
     app.send('addon-install:download-failed', formatInstallData(install));
   }
+};
+AddonManager.addInstallListener(installListener);
```
jaredhirsch, line 421: Could we collect these `add*Listener` methods into some kind of addon startup method? There's some loose initialization code around lines 126-141, too. Would be great to have a single, scannable place to grok the startup sequence, and I don't think it'd be _too_ much creep for this PR.

The `unload.when` callback below looks like a great candidate to extract out into a top-level shutdown method (yay symmetry).

idea_town/frontend/static-src/app/views/experiment-page.js
lmorchard, line 229: No, `evt` is a reference to [the parameter to `renderUninstallSurvey`](https://github.com/mozilla/idea-town/pull/377/files#diff-9a1984ab9661edeed8d9b7bed2951b8cR216)

idea_town/frontend/static-src/app/views/experiment-page.js
lmorchard, line 189: That sounds like a pretty hefty PR on its own, and would rather get this fix in ASAP

addon/index.js
lmorchard, line 210: During my debugging, exceptions thrown by the metrics ping broke the frontend client too. The exception was due to a typo, but having moved this doesn't hurt things either.

addon/index.js
lmorchard, line 421: Well, this whole `index.js` is pretty much a big startup method, really. Everything is initialization

jaredhirsch:
> Updated with a new addon package. It occurs to me, though that 1) we should automate that as part of the Docker build and 2) should we bump the version whenever that happens? and 3) is there anywhere we're already using this URL for addon installation? (maybe all concerns for #92)

It does seem weird to change the addon without at least bumping the patch number.

FWIW, I've been manually bumping the version number in the search addon, and it seems like a useful thing to have as a separate commit in git history, to help with bisecting when users start filing bugs. I'm not sure it's possible to fully automate this step, since humans need to decide whether it's a major, minor, or patch-level change.

addon/index.js
lmorchard, line 421: Might make sense to shuffle some things around - i.e. move all the functions & classes to the bottom. For example, I just now noticed the Metrics stuff got inserted between the `updatePrefs` declaration and the actual initialization of prefs. But, there's a lot going on here between prefs, metrics, toolbar button, addon/frontend message handling, and the AddonManager monitoring - that's pretty much the whole shebang in initialization :)

jaredhirsch:
@lmorchard Looks good overall, see comments. Couple of refactoring "choose your own adventure" suggestions; I think they would be useful to do now, but if you'd rather just get this patch landed, I can run a couple of sanity checks locally and then merge; let me know.

lmorchard:
> I'm not sure it's possible to fully automate this step, since humans need to decide whether it's a major, minor, or patch-level change.

I don't mean automating the version step so much as automating the build and rolling it into the Docker image, so we have an atomic artifact joining server / frontend / addon changes

lmorchard:
So, some shuffling I tried actually broke things. Hope this small reorg makes things a little more sense. Also bumped the addon version.

meandavejustice:
I just pulled this down and it's working well for me. The reorg of `index.js` looks good to me as well. I'm going to leave this to @6a68 to merge since they did most of the reviewing.

:+1:

lmorchard:
Rebased against master just to be sure things didn't break

PR #383, opened 2016-01-21
Issue #130: Localize page titles with L20n

lmorchard:
- Tweak page title handling in page-manager to use data-l10n-{args,id}
- Add l10n IDs for each page title string in major views
- Drop string concatenation for page title handling
- Add page title strings to en-US locale


meandavejustice:
:+1: 1:

PR #386, opened 2016-01-26
Add short_title field to experiments

### LINKED ISSUE #262: Truncate long experiment names and descriptions
It's currently a bit too easy to break the layout if you specify an offensively long experiment **title** (or **description**):

http://idea-town-dev.elasticbeanstalk.com/experiments/experiment-third-which-has-longer-title

idea_town
**Figure 1:** long **title** in the details page. This is super edge case for sure, but you can see how easy it is to overflow that "Enable {{experiment.name}}" button label, which is a lot more likely.

---

idea_town
**Figure 2:** The home page with the absurdly long description. We should probably trim to the first _X_ characters on this page (and make sure we strip out any potential HTML tags).

### LINKED ISSUE #301: Each experiment should have a short description and a long description
Sort descriptions should appear on the home page and be quite short (capped at 60 characters).

Long descriptions can be longer. These should appear on the experiment details page:

screen-shot-2015-11-19-at-7 43 33-pm

lmorchard:
- Add short_title to model, admin, and serializer
- Display short_title in experiment-row-view when it has content,
  otherwise fall back to title

Fixes #301, fixes #262

### DIFF
--- a/idea_town/experiments/admin.py
+++ b/idea_town/experiments/admin.py
@@ -13,7 +13,7 @@ class ExperimentAdmin(admin.ModelAdmin):
 
     list_display = ('id',
                     show_image('thumbnail'),
-                    'title', 'version', 'addon_id',
+                    'title', 'short_title', 'version', 'addon_id',
                     related_changelist_link('details'),
                     related_changelist_link('users'),
                     related_changelist_link('feedbacks'),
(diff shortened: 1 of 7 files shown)


lmorchard:
Hmm, test failure on travis, but passed locally... need to poke at that

meandavejustice:
need to update experiment_row_view test https://travis-ci.org/mozilla/idea-town/jobs/104995042#L758

lmorchard:
Added tests for the `short_title` fallback, passes locally so let's see if travis likes it

PR #390, opened 2016-02-02
Implement "Now Active" indicators for the experiment detail page.

### LINKED ISSUE #385: Implement "Now Active" indicator on Experiment detail page
This "Now Active" indicator seems to be unimplemented:

Now Active

Should get this implemented, if we still want it.

meandavejustice:
- added tests
- update locale string
- fixes #385

### DIFF
--- a/idea_town/frontend/static-src/app/templates/experiment-page.js
+++ b/idea_town/frontend/static-src/app/templates/experiment-page.js
@@ -5,6 +5,7 @@ export default `
       <div class="details-header-wrapper">
         <div class="details-header">
           <h1 data-hook="title"></h1>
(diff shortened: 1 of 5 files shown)


idea_town/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -5,6 +5,7 @@ export default `
       <div class="details-header-wrapper">
         <div class="details-header">
           <h1 data-hook="title"></h1>
+          <span class="now-active" data-hook="now-active">Now Active</span>
```
pdehaan: Will It Localize? Or do we need some fancy `data-l10n-id=""` spice on it?

meandavejustice: Ahh yes, thank you @pdehaan

PR #392, opened 2016-02-05
Model content localization

lmorchard:
- Add django-hvad 1.4.0 dependency
- Migrate existing model data to en-US translations
- Tweaks to get admin & API working with hvad
- Use django-mozilla-product-details to get supported locales, with some
  settings boilerplate stolen from mozilla/bedrock
- Revise demo_data.json for translations
- Remove natural key from ExperimentDetail because it did not make much sense
- Define order for experiments created in test, because hvad mixes up
  the previously implicit order

Issue #130

### DIFF
--- a/fixtures/demo_data.json
+++ b/fixtures/demo_data.json
@@ -119,129 +119,6 @@
   }
 },
 {
-  "model": "experiments.experiment",
-  "fields": {
-    "measurements": "- Uses per session\r\n- Index selected\r\n- Percentage of searches\r\n",
-    "version": "5.1.4",
-    "privacy_notice_url": "",
-    "contribute_url": "https://github.com/mozilla/universal-search-addon/blob/master/CONTRIBUTING.md",
-    "title": "Universal Search",
-    "measurements_markup_type": "markdown",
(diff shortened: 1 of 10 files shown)


lmorchard:
Putting this out there to finally get some eyes on it. This adds [django-hvad](https://github.com/KristianOellegaard/django-hvad) as a way of integrating translations into database content, which also works pretty well with the django-rest-framework API. There should be new language tabs at the top of the admin pages, and you can add new locales by creating directories under `locales/`

Some further important things to note about reviewing this PR:
- Because this adds new python dependencies, it needs a `docker-compose build` after any branch change from `master` to this PR and back.
- You'll also have to run migrations, but you'll lose data on the way back to master:
  - From `master` to this PR: 
    
      `docker exec ideatown_server_1 ./manage.py migrate`
  - From this PR back to `master`: 
    
      `docker exec ideatown_server_1 ./manage.py migrate experiments 0009`
- You should be able to reload demo data on either branch:
  - `docker exec ideatown_server_1 ./manage.py loaddata fixtures/demo_data.json`

meandavejustice:
:+1:

PR #400, opened 2016-02-10
Implement ui for addon feedback door hanger

### LINKED ISSUE #221: Implement Door Hanger and Feedback UI
Main page https://invis.io/8Z48QH33W
Net promoter page https://invis.io/WD4V3AJ5H

So the second page here should pop up at random somewhat. I'll figure out what random should mean here and add it to the bug.

meandavejustice:
- bump addon version
- fresh build of addon
- remove old addon build
- add ui in data/feeback.html
- remove call to /api/me endpoint in addon
- remove handling of user.id from addon object in metrics endpoint
- rename/refactor ping-server to auth-request
- add call to metrics for net promoter score feedback
- add panel.js, panel.css
- add mustache.js dependency for panel templating
- add panel templates in lib/templates.js
- fix typo in docs/DOCKER-README.md
- fixes #221

todo: 
- [ ] retina version for all new images
- [ ] get rid of scroll bars in popup

May need some tweaking from @johngruen. Still having a bit of an issue where a scroll bar is showing up within the popup.

wanna get some more :eyes: on this though.


lmorchard:
Needs a rebase since the project rename landed

lmorchard:
Hmm, this doesn't seem to be working for me.

I ran an `npm run watch` in the addon directory on this branch. That posted an XPI to my running firefox as expected, I see that 0.3.4 is installed in `about:addons`, but the addon toolbar button disappeared. Saw these messages in the browser console:

```
Could not read chrome manifest 'file:///Applications/FirefoxDeveloperEdition.app/Contents/Resources/chrome.manifest'.
testpilot-addon:Object { stack: "", 5 more… } runner.js:166
testpilot-addon:Object { stack: "", 5 more… } Promise-backend.js:936
```

lmorchard:
Ahh, I see, those messages are exceptions:

```
Module `./lib/templates` is not found at resource://testpilot-addon/lib/templates.js
```

There's no `addon/lib/templates` in this commit... maybe something left out of the check-in?

meandavejustice:
@lmorchard looks like that file got ignored somehow. It's added back in now :)

lmorchard:
Maybe some more files missing? Now I'm getting this exception:

```
Module `./lib/auth-request` is not found at resource://testpilot-addon/lib/auth-request.js
```

meandavejustice:
@lmorchard This should be working now

lmorchard:
Okay, build & installed successfully this time. I get a really short panel when I click the toolbar button, which eventually gets populated with experiments. Not totally sure what this _should_ be doing... are there mockups somewhere?

addon/package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "lint": "eslint .",
     "start": "jpm run -b /usr/bin/firefox --prefs dev-prefs.json",
+    "watch": "jpm watchpost --post-url http://localhost:8888",
```
lmorchard: It's a small thing, but this should probably be split off into a separate PR because it's not directly related to the task in this PR.

package.json
```diff
@@ -75,6 +75,7 @@
   "scripts": {
     "build": "gulp build",
     "lint": "gulp lint sass-lint",
+    "package": "cd addon && npm run package",
```
lmorchard: Same as the change for package.json for the addon - not directly related to this PR.

addon/lib/auth-request.js
```diff
@@ -19,36 +20,26 @@ function pingServer(config, title, data, addon) {
...
-        onComplete: function(resp) { // eslint-disable-line no-unused-vars
-          // console.error(resp);
-        }
-      }).post();
-    });
+    cb(headers);
   });
 }
 
-function getMe(headers, BASE_URL, cb) {
```
lmorchard, line 40: Why remove the call to `/api/me`? FWIW, I'm probably going to need it / move it to another spot for feature flipping, since `/api/me` contains the list of all installed experiments and those installations will list feature flags.

addon/lib/auth-request.js
```diff
@@ -19,36 +20,26 @@ function pingServer(config, title, data, addon) {
...
+      headers: headers,
+      onComplete: function(resp) { // eslint-disable-line no-unused-vars
+        // console.error(resp);
+      }
+    }).post();
+  });
 }
 
-function formatEvent(PREFIX, title, id, data, addon) {
+function formatEvent(PREFIX, title, data, addon) {
```
lmorchard, line 40: Hmm, though yeah we probably don't need the user ID here, since we could just pick that up on the server side if we really wanted it from the authenticated user.

lmorchard:
If I scroll through the panel, I can see all experiments and survey links for each... should I be seeing _all_ experiments, or just the ones I have installed?

addon/lib/auth-request.js
meandavejustice, line 40: We aren't using it anymore in this pr. How do you feel about adding it back in once we need it for feature flipping? That way we don't have unused code in here

addon/package.json
meandavejustice: I'll separate this out.

package.json
meandavejustice: I'll separate this out also :)

meandavejustice:
@lmorchard Mockups are in the issue #221. You should be seeing all of the experiments, but the installed/active experiments should have an indicator. It looks like my patch for that got lost in a rebase or something so I'll fix that up today.

lmorchard:
Needs a rebase against master to get tests green again. Beyond that, I think I'm fine with re-adding the request to `/api/me` when we get to it in feature flipping. And, things look like the mockups, so... what are the next steps here?

meandavejustice:
@lmorchard We should be able to just merge this. I will create issues for the two remaining todos.

addon/index.js
```diff
@@ -116,15 +129,66 @@ function setupApp() {
...
+});
+
+panel.port.on('launch-feedback', id => {
+  panel.port.emit('show', getFeedbackForm(id));
+});
+
+panel.port.on('feedback-submit', dataStr => {
+  const data = JSON.parse(dataStr);
+  data.tags = ['main-addon'];
+  authRequest.sendMetric(settings, 'user-feedback', data,
```
lmorchard, line 165: Also, we probably want to revisit this. Not sure we want to send feedback as a metric ping - we have [`/api/feedback`](https://github.com/mozilla/testpilot/blob/master/testpilot/experiments/views.py#L89) that we use with [the feedback form view](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/views/feedback-view.js#L81) on the main site. That API itself could switch to metrics pings at some point, but we should probably use the same thing to keep feedback responses together.

lmorchard:
Going to merge this as-is for now, and open a new issue to work out how we want to handle feedback in general.

PR #412, opened 2016-02-12
add circle.yml

clouserw:
Adding a circle.yml.  @mostlygeek -- this is unchanged because your default looked fine to me.  Can you add the values to circleci?  also, r? @mostlygeek :)  

This also says it only builds one container per day.  I wonder if that will be a problem... /me will think it through

/cc @lmorchard @meandavejustice

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,54 @@
+# These environment variables must be set in CircleCI UI
+#
+# DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
+# DOCKER_EMAIL   - login info for docker hub
+# DOCKER_USER
+# DOCKER_PASS
+#
+machine:
+  services:
+    - docker
+
+dependencies:
+  # make sure to keep the docker cache dir
+  cache_directories:
+    - "~/docker"
+
+  override:
+    - docker info
+    - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
+    - "[ ! -z $DOCKERHUB_REPO ]"
+
+    # build the container, use circleci's docker cache workaround
+    # only use 1 image per day to keep the cache size from getting
+    # too big and slowing down the build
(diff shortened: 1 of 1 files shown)


lmorchard:
Haven't read up on how circle.yml works, but there is one wrinkle to building our Docker container. We have to build the frontend client code first before rolling up the rest of the Django app into a container with dependencies. To simplify things (for a certain definition of simple), we build _another_ throwaway Docker container with gulp & node as a utility to do that.

[Here's the script I was using with Travis to build a container to push to Docker Hub](https://github.com/mozilla/testpilot/blob/master/bin/build-prod.sh)

circle.yml
```diff
@@ -0,0 +1,54 @@
...
+    - "[ ! -z $DOCKERHUB_REPO ]"
+
+    # build the container, use circleci's docker cache workaround
+    # only use 1 image per day to keep the cache size from getting
+    # too big and slowing down the build
+    - I="image-$(date +%j).tar"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; docker load -i ~/docker/$I; fi
+
+    # create a version.json
+    - printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s"}\n' "$CIRCLE_SHA1" "$CIRCLE_TAG" "$CIRCLE_PROJECT_USERNAME" "$CIRCLE_PROJECT_REPONAME" > version.json
+    - docker build -t app:build .
```
mostlygeek, line 29: this should probably be changed to @lmorchard's build script or something similar to it. I like the convention of tagging the container as `app:build` and then re-tagging it later in the `deployment` section.

circle.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    # Clean up any old images and save the new one
+    - I="image-$(date +%j).tar"; mkdir -p ~/docker; rm ~/docker/*; docker save app:build > ~/docker/$I; ls -l ~/docker
+
+test:
+  override:
+    - docker run app:build test
```
mostlygeek, line 36: This should match what we have in the `.travis.yml` file, ie: running the `tox` tests.

clouserw:
Closing in favor of #413

PR #413, opened 2016-02-13
add circle.yml to build docker containers

mostlygeek:
This circle.yml replicates the asset building container and tests in the travis build.

PR #420, opened 2016-02-18
Add CircleCI

mostlygeek:
Same as #413 but works and cleaner.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -24,7 +24,7 @@ dependencies:
     # build the container, use circleci's docker cache workaround
     # only use 1 image per day to keep the cache size from getting
     # too big and slowing down the build
-    - I="image-$(date +%j).tar"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; docker load -i ~/docker/$I; fi
+    - I="image-$(date +%j).tgz"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; gunzip -c ~/docker/$I | docker load; fi
 
     # create a version.json
     - printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s"}\n' "$CIRCLE_SHA1" "$CIRCLE_TAG" "$CIRCLE_PROJECT_USERNAME" "$CIRCLE_PROJECT_REPONAME" > version.json
@@ -33,13 +33,15 @@ dependencies:
     - npm install
     - npm run lint
     - npm run build
-    - npm test
+    # seems to be unstable on circle, commenting out
+    # until we can investigate more
+    #- npm test
 
     # build the actual deployment container
     - docker build -t app:build .
 
     # Clean up any old images and save the new one
-    - I="image-$(date +%j).tar"; mkdir -p ~/docker; rm ~/docker/*; docker save app:build > ~/docker/$I; ls -l ~/docker
+    - I="image-$(date +%j).tgz"; mkdir -p ~/docker; rm ~/docker/*; docker save app:build | gzip -c > ~/docker/$I; ls -l ~/docker
 
 test:
   override:
--- a/circle.yml
+++ b/circle.yml
@@ -44,7 +44,7 @@ dependencies:
 test:
   override:
     - >
-        docker run -it --rm --net=host 
+        docker run --net=host
         -e 'DEBUG=False' 
         -e 'SECRET_KEY=Foo' 
         -e 'ALLOWED_HOSTS=localhost' 
--- a/circle.yml
+++ b/circle.yml
@@ -24,6 +24,11 @@ dependencies:
 
   override:
     - docker info
+    - node --version
+    - npm --version
+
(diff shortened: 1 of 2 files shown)


mostlygeek:
@lmorchard take a look at this one. ~~It works (mostly). It should build and ship the container. I ran into some issues with `npm test` and the frontend testing. It [worked](https://circleci.com/gh/mostlygeek/testpilot/4) once and then it [didn't](https://circleci.com/gh/mostlygeek/testpilot/7).~~

With commit f3e8282 frontend is now tested in circle before static assets are built into the production container. Tests within the container also pass.

circle.yml
```diff
@@ -0,0 +1,72 @@
...
+    # build the actual deployment container
+    - docker build -t app:build .
+
+    # Clean up any old images and save the new one
+    - I="image-$(date +%j).tgz"; mkdir -p ~/docker; rm ~/docker/*; docker save app:build | gzip -c > ~/docker/$I; ls -l ~/docker
+
+test:
+  override:
+    - >
+        docker run -it --rm --net=host
```
lmorchard: FWIW, with the `--rm` here, I get permission errors on CircleCI:

```
Failed to destroy btrfs snapshot: operation not permitted
```

Also probably don't need the `-it` because that gives you an interactive terminal, and that doesn't get used here

package.json
```diff
@@ -29,7 +29,7 @@
   },
   "devDependencies": {
     "babel": "^5.8.21",
-    "babel-eslint": "^4.1.3",
+    "babel-eslint": "^4.1.8",
```
lmorchard, line 32: Is there a reason to update this version? (ie. is there a bug it fixes that we should know about that affects us beyond CircleCI?)

circle.yml
mostlygeek: Yes I was seeing that too. Good point I'll remove those options. I think I put them in due to muscle memory.

package.json
mostlygeek, line 32: There was two issues I ran into that this fixed:

```
> testpilot@1.0.0 test:client /home/ubuntu/testpilot
> browserify -t babelify testpilot/frontend/static-src/test/**/*.js | testling | faucet
⨯ Error: ReferenceError: Can't find variable: Promise on line 28250
```

I couldn't quite find the source of this bug but searching points in the direction of an ES6 native promise bug. 

Also I ran into [this bug](https://github.com/babel/babel-eslint/issues/243) on previous tests. Updating `bable-eslint` seemed to have resolved both of them.

lmorchard:
Other than those two nits, I'm all :+1: :+1: :+1: to merge this and continue tweaking as needed. 

Thinking we should probably have a follow up issue to move toward shutting off Travis and just doing everything through CircleCI in the near future so we don't have redundant testing services

package.json
mostlygeek, line 32: The tests all seem to pass on travis but break on circle. So it is definitely a circle-ci issue, maybe with the xvfb implementation.

lmorchard:
Cool, that all looks good to me... Once these checks finish, I think I'll just hit the merge button!

mostlygeek:
Sadness. It seems that updating the library didn't fix the `Error: ReferenceError: Can't find variable: Promise on line 28250` bug. I'll have to dig deep...

lmorchard:
Huh, weird... @meandavejustice do you happen to have an idea why that might happen with the frontend tests? (i'm not 100% on how the frontend tests are working, myself. need to do some reading)

lmorchard:
I wonder if there's a bit of a race condition here? On Travis, it looks like we have a 3-second wait before starting the frontend tests:

```
npm run lint
npm run build
sh -e /etc/init.d/xvfb start
sleep 3
npm test
```

Not sure how / when CircleCI starts up xvfb

pdehaan:
Weird. IIRC, Promise isn't native in Node 0.12.x, but I thought we were using the [**es6-promise**](https://github.com/stefanpenner/es6-promise) Promise polyfill.

meandavejustice:
@pdehaan I believe our promise support is included in babel. I'll do some digging into this

lmorchard:
FWIW, this actually pretty easy to get running on your own fork in CircleCI and you can ssh into the box. But, that said, I've been able to reproduce this failure on my own, so far :/

mostlygeek:
I wonder if this is an issue with CircleCI. Tests seem to pass/fail randomly on their system. I'll reach out to them and ask as well.

mostlygeek:
I added a few tweaks and the frontend tests seem to pass more consistently. I honestly do not know if adding the pre dependency with `nvm` is what did it. 

I rebuilt the tests a few times using the CircleCI cache and without and it has passed each time. I think we're good to merge this beast now.

lmorchard:
Did a quick rebase & squash, and merged 87b7c6eb9915f69976b5bc0701509c47409bc66a

PR #422, opened 2016-02-19
Foundations for feature flipping

### LINKED ISSUE #366: Feature flipping support for experiment addons
A notion for testing features in an experimental addon is to ship lots of alternate code branches and flip conditionally between them. We've done this a lot on Django sites at Mozilla with [django-waffle](https://github.com/jsocol/django-waffle).

What might be interesting is to have experiment addons talk to the Idea Town addon (e.g. via nsIObserverService) to obtain feature flipping configuration data - which in turn can come from the Idea Town server based on the authenticated user. This could let us do things like A/B testing, incremental roll out, etc.

Breaking this out into subtasks off the top of my head:
- [ ] Data model for features defined for a given experiment.
- [ ] Data model for feature flags for a particular user experiment installation
- [ ] Server-side API resource for a user's experiment feature flags
- [ ] Admin integration for managing features, both for defining features and for individual user/experiment flag state
- [ ] Any logic needed to come up with the feature flags for a particular user installation (e.g. segmentation, incremental rollout, etc. [django-waffle](https://github.com/jsocol/django-waffle) could be an inspiration)
- [ ] Metrics integration so that we can measure things based on the feature state for any given user
- [ ] Example experiment addon code to show how to fetch feature flag data from the main addon and how to use it in conditionally enabling features

lmorchard:
- Models for features, conditions, and per-installation state
- New feature flag state API resource at
  /experiments/{experiment_id}/installations/{client_id}/features
- Per-installation feature state added to UserInstallation serializer
- Extensible registry of condition operators
- Admin for features and conditions
- Migrations and tests

Closes #366

### DIFF
--- a/testpilot/experiments/admin.py
+++ b/testpilot/experiments/admin.py
@@ -3,13 +3,30 @@ from django.contrib import admin
 from hvad.admin import TranslatableAdmin, TranslatableTabularInline
 
 from .models import (Experiment, ExperimentDetail, UserInstallation,
-                     UserFeedback)
+                     UserFeedback, Feature, FeatureCondition)
 from ..utils import (show_image, parent_link, related_changelist_link,
                      translated)
 
 
 class ExperimentDetailInline(TranslatableTabularInline):
     model = ExperimentDetail
+    extra = 1
+
+
+class FeatureInline(admin.TabularInline):
+    model = Feature
+    extra = 1
+
+
+class FeatureConditionInline(admin.TabularInline):
+    model = FeatureCondition
+    extra = 1
+
+    def formfield_for_choice_field(self, db_field, request, **kwargs):
+        if db_field.name == 'operator':
+            kwargs['choices'] = FeatureCondition.objects.get_operator_choices()
+        return (super(FeatureConditionInline, self)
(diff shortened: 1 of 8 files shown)


lmorchard:
Okay, here's some server-side groundwork for feature flipping. It adds models for defining per-experiment features, programmatic conditions to enable features, and per-installation feature state. This stuff can be set up for experiments in admin, and there are new API resources for fetching state per installation.

Still to come: Code in the main site addon to fetch feature flag state and nsiObserver support to pass that state along to experiment addons. Thinking I might have a follow up PR with example experiment addon code showing how to use it. Oh, and we should work feature state into metrics pings. Will probably split all those things out of #366 as folow-up issues

But, this server side stuff looks like a big enough chunk of work to start looking through

lmorchard:
Also, for context: I tried adapting [django-waffle](https://github.com/jsocol/django-waffle) for this stuff, but couldn't quite get it to fit the per-experiment / per-installation model the way I wanted. Also looked into [gutter](https://github.com/disqus/gutter) as a generic python feature switch toolkit, but it has no python 3 support yet and didn't quite work here either

PR #423, opened 2016-02-22
Move our docs to /docs

clouserw:
We might want to de-dupe and clean up the Development part of the main README where it talks a bunch about Docker, but this will get the file system structure correct at least.

I didn't move TESTING, CODE_OF_CONDUCT, or CONTRIBUTING.  I know github expects the last one on the root, and I think our standard is for CODE_OF_CONDUCT to be there as well.


meandavejustice:
:+1:

PR #425, opened 2016-02-22
add convenience script for addon dev

meandavejustice:
- add `watch` command for addon watch/building to addon/package.json

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -18,7 +18,8 @@
   "scripts": {
     "lint": "eslint .",
     "start": "jpm run -b /usr/bin/firefox --prefs dev-prefs.json",
-    "package": "jpm xpi && mv @testpilot-addon-$npm_package_version.xpi dist/testpilot-addon-$npm_package_version.xpi"
+    "package": "jpm xpi && mv @testpilot-addon-$npm_package_version.xpi dist/testpilot-addon-$npm_package_version.xpi",
+    "watch": "jpm watchpost --post-url http://localhost:8888"
   },
   "license": "MPL-2.0",
   "devDependencies": {


package.json
```diff
@@ -75,6 +75,7 @@
   "scripts": {
     "build": "gulp build",
     "lint": "gulp lint sass-lint",
+    "package": "cd addon && npm run package",
```
pdehaan: I'm guessing this would only work if you did a `$ cd addon && npm install` first though, right?
I think I'd see an error if I just did this from a fresh `git clone`:

``` sh
npm install && npm run package
```

Of course, I should probably verify that theory, because there is nothing worse than some drive-by QA person starting two sentences with "I'm guessing" and "I think" like the lazy jerk they are... #thisguy

meandavejustice: hmm, yeah I guess this assumes that `npm install` has already been run. What do you recommend for this case?

pdehaan: Not sure of the best solution. If the command is only intended for local development, it probably isn't a big deal since we'd know that we need to run `npm install` in both places. If we're trying to be a bit more friendly, we could possibly do something a bit hackier like:

``` js
"package": "cd addon/ && npm install && npm run package",
```

Or, we could potentially do some super hacky-ness and add a "postinstall" solution to the root /package.json which does the CD into ./addon/ and runs `npm install`. But maybe that's getting too meta. But we have used that trick a few times in the past to auto-install Bower deps after you run `npm install`.

``` js
"package": "cd addon/ && npm run package",
"postinstall": "cd addon/ && npm install",
```

Or maybe even add a "prepackage" task which does `cd addon/ && npm install`, so it only attempts to install those deps if you specifically run `npm run package`.

But I'm definitely open to other suggestions from 6a68/Chuck/Les/Wil/anybody.

meandavejustice:
This is only failing because of the docs building stuff, which is removed in the master branch now.

package.json
lmorchard: Or, FWIW, you might even leave it out of the root `package.json` altogether... it's a little redundant with the same command in `addon/package.json`

meandavejustice: yeah, I'll just get rid of that one

PR #427, opened 2016-02-23
Add operations endpoints

### LINKED ISSUE #394: Add operations endpoints
The operations infrastructure is expecting some endpoints for each project to reply in a standardized fashion.  This will be expected for all Test Pilot experiments as well as Test Pilot itself.  They are handy for debugging too!
1. Respond to `/__version__` with a [JSON version object](https://docs.google.com/document/d/1rGVyiLYvZyKE2oHcSVx-vBmQRKhs1kLLgn7xeCs6qKs/edit)
2. Respond to `/__heartbeat__` with a HTTP 200 or 5xx on error
   
   this endpoint is used by external monitoring to check the health of the service
   it is recommended to check the app's dependencies (database, caches, etc) here
3. Respond to `/__lbheartbeat__` with an HTTP 200
   
   used by the load balancer to check if the server and application is OK
   do not include dependency checks as this check may trigger automatic node termination and replacement

lmorchard:
Add operations endpoints
- /__version__ endpoint
  - Responds with contents of version.json when available - i.e. when
    generated by circle.yml on Docker container build
  - Responds with canned fallback data when version.json is missing -
    i.e. in development
- /__lbheartbeat__ endpoint
  - responds with 200 OK when the Django app is available
- /__lbheartbeat__ endpoint
  - checks for database availability, responds with 500 Internal Server
    Error if there's a problem
  - responds with 200 OK otherwise
- Tests

Fixes #394

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ testpilot/frontend/static-src/vendor/
 testpilot/frontend/static-src/styles/vendor/
 docker-compose-s3.yml
 debug-config.json
+version.json
--- a/testpilot/base/tests.py
+++ b/testpilot/base/tests.py
@@ -1,0 +1,69 @@
+from unittest.mock import patch
+
+import io
+import json
+
+from django.core.urlresolvers import reverse
+from django.db import OperationalError
+
+from ..utils import TestCase
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+class OpsEndpointTests(TestCase):
+
+    def test_ops_heartbeat_ok(self):
+        """/__heartbeat__ should respond with 200 OK"""
+        url = reverse('ops-heartbeat')
+        self.assertEqual('/__heartbeat__', url)
+        resp = self.client.get(url)
+        self.assertEqual(200, resp.status_code)
+
+    def test_ops_heartbeat_db_failure(self):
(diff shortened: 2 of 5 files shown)


lmorchard:
So... this should probably be work-in-progress because the two heartbeat endpoints don't really do anything. But, I'm not sure what to make them do. The [dockerflow spec](https://github.com/mozilla-services/Dockerflow#a-containerized-app-must) says this:
- Respond to `/__heartbeat__` with a HTTP 200 or 5xx on error. This should dependent services like the database to also ensure they are healthy.
- Respond to `/__lbheartbeat__` with an HTTP 200. This is for load balancer checks and should not check any dependent services.

But, I think we'll respond with 500 on _any_ route if there's a database connection issue. So, that's `/__heartbeat__`. But, I'm not sure if we can _not_ respond with a 500 if Django's okay but not the database, which seems to be what `/__lbheartbeat__` wants. 

Will keep looking into that stuff, but ideas would be handy. `/__version__` works, at least.

lmorchard:
Actually, never mind all that. Just tried killing postgresql locally, and `__lbheartbeat__` still works. If I add a little code to select a count of experiments, I can make `__heartbeat__` respond with an error. Updated code coming in just a few...

meandavejustice:
Thanks for taking care of this @lmorchard

PR #431, opened 2016-02-25
Implement JSON log formatter for Cloud Services

### LINKED ISSUE #401: Change logging output to conform to Dockerflow
The [Dockerflow repo](https://github.com/mozilla-services/Dockerflow) says this about logging:

> Applications logs should be written to stdout and stderr in the [mozlog format](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=42895640). These will be sent to an Elastic Search cluster and accessible via Kibana. This feature is enabled on request. 

So, we need to change the logging setup in the Docker container to:
- [ ] Output to stdout & stderr
- [ ] Use the [mozlog format](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=42895640)

lmorchard:
- testpilot.base.logging.JsonLogFormatter class to produce JSON format
  https://mana.mozilla.org/wiki/display/CLOUDSERVICES/Logging+Standard
- Use "json" machine-readable format by default for prod
- Use "simple" text log format in docker-compose-base.yml for dev
- Remove all file-based logging and email alerts on errors, everything
  should be STDOUT
- Tests, with addition of testfixtures package

Closes #401


lmorchard:
@mostlygeek - You might want to peek at this, see if it properly meets the format spec. I _think_ it does, at least from my reading of the mana page. But, sometimes I'm not to be trusted

lmorchard:
(and if this _does_ meet the spec, I might file a follow up issue to extract this as a standalone package so other projects can use it)

mostlygeek:
@whd can you take a look and see if it meets the current heka mozlog spec? It looks good to me. 

@lmorchard yes, this would be great to have as a separate library other projects can pull into their python/django apps!

testpilot/base/logging.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def format(self, record):
+        """
+        Map from Python LogRecord attributes to JSON log format fields
+
+        * from - https://docs.python.org/3/library/logging.html#logrecord-attributes
+        * to - https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=42895640
+        """
+        out = dict(
+            Timestamp=record.created,
```
whd: The precision here needs to be ns (*1e9).

whd:
This looks like it should work, assuming LogRecords don't have any nested dicts in their attributes (looks like they don't). Using a json-schema validation library against the output and our schema would be a good way of verifying this passes validation.

lmorchard:
Updated commit to convert timestamp from fractional seconds to integer count of nanoseconds. Will look into running a JSON schema validator

lmorchard:
Updated the commit to include JSON schema validation of log messages in testing. Beyond actually seeing this in action on a production deploy, I think this is probably good to go unless there are any other loose ends?

lmorchard:
Updated commit to rebase against new master

lmorchard:
Updated again to fix a conflict

PR #432, opened 2016-02-26
Run app as non-root user in Docker container

### LINKED ISSUE #398: Rework server Docker container to run under non-root user
[The Dockerflow deployment pipeline from ops](https://github.com/mozilla-services/Dockerflow#dockerfile-guidelines) says this:

> - Set a non-root user and group app in the container and run everything as this user
>   - make the uid for the user 10001
>   - make the guid for the group 10001

But, our Docker container runs everything as root. This should be reworked.

lmorchard:
- Create "app" user with uid 10001, gid 10001
- Run app in docker container as USER app
- Move `collectstatic` command from `run-common.sh` to Dockerfile, so
  file modifications run as root and not lower-privilege app user
- Remove all file-based logging, because app user will not have
  permission to write files
- Escalate to root user in docker-compose for local dev server instance,
  so we can write uploads to the /app/media directory

Closes #398

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,3 +16,4 @@ LICENSE
 README.md
 .git
 .gitignore
+testpilot/frontend/static-src
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,9 @@ CMD ["./bin/run-prod.sh"]
 
 WORKDIR /app
 
+RUN groupadd --gid 10001 app && \
+    useradd --uid 10001 --gid 10001 --shell /usr/sbin/nologin app
+
 # Install & cache dependencies for Django/Python using peep with a supported
 # pinned version of pip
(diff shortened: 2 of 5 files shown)


lmorchard:
One thing I don't like in this PR is the hack to escalate from app user to root in local dev. The only reason is so that the local `media/` directory can be written to by the Django app when files are uploaded. That only seems to work for root, because the perms are readonly for the app user.

I might poke at that some more, though FWIW it doesn't actually affect production deployment. Prod should get a container that's a read-only filesystem for the app user, and file uploads should be configured to go to Amazon S3.

lmorchard:
<strike>Hmm, actually, a solution might be to advise devs to `mkdir media; chmod a+rw media` to be extra permissive.</strike> Nope, that didn't work on my machine.

meandavejustice:
lgtm

PR #435, opened 2016-02-27
Add contribute.json (Fixes #389)

### LINKED ISSUE #389: add contribute.json
https://www.contributejson.org/

[example](https://github.com/mozilla/universal-search-recommendation/pull/29)

robhudson:


testpilot/frontend/urls.py
```diff
@@ -1,8 +1,10 @@
-from django.conf.urls import url, patterns
+from django.conf.urls import patterns, url
 
 from . import views
 
+
 urlpatterns = patterns(
     '',
-    url(r'(?P<url>.*)', views.index, name="frontend_index")
+    url(r'^contribute\.json$', views.contribute_json, name='contribute_json'),
```
lmorchard: Might be better to add this to [`base/urls.py`](https://github.com/mozilla/testpilot/blob/master/testpilot/base/urls.py), where we've accumulated some other general project views. The `frontend` app has so far been mainly for the ampersand-based client frontend

testpilot/frontend/views.py
```diff
@@ -1,6 +1,12 @@
...
 from django.shortcuts import render
+from django.views import static
 
 
 def index(request, url=''):
     return render(request, 'testpilot/frontend/index.html',
                   {'user_id': request.user.email if not request.user.is_anonymous() else ''})
+
+
+def contribute_json(request):
```
lmorchard: Same thing, here - [`base/views.py`](https://github.com/mozilla/testpilot/blob/master/testpilot/base/views.py) might be a better home

testpilot/frontend/urls.py
robhudson: Ok, thanks. My eyes jumped right over "base" when I was looking for reasonable place to put this.

contribute.json
```diff
@@ -0,0 +1,30 @@
...
+        "django",
+        "docker",
+        "elasticsearch",
+        "javascript",
+        "jinja2",
+        "postgresql",
+        "python",
+        "redis"
+    ]
+}
```
robhudson, line 29: If there's any changes you'd like here let me know.

robhudson:
Updated to move urls/views to `base/`.

contribute.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "bugs": {
+        "list": "https://github.com/mozilla/testpilot/issues",
+        "report": "https://github.com/mozilla/testpilot/issues/new"
+    },
+    "keywords": [
+        "css",
+        "django",
+        "docker",
+        "elasticsearch",
```
lmorchard: Hmm, since you mention it... we're not actually using elasticsearch (yet?)

Other than that, I think this PR is r+ from me - thanks so much for doing it!

robhudson: Ok. Removing elasticsearch and pushing an update.

The schema also allows for a `urls` key for dev, prod, or stage... but I wasn't sure if there are any yet?

robhudson:
The schema says that these are required: name, description, repository. So I think I'll add a simple test to make sure this response with a 200 and is valid json and at least has those fields.

robhudson:
Added basic tests.

robhudson:
BTW, I'm assuming here that since I'm using Django's `static` that this will be served via Whitenoise and appropriate caching headers applied. I did check locally that subsequent requests have status code 304 and the last modified header is set.

lmorchard:
Works for me! We can circle back and tweak this if it doesn't work as expected with Whitenoise

PR #436, opened 2016-03-01
Packaging and Distributing the addon

### LINKED ISSUE #92: handle build and serving of addon

meandavejustice:
- create addon/sign script to handle signing and building addon
  if addon has already been signed we download it from amo
- bump jpm version
- update addon/README with dist instructions
- add proper keys to addon/package.json create update.rdf
- minimal server changes for hosting the addon and update.rdf files
- add `addon` task to gulpfile for moving into the static DIR
- fixes #92


lmorchard:
Skimming the changes and things look good. Going to try running this through its paces and see if I can produce a signed addon with my API keys. Will merge after that.

Thinking it should be pretty easy in a follow up PR to integrate this stuff into our CircleCI build.

addon/bin/sign
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+const version = require('../package.json').version;
+const resolve = require('path').resolve;
+const xpiPath = resolve(`dist/testpilot-addon-${version}.xpi`);
```
lmorchard: I tried `npm run dist` and got this error:

```
/Users/lmorchard/devel/testpilot/addon/bin/sign:4
const xpiPath = resolve(`dist/testpilot-addon-${version}.xpi`);
                        ^
SyntaxError: Unexpected token ILLEGAL
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
```

Looks like ES6 features aren't working here. FWIW, my host OS X machine and testpilot_frontend_watcher_1 in Docker are on node 0.12.7

lmorchard: Some more things:
- Looks like the addon moved from `dist/testpilot-addon...` to just `dist/addon.xpi`, so we don't need the version in the string here
- I think the `jpm` version in `package.json` needs an upgrade. The version I have (1.0.3) doesn't have `sign` as a command

Upgrading my local `jpm` and running the command manually gets me this far:

```
$ jpm sign --api-key $AMO_USER --api-secret $AMO_SECRET --xpi dist/addon.xpi
JPM [info] Signing XPI: dist/addon.xpi
JPM [error] Server response: Version already exists. ( status: 409 )
JPM [info] FAIL
```

Do we also need to bump the version every time we want to sign it?

lmorchard:
> Do we also need to bump the version every time we want to sign it?

Ah, hmm, yup [looks like we do](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/jpm#jpm_sign):

> Your add-on at this exact version number already exists, so it was not signed. Increment the version number in your package.json file and run the command again.

That command not being idempotent will probably throw a monkey wrench into CircleCI automation, but that's beyond the scope of this PR.

addon/bin/sign
chuckharmston: If you want to use the version in the filename, I have a small `sed` command in our Makefile to extract the version number from install.rdf (assume specifically-formatted XML, which shouldn't be a big problem):

https://github.com/mozilla/universal-search/blob/master/Makefile#L3

meandavejustice: Ahh I didn't realize we were on that old of a node version. I'll pull the template strings out.

> Do we also need to bump the version every time we want to sign it?

yes. I tried to get a `--silent` option upstream but it doesn't look like that will happen.
I wrote a wrapper `bin/sign` in order to catch those errors so we could still build, but I'm thinking it may be better to just check the version instead of rebuilding each time.

I'll bump jpm version in the package.json as well.

lmorchard: > Ahh I didn't realize we were on that old of a node version. I'll pull the template strings out.

Yeah, the version [is pinned in the Dockerfile](https://github.com/mozilla/testpilot/blob/master/Dockerfile-frontend-build#L1) for the frontend. We should probably upgrade that, but not as a part of this PR.

I think the problem is worse than just silencing errors, because we do need a signed .xpi with every CircleCI build. So, I also [filed an issue](https://github.com/mozilla-jetpack/jpm/issues/505) on `jpm sign` and trying to use it repeatedly on the same unmodified version to get the same signature. When I get around to trying CircleCI integration, it sounds like I might need to drop to the HTTP API instead of using the CLI tool :disappointed:

meandavejustice:
I have opened an issue for testing around addon updates https://github.com/mozilla/testpilot/issues/446
and to integrate addon building into circle ci https://github.com/mozilla/testpilot/issues/447

meandavejustice:
Travis seems to be ignoring my environment variables, local testing is working out. Since we are ditching travis for circleci we can probably just ignore it for now. Especially since we have an issue for circleci integration now.

lmorchard:
> Travis seems to be ignoring my environment variables

I think it's Tox, not Travis. You have to explicitly list env vars to pass into one of the testenvs, like [@robhudson just did to get coveralls to work](https://github.com/meandavejustice/testpilot/commit/bfd9006bacf8f53fee52894b81947efbeb46f519). We'll probably run into that in CircleCI at some point, if we flesh out tests there eventually to run the whole Tox gauntlet

tox.ini
```diff
@@ -51,4 +52,4 @@ commands =
     npm config set spin false
     npm install
     npm run lint
-    npm run package
+    npm run sign
```
lmorchard, line 55: Hmm, looks like `npm run sign` works on my dev machine, but even with the `passenv` line in `tox.ini` it seems like the env vars aren't getting seen.

Maybe revert this to `npm run package` for now? Packaging without signing is still useful for local dev, and will at least get us a sanity check on Travis as to whether the addon could be built.

lmorchard, line 55: That, or just comment out this step for now and we can follow up in #447 over on CircleCI. Ensuring `npm run lint` works is probably enough of a test on Travis for now

lmorchard, line 55: Actually, maybe I'll just go ahead and do that as I merge, since it's a one line change and this thing needs to get wrapped up!

meandavejustice, line 55: sounds good to me!

lmorchard:
Went ahead and merged, then added a commit to comment out the sign part - https://github.com/mozilla/testpilot/commit/da60deb049343ff154c2f4a81b9fc258d9cd8726

[Passed tests on my own branch in Travis](https://travis-ci.org/lmorchard/testpilot/jobs/114301355), so I'll say that's good enough

PR #449, opened 2016-03-03
Allow other domains for auto-activation (fixes #391)

### LINKED ISSUE #391: Allow other domains besides @mozilla.com for dev/stage
It will make our QA automated testing a lot easier if we can allow specific domains or accounts so QA doesn't have to try and manage LDAP passwords for @mozilla.com email addresses.

For other services we just create a bunch of @restmail.net accounts which are easy to create and check.

robhudson:


robhudson:
I think this is what #391 was asking for. I didn't go the regexp route b/c it didn't seem to add any value, but if you'd rather I can update it to use regexp.

lmorchard:
Yeah, this looks good enough for what was actually asked. Regex would be handy for supporting weird things, but we can also manually activate accounts whenever needed too. In the time it took to formulate with a fancy regex, we could probably search on accounts in admin and make them active.

PR #450, opened 2016-03-04
Fix coveralls reporting

robhudson:
I noticed coveralls reporting is not working. Looking at the [coveralls python lib docs](https://github.com/coagulant/coveralls-python) it has extra steps if using tox. I'm testing if the changes here actually fix the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
 services:
   - docker
 install:
-  - pip install tox coveralls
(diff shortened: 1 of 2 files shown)


robhudson:
:+1: This fixes coveralls reporting.  See: [coveralls build 679](https://coveralls.io/builds/5297855)

meandavejustice:
Thanks @robhudson !

PR #451, opened 2016-03-07
add build instructions for addon to circle.yml

### LINKED ISSUE #447: integrate addon build/signing into circle ci

meandavejustice:
- fixes #447

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,10 +1,12 @@
 # These environment variables must be set in CircleCI UI
 #
+# AMO_USER       - api user from addons.mozilla.com
+# AMO_SECRET     - api key secret from addons.mozilla.com
 # DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
 # DOCKER_EMAIL   - login info for docker hub
 # DOCKER_USER
 # DOCKER_PASS
-#
+
 machine:
   services:
     - docker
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -50,6 +52,15 @@ dependencies:
     # Clean up any old images and save the new one
     - I="image-$(date +%j).tgz"; mkdir -p ~/docker; rm ~/docker/*; docker save app:build | gzip -c > ~/docker/$I; ls -l ~/docker
 
+    # build addon
```
lmorchard: This stuff needs to go before [line 49](https://github.com/meandavejustice/testpilot/blob/447-circleci-addon/circle.yml#L49) where `docker build` happens, so the addon gets bundled into the container

lmorchard: Actually, strike that... I think it needs to go even before [line 43](https://github.com/meandavejustice/testpilot/blob/447-circleci-addon/circle.yml#L43), before the static asset build - because gulp copies it from static-src to static, right?

meandavejustice: Thanks les, I'll update

circle.yml
```diff
@@ -38,6 +40,14 @@ dependencies:
     # create a version.json
     - printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s"}\n' "$CIRCLE_SHA1" "$CIRCLE_TAG" "$CIRCLE_PROJECT_USERNAME" "$CIRCLE_PROJECT_REPONAME" > version.json
 
+    # build addon
+    - cd addon/
```
lmorchard: Something else I think I just learned after some googling of other folks' CircleCI usage: Each of these entries runs in the root directory of the project. It's not like a continuous shell script - each line is its own separate script. So, `cd` is undone right after you issue it.

Probably needs an ugly one-liner something more like `cd addon && npm config set spin false && npm install && npm run lint && npm run sign`

(The other commands should probably be joined into a few ugly one-liners, too, but luckily so far they all ran in the same directory)

meandavejustice: ugh, I've been trying to debug this for a while now. I didn't come across
that advice :) Thanks!

meandavejustice: it worked! @lmorchard

lmorchard:
\o/ :grinning:

Issue #234, opened 2015-11-11
Test Pilot Metrics Dashboard


johngruen:
What should this thing look like?

johngruen:
Here's a link to the working doc: https://docs.google.com/a/mozilla.com/document/d/1FxzBoA1ykMQNMpcgi8ZRR68sCs7lajAFnrwi9fGjKGs/edit?usp=sharing

Here's a mockup: 
https://invis.io/HB5BU02AP

johngruen:
https://wiki.mozilla.org/Test_Pilot/Metrics

clouserw:
RFC for draft plan @ https://docs.google.com/document/d/1D0QvPngp3Va7Q1AIR8Bp_4GYltszsYbUytvXw_gV8Ok/edit

clouserw:
Plan now lives at https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md

Bugs filed for telemetry hookups: https://bugzilla.mozilla.org/show_bug.cgi?id=1255184 and https://bugzilla.mozilla.org/show_bug.cgi?id=1255182

Next steps:  Document how tests will use our metrics pipeline.  File Issues for ensuring test pilot is hooked up to log in the formats detailed in the plan.  Ensure data is flowing into the pipeline and showing up in Redshift.  Build some visualizations from the data in Redshift (probably using Periscope).

chuckharmston:
> Next steps: Document how tests will use our metrics pipeline. File Issues for ensuring test pilot is hooked up to log in the formats detailed in the plan. Ensure data is flowing into the pipeline and showing up in Redshift. Build some visualizations from the data in Redshift (probably using Periscope).

Related: https://github.com/mozilla/universal-search/pull/79

chuckharmston:
> Next steps: Document how tests will use our metrics pipeline. File Issues for ensuring test pilot is hooked up to log in the formats detailed in the plan. Ensure data is flowing into the pipeline and showing up in Redshift. Build some visualizations from the data in Redshift (probably using Periscope).

We should also make sure that there is a client identifier that is passed on to Telemetry to allow calculations of retention rate and user type analysis. It should be a UUID derived from neither the test pilot user ID nor the FxA ID. Activity Stream is [doing something similar](https://github.com/mozilla/activity-streams/blob/master/data_dictionary.md) with their `client_id` property.

clouserw:
Telemetry has a unique ID

clouserw:
[...]

PR #464, opened 2016-03-08
Add README-METRICS (part of #234)

clouserw:
First version of our metrics doc.  Will have adjustments as we figure things out, but high level is correct.


docs/README-METRICS.md
```diff
@@ -0,0 +1,165 @@
...
+Data recorded by the log files will include standard system logging as well as:
+* Test Pilot User ID (if applicable)
+* Test ID (if applicable)
+* Feature Switch Set (if applicable)
+* User Agent
+* Campaign (marketing identifier) (if applicable)
+
+Using the [standard format][4] an example object will look like:
+
+```json
```
clouserw: @lmorchard I made this example JSON here which conforms to the spec.  I saw you already did the logging though -- is this the same format?

lmorchard: We have a Python logger that complies with that format in terms of the envelope / metadata, but we don't have anything yet logging that specific `request.summary` type. Probably need to write some Django middleware to do that.

Also, don't I think the `request.summary` type makes sense for these metric events. Seems like `request.summary` is meant to be the equivalent of an Apache `access.log` line generated on every request. 

For the specific metric events listed, we probably want to come up with our own separate type per event - e.g. `testpilot.newuser`, `testpilot.feedback`, `testpilot.main-install`, `testpilot.experiment-install`. Might also make some sense to make the `Fields` closely match whatever we put into Telemetry payloads in case we want to mingle the server / client streams.

docs/README-METRICS.md
```diff
@@ -0,0 +1,165 @@
...
+    "t": 256,
+
+    "test_id": "universal_search",
+    "feature_switches": "",
+    "campaign": ""
+  }
+}
+```
+
+The second source, **the client**, will use the existing Telemetry system (the
```
lmorchard, line 116: Still wrapping my head around how Telemetry works... I think we also want to declare / define our own "testpilot" ping that extends from the common format [like these other ones](https://gecko.readthedocs.org/en/latest/toolkit/components/telemetry/telemetry/index.html).

docs/README-METRICS.md
lmorchard: This is older stuff, but I think we want something like what Persona did. [They have 14 or so event types](https://github.com/mozilla/persona/blob/34c4c4c0ab191e4e94151fe4120d3730d3770a08/lib/logging/transports/filters/metrics-heka.js#L8), like `persona.signin`, `persona.complete_user_creation.success`, etc with `Fields` relevant to each event added when it's logged.

docs/README-METRICS.md
```diff
@@ -0,0 +1,165 @@
...
+
+Using the [standard format][4] an example object will look like:
+
+```json
+{
+  "Timestamp": 1457462527,
+  "Type": "request.summary",
+  "Logger": "TestPilotWeb",
+  "Hostname": "testpilot.servers.s3",
+  "EnvVersion": "2.0",  // We're conforming to the Fields spec
```
pdehaan, line 92: nit: JSON can't have comments (per the spec). So this will cause this red nonsense to appear in GitHub. We can change the code fence to use **```js** instead of **```json** and they should still render as expected.

Same w/ the block below on L133.

docs/README-METRICS.md
clouserw, line 116: I think that is done for us -- we define this json blob, and "they" write that page is my understanding...

docs/README-METRICS.md
clouserw: Good points.  I like the separate event types.  Let me see about putting those in here...

docs/README-METRICS.md
clouserw, line 92: will fix

docs/README-METRICS.md
```diff
@@ -0,0 +1,165 @@
...
+    * feature switch status
+  * User Agent
+  * Test Pilot User ID
+
+An example payload (within the full ping) would look like:
+```js
+{"tests":
+  {"universal_search":
+    {"last_enabled": 1457462200,
+     "last_disabled": 1457461100,
```
robhudson, line 136: This doesn't have the "state" as mentioned above in the docs. Unless it is assumed that it is derived from the latest timestamp?

PR #469, opened 2016-03-11
update add-on name

clouserw:
This is low priority but:

1) Changes name from "Test Pilot Addon" to "Test Pilot"
2) Updates description
3) Updates the home page now that we have one
4) Changes visible references to "addon" to "add-on" which is technically correct.  The best kind of correct.

### DIFF
--- a/addon/README.md
+++ b/addon/README.md
@@ -1,5 +1,5 @@
 # Test Pilot
-The addon where ideas come to idea
+The add-on where ideas come to idea
 
 ## installation
 
@@ -51,8 +51,8 @@ with the add-on installed.
 
 ## distributing
 
-We serve the addon from the `/static/addon/` directory. We will need
-to get the addon signed via [AMO](http://addons.mozilla.org/) and move
+We serve the add-on from the `/static/addon/` directory. We will need
+to get the add-on signed via [AMO](http://addons.mozilla.org/) and move
 it into the correct directory. This is all packaged into a script,
 `npm run sign` in the [package.json](./package.json).
 
@@ -117,9 +117,9 @@ The event `addon-install:install-ended` will include some extra properties:
 }
 ```
 
-#### Talking to the addon
+#### Talking to the add-on
 
-You will need to setup the following function (or an equivalent) to send messages to the addon.
+You will need to setup the following function (or an equivalent) to send messages to the add-on.
 
 ``` javascript
 function sendToAddon (data) {
--- a/addon/sign
+++ b/addon/sign
@@ -6,7 +6,7 @@
  */
 /*
  * This script is for signing, packaging, and distributing the testpilot
- * addon. If the current package version has already been signed, we generate
+ * add-on. If the current package version has already been signed, we generate
  * an update.rdf, download the signed xpi from amo, and move the files into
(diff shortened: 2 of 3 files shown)


addon/README.md
```diff
@@ -16,7 +16,7 @@ dev: [dev-prefs.json](dev-prefs.json)
 
 You may need to set both `xpinstall.signatures.required` and
 `xpinstall.whitelist.required` flags to false in `about:config`
-if the addon has not yet been signed.
+if the add-on has not yet been signed.
```
pdehaan: Do we want to remove this section now that the ~~addon~~ add-on is signed, or is it still relevant for local development?

addon/README.md
```diff
@@ -1,5 +1,5 @@
 # Test Pilot
-The addon where ideas come to idea
+The add-on where ideas come to idea
```
pdehaan, line 2: :+1: to the comedic historical tagline. But, not sure if we want to  update it to be more descriptive. Stealing from the package.json `description`:

> _Test Pilot is a privacy sensitive user research program focused on getting new features into Firefox faster._

addon/README.md
lmorchard: Still relevant for local development - we have to bump the version each and every time we want a new signed XPI, so that's totally unworkable for local dev.

So, we use Fx Dev Edition, turn off sig checking, and [run `jpm watchpost` via `npm run watch`](https://github.com/mozilla/testpilot/blob/master/addon/package.json#L24) along with the [Extension Auto-installer](https://addons.mozilla.org/en-US/firefox/addon/autoinstaller/).

meandavejustice:
@clouserw Could you rebase this with master when you get a free moment?

addon/sign
```diff
@@ -49,13 +49,13 @@ function signCb(err, resp, body) {
...
-// if we need to sign and distribute our addon, we want to use this method
+// if we need to sign and distribute our add-on, we want to use this method
 function distAddon() {
   // generate our xpi and update.rdf
   packageAddon(function(err) {
     if (err) return console.error(err);
 
-    // sign our addon
+    // sign our add-on
     var generatedXpi = '@testpilot-addon-' + version + '.xpi';
```
meandavejustice: This will need to be changed to `testpilot-` from `testpilot-addon`

addon/sign
```diff
@@ -80,17 +80,17 @@ function distAddon() {
...
+// if we download our signed add-on we want to use this method of distribution
 function distAddonSigned() {
   // generate our xpi and update.rdf
   packageAddon(function(err) {
     if (err) return console.error(err);
-    console.log('addon packaged.');
+    console.log('add-on packaged.');
 
     // move generated files into /dist directory
     var updateFile = '@testpilot-addon-' + version + '.update.rdf';
```
meandavejustice: This will need to be changed to `testpilot-` from `testpilot-addon-`

addon/README.md
ghost, line 2: I left it for whimsy's sake.  Being in a readme is a lot different than someone seeing it in about:addons. :)

clouserw:
rebased w/ master

PR #480, opened 2016-03-14
[fixes #463] Create mechanism to allow for content detection

### LINKED ISSUE #463: Create mechanism to allow for content detection
There is a request to give owned channels (about:home, moz.org) the ability to detect whether or not a user is a participant in Test Pilot. I don't know if such a thing exists (or is possible), but this issue is opened to discuss, and implement if appropriate.

ckprice:
Just tried to do what @lmorchard suggested in the issue. Works locally using `npm run once`.

fixes #463

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -8,7 +8,8 @@
 
 const settings = {};
 
-let currentPageMod;
+let setInstalledFlagPageMod;
+let messageBridgePageMod;
 let button;
 let app;
 
@@ -47,16 +48,19 @@ const SERVER_ENVIRONMENTS = {
   local: {
     BASE_URL: 'http://testpilot.dev:8000',
     TESTPILOT_PREFIX: 'testpilot.addon.LOCAL.',
+    WHITELIST_URLS: 'https://www.mozilla.org/*,about:home',
     BADGE_COLOR: '#AA00AA'
   },
   stage: {
     BASE_URL: 'https://testpilot.stage.mozaws.net',
     TESTPILOT_PREFIX: 'testpilot.addon.STAGE.',
+    WHITELIST_URLS: 'https://www.mozilla.org/*,about:home',
     BADGE_COLOR: '#A0AAA0'
   },
   production: {
     BASE_URL: 'https://testpilot.firefox.com',
     TESTPILOT_PREFIX: 'testpilot.addon.MAIN.',
+    WHITELIST_URLS: 'https://www.mozilla.org/*,about:home',
     BADGE_COLOR: '#00AAAA'
   }
 };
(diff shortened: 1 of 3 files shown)


addon/data/message-bridge-readonly.js
```diff
@@ -0,0 +1,12 @@
+/*
```
lmorchard: As a nitpick, I'd say this shouldn't be named "message-bridge" - because it's not actually doing the addon port / content event bridging. Maybe something more like `set-installed-flag`

Also, to reduce redundant work, if we're setting the flag here we should probably take that side effect out of `message-bridge.js` and just load this same PageMod on the Test Pilot site, too

addon/index.js
```diff
@@ -71,13 +72,26 @@ function updatePrefs() {
...
+  if (readPageMod) { readPageMod.destroy(); }
+  if (writePageMod) { writePageMod.destroy(); }
+
+  // Set up new PageMod for read access to detect Test Pilot installation.
+  readPageMod = new PageMod({
+    include: settings.ALLOWED_ORIGINS_READONLY.split(','),
+    contentScriptFile: './message-bridge-readonly.js',
+    contentScriptWhen: 'start',
+    attachTo: ['top', 'existing'],
+    onAttach: setupApp
```
lmorchard: This shouldn't be called here, since `setupApp` is called in `onAttach` for the new `writePageMod` too. Also `setupApp` depends on the existence of `writePageMod`

addon/index.js
```diff
@@ -71,13 +72,26 @@ function updatePrefs() {
   Object.assign(settings, {
     BASE_URL: env.BASE_URL,
     ALLOWED_ORIGINS: env.BASE_URL + '/*',
+    ALLOWED_ORIGINS_READONLY: 'https://www.mozilla.org/*,about:home',
```
lmorchard: Rather than being hardcoded here, I think these origins should appear [as a default preference in `package.json`](https://github.com/ckprice/testpilot/blob/issue-463-content-detection/addon/package.json#L37) and get picked up [from `simplePrefs.prefs`](https://github.com/ckprice/testpilot/blob/issue-463-content-detection/addon/index.js#L67). Then, we can change them without rebuilding the addon - e.g. for QA & integration testing.

That, or include them in [the list of canned environments](https://github.com/ckprice/testpilot/blob/issue-463-content-detection/addon/index.js#L47) so we can swap between origin lists along with server environments.

addon/package.json
```diff
@@ -36,6 +36,13 @@
   },
   "preferences": [
     {
+      "name": "WHITELIST_URLS",
```
ckprice: Not sure if this is what you meant @lmorchard. Seems odd to expose this to the user.

meandavejustice:
lgtm gunna let @lmorchard do the merging since he's been working with you on the review.

addon/package.json
lmorchard: Well, that's why I also mentioned [sticking this in the list of canned server environments](https://github.com/ckprice/testpilot/blob/issue-463-content-detection/addon/index.js#L47). Main reason to modify this is for testing, and we expose the local/stage/prod choice to the user. So, that might be a happy medium rather than exposing the list completely.

addon/index.js
```diff
@@ -66,18 +67,32 @@ function updatePrefs() {
   const envName = simplePrefs.prefs.SERVER_ENVIRONMENT;
   const env = (envName in SERVER_ENVIRONMENTS) ?
     SERVER_ENVIRONMENTS[envName] : SERVER_ENVIRONMENTS.production;
+  // List of URLs which can view the installed flag.
+  const whitelistURLs = simplePrefs.prefs.WHITELIST_URLS;
```
lmorchard: Oh also, this pref won't get picked up when it changes, because this `updatePrefs()` function [only gets called when the SERVER_ENVIRONMENT pref changes](https://github.com/ckprice/testpilot/blob/issue-463-content-detection/addon/index.js#L106). (Well, and also on browser startup) So, might as well put this `WHITELIST_URLS` list into `SERVER_ENVIRONMENTS`.

ckprice: @lmorchard thanks! updated

addon/index.js
```diff
@@ -47,16 +48,19 @@ const SERVER_ENVIRONMENTS = {
   local: {
     BASE_URL: 'http://testpilot.dev:8000',
     TESTPILOT_PREFIX: 'testpilot.addon.LOCAL.',
+    WHITELIST_URLS: 'https://www.mozilla.org/*,about:home',
```
lmorchard, line 51: Cool! Seems silly to duplicate them right now. But, that's fine: hopefully once we have some integration testing that involves the add-on, we can patch this to include local URL origins relevant to those tests.

lmorchard:
Thanks for bearing with me through all the nits, and thanks for tackling this! Going to give this a quick run through on my machine and then probably merge

lmorchard:
`npm run once` worked. Visited mozilla.com, checked `navigator.testpilotAddon` and it was `true`. Went to google.com, and it was undefined. :ship:

PR #488, opened 2016-03-15
Sign addon with hashed version only on master or release

### LINKED ISSUE #484: Add-on builds result in stale code when version is not updated
Currently, add-on signing happens [on every Circle CI run](https://github.com/mozilla/testpilot/blob/master/circle.yml#L48) - this includes PR review test runs. This is problematic, because AMO only supports signing any given version once.

That means if a PR is submitted, the code gets built & signed on initial submission with the version string in `package.json`. If add-on code changes are made in response to a review - but the version isn't changed - the old code from the initial submission is what sticks around. Our build automation [just downloads the previously signed version](https://github.com/mozilla/testpilot/commit/e4659e9351ebf45b5f227d03f42c28fd7af8b48a#diff-8894e422ed2d6af58eed4df3575eb027R1), because attempting to re-sign an already used version results in an AMO API error.

But, manually bumping the add-on version number with every edit to a PR seems problematic - both in terms of remembering to do it and in terms of arbitrary version number inflation for unreleased changes.

We could investigate a way to run `npm run sign` only when we **mean** it. That is, only when building a Docker container from master branch intended for deployment. I suspect that will really complicate our `circle.yml` file, though.

Alternatively, we could find a way to automatically vary the version string based on the code - e.g. include the git commit in the version string. That would directly tie the add-on version to code changes. Are there any drawbacks to that? Could un-reviewed / un-approved yet signed versions of the XPI get accidentally released?

lmorchard:
- In Circle CI, only use `npm run sign` when we're on master branch or
  are building a release. Use `npm run package`, otherwise.
- As part of `npm run sign`, modify the addon version to include current
  git revision hash or current Circle CI release tag.

Fixes #484

### DIFF
--- a/addon/bin/sign
+++ b/addon/bin/sign
@@ -17,7 +17,9 @@ var fs = require('fs');
 var path = require('path');
 var request = require('request');
 var jwt = require('jsonwebtoken');
-var manifest = require('./package.json');
+
+var manifestFilename = __dirname + '/../package.json';
+var manifest = require(manifestFilename);
 var version = manifest.version;
 var apiKey = process.env['AMO_USER'];
 var apiSecret = process.env['AMO_SECRET'];
--- a/addon/bin/update-version
+++ b/addon/bin/update-version
@@ -1,0 +1,29 @@
+#!/usr/bin/env node
+
+var fs = require('fs');
+var util = require('util');
+var childProcess = require('child_process');
+
+// Load the current manifest.json
+var manifestFilename = __dirname + '/../package.json';
+var manifest = require(manifestFilename);
+
+// Come up with a version tag suffix based on CIRCLE_TAG, or fall back to
+// looking up the current git commit hash.
+var tag;
+if (process.env['CIRCLE_TAG']) {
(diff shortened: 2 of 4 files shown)


lmorchard:
FWIW, looks like this successfully used `npm run package` in Circle CI for checks on this PR. Not sure exactly how to verify it does the right thing on master without merging it :/ Hmm, unless maybe I tag a release from my own fork and my personal Circle CI account does it...

lmorchard:
Ah, here we go:
- [Circle CI used `npm run package` on my PR branch](https://circleci.com/gh/lmorchard/testpilot/52)
- [Circle CI used `npm run sign` on a release I tagged from my fork](https://circleci.com/gh/lmorchard/testpilot/55) - (which only worked because I have my AMO API credentials in my Circle CI config for Test PIlot)

Seems like it works, but AMO signing is very picky about the version being < 32 characters. So, `0.4.10-tag-2016-03-15` ends up being 21 characters - we should be okay if we don't get more adventurous than that with overall project release tags.

meandavejustice:
This is great! thanks @lmorchard

PR #490, opened 2016-03-15
Retire Travis CI in favor of Circle CI

lmorchard:
We're now running all the same tests in Circle CI that we have been in Travis CI. But, Circle CI does us one better by running Django tests inside the actual Docker container that we build for deployment. So, we should probably retire Travis CI as redundant.


lmorchard:
Follow up task would be to turn off Test Pilot in Travis CI, because I don't think this just does that for us.

PR #495, opened 2016-03-17
Experiment Gradients

### LINKED ISSUE #461: Assign every test a color-scheme
In the new UX, every test should be assigned two hex values and an angle.

screen shot 2016-03-08 at 12 52 15 pm

meandavejustice:
- add django-colorfield dependency
- add gradient-start and gradient-stop fields to experiment model
- update experiment model test and serializer
- update experiment-row-view.js bindings to apply gradient
- update experiment-page-view.js and it's template to apply gradient
- add colorfield to INSTALLED_APPS in settings.py
- add migration for new experiment model fields
- fixes #461

cc @johngruen


lmorchard:
:shipit: r+

Issue #476, opened 2016-03-14
Test Pilot title should link to the home page


johngruen:
Right now the test pilot title is just an h1, we should make it a link to the most appropriate page (home if you're logged in, sign up, or get the add on)

jaredhirsch:
@johngruen Stealing for a new contributor :-)

johngruen:
@ktbee hey, WRT styles, no need to underline or change color or anything...just make sure `cursor:pointer`

pdehaan:
I think this was fixed via #510. Closing.

PR #510, opened 2016-03-19
476-home-page-link

ktbee:
Adds homepage or sign up link to title h1 text depending on whether the user is signed in or not. This is a solution for [issue 476](https://github.com/mozilla/testpilot/issues/476)

### DIFF
--- a/testpilot/frontend/static-src/app/templates/header-view.js
+++ b/testpilot/frontend/static-src/app/templates/header-view.js
@@ -2,7 +2,9 @@ export default `
   <header id="main-header" class="content-wrapper" data-hook="active-user">
     <h1>
       <span class="firefox-logo"></span>
-      <span data-hook="title"></span>
+      <a href="/">
+        <span data-hook="title"></span>
+      </a>
     </h1>
     <div data-hook="settings"></div>
   </header>
--- a/testpilot/frontend/static-src/app/templates/landing-page.js
+++ b/testpilot/frontend/static-src/app/templates/landing-page.js
@@ -5,7 +5,9 @@ export default `
       <header id="main-header" class="content-wrapper">
         <h1>
           <span class="firefox-logo"></span>
-          <span data-l10n-id="siteName">Firefox Test Pilot</span>
+          <a href="/accounts/login/?next=/">
+            <span data-l10n-id="siteName">Firefox Test Pilot</span>
+          </a>
         </h1>
         <a data-l10n-id="landingFxaAlternateButton" href="/accounts/login/?next=/" class="button outline">Sign in</a>
       </header>


jaredhirsch:
@meandavejustice Hey, you might be a better reviewer for this, I haven't touched the testpilot code in ages. I'm not sure what's up with CircleCI, it seems to be pointing to a test run that doesn't exist?

jaredhirsch:
I mean, this code looks fine to me, but not sure what exactly y'all had in mind here

testpilot/frontend/static-src/styles/modules/_main-header.scss
```diff
@@ -7,6 +7,8 @@
     font-weight: 700;
     margin: 10px 0;
 
+    a { cursor: pointer; }
```
jaredhirsch: @ktbee I'm surprised this was needed. Does the link not get the pointer style without this?

ktbee: Hmm, good question @6a68 . Looking at it again, I don't think it's necessary.  When I tested this styling earlier, I had thought my cursor stayed an arrow icon instead of the pointer hand image. Now however, I can see that the cursor has pointer styling whether I declare it or not. I think initially my browser was slow to adjust to my cursor hovering over the link, and so I thought it wasn't displaying the style it needed.

Should I do a separate PR without this style declaration?

jaredhirsch: @ktbee Cool, I wasn't sure if the pointer styles seemed off, or something. To update the PR with this line changed, you just need to update the commit, then force-push to the existing branch.

There are many ways to alter a git commit, but the simplest approach here is probably to check out this branch, remove that line from the stylesheet, then 'git add' the file, and 'git commit --amend --no-edit' to apply the new changes to the changes in the old commit. After doing this, git log should just show the one commit, without the style changes.

Once you've got the commit looking good locally, you can use 'git push -f' to the existing branch, and it'll overwrite the remote (github's copy of the branch) with the new commit.

Feel free to ping me with any questions. Git has a steep learning curve, but it's an amazing tool once you get it. The pro git book is a great learning resource: https://progit.org

jaredhirsch:
The 'details' CircleCi link is broken. I find this baffling. @meandavejustice is this a known thing?

The link seems to point to https://circleci.com/gh/mozilla/testpilot/110, which doesn't exist.

meandavejustice:
@6a68 weird, that link was broken for me on mobile, but is working now on my laptop.

The build was broken because of a caching issue(out of @ktbee's control). We updated our node version last week and needed to bust cache so that the native dependency `node-sass` could be rebuilt. I triggered a new build with the `build without cache` option in circleci.

@ktbee This looks good, thanks for the pull request. I'll merge this as soon as tests are passing.

PR #521, opened 2016-03-23
remove tour link from settings

### LINKED ISSUE #279: Implement "Tour Idea Town" flow
Currently it doesn't look like the ["Tour Idea Town"](https://github.com/mozilla/idea-town/blob/8e98c7963b58349cd4a1638df9b96f3672dd7a2f/idea_town/frontend/static-src/app/views/settings-menu-view.js#L16) link in the settings drop down menu on http://idea-town-dev.elasticbeanstalk.com/home does anything.

Not sure if we want to add that before Mozlando, or maybe just hide the option.

johngruen:
Fixes #279...
Lets clean this up for now and see what UR comes back with to decide if we actually need a tour.

### DIFF
--- a/locales/en-US/app.l20n
+++ b/locales/en-US/app.l20n
@@ -14,12 +14,9 @@
 
 <home "Home">
 
-<menuTour "Tour Test Pilot">
 <menuWiki "Test Pilot Wiki">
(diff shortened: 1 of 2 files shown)


johngruen:
#279

Issue #479, opened 2016-03-14
Build a wrapper for submitExternalPing()


clouserw:
The way the metrics plan is developing is to use Telemetry to record our measurement data.  This means using [Telemetry's submitExternalPing()](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/TelemetryController.jsm#192).  This issue is to add a function to TestPilot which allows test pilot clients to easily record data with us.  (basically a light wrapper around submitExternalPing())

The "type" there should be 'testpilot' and we do want to enable the clientID and environment.  [The documentation](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md) shows what a ping should look like.  I think we'll also want TestPilot to add the user agent and a version along with whatever the client sends.

This issue also includes writing documentation for using the new function.  I'd suggest putting it at the top of https://wiki.mozilla.org/Test_Pilot/Metrics

mostlygeek:
There are several ways to get data in. @whd would be the authoritative source for recommendations on how best to send metrics. 
1. via telemetry, as part of the regular FF telemetry upload
2. http => testpilot => statsd (datadog) 
3. http => testpilot => heka => ElasticSearch => Kibana
4. http => testpilot => heka => (pipeline?) => ??

Figuring out which of these paths data will flow will help getting the backing service setup.

lmorchard:
This issue is just about using Telemetry on the client side from experiments. [More complete details about our collection strategy are over here in README-METRICS.md](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md#data-collection-details).

lmorchard:
Digging into this and thinking out loud, I think we're missing some detail in how this should work for individual experiments. In fact, the README-METRICS says this:

> This document is about metrics about Test Pilot, not metrics about tests within Test Pilot. If you're looking for more information about measuring tests themselves please [read the wiki](https://wiki.mozilla.org/Test_Pilot/Metrics).

There's [the main `testpilot` summary ping](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md#data-collection-details), so I don't think we want individual experiments to also each send their own `testpilot` ping via `submitExternalPing`. Those pings would collide with the main daily summary. 

From reading about Telemetry so far, I think experiments should contribute data to the main summary ping over the course of 24 hours. Something like this (with totally made-up measurements):

``` json
  {"tests":
    {"universal_search":
      {"last_enabled": 1457462200,
[...]

PR #538, opened 2016-03-24
Build a wrapper for submitExternalPing()

lmorchard:
- Add small example add-on showing how to use the ping messaging
- Switch to sdk/system/events for cleaner access to nsIObserver
- Remove lib/authRequest.js for now, because metrics pings do not go to
  the Test Pilot server any more. (Feedback also now does nothing)
- Do not ping for every message that the addon sends to web content
- Small bugfix for isTestpilotAddonID
- Misc reorganizing in require()'s at the top

Issue #479


lmorchard:
So, I think this works. But, a couple open questions:
- This does pass a message event from an example experiment add-on to the Test Pilot add-on, but [my pings are probably bogus examples](https://github.com/mozilla/testpilot/pull/538/files#diff-e5efab61baef2675973f8a95bc256e6cR28). Guessing this is where each experiment should define a consistent ping schema and stick with it.
- No docs for this yet, but I wanted to make sure the code was solid before writing them.
- I was tempted to (re)start a shared library for the `sendMetricsPing` function, but it's very tiny.

lmorchard:
<strike>Oh yeah, and this doesn't include the once-a-day `testpilot` ping</strike>

lmorchard:
Just added code to send the once-a-day `testpilot` ping, with payload kept up to date with experiment installs & uninstalls.

lmorchard:
@chuckharmston - you might want to peek at [the `sendMetricsPing` function in the example add-on](https://github.com/mozilla/testpilot/pull/538/files#diff-e5efab61baef2675973f8a95bc256e6cR5) in this PR to see how the ping will work. What to send as the actual ping payload is still kind of a moving target, though.

lmorchard:
@meandavejustice / @chuckharmston - have time to peek at this, have any thoughts?

meandavejustice:
This looks great to me, and the addon builds locally. We'll review the actual pings to telemetry once we get that hooked up from the otherside. :+1:

PR #540, opened 2016-03-25
Implement addon install flow

### LINKED ISSUE #440: update add-on install flow
Demo: https://people.mozilla.org/~jgruen/test-pilot/get-add-on.html

add-install

meandavejustice:
- rename `data/feedback.html` to `data/base.html`
- create base template in `lib/templates.js` (will possibly port
  `data/base.html` over in a later commit)
- create installed dialog template
- add copter.png and copter@2x.png to addon directory
- add `data/link-catch.js` for installed dialog
- add copter animations and install dialog styles
- add handling of new install dialog to `index.js`
- remove 'reload when addon installed`from`app/main.js`
- update minor version in `package.json`
- update 'watch' task in `gulpfile.js` to include `addon` task
- update `templates/landing-page.js` to handle logged-in/no addon state
- update `views/landing-page.js` for install progress

Fixes #440 #545


addon/index.js
```diff
@@ -136,7 +147,46 @@ function setupApp() {
...
+
+      installMsgPanel.show({width: INSTALLED_PANEL_WIDTH,
+                            height: INSTALLED_PANEL_HEIGHT,
+                            position: button});
+
+      // make sure we don't show the installed dialog more than once
+      showInstalledDialog = false;
+
+      // allow us to hide the panel from a landing page interaction
+      app.on('hide-installed-panel', () ={
```
lmorchard: [Looks like Circle CI lint caught a syntax error, here.](https://circleci.com/gh/mozilla/testpilot/138)

addon/index.js
```diff
@@ -136,7 +147,46 @@ function setupApp() {
...
+        }
+      });
+
+      installMsgPanel.port.on('link', (url) => {
+        const tabExists = false;
+
+        // reload the testpilot tab now that we are logged in with the
+        // addon installed
+        for (let tab of tabs) {
+          if (tab.url.indexOf(`${settings.BASE_URL}`) > -1) {
```
pdehaan: Maybe I'm missing something here, but do we need the variable interpolation here, or can we just do something like:

``` js
if (tab.url.indexOf(settings.BASE_URL) > -1) {
```

Or if we know we're in a recent version of Firefox, we could just use the new `.includes()` method:

``` js
if (tab.url.includes(settings.BASE_URL)) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

meandavejustice: Thanks @pdehaan forgot to update that :)

ckprice:
@meandavejustice @lmorchard FYI - this does not block the public launch. So, you can decide how to handle this PR (e.g. finish if it's close to merging vs. if we should close and address later).

We triaged our priority issues into https://github.com/mozilla/testpilot/milestones/2016-04-08

testpilot/frontend/static-src/app/views/landing-page.js
```diff
@@ -5,6 +5,8 @@ import ExperimentListView from './experiment-list-view';
 
 // TODO back out this commit (78bd0a93a1q) before launch.
 import queryString from 'query-string';
+const installedMsgTmp = `<h2 class="emphasis">Test Pilot installed!</h2>
+<a href="/" class="button large default quick-pop">Let's Go</a>`;
```
pdehaan: I'm not 100% certain on our l20n flow, but will these work with our localization flow, or is this just temporary text?

testpilot/frontend/static-src/app/views/landing-page.js
```diff
@@ -29,6 +36,21 @@ export default PageView.extend({
...
+      wrapper.querySelector('.button').onclick = () => {
+        app.webChannel.sendMessage('hide-installed-panel', {});
+      };
+    });
+  },
+
+  installClicked() {
+    const btn = this.query('.button.default');
+    btn.classList.add('state-change');
+    btn.querySelector('span').textContent = 'Installing Test Pilot...';
```
pdehaan: Same comment as above, will this play nice with l20n?

addon/index.js
```diff
@@ -136,7 +147,46 @@ function setupApp() {
...
+          }));
+        }
+      });
+
+      installMsgPanel.port.on('link', (url) => {
+        const tabExists = false;
+
+        // reload the testpilot tab now that we are logged in with the
+        // addon installed
+        for (let tab of tabs) {
```
pdehaan:
```
166:18  error  `tab` is never modified, use `const` instead  prefer-const
```

addon/index.js
```diff
@@ -136,7 +147,46 @@ function setupApp() {
...
+          if (tab.url.indexOf(`${settings.BASE_URL}`) > -1) {
+            tabExists = true;
+            tab.reload();
+          }
+        }
+
+        // If we cannot find an open testpilot tab, just open one
+        if (!tabExists) tab.open(url);
```
pdehaan: I think `tab` is out of scope here.

```
174:25  error  "tab" is not defined                          no-undef
```

addon/index.js
```diff
@@ -136,7 +147,46 @@ function setupApp() {
...
+      installMsgPanel.show({width: INSTALLED_PANEL_WIDTH,
+                            height: INSTALLED_PANEL_HEIGHT,
+                            position: button});
+
+      // make sure we don't show the installed dialog more than once
+      showInstalledDialog = false;
+
+      // allow us to hide the panel from a landing page interaction
+      app.on('hide-installed-panel', () => {
+        installMsgpanel.destroy();
```
pdehaan: Typo: `installMsgPanel.destroy();` (upper "P")

```
187:9   error  "installMsgpanel" is not defined              no-undef
```

testpilot/frontend/static-src/app/views/landing-page.js
lmorchard: Yeah, we don't have any translations yet - but, any new text elements really should be marked with a `data-l10n-id` and added to [locales/en-US/app.l20n](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.l20n)

testpilot/frontend/static-src/app/views/landing-page.js
lmorchard: Nope, this should probably instead be a hidden element that gets revealed, with content marked with `data-l10n-id` and a corresponding string in `app.l20n`. We can't directly set text content like this with L20n

addon/lib/templates.js
```diff
@@ -71,3 +71,24 @@ module.exports.feedback = `
   </div>
   <div class="clear"></div>
 </div>`;
+
+module.exports.installed = `
+<div id="browser-doorhanger">
+  <div class="copter-wrapper">
+    <div class="copter fly-down"></div>
+  </div>
+  <p>Test Pilot lets you test out experimental features in Firefox. Ready for takeoff?</p>
```
lmorchard, line 80: This copy also needs a `data-l10n-id` and to be added to `app.l20n`. Same with the line under it

johngruen:
@meandavejustice hey dave...seeing some UI weirdness here right now.
screen_shot_2016-03-28_at_4_10_02_pm
screen_shot_2016-03-28_at_4_10_18_pm
screen_shot_2016-03-28_at_4_16_27_pm

addon/lib/templates.js
meandavejustice, line 80: Thanks for catching these @pdehaan  @lmorchard I'll get them fixed up

lmorchard, line 80: Yeah, it's kind of annoying - and until @pdehaan noticed the issue, I'd mostly forgotten about L20n myself. :frowning: 

I wonder if there's any kind of validation or linting or something we can do to be sure we don't miss strings? I can't think of what it would be off the top of my head. Also doesn't help that we're not totally hooked up to a localizing interface to make this stuff actually useful

pdehaan, line 80: I cant think of any linting that would catch these. I was only finding issues before when I added a second, fake dummy locale and then switched my browser's locale to the new locale and manually looked for untranslated strings. ¯\_(ツ)_/¯

meandavejustice, line 80: These strings specifically are a part of the addon, which we are not using l20n with yet.

lmorchard, line 80: Oh yeah, derp - we should probably add l20n to the add-on. Filed #557 for that.

lmorchard:
Hmm, running through this, and it doesn't ever seem to leave the "installing test pilot..." state. 

I do have to run through the door-hanger permission stuff - which is built into Firefox and we can't change. Though, I think it does go away when you have a signed addon on a *.firefox.com domain. 

But, on installation, the addon is set to point at production. So, I go into about:addons and change that in prefs. Thinking maybe the dev side of things just never catches the install event because of that?

Maybe I'm missing a trick to getting this to work on dev?

meandavejustice:
@lmorchard that is correct, testing this has been a bit tough on dev... I've basically been using `jpm watchpost` while removing the loadReason==='install' check(firefox caches removed addons so the loadReason after initial install becomes 'downgrade').

lmorchard:
I wonder if something simpler / hackier might work better here? That is, on the install page, start a setInterval() to watch for `navigator.testpilotAddon === true` every second or so. And when that flips to true, then do the post-install stuff. That would even catch the pref change from prod to local

meandavejustice:
@lmorchard hmm that may be the better solution in this case. I'll think it over and try and fix that up in the morning.

johngruen:
@meandavejustice seeing a few nits:
1. no animation on the installing test pilot button
2. looks like the header is getting passed into the view and is rendered on the bottom of the page below the footer
3. the button on the door hanger should have larger text (at least 16px)
4. if you dismiss the door hanger, the "let's go" button should show back up on the page itself

lmorchard:
Don't have anything to add to the nits - I saw all those same things, too. But, the flow does work for me, hooray! 

It was a little weird because I had to triple check I had packaged the right add-on version. But, that's not something specific to this PR. Might be something we should think about ways to make smoother in dev.

johngruen:
@lmorchard yeah i did too, i think i wound up manually copying the newly packaged add-on to `/static`

meandavejustice:
@johngruen I've addressed 1, 3, and 4. I'm not seeing the header issue. Would you mind posting a screenshot?

lmorchard:
Ran through this again and everything seems to be working. The last thing seems like that header-below-the-footer issue, but I can't see it myself either.

r+ from me, ready to merge if that footer thing seems like a non-issue

meandavejustice:
Thanks @lmorchard

PR #543, opened 2016-03-28
Emit request.summary log entries on every request

### LINKED ISSUE #541: Implement request logging
Using the new JSON logging module and (probably) Django middleware, start producing [`request.summary`](https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md) messages

lmorchard:
- Switch from simple to json log format on dev
- Middleware to emit request.summary type log events
- Tweaks to settings to use the new middleware

Fixes #541


lmorchard:
@mostlygeek - I think this should do the job of sending JSON log events to the console on every request. Let me know if anything looks weird in there!

meandavejustice:
looks :ok_hand: to me

lmorchard:
Passes tests & has a least one thumbs up, so I'll just merge this. That way, we can get it out to stage / prod when we deploy next and see if it does the right thing.

PR #552, opened 2016-03-29
Add metrics flows image to docs

lmorchard:
Seems like this is a fairly accurate image to represent metrics flows.

[Original drawing is here on Google Docs](https://docs.google.com/a/mozilla.com/drawings/d/1McTKdbMog0OQ9Mg29DCPRoGL43NODOs2ZqS1bAkvhQ0/edit?usp=sharing) (sorry, Mozilla org folks only). Let's correct the image if there's anything inaccurate about it. Otherwise, this should be something good to include in the docs.


meandavejustice:
lgtm, gunna let @clouserw look this over before merging since I haven't been as involved in the metrics discussions

ghost:
r+  this looks good, with the caveat that things like Periscope are just being tested and may or may not be the final choice.  Thanks!

PR #561, opened 2016-03-30
Fix UUID in addon

### LINKED ISSUE #560: new client UUID is breaking addon
using `require('sdk/util/uuid').uuid()` breaks the current addon as it returns an object

meandavejustice:
- fixes #560


lmorchard:
Ugh, thought I'd fixed that!

PR #564, opened 2016-03-31
make now active animation smooth

### LINKED ISSUE #546: Now Active label doesn't animate very smoothly
This is weird because it looks "Not Great (tm)" on my laptop, and when I create an .MOV via QuickTime (but when I convert to GIF, it looks slightly better -- so it's a bit difficult to capture in the wild).

johngruen:
@pdehaan one glorious line!

Fixes #546


johngruen:
#546

Issue #555, opened 2016-03-29
Refresh demo_data.json with up to date info


ckprice:
So we can stop losing things when instances are rebuilt. And generally probably good to poke at this thing every couple of months.

Ref: [/fixtures/demo_data.json](https://github.com/mozilla/testpilot/blob/master/fixtures/demo_data.json)

pdehaan:
:+1: to marking this "faux-critical" so we can have consistent data on dev/stage/prod for testing.

pdehaan:
Ref: #493

ckprice:
https://testpilot.dev.mozaws.net has been updated with the new freshness. Will use that as base.

lmorchard:
Copying over a comment from #493 since it seems to make sense here too:

I'm not sure `demo_data.json` should actually match any particular environment. Data will differ between prod, stage, and dev depending on what kind of tests / research we want to run there. 

`demo_data.json` is mainly there to offer some sample data to exercise features during local dev, not to replicate what might be in production. I don't think we should spend effort to try to ensure it stays in synch reflecting prod, just that it has enough data to trigger all the site features and/or special cases

meandavejustice:
I think we should make a practice of updating demo_data.json when adding a new field or changing data shape. Just as we have to change the shape of data in our tests to get them to pass when we make changes to a model. This is probably a given, but that is my only opinion

lmorchard:
> I think we should make a practice of updating demo_data.json when adding a new field or changing data shape.

Yes, this. Update it to exercise features, not to match any deployed environment.

ckprice:
This is still on me, but won't be this week.

For new fields it seems pretty straight forward from the testing I've done. For example, `gradient_start` isn't currently in `demo_data.json`, but just adding it to the file automagically saves it to the DB (tested with `docker exec testpilot_server_1 ./manage.py loaddata fixtures/demo_data.json`).

As far as getting the fieldname, I guess just troll through [`testpilot/experiments/migrations/`](https://github.com/mozilla/testpilot/tree/master/testpilot/experiments/migrations) unless there's a more elegant method.

lmorchard:
For what it's worth, there's also `./manage.py dumpdata` which does the opposite of `loaddata`. The problem with that, though, is it's the firehose of _all_ data in the database. You can pare it down and prettify it with options (`dumpdata --help`), but it's best to open the fixture and a data dump side-by-side and selectively copy things over to the fixture from the raw dump.

[...]

PR #572, opened 2016-04-05
Update Universal Search fixture data.

### LINKED ISSUE #571: Update the version of UniversalSearch XPI that's currently on dev
Apparently the version being served by testpilot.dev.mozaws.net is a couple months old, so we should check the XPI path to make sure it's pointing to the latest version.

chuckharmston:
Closes #571 

r? @pdehaan

### DIFF
--- a/fixtures/demo_data.json
+++ b/fixtures/demo_data.json
@@ -181,9 +181,9 @@
 {
   "model": "experiments.experiment",
   "fields": {
-    "addon_id": "universal-search-addon@mozilla.com",
-    "version": "5.1.4",
-    "changelog_url": "https://github.com/mozilla/universal-search-addon/commits/master",
+    "addon_id": "universal-search@mozilla.com",
+    "version": "1.0.0b2",
+    "changelog_url": "https://github.com/mozilla/universal-search/commits/master",
     "privacy_notice_url": "",
     "contributors": [ [ "lorchard-demo" ], [ "jgruen-demo" ], [ "charmston-demo" ], [ "jhirsch-demo" ] ],
     "order": 0,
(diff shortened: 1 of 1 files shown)


fixtures/demo_data.json
```diff
@@ -191,8 +191,8 @@
     "thumbnail": "demo/experiments_experiment/a/8/a8182b008f6434ba702ab177232625e2_thumbnail_1449675941_0564.png",
     "created": "2015-12-09T15:45:42.395Z",
     "slug": "universal-search",
-    "xpi_url": "https://people.mozilla.org/~jhirsch/universal-search-addon/addon.xpi",
-    "contribute_url": "https://github.com/mozilla/universal-search-addon/blob/master/CONTRIBUTING.md"
+    "xpi_url": "https://s3-us-west-2.amazonaws.com/universal-search/universal-search-1.0.0b1.xpi",
```
pdehaan: Why a specific filename vs always just renaming to "universal-search.xpi" or "latest.xpi"?

lmorchard:
FWIW, this will not actually fix #571 - dev is not auto-updated with this fixture data. It's a manual Django loaddata operation. 

This should probably also be rolled into #555, because I think @ckprice is currently updating the fixture overall.

fixtures/demo_data.json
lmorchard: Yeah, you might want to use a consistent URL pointing to the latest, so we don't need to update all the URLs with every release

chuckharmston: Yep, that was my bad. Copy/pasted from the wrong thing during a meeting.

pdehaan: But note, we'll still need to update the version number on every release manually through the /admin/ (and possibly in the fixtures/demo_data.json) for consistency maybe.

lmorchard:
#571 was manually addressed, but will merge this to update data for local installs. We should make sure it gets carried over in #555

pdehaan:
**NOTE:** I also tweaked the data on the admin site on http://testpilot.dev.mozaws.net/experiments/universal-search to reflect the latest changes.

PR #581, opened 2016-04-07
Auto-select last visited server env on add-on startup

### LINKED ISSUE #580: Add-on should be automatically set to the appropriate env on install
This is sort of a nit, but it makes user testing and debugging the add-on install flow quite difficult.

lmorchard:
Add initServerEnvironmentPreference helper in addon, which searches
through browser history for known server environments and finds which
one was most recently visited to auto-select environment.

Fixes #580

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -27,6 +27,7 @@ const {ToggleButton} = require('sdk/ui/button/toggle');
 const request = require('sdk/request').Request;
 const simplePrefs = require('sdk/simple-prefs');
 const URL = require('sdk/url').URL;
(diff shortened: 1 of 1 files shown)


lmorchard:
This could be a bad idea, but I can't think of a reason why. It should auto-select a server environment only on add-on startup or reinstall, i.e. during normal installation flow on any environment. Manual pref change should still apply after installation.

meandavejustice:
This is clever @lmorchard I'm pulling down for a manual test before merging.

johngruen:
@lmorchard this is dope.

PR #582, opened 2016-04-08
add title attr for experiment user counts

### LINKED ISSUE #570: Make it more obvious what the number of active users means on the experiment page
Currently it isn't super obvious what the "2" means next to the profile image. Not sure if we want a tooltip or something explaining that it's the current active number of users.

firefox_test_pilot_-_universal_search

meandavejustice:
Fixes #570

### DIFF
--- a/locales/en-US/app.l20n
+++ b/locales/en-US/app.l20n
@@ -83,3 +83,4 @@
 <lastUpdate "Last Update">
 <contribute "Contribute">
 <nowActive "Now Active">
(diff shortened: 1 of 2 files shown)


lmorchard:
r+ from me as long as the UX is kosher

meandavejustice:
lemme post a screenshot 4 u

meandavejustice:
title-attr

my screenshotting tool erases my cursor, but you get the idea.

PR #586, opened 2016-04-08
Fix cursor on download Firefox button

pdehaan:
Sets the `cursor` style to "auto", which gives the correct hand cursor for the download button (when using non-Firefox browser).

firefox_test_pilot

r? @johngruen @meandavejustice @lmorchard


meandavejustice:
Thanks @pdehaan

PR #577, opened 2016-04-07
[fixed #481] Wire GA events

ckprice:
@meandavejustice @lmorchard r?

PR #587, opened 2016-04-11
Check for existence before calling Google Analytics in index template

lmorchard:
Submitting this, since I think I prematurely merged PR #577 and this looked like one of the last issues

### DIFF
--- a/testpilot/frontend/templates/testpilot/frontend/index.html
+++ b/testpilot/frontend/templates/testpilot/frontend/index.html
@@ -24,7 +24,14 @@
     m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
     })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-    ga('create', 'UA-49796218-34', 'auto');
+    if (typeof(ga) !== 'undefined') {
+        ga('create', 'UA-49796218-34', 'auto');
+    } else {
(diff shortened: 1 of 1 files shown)


testpilot/frontend/templates/testpilot/frontend/index.html
```diff
@@ -24,7 +24,9 @@
     m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
     })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-    ga('create', 'UA-49796218-34', 'auto');
+    if (typeof(ga) !== 'undefined') {
+        ga('create', 'UA-49796218-34', 'auto');
+    }
```
meandavejustice, line 34: Would maybe be nice to those an `else console.warn('You have google analytics blocked. We understand. Take a look at our privacy policy to see how we handle your data.')`

meandavejustice:
Can we also add this check here https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/main.js#L71 ?

lmorchard:
That check is already [here](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/main.js#L66), though. (line 66, before that ga() call on 71)

meandavejustice:
@lmorchard Wow, I totally over looked that, sorry!

PR #598, opened 2016-04-13
Faking querystring for marionette

pdehaan:
@silne30 was saying that marionette was appending a "/" onto the end of the query string arg so we were having some automation issues.
This is a hacky fix to bypass a hacky marionette thing, all of which will be moot once we open the flood gates.

### DIFF
--- a/testpilot/frontend/static-src/app/views/landing-page.js
+++ b/testpilot/frontend/static-src/app/views/landing-page.js
@@ -21,7 +21,7 @@ export default PageView.extend({
   render() {
     const isLoggedIn = !!app.me.user.id;
     const query = queryString.parse(location.search);
-    this.isMoz = query.hasOwnProperty('butimspecial');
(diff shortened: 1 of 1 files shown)


pdehaan:

Sorry, this is just a hack on top of a marionette bug. It's turtles all the way down, man...

meandavejustice:
@pdehaan We don't use marionette. Do you mean ampersand? Can you link to the issue?

pdehaan:
@meandavejustice: The fabulous Mr. @silne30 is writing some e2e tests in marionette which we'll summon via our QA Jenkins pipeline for Continual Awesomeness (CA). This [admittedly ugly] PR will help unblock this marionette bug and let Dorlus unleash his inner Marionette.

meandavejustice:
Oh okay, My bad. I thought you were talking about http://marionettejs.com/

PR #601, opened 2016-04-13
Fix sigin and xpi download links for landing page

### LINKED ISSUE #597: Addon download button broken
addon installation was broken when we switch to `<button>` from `<a>` in this commit https://github.com/mozilla/testpilot/commit/00a8d90be21758ce6672d1ffc0e0949a87c6df91

meandavejustice:
fixes #597

### DIFF
--- a/testpilot/frontend/static-src/app/templates/landing-page.js
+++ b/testpilot/frontend/static-src/app/templates/landing-page.js
@@ -38,7 +38,7 @@ export default `
                 <span/>
               {{/isFirefox}}
               {{#isFirefox}}
-                <button data-l10n-id="landingFxaAlternateButton" class="button outline" data-hook="signin">Sign in</button>
+                <a data-l10n-id="landingFxaAlternateButton" class="button outline" data-hook="signin">Sign in</a>
               {{/isFirefox}}
             </header>
             <div class="split-banner responsive-content-wrapper">
@@ -67,7 +67,7 @@ export default `
                 </a>
               {{/isFirefox}}
               {{#isFirefox}}
-              <button data-l10n-id="landingFxaGetStartedButton" class="button large primary" data-hook="get-started-with-account">Get started with a Firefox Account</button>
+              <a data-l10n-id="landingFxaGetStartedButton" class="button large primary" data-hook="get-started-with-account">Get started with a Firefox Account</a>
               {{/isFirefox}}
             </div>
             <p data-l10n-id="landingLegalNotice" class="legal-information delayed-fade-in-up">By proceeding, you agree to the <a href="https://www.mozilla.org/about/legal/terms/services/">Terms of Service</a> and <a href="https://www.mozilla.org/privacy/firefox-cloud/">Privacy Notice</a> of Test Pilot</p>
@@ -96,7 +96,7 @@ export default `
                   </a>
                 {{/isFirefox}}
                 {{#isFirefox}}
-                <button data-l10n-id="landingFxaGetStartedButton" class="button large primary" data-hook="get-started-with-account">Get started with a Firefox Account</button>
+                <a data-l10n-id="landingFxaGetStartedButton" class="button large primary" data-hook="get-started-with-account">Get started with a Firefox Account</a>
                 {{/isFirefox}}
               </div>
               <div data-hook="footer-view"></div>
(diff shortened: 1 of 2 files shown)


ckprice:
This will also break on the [experiment listings pages](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/views/experiment-page.js#L205) (and other pages).

I think the problem is in the [helper function in `main.js`](https://github.com/mozilla/testpilot/blob/94845efe2459e9b5b90980ee34d5dc434cfba8e3/testpilot/frontend/static-src/app/main.js#L65)

```
// Send to GA
sendToGA(type, data) {
    if (typeof(ga) !== 'undefined') {
      data.hitType = type;
      if (data.outboundURL) {
        data.hitCallback = () => document.location = data.outboundURL;
      }
      ga('send', data);
    }
}
```

If ga is `undefined`, it never executes the `hitCallback`. I'm trying to find a more reliable way to detect if ga is loaded or not (it's pretty hit and miss with TP on/private browsing).

Ideally we can just update this one function and fix it. @meandavejustice since I created this thing, I'm happy to take a stab at trying to fix?

lmorchard:
Sounds like there's still some work to do here? I think we're also trying to self-assign on PRs and issues when we work on them, so whoever's best suited can tackle

lmorchard:
Also, hadn't quite realized the sendToGA function was setting document.location. Can we switch that to just sending metrics as a side-effect, and have links work normally? I might be overthinking it, but we also have history state stuff going on in the Ampersand router, so this might clash with that? (or does history state handle document.location assignment gracefully?)

ckprice:
@meandavejustice I think I have a solution in [my local branch](https://github.com/ckprice/testpilot/commit/538a5841168ca9c8234ba58e6eacf1e61ae1e6a9) that doesn't touch every file.

Let me know if you think this may be better, and I'll go ahead and open a PR.

/edit I linked to an [MDN article](https://developer.mozilla.org/en-US/Firefox/Privacy/Tracking_Protection) in my branch that describes this problem.

ckprice:
> Can we switch that to just sending metrics as a side-effect, and have links work normally?

I would have loved to do that :( But GA's send function is async. So, if you use `href`, and have, say, `onclick=sendMetric`, the `sendMetric` never takes.

lmorchard:
> I would have loved to do that :( But GA's send function is async. So, if you use href, and have, say, and onclick=sendMetric, the sendMetric never takes.

I'm not suggesting that, so much as it might be better to put the metrics calls [in the central page router](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/lib/router.js) instead of on every link

ckprice:
> I'm not suggesting that, so much as it might be better to put the metrics calls in the central page router instead of on every link

I'm sure it could be. But I'd like to see if we can get a quick fix on this as dev (/edit and stage) is currently broked. Also, not sure how we could do clicks to Signin/Logout/Install links.

lmorchard:
> @meandavejustice I think I have a solution in my local branch that doesn't touch every file.

Yeah, I think this quick tweak is simple enough. 

Feels like there's more we could do to make GA work less intrusively, but I haven't really played enough with it myself to say for sure. Did a bit more reading, and my worry about setting `document.location` & history state is a red herring - it should work fine. And we're already building a single-page-app here, so my fuzzy thoughts about "links working normally" are also kind of off track :)

lmorchard:
Looking through this PR, there's some nice cleanup in there - but I don't think it resolves the main issue with GA being unavailable & breaking links like [@ckprice's branch](https://github.com/ckprice/testpilot/commit/538a5841168ca9c8234ba58e6eacf1e61ae1e6a9) does.

jaredhirsch:
Here's a suggestion: rather than intercept every link click, why not just fire a pubsub event when a link is clicked, and if GA has loaded successfully, it can listen for that event?

ckprice:
Chatted with @meandavejustice. Going to close this, and open a PR from my branch.

> Here's a suggestion: rather than intercept every link click, why not just fire a pubsub event when a link is clicked, and if GA has loaded successfully, it can listen for that event?

Without digging too much into the possibilities here, I'll note two things:
- The GA send is async, so `<a href="/experiments" onclick="sendToGA()">` doesn't track the event.
- We do have links that go to external pages (e.g. Accounts links) outside of the Ampersand routing.

Opening a new PR, we can carry on the discussion there.

PR #603, opened 2016-04-14
[fixes #597] Fix GA helper function

### LINKED ISSUE #597: Addon download button broken

ckprice:
Replaces #601.

### DIFF
--- a/testpilot/frontend/static-src/app/main.js
+++ b/testpilot/frontend/static-src/app/main.js
@@ -63,15 +63,18 @@ app.extend({
 
   // Send to GA
   sendToGA(type, data) {
-    if (typeof(ga) !== 'undefined') {
+    if (window.ga && ga.loaded) {
       data.hitType = type;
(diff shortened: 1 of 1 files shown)


meandavejustice:
Continuing the discussion from #601:

I have the same concerns about `ampersand.history`. I think the correct solution will be implementing this as a side effect in the router as @6a68 @lmorchard proposed. I have opened up a follow up issue, #604 so I'll merge this now to fix dev env and fix with a more long term solution soon.

ckprice:
@pdehaan wondering if you could help test dev that 1. GA is still sending in non PBM and 2. If PBM is on, buttons are still navigable.

(WFM, but would like a 2nd pair of eyes)

Issue #585, opened 2016-04-08
Update Python dependencies


pdehaan:
https://requires.io/github/mozilla/testpilot/requirements/?branch=master

Specifically https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst

PR #630, opened 2016-04-22
Additional SHA hashes for Pillow==3.2.0

lmorchard:
Now includes *.tar.gz and *.zip versions, as well as the Mac OS X binary

Follow up to #594 and #585

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -113,6 +113,8 @@ requests==2.7.0
 requests_oauthlib==0.5.0
 
 # sha256: ZLCgVyEMSArqmUBsk5EYDNhm_A_Y8LUzZ-OvIbGVeEo
+# sha256: zETFS0MrJ8WePQvASKWnWBsTlEwUxWyUY0z-jV6rreU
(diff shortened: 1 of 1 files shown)


lmorchard:
Not sure why this surfaced as a problem, but it looks like sometimes installation would get a *.zip file and sometimes a *.tar.gz file. Both had the same contents, so the package looks okay, but the SHA hashes would differ due to the different archive formats.

ckprice:
This fixes the error I was getting on `docker-compose build`

> THE FOLLOWING PACKAGES DIDN'T MATCH THE HASHES SPECIFIED IN THE REQUIREMENTS
> FILE. If you have updated the package versions, update the hashes. If not,
> freak out, because someone has tampered with the packages.
> 
> ```
> Pillow: expected ZLCgVyEMSArqmUBsk5EYDNhm_A_Y8LUzZ-OvIbGVeEo
>              got zETFS0MrJ8WePQvASKWnWBsTlEwUxWyUY0z-jV6rreU
> ```

Not sure what other review this needs, but looks okay to me.

PR #632, opened 2016-04-22
remove extra title on add on install

### LINKED ISSUE #620: Odd v-scrolling on "Test Pilot installed!" page
Found in http://testpilot.dev.mozaws.net/__version__ (6b009172185f98271ce8163e24b65e266379da63):

``` json
{ "version": "",
  "commit": "6b009172185f98271ce8163e24b65e266379da63",
  "source": "https://github.com/mozilla/testpilot" }
```

---

By complete accident I noticed that when you sign up for Test Pilot, you end up on the "Test Pilot installed!" page, and you can scroll vertically and a Firefox logo appears beneath the fold.

firefox_test_pilot_and_canceled_event___tab_center__launch_status_checkin___wed_apr_20__2016_12_30pm_-_12_45pm__pdehaan_mozilla_com__-_pdehaan_mozilla_com_-_mozilla_mail_and_irccloud_and_1__docker-compose_up__docker-compose_

johngruen:
Fixes #620

### DIFF
--- a/testpilot/frontend/static-src/app/views/landing-page.js
+++ b/testpilot/frontend/static-src/app/views/landing-page.js
@@ -31,7 +31,8 @@ export default PageView.extend({
     if (!this.loggedIn && this.isMoz) {
       this.renderSubview(new ExperimentListView({
         loggedIn: this.loggedIn,
-        isFirefox: this.isFirefox
+        isFirefox: this.isFirefox,
+        skipHeader: true
       }), '[data-hook="experiment-list"]');
     }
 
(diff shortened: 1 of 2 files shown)


testpilot/frontend/static-src/styles/modules/_main-header.scss
```diff
@@ -1,6 +1,12 @@
 #main-header {
   @include flex-container(row, space-between, center);
 
+  // overrides the page view inserting the header
+  // on the landing page
+  [data-hook=landing-page] &:nth-child(2) {
```
lmorchard: This would probably be better done in the view JS, just don't add the elements to the DOM rather than hide them after the fact

johngruen: Yeah, but this way was faster! Will go back and do it right :)

johngruen:
@lmorchard okay, should be better now :)

PR #635, opened 2016-04-25
Pin last good versions of node-sass & sass-lint

### LINKED ISSUE #633: sass-lint is running into fatal errors with a fresh container rebuild
TL;DR: Something changed in `sass-lint` that doesn't like our `.scss` files now. It's breaking my local dev machine.

Because `gulp-sass` and `gulp-sass-lint` both have fuzzy version requirements for their dependencies, we end up getting incompatible upgrades to their dependencies that don't like our code.

This problem won't be seen on Circle CI until/unless we clear the dependency caches, but you'll see it if you rebuild all your docker containers. You should also see if it you install gulp from scratch locally on the host (e.g. `rm -rf node_modules && npm install`)

More details:

I just rebuilt all my docker containers totally from scratch. I removed all the containers and even deleted & recreated my docker-machine on OS X. Now, I'm getting errors like this in gulp:

```
frontend_watcher_1 | [21:01:07] Starting 'sass-lint'...
frontend_watcher_1 |
frontend_watcher_1 | testpilot/frontend/static-src/styles/_layout.scss
frontend_watcher_1 |   60:1  error  Please check validity of the block starting from line #60  Fatal
frontend_watcher_1 |
frontend_watcher_1 | ✖ 1 problem (1 error, 0 warnings)
frontend_watcher_1 |
frontend_watcher_1 |
frontend_watcher_1 | testpilot/frontend/static-src/styles/_state.scss
frontend_watcher_1 |   128:1  error  Please check validity of the block starting from line #128  Fatal
frontend_watcher_1 |
frontend_watcher_1 | ✖ 1 problem (1 error, 0 warnings)
```

This doesn't happen on Circle CI, so I started a build with ssh and looked at what was different. Here's what I think happened:

`gulp-sass-lint` depends on `sass-lint`. And, until recently, `gulp-sass-lint` was pulling in `sass-lint` v1.5.1. But, the `package.json` in `gulp-sass-lint` specifies `"sass-lint": "^1.3.2"`. So, since `sass-lint` v1.6.0 is now available and the '^' version spec means an upgrade is okay, so fresh installations use the newest version.

Unfortunately, it looks like something changed between v1.5.1 and v1.6.0 that doesn't agree with our `.scss` files.

lmorchard:
Bandaid fix for breakage in frontend build

Fixes #633

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,8 @@
     "gulp-eslint": "^1.0.0",
     "gulp-if": "^1.2.5",
     "gulp-imagemin": "^2.3.0",
+    "node-sass": "3.4.2",
+    "sass-lint": "1.5.1",
     "gulp-sass": "^2.0.4",
     "gulp-sass-lint": "^1.1.0",
     "gulp-sourcemaps": "^1.5.2",


lmorchard:
This seems to fix the problem for now. As a followup, we'll need to try an upgrade and see where 1) sass itself is breaking from upstream bugs and 2) our sass in particular has problems.

Edit: https://github.com/mozilla/testpilot/issues/636

lmorchard:
To test this, at a minimum you'll need to rebuild your frontend watcher container:
`docker-compose build frontend_watcher`

To be really sure, do something like this:
`docker-compose rm frontend_watcher && docker-compose build frontend_watcher`

If you're using gulp on the host side, something like this should do:
`rm -rf node_modules && npm install`

And the result should be that everything just works like it did before, and you should see no errors in the frontend build.

PR #637, opened 2016-04-25
make sure people see the privacy notice

### LINKED ISSUE #575: Display experiment-level Privacy and Legal Notices (reboot)
We closed [#501](https://github.com/mozilla/testpilot/issues/501#issuecomment-200988955) thinking that this was enough of an integration for displaying legal stuff at the experiment level. After a discussion with legal, this needs to be revisited.

This also deserves some fairly immediate attention from @johngruen as it could affect UI. **Next steps** is for John/Javaun to work together to mock up a recommendation to show legal.

Per an email from @javaun (with image):

> We just had a meeting w/ Legal/policy and revisited an ask they made a few weeks ago.  On each TxP experiment page, we have the “Measurements” (i.e. data disclosure) section in the left rail, (1.) on my screencap below. The literal ask is that the summary privacy info be in closer proximity to the “Enable” button (2.), but upon clarification the real ask of legal is that user has reasonably seen the summary privacy info at the time they enable. For illustration purposes only, that could mean that the privacy disclosure highlights when a user clicks “Enable”. The ask is also that the “friendly disclosure” is somewhere on this page, not on another linked page. 
> 
> For Activity stream, we have two docs. There is a  “[Friendly Privacy Disclosure](https://docs.google.com/document/d/1M0GgTZsrpHMRsIWnn7XBMryM7hSJOZ_ZqVvGBditcy8/edit)”. This is what they want to user to see on the page. The text seems like more than what will fit in the left rail. Then there is an advanced user [Data Dictionary](https://github.com/mozilla/activity-streams/blob/master/data_dictionary.md). This is something we can link to from the Friendly Disclosure, or bury in the description somewhere else. The Advanced/Data Dictionary shouldn’t be more prominent than the friendly statement.
> 
> We batted ideas around the week before last. The friendly disclosure is intended to be highly readable. It makes our data collection approachable to normal people. That said, the ideal solution for us is whatever looks good and is as close to zero effort as possible.

javaun image

johngruen:
Fixes #575

### DIFF
--- a/locales/en-US/app.l20n
+++ b/locales/en-US/app.l20n
@@ -93,9 +93,8 @@
 <retirePageMessage "Before you go, we want to thank you for contributing to Test Pilot.<br>You're welcome back any time!">
 <retirePageSurveyButton "Take a quick survey">
 
-<measurements "Measurements">
-<measurementsDisclaimer "All data is collected anonymously and used only to help us improve this test.">
-<experimentPrivacyNotice "{{title}} Privacy Notice">
+<measurements "Measurements &amp; Privacy">
+<experimentPrivacyNotice "You can learn more about the data collection for {{title}} here.">
 <contributorsHeading "Brought to you by">
 <detailsHeading "Details">
 <version "Version">
@@ -104,8 +103,9 @@
 <contribute "Contribute">
 <nowActive "Now Active">
 <userCount title:"Active Users of this Experiment">
-
 <tourOnboardingTitle "{{title}} enabled!">
 <tourStartButton "Take the Tour">
 <tourCancelButton "Skip">
 <tourDoneButton "Done">
+<userCountContainer "There are <span class=\"bold\">{{installation_count}}</span> people trying {{title}} right now!">
+<highlightPrivacy "Your privacy">
--- a/testpilot/frontend/static-src/app/templates/experiment-page.js
+++ b/testpilot/frontend/static-src/app/templates/experiment-page.js
@@ -12,11 +12,10 @@ export default `
           <h1 data-hook="title"></h1>
           <span class="now-active" data-hook="now-active" data-l10n-id="nowActive">Now Active</span>
           <div class="experiment-controls">
+            <a data-hook="highlight-privacy" class="highlight-privacy" data-l10n-id=highlightPrivacy>Your privacy</a>
             <button data-l10n-id="giveFeedback" data-hook="feedback" id="feedback-button" class="button default">Give Feedback</button>
             <button data-hook="uninstall" id="uninstall-button" class="button secondary"><span class="state-change-inner"></span><span data-l10n-id="disableExperimentTransition" class="transition-text">Disabling...</span><span data-l10n-id="disableExperiment" class="default-text">Disable <span data-hook="title"></span></span></button>
             <button data-hook="install" id="install-button" class="button default"><span class="state-change-inner"></span><span data-l10n-id="enableExperimentTransition" class="transition-text">Enabling...</span><span data-l10n-id="enableExperiment" class="default-text">Enable <span data-hook="title"></span></span></button>
-            <span data-l10n-id="userCount" data-hook="install-count"></span>
-            <span data-l10n-id="userCount" class="user-count"></span>
           </div>
         </div>
       </div>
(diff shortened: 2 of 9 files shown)


locales/en-US/app.l20n
```diff
@@ -104,8 +103,10 @@
 <contribute "Contribute">
 <nowActive "Now Active">
 <userCount title:"Active Users of this Experiment">
-
 <tourOnboardingTitle "{{title}} enabled!">
 <tourStartButton "Take the Tour">
 <tourCancelButton "Skip">
 <tourDoneButton "Done">
+<userCountContainerIntro "There are">
+<userCountContainerOutro "people trying {{title}} right now!">
```
pdehaan: Random L10N, but does intro+outro play nice w/ rtl languages? Or should this be a single string with another variable somewhere in the middle?

johngruen:
Fixes #575

locales/en-US/app.l20n
johngruen: I think i'd just need to pass the span in, b/c The user count comes from the model

lmorchard: Yeah, this should be put together as one string. Both the title and the count can come from the model. The problem is that other languages will need to shuffle this into an entirely different order than English

johngruen: okay fixing now

locales/en-US/app.l20n
```diff
@@ -104,8 +103,9 @@
 <contribute "Contribute">
 <nowActive "Now Active">
 <userCount title:"Active Users of this Experiment">
-
 <tourOnboardingTitle "{{title}} enabled!">
 <tourStartButton "Take the Tour">
 <tourCancelButton "Skip">
 <tourDoneButton "Done">
+<userCountContainer "There are <span data-l10n-id="userCount" class="bold" data-hook="install-count"></span> people trying <span data-hook="title"></span> right now!">
```
pdehaan: I don't know l20n, but would something like this work?

```
"There are {{userCount}} people trying {{title}} right now!"
```

Adding too much markup in .l20n files seems risky for translators.

johngruen: no

johngruen: A bit less terse, the user count number comes from the ampersand model, which gets picked up by the data-hook in the span. I could use the {{title}} template tag, but for consistency sake, I'm bringing that in from the model as well.

pdehaan: Also, because it's probably my job to make stupid comments, but do we care about the singular case?

:+1: "There are 0 people using {{title}} right now!"
:-1: "There are 1 people using {{title}} right now!" [are => is, people => person]
:+1: "There are 2 people using {{title}} right now!"

I'd almost lean towards "meh, f-it", because it's a pretty small and blunt edge and not worth the added complexity, and words are hard.
... But what do I know, I'm just one people.

---

**NOTE:** If it **_isn't_** my job to make stupid comments, I'd like to apologize for the past 2.5 years of stupidity and GIFs.

chuckharmston: GIFs are never stupid comments.

johngruen: @pdehaan can we do this as a follow-on bug :)

pdehaan: > _"GIFs are never stupid comments."_

@chuckharmston: Explain that to all the repo owners who keep seeing this GIF I keep recycling today:

lmorchard: Hmm, this is not quite working. Two problems:
1. The l20n string needs to have the quotes escaped.
2. Localized string replacement in L20N.js happens _after_ Ampersand renders, so Ampersand never gets a chance to populate these `data-hook` elements.

So, you need something more like this:

```
<userCountContainer "There are <span class=\"bold\">{{installation_count}}</span> people trying {{title}} right now!">
```

We populate the L20N variable context with all the properties of the current page model, so you can access all those from within strings.

johngruen: @lmorchard fixed, thanks!

lmorchard:
And, yeah, the plural thing matters but... ugh, I don't think we've started using that anywhere in any L20N strings. I'm okay with looking into this as a follow up issue. I might even try it myself just so I know how it works.

r+, will merge as soon as Circle CI gets done doing its thing

pdehaan:
FWIW, I was playing around w/ l20n last week and added a new locale locally to make sure everything was translated that I'd expect to be translated and everything [at the time] looked great. So I don't think we were missing any data-lion-id attributes.

pdehaan:
> I'm okay with looking into this as a follow up issue.

Filed as #639

testpilot/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -28,16 +27,19 @@ export default `
...
-                <section>
+                <section class="contributors-section">
                   <h3 data-l10n-id="contributorsHeading">Brought to you by</h3>
                   <ul class="contributors"></ul>
                 </section>
+                <section data-hook="measurements-container" class="measurements">
+                  <h3 data-l10n-id="measurements">Measurements &amp; Privacy</h3>
+                  <div data-hook="measurements-html" class="measurement"></div>
+                  <a class="privacy-policy" data-l10n-id="experimentPrivacyNotice" data-hook="privacy-notice-url">You can learn more about the data collection for <span data-hook="title"></span> here.</a>
+                </section>
```
pdehaan, line 42: @johngruen: Just looking at this now on dev.mozaws, but why did we swap the order of contributors and privacy?

Not that I'm arguing that data privacy is more important than a bunch of avatars... but data privacy seems more important than a bunch of avatars.

meandavejustice, line 42: lol agreed.

johngruen, line 42: @pdehaan glad you asked. Importance is a relative term. The heaviness of the privacy stuff made the layout of the page unbalanced visually when it was right up top. It also sort of winds up looking like threatening legal bullshit. Putting it lower and combining it with the very clear link right next to the primary call to action makes it less in your face, while making it easily discoverable.

pdehaan, line 42: Bury privacy notices in the bowels of the webpage. Got it.

johngruen, line 42: @pdehaan snark-o-meter index: ∞

pdehaan, line 42: @johngruen

PR #679, opened 2016-05-02
Handle manual enabling/disabling by the user

### LINKED ISSUE #670: Test Pilot thinks disabled add-ons are active
This is a little edge-casey but was showed to me by someone who had old versions of tab center and activity stream installed but disabled, and then installed Test Pilot who saw the add-ons were installed and said "Now Active" for them.

So, basically we shouldn't be looking at if an add-on is installed, but if it is installed _and active._

meandavejustice:
- fixes #670
- adds `addon-manage:enabled` and `addon-manage:disabled` messages so we can handle these edge
  cases on the front end as well.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -246,7 +246,9 @@ function getExperimentList(availableExperiments, installedAddons) {
   return Mustache.render(templates.experimentList, {
     base_url: settings.BASE_URL,
     experiments: Object.keys(availableExperiments).map(k => {
-      availableExperiments[k].active = Boolean(installedAddons[k]);
+      if (installedAddons[k]) {
+        availableExperiments[k].active = installedAddons[k].active;
+      }
       return availableExperiments[k];
     })
   });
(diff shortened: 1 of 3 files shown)


addon/index.js
```diff
@@ -445,6 +448,28 @@ function requestAPI(opts) {
...
 const addonListener = {
+  onEnabled: function(addon) {
+    if (isTestpilotAddonID(addon.id)) {
+      store.installedAddons[addon.id] = addon;
+      app.send('addon-manage:enabled', {
+        id: addon.id,
+        name: addon.name,
+        version: addon.version
+      });
+      Metrics.experimentEnabled(addon.id);
```
lmorchard, line 459: Oh, good catch on the metrics calls, I think I would have missed them

lmorchard:
Small merge conflict in the addon version number. I went ahead and resolved it:

a61da8edbfb7c68868825467e5fbae77d94e5a31

Issue #615, opened 2016-04-19
Make sure all the things are GA'd


ckprice:
We gave GA an initial pass in #577.

This issue is opened to make sure all of the functionality landed since then are also wired correctly.
- [x] #614 retire flow
- [x] #438 onboarding
- [ ] #609 subscribe/unsubscribe to emails
- [x] #611 surveygizmo

lmorchard:
We should think about adding this as a concern to every new feature. I'm not sure what our GA strategy is, what we want to measure, etc. This stuff should also probably be documented in [docs/README-METRICS.md](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md) in terms of what we specifically measure and why, both for transparency and so we're all on the same page about the GA strategy

ckprice:
We're keeping track of some of these things in the [GA Reqs document](https://docs.google.com/document/d/1SIOaZLHsMGNBNmAxphon4cxF_F07Q1ctBkU6gkP2SD0/edit?ts=56fc371c). Once this is cleaned up we'll be sure to add to the readme.

meandavejustice:
@ckprice I'm going to try and get this finished up today. #438 #609 and #611 are done, but none of them included a ping to GA. To finish this up. I'm going to go through the GA reqs doc linked above and make sure we have the calls in proper places and update readme-metrics.

ckprice:
Closed via #690.

PR #681, opened 2016-05-02
[fixes #609] Add subscribe and unsubscribe to basket

### LINKED ISSUE #609: Send subscribe and unsubscribe pings to basket on install and disable
References:
- [Bug 1263675](https://bugzilla.mozilla.org/show_bug.cgi?id=1263675) `Email integration with Test Pilot` [**blocker**] we need the newsletter ID to get this bug going (/cc @bniolet1).
- #505 `Add the ability to email Test Pilot users when new experiments become available` lots of breakdown/discussion here, as well as basket integration specifications.
- #499 `Implement 'Retire' flow to allow users to uninstall Test Pilot` unsubscribe ping will have to be sent when the user retires.
- #589 `Design email template for new experiment notifications` @johngruen will be designing the email, and our email marketing agency will be responsible for coding/deploying it. Note: _not a blocker_ for launch, but added to the [post-launch milestone](https://github.com/mozilla/testpilot/milestones/2016-05-20%20%28post-launch%29) as a priority.

ckprice:
Requires that `BASKET_API_KEY` be added to the environment.

This is still pretty rough, but ready for a look from @meandavejustice or @lmorchard.

Ref: http://basket.readthedocs.io/newsletter_api.html

### DIFF
--- a/testpilot/frontend/static-src/app/main.js
+++ b/testpilot/frontend/static-src/app/main.js
@@ -79,6 +79,22 @@ app.extend({
     } else {
       hitCallback();
     }
+  },
+
+  // Subscribe to basket
+  subscribeToBasket(email, callback) {
+    const url = 'https://basket.mozilla.org/news/subscribe/';
+    fetch(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded'
+      },
+      body: 'newsletters=test-pilot&email=' + email
+    }).then(callback)
+    .catch(err => {
+      // for now, log the error in the console & do nothing in the UI
+      console && console.error(err); // eslint-disable-line no-console
+    });
   }
 });
 
--- a/testpilot/frontend/static-src/app/views/landing-page.js
+++ b/testpilot/frontend/static-src/app/views/landing-page.js
@@ -47,11 +47,13 @@ export default PageView.extend({
     this.query('[data-hook=install]').classList.add('state-change');
     this.query('.default-btn-msg').classList.add('no-display');
     this.query('.progress-btn-msg').classList.remove('no-display');
-    app.sendToGA('event', {
-      eventCategory: 'HomePage Interactions',
-      eventAction: 'button click',
-      eventLabel: 'Install the Add-on',
-      outboundURL: downloadUrl
+    app.subscribeToBasket(app.me.user.id, () => {
+      app.sendToGA('event', {
+        eventCategory: 'HomePage Interactions',
+        eventAction: 'button click',
+        eventLabel: 'Install the Add-on',
+        outboundURL: downloadUrl
+      });
     });
 
     const interval = setInterval(() => {
--- a/testpilot/users/models.py
+++ b/testpilot/users/models.py
@@ -1,8 +1,10 @@
 from django.db import models
 from django.contrib.auth.models import User
+from django.conf import settings
 
 from ..utils import HashedUploadTo
 
+import requests
 
 avatar_upload_to = HashedUploadTo('avatar')
 
@@ -35,5 +37,30 @@ class UserProfile(models.Model):
 
     def natural_key(self):
         return self.user.natural_key()
+
+    def unsubscribe(self):
+        """Unsubscribe this email from basket. Lookup-user call for token, then the unsubscribe"""
+        try:
+            if settings.BASKET_API_KEY is None:
+                return 'API Key not found'
+            email = self.user.email
+            url = settings.BASKET_LOOKUP_USER_URL
+
+            params = {'api_key': settings.BASKET_API_KEY, 'email': email}
+            r = requests.get(url, params=params)
+            resp = r.json()
+
+            if 'token' not in resp:
+                return'email not found in basket'
+            user_token = resp['token']
+
+            url = settings.BASKET_UNSUBSCRIBE_URL + user_token + '/'
+            payload = {'newsletters': 'test-pilot', 'email': email}
+            r = requests.post(url, data=payload)
+            resp = r.json()
+
+            return resp
+        except Exception as e:
+            return e
 
     natural_key.dependencies = ['auth.user']
--- a/testpilot/users/views.py
+++ b/testpilot/users/views.py
@@ -47,7 +47,8 @@ class MeViewSet(ViewSet):
 @permission_classes([IsAuthenticated])
 def user_retire(request):
     user = request.user
-    result = {"id": user.id, "username": user.username}
+    profile = UserProfile.objects.get_profile(user)
+    result = {'id': user.id, 'username': user.username, 'unsubscribe': profile.unsubscribe()}
     logout(request)
     user.delete()
     return Response(result, status=status.HTTP_200_OK)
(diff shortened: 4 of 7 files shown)


testpilot/users/views.py
```diff
@@ -42,12 +42,12 @@ def list(self, request):
                 context={'request': request}).data
         })
 
-
```
meandavejustice, line 45: pep8 is failing because of this

testpilot/frontend/static-src/app/main.js
```diff
@@ -79,6 +79,15 @@ app.extend({
...
     }
+  },
+
+  // Subscribe to basket
+  subscribeToBasket(email) {
+    const url = 'https://basket.mozilla.org/news/subscribe/';
+    const params = 'newsletters=test-pilot&email=' + email;
+    const request = new XMLHttpRequest();
+    request.open('POST', url);
+    request.send(params);
```
pdehaan: We're using the `fetch()` api elsewhere, not sure if it makes sense to use here too.
https://github.com/mozilla/testpilot/blob/c4c41e893120132639347e41620fc84d054670f4/testpilot/frontend/static-src/app/views/settings-menu-view.js#L64

lmorchard: Yeah, this should probably use the `fetch()` API. It wraps XMLHttpRequest in a nicer API. There's native support in a bunch of browsers, and we've imported a polyfill for the ones that don't have support.

ckprice: Cool will do thanks @pdehaan and @lmorchard

lmorchard:
> Requires that BASKET_API_KEY be added to the environment.

Is there a semi-throwaway API key that we can share publicly and use for local dev machines (i.e. is the staging key okay for this)? [For example, we have some credentials rolled into the Docker environment so it works out of the box with FxA](https://github.com/mozilla/testpilot/blob/master/docker-compose-base.yml#L37). That's where this API key needs to be added to the environment for local Docker containers.

If not, we should consider making this stuff a no-op [in the default case](https://github.com/mozilla/testpilot/pull/681/files#diff-461aaf29fb21fa9e18300c35ee7853ccR152) where the Basket API key will be `None`

testpilot/users/models.py
```diff
@@ -36,4 +38,26 @@ class UserProfile(models.Model):
     def natural_key(self):
         return self.user.natural_key()
 
+    def unsubscribe(self):
```
lmorchard, line 41: This model method could use a test with the Basket API mocked out. For example, [here's how we test code checking whether a user is a vouched Mozillian against a mocked-out mozillians.org API](https://github.com/mozilla/testpilot/blob/c4c41e893120132639347e41620fc84d054670f4/testpilot/users/tests.py#L304)

meandavejustice:
also need to add call to `app.sendToGA` before merging so we can close out https://github.com/mozilla/testpilot/issues/615

ckprice:
> Is there a semi-throwaway API key that we can share publicly and use for local dev machines (i.e. is the staging key okay for this)?

@lmorchard so I spoke briefly to @pmclanahan and it turns out we don't really have this right now. We could work with him to set up something for the basket-dev environment, but it's not totally sync'd with stage/prod and probably wouldn't have our newsletter. I think this may also put a wrench into your [testing comment](https://github.com/mozilla/testpilot/pull/681#discussion_r61761200) as well.

Could we go to launch without these?

lmorchard:
> @lmorchard so I spoke briefly to @pmclanahan and it turns out we don't really have this right now. We could work with him to set up something for the basket-dev environment, but it's not totally sync'd with stage/prod and probably wouldn't have our newsletter.

Then we need some conditionals to disable the Basket calls when the API key is `None`, since that will be the default case for local dev.

> I think this may also put a wrench into your testing comment as well.

It shouldn't... The test should mock out the requests to the API, so it shouldn't need real API keys because it won't actually be hitting the Basket API. It should just exercise our own code, assert that the expected HTTP requests were attempted and expected things were done with the mock results. [That's what we do in the test that exercises code using the Mozillians API](https://github.com/mozilla/testpilot/blob/c4c41e893120132639347e41620fc84d054670f4/testpilot/users/tests.py#L304) - it never actually calls out to the network.

ckprice:
Updated
1. [x] Check for existence of API key in env before unsubscribe.
2. [x] Switch to `fetch()`.
3. [x] Call `sendToGA` when "Retire" called from settings menu.
4. [x] Call `sendToGA` when the user confirms in modal.
5. [x] Add callback to `subscribeToBasket` because things were getting weird without it.
6. [ ] Add unit test per [#issuecomment-216365309](https://github.com/mozilla/testpilot/pull/681#issuecomment-216365309).

testpilot/frontend/static-src/app/main.js
```diff
@@ -79,6 +79,23 @@ app.extend({
...
+  // Subscribe to basket
+  subscribeToBasket(email, callback) {
+    const url = 'https://basket.mozilla.org/news/subscribe/';
+    fetch(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded'
+      },
+      body: 'newsletters=test-pilot&email=' + email
+    }).then(function subscribeReturn() {
```
lmorchard: Could this line maybe just be `}).then(callback)`? Or does the callback just absolutely have to receive no parameters?

testpilot/frontend/static-src/app/main.js
```diff
@@ -79,6 +79,23 @@ app.extend({
...
+      callback();
+    }).catch((err) => {
```
lmorchard: Nit: if you have a single parameter, you can leave the parens out of the arrow function like so:

```
}).catch(err => {
```

testpilot/frontend/static-src/app/views/landing-page.js
```diff
@@ -47,11 +47,13 @@ export default PageView.extend({
     this.query('[data-hook=install]').classList.add('state-change');
     this.query('.default-btn-msg').classList.add('no-display');
     this.query('.progress-btn-msg').classList.remove('no-display');
-    app.sendToGA('event', {
-      eventCategory: 'HomePage Interactions',
-      eventAction: 'button click',
-      eventLabel: 'Install the Add-on',
-      outboundURL: downloadUrl
+    app.subscribeToBasket(app.me.user.id, function subscribeCallback() {
```
lmorchard: If the user is already subscribed, does this result in an error and no GA event?

Also a nit: could use an arrow function here, eg. `() => {` instead of `function subscribeCallback() {`

ckprice:
Added @lmorchard's comments. Haven't started digging in (learning) the Mock stuff. Anyway we can slide this in and open a PR for that post-launch? 😬

lmorchard:
I can write the test, if that helps. It's the kind of stuff I'd rather not let slip under the rug because we will never get around to it later.

testpilot/frontend/static-src/app/views/landing-page.js
ckprice: > If the user is already subscribed, does this result in an error and no GA event?

Nope, response from basket is the same.

lmorchard:
Closing in favor of https://github.com/mozilla/testpilot/pull/697

PR #689, opened 2016-05-02
Addon install message follow up

### LINKED ISSUE #659: Follow up with addon install UI (double `Let's go` buttons)
We are _still_ having some inconsistency with showing the dialog when the addon is installed.
When this happens I'm not seeing anything in either the browser toolbox or browser console.
This means one of two things:
1. I am overlooking a race condition/hazard
2. There is a serious bug in [unsafeWindow](https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/Content_Scripts/Interacting_with_page_scripts).    ([our usage](https://github.com/mozilla/testpilot/blob/9f323963554d2267de4dc0e84b84ce58a9b75e7b/addon/data/message-bridge.js))

Going ahead with @johngruen 's fix in #656 for the time being, Until I get some extra time to take a look at this.

Since users can get into site with #656 without a refresh. The priority is low.

meandavejustice:
- fixes #659
- remove "Let's go" site link from installed message panel
- adjust installed message panel height
- bump `addon/package.json` version to `0.5.4`
- remove call to `Mustache.render` for installed template,
  since we no longer have data passed to it.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -41,7 +41,7 @@ const survey = require('./lib/survey');
 const PANEL_WIDTH = 300;
 const FOOTER_HEIGHT = 50;
 const EXPERIMENT_HEIGHT = 80;
-const INSTALLED_PANEL_HEIGHT = 370;
+const INSTALLED_PANEL_HEIGHT = 300;
 
 // Canned selectable server environment configs
 const SERVER_ENVIRONMENTS = {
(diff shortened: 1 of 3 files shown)


lmorchard:
r+ from me & good to merge if the UX is copacetic

johngruen:
checking

PR #694, opened 2016-05-02
Enable full Telemetry on install/enable

### LINKED ISSUE #655: Enable Full Telemetry for users who enroll in Test Pilot
TL;DR tasks:
- [x] Preserve initial values of `toolkit.telemetry.enabled` and `datareporting.healthreport.uploadEnabled` preferences.
- [x] Set those two prefs to `true` on main add-on install / enable
- [x] Restore initial values of those two prefs on main add-on uninstall / disable

---

To enable us to answer behavioral questions about how users interact w/ experiments and whether they actually change browsing behavior, we want to enroll all Test Pilot users in full telemetry (opt-out + opt-in). 

Some measure of this is required for v1 launch.

**ENG Requirements**
- Full data: Test Pilot users send full Telemetry dataset to Telemetry system. (TBD on whether we just flip opt-in Telemetry on, or whether we send this data some other way, such as stuffing it into the custom Test Pilot Ping)
- Un-enrolling: If a user completely leaves Test Pilot, we need to restore their Telemetry to the prior state.
- Privacy needs: Telemetry ID cannot be mapped to their FxA account. (Illustrative purposes only: This might be achieved multiple ways, for example via siloed data storage, or by injecting some new identifier in place of the Telemetry ID, i.e. a one-way hash) 

**Corner cases:**
- If a user uninstalls Test Pilot via about:add-ons, we still need to clean up their Telemetry settings (if possible). 
- At launch, if a user manually turns off Telemetry (opt-out, opt-in, etc.), Test Pilot still works, Mozilla just doesn’t get data. (specifying this as a test case only, since I assume this is the current behavior. If this is widely abused, we may have to build a feature to notify users that Test Pilot requires telemetry)

lmorchard:
- Preserve original preferences on enable / install, then overrides
  preferences to enable full Telemetry collection. Original preferences
  are restored on disable / uninstall.
- Bump addon to v0.5.4

Closes #655

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -538,12 +538,19 @@ const installListener = {
 };
 AddonManager.addInstallListener(installListener);
 
-exports.main = function() {
+exports.main = function(options) {
+  const reason = options.loadReason;
+
   if (!store.clientUUID) {
     // Generate a UUID for this client, so we can manage experiment
     // installations for multiple browsers per user. DO NOT USE IN METRICS.
     store.clientUUID = require('sdk/util/uuid').uuid().toString().slice(1, -1);
   }
+
+  if (reason === 'install' || reason === 'enable') {
+    Metrics.onEnable();
+  }
+
   initServerEnvironmentPreference();
   Metrics.init();
 };
@@ -555,6 +562,11 @@ exports.onUnload = function(reason) {
   button.destroy();
   Metrics.destroy();
   survey.destroy();
+
+  if (reason === 'uninstall' || reason === 'disable') {
+    Metrics.onDisable();
+  }
+
   if (reason === 'uninstall') {
     if (store.installedAddons) {
       Object.keys(store.installedAddons).forEach(id => {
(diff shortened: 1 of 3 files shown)


lmorchard:
Hopefully this does it. Some unnecessarily verbose manual verification steps:
1. Uninstall Test Pilot
2. Go to about:config, search for the `toolkit.telemetry.enabled` and `datareporting.healthreport.uploadEnabled` preferences, set them each to `false`
3. Install Test Pilot
4. Go to about:config, search for the preferences, they should both be `true`
5. Disable Test Pilot in about:addons
6. Go to about:config, search for the preferences, they should both be `false`
7. Enable Test Pilot in about:addons
8. Go to about:config, search for the preferences, they should both be `true`
9. Uninstall Test Pilot in about:addons
10. Go to about:config, search for the preferences, they should both be `false`

pdehaan:
Trying to verify this, but I'm getting blocked by https://github.com/mozilla/testpilot/issues/703 on testpilot.dev.mozaws.net.

pdehaan:
@ckprice: Verified @lmorchard's steps from https://github.com/mozilla/testpilot/pull/694#issuecomment-216359670 on http://testpilot.dev.mozaws.net/ :+1:

PR #695, opened 2016-05-03
s/test/experiment/.  Also add docs about schemas

clouserw:
Just some updates to the docs.  We're calling them "experiments" now and I put in that schema from https://bugzilla.mozilla.org/show_bug.cgi?id=1255184#c15


lmorchard:
lgtm, r+

PR #697, opened 2016-05-03
Add subscribe and unsubscribe to basket

lmorchard:
This is PR #681 with added tests and a few small tweaks


lmorchard:
Will probably merge this after (hopefully) Circle CI passes. Added a handful of tests to make sure to run the unsubscribe method through all its paces

PR #705, opened 2016-05-04
Bug fix for sendToGA call in landing-page

lmorchard:
- Check for existence of app.me.installed and use key count instead of
  `length`
- Tests to trigger the error and exercise other conditional paths

Fixes #703

PR #726, opened 2016-05-05
Send installed experiments to external survey

### LINKED ISSUE #667: Finalize all survey content
Survey content for experiments is currently just filler data. Final content will come before launch.

We'll also need content for the retire survey.

meandavejustice:
- fixes #667
- bump addon version to 0.5.5
- sends installed experiments to external survey under key `installed`

cc/ @ckprice

### DIFF
--- a/testpilot/frontend/static-src/app/views/experiment-page.js
+++ b/testpilot/frontend/static-src/app/views/experiment-page.js
@@ -152,12 +152,6 @@ export default PageView.extend({
       type: 'attribute',
       name: 'src',
       hook: 'thumbnail'
-    },
-
-    'model.survey_url': {
-      type: 'attribute',
-      name: 'href',
-      hook: 'feedback'
     }
   },
 
@@ -273,7 +267,10 @@ export default PageView.extend({
       eventLabel: 'Disable Experiment'
     });
 
-    const queryParams = querystring.stringify({ref: 'disable', experiment: this.model.title});
+    const installed = Object.keys(app.me.installed);
+    const queryParams = querystring.stringify({ref: 'disable',
+                                               experiment: this.model.title,
+                                               installed: installed});
 
     this.uninstall(evt);
 
@@ -311,12 +308,18 @@ export default PageView.extend({
     this.didScroll = false;
   },
 
-  feedback() {
-    // Survey link is opened via href link in the template
+  feedback(evt) {
+    evt.preventDefault();
+    const installed = Object.keys(app.me.installed);
+    const queryParams = querystring.stringify({ref: 'givefeedback',
+                                               experiment: this.model.title,
+                                               installed: installed});
+
     app.sendToGA('event', {
       eventCategory: 'ExperimentDetailsPage Interactions',
       eventAction: 'button click',
-      eventLabel: 'Give Feedback'
+      eventLabel: 'Give Feedback',
+      outboundURL: `${this.model.survey_url}?${queryParams}`
     });
   }
 });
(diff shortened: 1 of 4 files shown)


ckprice:
FWIW I think this is a pretty lower priority issue in general.

That said, here is my testing with _Activity Stream and Universal Search enabled_.

## 

Click `Give Feedback` from Activity Stream page =>

https://qsurvey.mozilla.com/s3/activity-stream?ref=givefeedback&experiment=Activity%20Stream&installed=universal-search-addon%40mozilla.com&installed=%40activity-streams

Note: `&installed=universal-search-addon%40mozilla.com&installed=%40activity-streams`

## 

Click `Give Feedback` from Universal Search page =>

https://qsurvey.mozilla.com/s3/universal-search?ref=givefeedback&experiment=Universal%20Search&installed=universal-search-addon%40mozilla.com&installed=%40activity-streams

Note: `&installed=universal-search-addon%40mozilla.com&installed=%40activity-streams`

## 

Click `Take survey` after Disable from Universal Search page =>

https://qsurvey.mozilla.com/s3/universal-search?ref=disable&experiment=Universal%20Search&installed=universal-search-addon%40mozilla.com&installed=%40activity-streams

Note: `&installed=universal-search-addon%40mozilla.com&installed=%40activity-streams`

## 

I don't know if this is the expected/desired behaivor, but it looks funky to me. @mattgrimes?

Also there appears to be some [`app.me.installed` stuff](https://github.com/mozilla/testpilot/pull/726/commits/74a3c5632c059a0c9b4da52e064136a58022c63a#diff-59f5e52c87ff47eddc3834d7e5293990R266) happening which I failed to review sufficiently in the last PR.

I'm going to unassign myself for now.

MattGrimes:
@ckprice that all looks correct to me. That way if you leave feedback about Activity Steam and say it's making Firefox slow we can see if you have other experiments that might be causing conflicts. Are you just concerned that the URL's are long? I don't think that's an issue personally.

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -273,7 +267,10 @@ export default PageView.extend({
       eventLabel: 'Disable Experiment'
     });
 
-    const queryParams = querystring.stringify({ref: 'disable', experiment: this.model.title});
+    const installed = Object.keys(app.me.installed);
```
lmorchard, line 270: Wait, [I thought `app.me.installed` was an array](https://github.com/mozilla/testpilot/pull/705#issuecomment-216991318).

lmorchard, line 270: Oh, I see what's up... `app.me.installed` gets set first by [the server-side API result (as an Array)](https://github.com/mozilla/testpilot/blob/master/testpilot/users/views.py#L40) and later by [a message from the add-on (as an Object)](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/models/me.js#L45). I think we need to change one of these sources to be consistent with the other

meandavejustice, line 270: @lmorchard yeah, I thought so too, initially. I think that https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/models/me.js#L16 might still be forcing it to into an object.

lmorchard:
Hmm, I'll go ahead and merge this since it seems to be working. I think the add-on always wins via `sync-installed` at the end of the `app.me.fetch()` call. I'll throw together a fix for the server-side of things as a followup

meandavejustice:
@lmorchard I see what you mean. I'll investigate that possible race condition as well

I've opened #742 to follow up with this inconsistent data shape

Issue #717, opened 2016-05-05
Signing up for a new Firefox Account leaves you at a deadend


ghost:
This was reported to me:

If you hit Test Pilot as a new user with no Firefox Account, and go through the sign up flow, once you confirm your email address you are shown a dead end page.

I asked rfkelly and he said we should be passing in verification_redirect=always in our oauth request query params.

Not to side-track my own issue, but while I'm here:  it looks like we are using django-allauth,  but then copy/paste the FxA specific code into our code instead of using the version in django-allauth: https://github.com/mozilla/testpilot/tree/master/testpilot/users/providers/fxa .  Why is that?

lmorchard:
> Not to side-track my own issue, but while I'm here: it looks like we are using django-allauth, but then copy/paste the FxA specific code into our code instead of using the version in django-allauth: https://github.com/mozilla/testpilot/tree/master/testpilot/users/providers/fxa . Why is that?

Trying to dig up the history to say for sure, but off the top of my head: FxA in django-allauth didn't work out of the box, and [we needed configurable endpoints](https://github.com/mozilla/testpilot/blob/master/testpilot/users/providers/fxa/views.py#L19) to use non-prod FxA for local/dev/staging. 

Looks like allauth might support those things now ([yup, as of Feb 16](https://github.com/pennersr/django-allauth/commit/80351a2da6aa7c3b7a2e5221453676da6bf68277)), so maybe we can back out this local custom code at some point.

pdehaan:
This feels weird to me, considering I signed up to stage+prod today with a shiny new restmail.net account and don't remember seeing this.
I'll try again.

ghost:
Thanks.  Original report from @ckarlof .

pdehaan:
@ckarlof I'm unable to reproduce this on stage with my shiny new ckarl@restmail FxA account.

---

https://testpilot.stage.mozaws.net/__version__ (ca8af43986ea2e0f8f8ecb08625b26f1aae7a6e5):

``` json
{ "build": "https://circleci.com/gh/mozilla/testpilot/325",
  "commit": "ca8af43986ea2e0f8f8ecb08625b26f1aae7a6e5",
  "source": "https://github.com/mozilla/testpilot",
  "version": "2016-05-04" }
```

ckarlof:
Still feels like a dead end to me. What do I click on to get back to Test Pilot?

[...]

PR #731, opened 2016-05-06
Replace old testpilot.base.html template to fix URL route errors

lmorchard:
This does not fix issue #717, but it does at least fix the Django errors that occur when the site attempts to render the django-allauth authentication_error page on failed OAuth flows.


lmorchard:
Closing, because I think I might add a redirect to home page on the error page. That might help resolve most common errors (e.g. caused by simultaneous signins after email verification)

PR #770, opened 2016-05-11
fix terms link to use non-specific locale

learnedhoof:


pdehaan:
:+1: Good catch, thanks!

PR #790, opened 2016-05-11
Install django-waffle and django-constance

### LINKED ISSUE #506: Implement Django admin access to control some server settings
Most Django server settings are related to ops environment and connections to other services. Some settings are related to product features, and not necessarily ops concerns - e.g. [ACCOUNT_AUTOACTIVATION_DOMAINS](https://github.com/mozilla/testpilot/blob/master/testpilot/settings.py#L159) and [ACCOUNT_INVITE_ONLY_MODE](https://github.com/mozilla/testpilot/blob/master/testpilot/settings.py#L157). 

However, _all_ settings are currently controlled by server env vars, which require ops intervention to change. So, we might want to make _just_ product features (and not ops environment) easier to control via the Django admin.

Two options that could help with this in different ways:
- [django-constance](http://django-constance.readthedocs.org/en/latest/) - Django admin control panel to manage a defined subset of server settings
- [django-waffle](https://github.com/jsocol/django-waffle) - feature flipping for the site itself (as opposed to experiments)

lmorchard:
Install django-waffle and django-constance
- Add dependencies to requirements.txt
- Install django-constance settings and backend
- Install django-waffle settings, URLs, and JS data in frontend template

Fixes #506

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -158,3 +158,12 @@ jsonschema==2.5.1
 
 # sha256: 2jgtD0CR_QWvc0NNXdXZA1hnZB5sUMcEtTLes4JAf-M
 django-colorfield==0.1.10
+
+# sha256: kUtLh0-mJQoIl7wwtn4CA02SpyNatyqRv22km3F1HeQ
+django-waffle==0.11
+
+# sha256: VIn-8WTeQ3JSQtVuZeAWE3098NGgBnK9py2Af1srDZk
+django-picklefield==0.3.2
+
+# sha256: jtumjPR776LS5tnrqY_3TcmIHghr_KwDN_6BmNXt_78
+django-constance==1.1.2
--- a/testpilot/frontend/templates/testpilot/frontend/index.html
+++ b/testpilot/frontend/templates/testpilot/frontend/index.html
@@ -16,6 +16,7 @@
 
   <div data-hook="page-container"></div>
 
+  <script src="{{ url('wafflejs') }}"></script>
   <script src="{{ static('app/app.js') }}"></script>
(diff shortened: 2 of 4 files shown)


lmorchard:
[This commit may come in handy later](https://github.com/lmorchard/testpilot/commit/7f0525f424f07a0ff10c20fde14c68e5a2057d63), if only as an example of how to use this stuff. But, I snipped it out of this PR because we've taken the query string gate away for good and don't need a switch to manage it.

You should see the django-waffle addition to the admin pages, but there won't be anything that uses it yet.

lmorchard:
Also just tossed in django-constance, which will add another admin section, but also blank because nothing's using it yet. Overall, this stuff should add new runtime configuration options and not break anything.

requirements.txt
```diff
@@ -158,3 +158,12 @@ jsonschema==2.5.1
 
 # sha256: 2jgtD0CR_QWvc0NNXdXZA1hnZB5sUMcEtTLes4JAf-M
 django-colorfield==0.1.10
+
+# sha256: kUtLh0-mJQoIl7wwtn4CA02SpyNatyqRv22km3F1HeQ
+django-waffle==0.11
+
+# sha256: VIn-8WTeQ3JSQtVuZeAWE3098NGgBnK9py2Af1srDZk
+django-picklefield==0.3.2
```
meandavejustice, line 166: is this being used for anything?

lmorchard, line 166: Yeah, that's a pre-requisite for django-constance and its database backend

meandavejustice, line 166: cool, thanks. This looks good to me and tests are passing. so I'm merging

Issue #810, opened 2016-05-12
Triage of recent outage and changes to make to improve scaling


mostlygeek:
While triaging the recent outage we discovered that: 
- when `DEBUG=false` 
  - the app returns a 500 error code for all requests. This includes `/` and `/__lbheartbeat__` which means it can never join the load balancer.
  - whitenoise defaults to `Cache-Control: public, max-age=0` for static assets which disables nginx proxy caching. 
- ~~~the app appears to be single threaded, able to handle only a single request at a time. When the python process is overloaded the `__lbheartbeat__` endpoint becomes unresponsive and the server is removed from the load balancer and terminated.~~~

Changes to make: 
- [x] Determine and fix the app to work in when `DEBUG=false`
- [x] Provide a better default for static assets, even 60s would improve the performance of the site substantially 
- [x] ~~Fix why we basically get single threaded concurrency. Is it not an evented engine in the docker container? Can we configure how many child processes to start up?~~
- [x] ~~On [this line](https://github.com/mozilla/testpilot/blob/master/bin/run-prod.sh#L4) it should be `exec gunicorn...` so bash replaces itself and process signals can be sent to gunicorn instead of being lost~~ (ref #816)

cc @lmorchard @relud

mostlygeek:
I figuree out the concurrency issue and will fix on the deployment side. Basically set `WEB_CONCURRENCY` to 4\* number of CPUs in the environment.

chuckharmston:
The default value for `WHITENOISE_MAX_AGE` is `60 if not settings.DEBUG else 0`, so it seems likely that being able to turn `DEBUG` to `False` will solve the problems.

http://whitenoise.evans.io/en/stable/django.html#WHITENOISE_MAX_AGE

That said, there's no reason we couldn't set it to `60 * 60 * 24` or something moderately aggressive, since the upstream cache is flushed with each deploy.

mostlygeek:
> That said, there's no reason we couldn't set it to 60 \* 60 \* 24 or something moderately aggressive, since the upstream cache is flushed with each deploy.

The `Cache-Control` value is sent to the user as well, so we have to consider the browser's cache too. I think 60s will already make a huge performance improvement on the server side.

lmorchard:
This Django app is not evented and I don't think multi-threading support is great in Django & Python - though my understanding might be outdated.

[...]

PR #816, opened 2016-05-12
Switch to `exec gunicorn` on prod for better signal handling

lmorchard:
> <mostlygeek> could you make it "exec gunicorn ..." instead
> <mostlygeek> that will cause bash to replace itself with gunicorn and
>              we can send SIGHUP/SIGKILL/SIG\* to it without any
>              timeouts

Issue #810

### DIFF
--- a/bin/run-prod.sh
+++ b/bin/run-prod.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 ./bin/run-common.sh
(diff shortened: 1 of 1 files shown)


lmorchard:
Will probably just self-merge this after CircleCI passes

mostlygeek:
exactly what I wanted. :)

lmorchard:
I'll just merge this and we can keep iterating as needed

PR #833, opened 2016-05-12
fix milliseconds calculation

clouserw:
Pretty sure this is wrong

### DIFF
--- a/testpilot/base/middleware.py
+++ b/testpilot/base/middleware.py
@@ -19,7 +19,7 @@ class RequestSummaryLogger(object):
 
     def _build_extra_meta(self, request):
         td = datetime.datetime.utcnow() - request._logging_start_dt
-        t = int(td.total_seconds() * 10**6)
+        t = int(td.total_seconds() * 1000)  # in ms
         return {
             "errno": 0,
             "agent": request.META.get('HTTP_USER_AGENT', ''),


testpilot/base/middleware.py
```diff
@@ -19,7 +19,7 @@ def process_request(self, request):
 
     def _build_extra_meta(self, request):
         td = datetime.datetime.utcnow() - request._logging_start_dt
-        t = int(td.total_seconds() * 10**6)
+        t = int(td.total_seconds() * 1000) # in ms
```
lmorchard: The `10**6` thing ([nanoseconds since epoch](https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md#top-level-properties)) is specifically what ops asked for.

lmorchard:
Comment got hidden with the commit update: The `10**6` thing ([nanoseconds since epoch](https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md#top-level-properties)) is specifically what ops asked for.

testpilot/base/middleware.py
clouserw: I see what you're saying, but that's for the `Timestamp` field.  This is `t` which is defined further down the page in milliseconds

clouserw:
Also hidden :) :  I see what you're saying, but that's for the Timestamp field. This is t which is defined further down the page in milliseconds

lmorchard:
Ohhhhhh... derp.

PR #834, opened 2016-05-12
Fix 500 errors for prod when DEBUG=False

lmorchard:
- Use STATIC_URL setting to construct locale URL, because static()
  causes 500 ISEs when the path doesn't lead to a real file.

Issue #810

### DIFF
--- a/testpilot/frontend/templates/testpilot/frontend/index.html
+++ b/testpilot/frontend/templates/testpilot/frontend/index.html
@@ -9,7 +9,7 @@
     <meta name="defaultLanguage" content="en-US">
     <meta name="availableLanguages" content="en-US">
     <meta name="viewport" content="width=device-width">
-    <link rel="localization" href="{{ static('locales/{locale}/app.l20n') }}">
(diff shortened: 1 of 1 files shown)


lmorchard:
Updated to resolve conflict with #833

chuckharmston:
f+, I guess? Code looks great, but I don't have my head around exactly how all of these things added up to the errors we were seeing. Feel free to self-r+ and :shipit: *

\* If you can get the build passing

lmorchard:
Er, that's an odd test failure. 😞 Oh, but I can reproduce locally, so all is not lost

lmorchard:
Actually it was probably good to look through this thing again... I found the 500 error caused by `static()` last, after a handwavy attempt to upgrade whitenoise. Turns out the whitenoise upgrade was probably a bad idea. Seems that caused caused the 500 errors in the logging middleware and now the test failure.

So... TL;DR: I stripped the commit down to just the `static()` fix, which seems to have done the trick. Now, let's see if Circle CI is happy with me

mostlygeek:
I deployed this in stage and the 500 errors still exist.

PR #846, opened 2016-05-13
Further fixes for 500 errors on prod when DEBUG=False

lmorchard:
- Bug fixes for 500 ISEs caused by logging middleware errors.
- Use STATIC_URL setting to construct locale URL, because static()
  causes 500 ISEs when the path doesn't lead to a real file.

Issue #810


lmorchard:
This includes a fix I pulled out of an earlier PR, plus a second spot with the same fix from my previous PR. Planning to self-merge this if/when CircleCI passes and see if this fixes the 500 error on sign-in I found on dev

PR #858, opened 2016-05-16
Custom S3 domain setting to workaround storages/boto URL construction bug

lmorchard:
New AWS_S3_CUSTOM_DOMAIN env var. When set, it skips using the boto URL
construction code, which erroneously adds a `x-amz-security-token`

See also:
- https://github.com/boto/boto/issues/1477
- http://stackoverflow.com/a/28749849


lmorchard:
Will probably self-merge this after Circle CI passes to move things along for #810

Issue #761, opened 2016-05-10
Serve up social metadata so Facebook & friends (and search) show appropriate previews for experiments


lmorchard:
Related to #419.

But, more than just adding views to our client-side app. This issue requires that we [add meta headers to our HTML on experiment URL routes on the server side so that Facebook's open graph spider picks them up](https://developers.facebook.com/docs/sharing/best-practices#tags).

pdehaan:
via @edmoz:

![](http://people.mozilla.org/~ewong2/fb.png)

ckprice:
Moving into this sprint.

ckprice:
Quick note that this also effects SERPs (search engine results pages).

![SERP](https://s3.amazonaws.com/f.cl.ly/items/0T2R1S2s0b3W2X3T3N2b/Image%202016-05-10%20at%202.45.05%20PM.png?v=a06eafc8)

ckprice:
> But, more than just adding views to our client-side app. This issue requires that we add meta headers to our HTML on experiment URL routes on the server side so that Facebook's open graph spider picks them up.

I'd like to bring down the scope of this to just simple meta tags (title, description, image) on [`/frontend/templates/testpilot/frontend/index.html`](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/templates/testpilot/frontend/index.html) if we can.

This would generate the same preview for all pages, yes, but until #419 is completed sharing out experiment URLs is going to be funky anyway. Not optimal, but right now we don't have anything coming up and I've received a couple of reports about this.

@lmorchard thoughts?

@johngruen can we get a preview image? [Facebook's sharing best practices](https://developers.facebook.com/docs/sharing/best-practices#tags) says

> Image Sizes
> Use images that are at least 1200 x 630 pixels for the best display on high resolution devices. At the minimum, you should use images that are 600 x 315 pixels to display link page posts with larger images. Images can be up to 8MB in size.

lmorchard:
> I'd like to bring down the scope of this to just simple meta tags (title, description, image) on /frontend/templates/testpilot/frontend/index.html if we can.

Yeah, I think that would be a good first step, at least. Only need to change one spot, don't need to add per-experiment routes. 

We could iterate on adding those later and get _something_ better now even if it's the same for all URLs. That will be kind of a moot point until #419 is resolved, anyway.

pdehaan:
[...]

PR #872, opened 2016-05-18
Add basic site-wide metadata tags for social media

lmorchard:
Issue #761


lmorchard:
I think this might need some iteration & help:
- For the site thumbnails, I just copied the copter. But, [Facebook wants a 1200 x 630 or 600 x 315 image](https://developers.facebook.com/docs/sharing/best-practices#images) and Twitter wants a 60x60 image (though it will use the Facebook image in a pinch).
- The tags I'm using are pretty redundant, but my SEO-fu is not strong. Need some recommendations here. Maybe we can drop some of these? Maybe we need all of these?

/cc @johngruen @ckprice

testpilot/frontend/templates/testpilot/frontend/index.html
```diff
@@ -11,6 +10,26 @@
     <meta name="viewport" content="width=device-width">
     <link rel="localization" href="{{ settings.STATIC_URL }}locales/{locale}/app.l20n">
 
+    {% set meta_title = "Firefox Test Pilot" %}
+    {% set meta_description = "Test new Features. Give us feedback. Help build Firefox." %}
+    {% set meta_url = request.build_absolute_uri() %}
+
```
pdehaan, line 16: Do we need/want to add something like `<meta charset="UTF-8">` here (or at the very top of `<head>`)?

ckprice:
I chatted briefly with @chrismore, and the things you've laid out are still the most important meta tags to have. I'm going to go ahead and merge this one, then we can see it in action on [the Facebook Open Graph debugger](https://developers.facebook.com/tools/debug/) on development.

PR #898, opened 2016-05-23
Build separate frontend app and vendor JS bundles

### LINKED ISSUE #665: Generate separate vendor & app JS bundles
Noticing that the scripts task in gulp takes awhile, these days - between 7 and 18 seconds sometimes.

It might be handy to generate separate JS bundles for vendor and app code. The vendor bundle would contain stuff from `package.json` while the app bundle would contain our own code. Just rebuilding the app bundle with project code changes would probably speed up file watcher cycles.

For an example, [I did something like this on a side project](https://github.com/lmorchard/arboretum/blob/master/gulpfile.babel.js#L15). Might be handy to find a way to auto-generate the external vendor modules list from `package.json` rather than hand-coding it like I did though

lmorchard:
- fix version on ampersand-event dependency
- move some devDependencies found in dependencies in package.json
- scripts-app task to build app bundle, watches for changes to app
- scripts-vendor task to build vendor bundle, only watches for changes
  to package.json
- add vendor.js script tag to frontend template
- refactor & simplify browserify build chain
- remove now-unused globby dependency, because we can give browserify a
  single app/main.js entry point and it finds other dependencies
- stop applying babel to vendor modules
- auto-assemble list of vendor dependencies from package.json

Fixes #665

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,7 +6,6 @@ const buffer = require('vinyl-buffer');
 const connect = require('gulp-connect');
 const del = require('del');
 const eslint = require('gulp-eslint');
-const globby = require('globby');
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
@@ -23,6 +22,9 @@ const uglify = require('gulp-uglify');
 const tryRequire = require('try-require');
 const Remarkable = require('remarkable');
 const md = new Remarkable();
+
+const packageJSON = require('./package.json');
+const vendorModules = Object.keys(packageJSON.dependencies);
 
 const IS_DEBUG = (process.env.NODE_ENV === 'development');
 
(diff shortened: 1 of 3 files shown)


lmorchard:
This one could use a good manual QA beating. Turns out the tests bundle the JS differently, so they won't catch issues here. Seems okay to me after downgrading ampersand-events to what all the other ampersand-\* modules seemed to expect - v2.0.1 was throwing errors when included in the vendor bundle.

lmorchard:
Also, it looks like this brings down app building time to around 2-3 seconds. The vendor bundle is pretty speedy to build too, because we were passing it all through babel when we shouldn't have been

meandavejustice:
Thank you @lmorchard ! pulling this down now to test

meandavejustice:
@pdehaan I'd like to get this. In today, would you mind banging on this with me?

meandavejustice:
I'm seeing an error which holds up the experiment detail page.

Steps to reproduce:
1. login with firefox account
2. navigate to activity-stream from experiment list page
3. page is all white with error in console(error pasted below)
4. refresh page and it renders without errors.

Error: `TypeError: this.config is undefined                  vendor.js:8:3736`

stack trace:

```
require<["ampersand-view-switcher"]</r.prototype._onViewRemove()
vendor.js:8
require<["ampersand-events"]</l.once/a<()
vendor.js:7
require<[188]</r/<()
vendor.js:2
require<[94]</n.triggerEvents()
vendor.js:1
require<["ampersand-events"]</l.trigger()
vendor.js:7
require<["ampersand-view"]</m<.derived.rendered.fn()
vendor.js:8
require<["ampersand-state"]</<._initDerived/</o()
vendor.js:7
require<["ampersand-state"]</<._initDerived/</<()
vendor.js:7
forEach()
self-hosted
require<["ampersand-state"]</<._initDerived/<()
vendor.js:7
require<[94]</n.triggerEvents()
vendor.js:1
require<["ampersand-events"]</l.trigger()
vendor.js:7
require<["ampersand-state"]</<.set/<()
vendor.js:7
forEach()
self-hosted
require<["ampersand-state"]</<.set()
vendor.js:7
require<["ampersand-state"]</o/<.set()
vendor.js:7
require<["ampersand-view"]</<._remove()
vendor.js:8
[21]</n.default<.remove()
app.js:2
require<["ampersand-view"]</<._setRemove/<.set/this._remove()
vendor.js:8
require<["ampersand-view-switcher"]</r.prototype._hide()
vendor.js:8
require<["ampersand-view-switcher"]</r.prototype.set()
vendor.js:8
[5]</k<.value()
app.js:1
bound ()
self-hosted
require<[94]</n.triggerEvents()
vendor.js:1
require<["ampersand-events"]</l.trigger()
vendor.js:7
[6]</n.default<.experimentDetail()
app.js:1
require<["ampersand-router"]</<.execute()
vendor.js:7
require<["ampersand-router"]</<.route/<()
vendor.js:7
require<[705]</<.loadUrl/<()
vendor.js:5
some()
self-hosted
require<[705]</<.loadUrl()
vendor.js:5
require<[705]</<.navigate()
vendor.js:5
require<["ampersand-router"]</<.navigate()
vendor.js:7
[24]</n.default<.openDetailPage()
app.js:2
n()
vendor.js:6
require<[821]</n.bind/<()
vendor.js:6
vendor.js:8:3736
```

lmorchard:
> I'm seeing an error which holds up the experiment detail page.

Did you do an `npm install`? This is the error that the [ampersand-events dependency change](https://github.com/mozilla/testpilot/pull/898/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R16) was supposed to have fixed

lmorchard:
Rebased against master

meandavejustice:
@lmorchard Looks like I just needed a fresh npm install as you suggested, works for me now.

PR #900, opened 2016-05-23
Add meta charset

ckprice:
@pdehaan [noted that we should add `<meta charset="utf-8" />`](https://github.com/mozilla/testpilot/pull/872#discussion_r63760415) to the head. I merged this in before the comment was addressed.

It was working locally, but dev is throwing a 500 error with `The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol.` This ~~should~~ may fix it. /edit I can't confirm it wasn't broken before.

### DIFF
--- a/testpilot/frontend/templates/testpilot/frontend/index.html
+++ b/testpilot/frontend/templates/testpilot/frontend/index.html
@@ -1,6 +1,7 @@
 <!doctype html>
 <html>
 <head>
+    <meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


pdehaan:
Not sure why Circle isn't running against some of these PRs today, but I think GitHub may have been having some issues earlier.

PR #903, opened 2016-05-24
fix issue with gulp-imagemin skipping files

meandavejustice:
https://github.com/sindresorhus/gulp-imagemin/issues/183

optimized images and commented out gulp-imagemin code as a hot fix.

cc/ @pdehaan


ghost:
Seems reasonable.  Let's see what it looks like on -dev!

Issue #719, opened 2016-05-05
Test pilot icon appears white


past:
Just installed test pilot add-on on latest nightly, OS X, default theme:

<img width="210" alt="white-button" src="https://cloud.githubusercontent.com/assets/49118/15052123/ac904740-1304-11e6-8b12-1156ad850022.png">

I already had activity-stream and universal-search installed, nothing peculiar in the browser console.

meandavejustice:
@past Thanks for reporting this, I'm also seeing this with the light theme.

meandavejustice:
@past would you mind going into `about:config` and looking for the `devtools.theme` key? I've seen a weird state with firefox stable where `devtools.theme` is set to dark, but the address bar/toolbar area still has a light background. Lemme know if this is what you are seeing.

callahad:
On my end, toggling the DevTools theme from light to dark changes the Test Pilot icon to the inverse.

Which would be appropriate if I was using the DevEdition Firefox theme, but I'm not.

past:
@meandavejustice I see the same thing as @callahad: when I switch to the light devtools theme, I get the right test pilot icon. So this looks like what's going on: the icon is toggled depending on the selected devtools theme, regardless of the actual theme that is applied in the browser chrome.

ckprice:
Removing this as a launch blocker.

meandavejustice:
Talking to bwinton about this today. We were originally checking against the `devtools.theme` pref. It seems we need to be checking something more like this: https://dxr.mozilla.org/mozilla-central/source/browser/themes/linux/browser.css#1565

leplatrem:
I have a similar problem with GTK and the dark theme:
![capture d ecran de 2016-05-10 12-49-08](https://cloud.githubusercontent.com/assets/546692/15144111/9dade5c8-16ad-11e6-921f-ef9fc522e2a5.png)

![capture du 2016-05-10 12-50-02](https://cloud.githubusercontent.com/assets/546692/15144133/d61cf71e-16ad-11e6-9f42-13eb7fdb834e.png)

rtanglao:
i have this problem in developer edition and so do users:

https://twitter.com/uimanishere/status/731162989735350273

warpr:
Looks correct for me in Developer Edition on Fedora 23, when it is near the address bar.  But it stays white when dragged into hamburger menu using customize:

[...]

Issue #53, opened 2015-08-31
Support versioning the API


lmorchard:
Let's make the API versionable - e.g. `/api/0.2.3/experiments/1`
- [Django REST Framework has lots of options for versioning](http://www.django-rest-framework.org/api-guide/versioning/), could support the above.
- Use [semver](http://semver.org/) - [there's a Python module to help compare versions](https://pypi.python.org/pypi/semver)
- We can use version ranges to switch behaviors / serializers / etc over time, keeping older stuff around until / unless deprecation happens.
- Model changes could be tricky, but let's burn that bridge when we cross it

PR #917, opened 2016-05-25
Fix inconsistency in app.me.installed

### LINKED ISSUE #742: audit app.me.installed data
https://github.com/mozilla/testpilot/pull/726#issuecomment-217892254

`/api/me` installed key returns addon objects in an array
`sync-installed` event from addon returns objects in the shape of:

``` json
"univeral-search@mozilla.com" {
  "some-data": "for the addon"
}
```

meandavejustice:
- fixes #742
- users/serializers Me now returns installed in format:
  {addon_id: {addon: data}}
- fix consumption of me.installed data in addon/index.js
- bump addon package version to 0.6.4
- update experiments/tests to match new data shape
- fix default obj for installed in me.js model
- catch and log occasional error in me.js when sync-installed fails
- update experiment-page.js tests for installed shape update
- update users view and tests to send data correctly
- reformate line in experiments/serializers.py for pep8

I think this will take care of the white screen of death issue as well. refs #719 
There was an issue where `sync-installed` would timeout/fail and we did nothing with the error except reject. Which in turn threw an error in the `async.all` call in main.js, essentially sabotaging our whole plot to ship new features.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -349,9 +349,13 @@ function updateExperiments() {
 function syncAllAddonInstallations(serverInstalled) {
   const availableIDs = Object.keys(store.availableExperiments);
   const clientIDs = Object.keys(store.installedAddons);
-  const serverIDs = serverInstalled
-    .filter(item => item.client_id === store.clientUUID)
-    .map(item => item.addon_id);
+  const serverIDs = [];
+
+  for (let key in serverInstalled) { // eslint-disable-line prefer-const
+    if (serverInstalled[key].client_id === store.clientUUID) {
+      serverIDs.push(key);
+    }
+  }
 
   return Promise.all(availableIDs.filter(id => {
     const cidx = clientIDs.indexOf(id);
--- a/testpilot/experiments/tests.py
+++ b/testpilot/experiments/tests.py
@@ -308,7 +308,7 @@ class FeaturesBasicTests(BaseTestCase):
     def test_me_list(self):
         """/api/me installation listing should include feature flags"""
         result = self.jsonGet('me-list')
-        self.assertEqual(result['installed'][0]['features'], {
+        self.assertEqual(list(result['installed'].values())[0]['features'], {
             self.feature1_title: True,
             self.feature2_title: True
         })
--- a/testpilot/frontend/static-src/app/models/me.js
+++ b/testpilot/frontend/static-src/app/models/me.js
@@ -13,7 +13,7 @@ export default Model.extend({
   props: {
     user: 'object',
     clientUUID: 'string',
-    installed: {type: 'object', default: () => []},
+    installed: {type: 'object', default: () => {}},
     hasAddon: {type: 'boolean', required: true, default: false},
     addonTimeout: {type: 'number', default: 1000}
   },
@@ -42,10 +42,29 @@ export default Model.extend({
       this.hasAddon = Boolean(window.navigator.testpilotAddon);
       if (!this.hasAddon) { return false; }
 
-      return app.waitForMessage('sync-installed', userData.installed)
+      let installedData;
+      if (!window.navigator.testpilotAddonVersion) {
+        // Previous add-ons didn't expose version info, so assume
+        // old API data
+        installedData = [];
+        for (let k in userData.installed) { // eslint-disable-line prefer-const
+          if (userData.installed.hasOwnProperty(k)) {
+            installedData.push(userData.installed[k]);
+          }
+        }
+      } else {
+        // TODO: Need a semver matching check here someday, but for
+        // now we can assume
+        // just the presence of a version means we're in the future.
+        installedData = userData.installed;
+      }
+
+      return app.waitForMessage('sync-installed', installedData)
         .then(result => {
           this.clientUUID = result.clientUUID;
           this.installed = result.installed;
+        }).catch((err) => {
+          console.error('sync-installed failed', err); // eslint-disable-line no-console
         });
     });
   },
--- a/testpilot/users/tests.py
+++ b/testpilot/users/tests.py
@@ -267,7 +267,7 @@ class MeViewSetTests(TestCase):
                     'username': 'johndoe'
                 },
                 'addon': self.addonData,
-                'installed': []
+                'installed': {}
             }
         )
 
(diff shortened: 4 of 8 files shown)


testpilot/frontend/static-src/app/models/me.js
```diff
@@ -46,6 +46,8 @@ export default Model.extend({
         .then(result => {
           this.clientUUID = result.clientUUID;
           this.installed = result.installed;
+        }).catch((err) => {
+          console.error('sync-installed failed', err);
```
pdehaan:
```
testpilot/frontend/static-src/app/models/me.js
  50:11  warning  Unexpected console statement  no-console

✖ 1 problem (0 errors, 1 warning)
```

Not sure if we want to just inline disable that rule for the `console.error()` (ie: `// eslint-disable-line no-console`), or maybe send the error event through GA or some other monitoring service. Or whatever.

testpilot/experiments/tests.py
```diff
@@ -288,7 +288,8 @@ def test_features_list(self):
     def test_me_list(self):
         """/api/me installation listing should include feature flags"""
         result = self.jsonGet('me-list')
-        self.assertEqual(result['installed'][0]['features'], {
+        print(list(result['installed'].values()))
```
lmorchard: Nit: Should probably take out this print statement, though it's in a test and won't affect the server

testpilot/users/tests.py
```diff
@@ -277,19 +277,23 @@ def test_get_logged_in(self):
         resp = self.client.get(self.url)
         result_data = json.loads(str(resp.content, 'utf-8'))
 
+        print(result_data)
```
lmorchard: Same for this print statement

testpilot/frontend/static-src/app/models/me.js
lmorchard: Might be worth leaving the warning there as a reminder to circle back if/when we have a proper error reporting thing like Sentry for the front end. We have a few places in the code where a `.catch(err =>` just ends up in a ¯\_(ツ)_/¯  `console.error` line

addon/index.js
```diff
@@ -346,9 +346,13 @@ function updateExperiments() {
 function syncAllAddonInstallations(serverInstalled) {
   const availableIDs = Object.keys(store.availableExperiments);
   const clientIDs = Object.keys(store.installedAddons);
-  const serverIDs = serverInstalled
-    .filter(item => item.client_id === store.clientUUID)
-    .map(item => item.addon_id);
+  const serverIDs = [];
```
lmorchard, line 352: Oof, you know something just occurred to me: After this makes it to prod, we will have a period where folks are running the previous add-on against the site until their Firefox looks for an update. Will the API change break the old version of the add-on in a nasty way? If so, we may need to consider [versioning the API](https://github.com/mozilla/testpilot/issues/53).

lmorchard, line 352: Something we could do short of versioning the API for now: That `serverInstalled` data [comes from the web frontend](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/models/me.js#L45) in the `sync-installed` message - the add-on doesn't hit the API for that data itself.

So, on the frontend, we could convert that installation object into the array the existing add-on version expects, and hold off on the add-on change until/unless we get a versioned API

meandavejustice, line 352: @lmorchard How long would this period be before users would be updated to the new addon?

lmorchard, line 352: I'd guess 24 to 48 hours for addon updates, assuming everyone's Firefoxes are running at some point during that period to wake up & check

meandavejustice, line 352: How involved is it currently to create a versioned api? (time estimate)

lmorchard, line 352: Not really sure, I need to dig into DRF and see what that entails. Could be simple, but need to spend some hours figuring that out. (Issue #53) That's why I suggest doing a tweak on the frontend for now to convert to the old format for the add-on's sake until that's worked out.

lmorchard:
Okay, so after some poking, I realized that [enabling versioning on the API is pretty easy](https://github.com/mozilla/testpilot/pull/933). But, in attempting to use that, I realized the problem is more complicated: 

The _frontend_ makes the initial request to the API for `/api/me`, not the add-on. But, the frontend gets deployed along with the API, so there's no version mismatch there. It's just the add-on that can lag behind the rest of the site.

The add-on gets a message from the frontend with that data. That means, the _frontend_ needs to know the add-on version - so that it can branch send the expected format. (i.e. an array for old add-on versions, an object for future versions)

So, along with enabling API versioning, I also made a tweak to the add-on in [that PR](https://github.com/mozilla/testpilot/pull/933) to expose the add-on version as `window.navigator.testpilotAddonVersion`.

With that, you can do something like this [in `app/models/me.js`](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/models/me.js#L45):

```
let installedData;
if (!window.navigator.testpilotAddonVersion) {
  // Previous add-ons didn't expose version info, so assume old API data
  installedData = Object.values(userData.installed); // Maybe Object.values needs a polyfill?
} else {
  // TODO: Need a semver matching check here someday, but for now we can assume
  // just the presence of a version means we're in the future.
  installedData = userData.installed;
}

return app.waitForMessage('sync-installed', installedData)
.then(result => {
  this.clientUUID = result.clientUUID;
  this.installed = result.installed;
});
```

meandavejustice:
@lmorchard This is perfect. Thanks for figuring this out. I'll look over your versioning pr and get this updated.

testpilot/frontend/static-src/app/models/me.js
```diff
@@ -1,6 +1,8 @@
 import app from 'ampersand-app';
 import cookies from 'js-cookie';
 import Model from 'ampersand-model';
+// Object.values only has support in ff 47 and ^
+import values from 'object.values';
```
lmorchard: Hmm, something missing from package.json? Might be worth doing something like:

```
installedData = [];
for (let k in userData.installed) {
  installedData.push(userData.installed[k]);
}
```

meandavejustice: I had some weird issues on my rebase, I'll replace with this^

lmorchard:
Tests are passing, things look good on a skim-through, going to give it a run through and probably merge very soon

Issue #738, opened 2016-05-06
Move legal documents to /mozilla/legal-docs


ckprice:
Our legal documents are currently in https://github.com/mozilla/testpilot/tree/master/legal-copy.

We should move them over to https://github.com/mozilla/legal-docs someday, since that's where the rest are (and will help l10n at some point).

pdehaan:
Ref: https://github.com/mozilla/legal-docs/pull/509
I made a pull request against **mozilla/legal-docs** to add a package.json, which will let us do something like this and add the legal-docs repo :

``` bash
$ npm install mozilla/legal-docs -D
```

Then I think we'll just need to move over our legal *.md files and blah, blah, blah.

---

In **fxa-content-server** we used the [bower.json](https://github.com/mozilla/fxa-content-server/blob/4450796bed0224a617b5cbd201895a22647c9d77/bower.json#L37) file to specify the **mozilla/legal-docs** as a dependency, but I'd rather not add Bower to this stack just for that one repo.

pdehaan:
Upstream **legal-docs** package.json fix was merged (thanks @clouserw!), in case anybody wants to jump on this before I get around to it.

ckprice:
~~Received the following request from Brian Smith in legal (also updated bug title, and opening):~~

> I noticed that the Test Pilot privacy notice webpage does not have the sections collapsed like our other [privacy notices](https://www.mozilla.org/en-US/privacy/firefox/). Can we please revise to make the two sections of the notice collapse/expand? Denelle does not like our notices to just be a wall of text, so this is a priority for us.
> 
> Casey (cc'd) can help the relevant person on your team with the css/html formatting as needed.

/cc @mozcmcgill

/edit addressing in #877.

meandavejustice:
@ckprice This sounds like a separate issue, maybe to be made as a follow up to this one.

pdehaan:
> _This sounds like a separate issue, maybe to be made as a follow up to this one._

Filed as #877

pdehaan:
Another legal hiccup, but in the [**mozilla/legal-docs** firefox_cloud_services_PrivacyNotice](https://github.com/mozilla/legal-docs/tree/master/firefox_cloud_services_PrivacyNotice) we have localized versions of the legal docs. So we'll need something like dis:

```
mozilla/legal-docs/
  - testpilot_privacy-notice/
    - de.md
    - en-US.md
    - es-ES.md
    - fr.md
    - it.md
    - ...

[...]

PR #923, opened 2016-05-26
Fix #877 Legal Docs Privacy Notice sections should be collapsable

### LINKED ISSUE #877: Legal Docs Privacy Notice sections should be collapsable
Ref: https://github.com/mozilla/testpilot/issues/738#issuecomment-220063420

---

@ckprice received the following request from Brian Smith in legal (also updated bug title, and opening):

> I noticed that the [Test Pilot privacy notice](https://testpilot.firefox.com/privacy) webpage does not have the sections collapsed like our other [privacy notices](https://www.mozilla.org/en-US/privacy/firefox/). Can we please revise to make the two sections of the notice collapse/expand? Denelle does not like our notices to just be a wall of text, so this is a priority for us.
> 
> Casey (cc'd) can help the relevant person on your team with the css/html formatting as needed.

/cc @mozcmcgill

---

/edit adding some specifications

Ref: https://docs.google.com/document/d/14HB8hPnuH5YlO3HtZeIATmFCMIhiHIPBNEgWN66XkDQ/edit?usp=sharing

The two _Learn More_ links should expand out the bullet points underneath them. The bullets start collapsed.

kyoshino:
Fixes #877

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -158,6 +158,12 @@ gulp.task('addon', function localesTask() {
     .pipe(gulp.dest(DEST_PATH + 'addon'));
 });
 
+// Copy the static legal.js file to dest
+gulp.task('static-script-copy', function staticScriptCopyTask() {
+  return gulp.src(SRC_PATH + 'scripts/**/*')
+    .pipe(gulp.dest(DEST_PATH + 'scripts'));
+});
+
 gulp.task('build', function buildTask(done) {
   runSequence(
     'clean',
@@ -167,6 +173,7 @@ gulp.task('build', function buildTask(done) {
     'locales',
     'addon',
     'legal',
+    'static-script-copy',
     done
   );
 });
--- a/testpilot/frontend/static-src/scripts/legal.js
+++ b/testpilot/frontend/static-src/scripts/legal.js
@@ -1,0 +1,62 @@
+window.addEventListener('DOMContentLoaded', function (event) {
+  'use strict';
+
+  // Add an accordion to the Privacy Notice page
+  if (location.pathname === '/privacy') {
+    var ps = document.querySelectorAll('h2 ~ p');
+    var tablist = ps[0].parentElement;
+
+    tablist.setAttribute('role', 'tablist');
+    tablist.setAttribute('aria-multiselectable', 'true');
+
+    Array.prototype.map.call(ps, function (p, i) {
+      var expanded = false;
+      var label = document.createElement('span');
+      var tab = document.createElement('a');
+      var tabpanel = p.nextElementSibling;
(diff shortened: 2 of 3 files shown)


gulpfile.js
```diff
@@ -148,6 +148,11 @@ gulp.task('images', function imagesTask() {
     .pipe(gulp.dest(DEST_PATH + 'images'));
 });
 
+gulp.task('scripts', function imagesTask() {
```
lmorchard: We already have [a `scripts` task that builds our minified app.js bundle](https://github.com/kyoshino/testpilot/blob/3f98f4a20ece73fbdc8600a2f82d59412381aea7/gulpfile.js#L83). We should rename this new one, and change the `imagesTask` function name as a nitpick.

testpilot/frontend/static-src/scripts/legal.js
```diff
@@ -0,0 +1,62 @@
+window.addEventListener('DOMContentLoaded', function (event) {
+  'use strict';
+
+  // Add an accordion to the Privacy Notice page
+  if (location.pathname === '/privacy') {
+    var ps = document.querySelectorAll('h2 ~ p');
+    var tablist = ps[0].parentElement;
```
pdehaan, line 7: I don't know if it helps, but we currently "own" the Privacy Notice and other legal docs in [./legal-copy](https://github.com/mozilla/testpilot/tree/master/legal-copy), so we could probably inject some HTML with classnames/IDs into the Markdown, if that makes selectors and DOM grabbing any easier.

meandavejustice, line 7: Aren't we planning to move them to the mozilla legal repo?

pdehaan, line 7: Yep (ref #738). But for now they're in this repo, so we can easily tweak the templates without having to get it reviewed and r+'ed by a different team.

gulpfile.js
kyoshino: I should have noticed this. Fixed!

testpilot/frontend/static-src/scripts/legal.js
kyoshino, line 7: On [legal-docs](https://github.com/mozilla/legal-docs), we are using the [Python Markdown extension](https://pythonhosted.org/Markdown/extensions/attr_list.html) syntax for www.mozilla.org to inject attributes, IDs and class names. I just checked the [remarkable](https://github.com/jonschlinkert/remarkable) library used on this textpilot repo but looks like it doesn't support such attribute injections unfortunately.

Since this Privacy Notice is pretty simple, I think it's safe to rely on JS to do tweaks.

PR #927, opened 2016-05-26
Switch to shared mozilla-cloud-services-logger package

### LINKED ISSUE #444: Package ops-friendly JSON logger as a standalone Python package
So, [we just merged a JSON log formatter](https://github.com/mozilla/testpilot/pull/431#issuecomment-188993849) that seems to conform to cloudops' specifications. This should be generally useful to Python apps, so let's split it off into a standalone PyPi package for use by any project

lmorchard:
- Moved code to produce Cloud Services' preferred JSON format into a new
  shared package on PyPi:
  
  https://pypi.python.org/pypi/mozilla-cloud-services-logger/
- Add mozilla-cloud-services-logger v1.0.1 to requirements.txt
- Switch to using the shared log formatter and Django middleware
- Remove old logger tests from testpilot.base
- Remove jsonschema dependency, only used in log formatter tests
- Switch to new shared log formatter in other tests

Fixes #444


lmorchard:
For bonus points, I'd like to hear from anyone who wants to be added as owners / maintainers on [the PyPi package](https://pypi.python.org/pypi/mozilla-cloud-services-logger/) to reduce my bus factor so others can work on [the package repo](https://github.com/mozilla/mozilla-cloud-services-logger) and push releases.

mostlygeek:
i prefer "lotto" factor to "bus factor" :)

mostlygeek:
Is there a need to name it `JsonLogFormatter`, can we name it `MozlogFormatter` instead?

lmorchard:
Yeah, I guess that's a bit generic. But, turns out there's already [a `mozlog`](https://pypi.python.org/pypi?%3Aaction=search&term=mozlog&submit=search) out there, and it doesn't do what we want here. So, I split the difference with a more verbose base package name

PR #929, opened 2016-05-27
Change baseimage to python:3-alpine

mostlygeek:
- reduces image size from 846MB => 195MB
- improve bin/run-prod.sh so gunicorn can receive signals from
  supervisor correctly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,25 @@
-FROM python:3
+FROM python:3-alpine
 
 EXPOSE 8000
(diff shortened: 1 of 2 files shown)


mostlygeek:
@lmorchard  r? please 

A nice plus of this is that it cuts about a minute off the circle-ci build :)

lmorchard:
Whoa. Changing the base image freaks me out. But, I can't find anything wrong and all tests pass. So, umm... 🚢 it

PR #930, opened 2016-05-27
Enable Content Security Policy

### LINKED ISSUE #882: Implement Content Security Policy
[CSP](https://developer.mozilla.org/en-US/docs/Web/Security/CSP) lets us tell the browser what we consider as valid resources to load for a page, restricting by domain & etc. We should implement this for our site.

lmorchard:
- Enable django-csp middleware
- Add Google & Gravatar origins to CSP policy
- Add /__cspreport__ endpoint to log CSP violation reports
- Move Google Analytics snippet out of inline HTML
- Create scripts/legal.js to contain inline JS from legal templates
- Add extra-scripts gulp task to minify & move scripts/*js

Fixes #882

### DIFF
--- a/testpilot/frontend/static-src/scripts/legal.js
+++ b/testpilot/frontend/static-src/scripts/legal.js
@@ -1,3 +1,22 @@
+/* global ga*/
+/*eslint-disable*/
+// HACK: Google Analytics lives here, because CSP won't let it live inline
+(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)
+},i[r].l=1*new Date();a=s.createElement(o),
+m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+})(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+/*eslint-enable*/
+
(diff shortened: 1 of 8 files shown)


testpilot/frontend/static-src/scripts/legal.js
```diff
@@ -0,0 +1,18 @@
...
+/*eslint-enable*/
+
+if (typeof(ga) !== 'undefined') {
+  ga('create', 'UA-49796218-34', 'auto');
+} else {
+  console.warn(
+    'You have google analytics blocked. We understand. Take a ' +
+    'look at our privacy policy to see how we handle your data.'
+  );
+}
```
pdehaan, line 18: I'm confused. This looks like the same copy-pasta from _testpilot/frontend/static-src/app/main.js_ above, and not legal-specific.

lmorchard, line 18: Yeah, it has to be duplicated, because the legal pages don't use the same frontend template as the rest of the site and so don't use the app/main.js code

meandavejustice:
Looks like this needs a rebase

lmorchard:
Yeah, it's bitrotted since the legal PR got merged, need to rework part of it

lmorchard:
Rebased against master

meandavejustice:
@lmorchard Looks like this needs another rebase

lmorchard:
Rebased against master

PR #938, opened 2016-06-02
Bugfixes for telemetry ping timing

### LINKED ISSUE #815: Telemetry pings are underreported
This is https://bugzilla.mozilla.org/show_bug.cgi?id=1272395 .  I don't know if it's in the code or in the pipeline.  I think we're just doing a simple call to submitExternalPing() but this issue is to double check that code and make sure we're not doing anything obviously wrong.  We'll come at it from the other side too.

### LINKED ISSUE #936: TypeError: can't define property "pingTimer": Object is not extensible
Randomly spotted in Test Pilot Browser Console:

> TypeError: can't define property "pingTimer": Object is not extensible [[/addon/lib/metrics.js:85:22](https://github.com/mozilla/testpilot/blob/3eef36e8365dce6693f5de2f8f9b903cf1a3b69b/addon/lib/metrics.js#L85-L88)]

``` js
    this.pingTimer = setTimeout(
      () => this.maybePingTelemetry(),
      PING_CHECK_INTERVAL
    );
```

lmorchard:
- Perform initial ping immediately on initialization rather than 24
  hours + 10 min later.
- `store.telemetryPing` -> `store.telemetryPingPayload` in init.
- Ensure `pingTimer` property exists in metrics.js module.

Fixes #936, Fixes #815

### DIFF
--- a/addon/dev-prefs.json
+++ b/addon/dev-prefs.json
@@ -1,5 +1,4 @@
 {
-  "BASE_URL": "http://testpilot.dev:8000",
-  "HOSTNAME": "testpilot.dev",
-  "ALLOWED_ORIGINS": "http://testpilot.dev:8000/*,https://testpilot.firefox.com/*"
+  "toolkit.telemetry.server": "http://127.0.0.1/",
+  "extensions.@testpilot-addon.SERVER_ENVIRONMENT": "local"
 }
--- a/addon/lib/metrics.js
+++ b/addon/lib/metrics.js
@@ -10,7 +10,7 @@ const {Cu} = require('chrome');
 
 Cu.import('resource://gre/modules/TelemetryController.jsm');
 
-const { setTimeout } = require('sdk/timers');
+const { setTimeout, clearTimeout } = require('sdk/timers');
 const Events = require('sdk/system/events');
 const store = require('sdk/simple-storage').storage;
 const PrefsService = require('sdk/preferences/service');
@@ -31,11 +31,13 @@ const PREFERENCE_OVERRIDES = {
   'datareporting.healthreport.uploadEnabled': true
 };
 
+let pingTimer = null;
+
 module.exports = {
 
   init: function() {
-    if (!store.telemetryPing) {
-      store.lastTelemetryPingTimestamp = Date.now();
+    if (!store.telemetryPingPayload) {
+      store.lastTelemetryPingTimestamp = false;
       store.telemetryPingPayload = {
         version: 1,
         tests: {}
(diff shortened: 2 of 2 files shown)


lmorchard:
For extra bonus points:
- [gzipServer](https://github.com/vdjeric/gzipServer) is a little Python web server that works as a local Telemetry end point. Convenient instructions on how to make Firefox use it are in the README
- If you're running that server, and about:config shows "toolkit.telemetry.server" == "http://127.0.0.1/", you _should_ see a Telemetry ping show up in the gzipServer console output after about 5 min of the browser starting up with `npm run once`

addon/dev-prefs.json
```diff
@@ -1,5 +1,4 @@
 {
-  "BASE_URL": "http://testpilot.dev:8000",
-  "HOSTNAME": "testpilot.dev",
-  "ALLOWED_ORIGINS": "http://testpilot.dev:8000/*,https://testpilot.firefox.com/*"
+  "toolkit.telemetry.server": "http://127.0.0.1/",
+  "extensions.@testpilot-addon.SERVER_ENVIRONMENT": "local"
```
pdehaan, line 3: Well, these changes look very different...
Planned change, or local hacking gone public?

meandavejustice, line 3: These are in `dev-prefs.json` so it seems appropriate

lmorchard, line 3: Yeah, it's been awhile since we looked at these settings. Really should have updated them back when [I implemented the simple env drop-down selector](https://github.com/mozilla/testpilot/commit/7339a3a5885f3f308537318fbacc293353b416ba), because none of the settings in there now do anything.

addon/dev-prefs.json
```diff
@@ -1,5 +1,4 @@
 {
-  "BASE_URL": "http://testpilot.dev:8000",
-  "HOSTNAME": "testpilot.dev",
-  "ALLOWED_ORIGINS": "http://testpilot.dev:8000/*,https://testpilot.firefox.com/*"
+  "toolkit.telemetry.server": "http://127.0.0.1/",
```
lmorchard, line 2: This may or may not be a good thing, though. I stuck in in there, because along with [gzipServer](https://github.com/vdjeric/gzipServer), this lets me see Telemetry pings go to my local server on `npm run once`. We might want to change the port, though, since port 80 on localhost might already be taken

ghost:
Please let me know if you need anything on this.  If we need more people testing, I'm happy to install an add-on.

pdehaan:
I'm still voting to just merge it and we all test it live on the dev server.

lmorchard:
Don't see any objections and no one else has merged yet, so I'll self-merge

Issue #433, opened 2016-02-26
Need a way to connect WebExtension tests to the Test Pilot add-on.


bwinton:
The current solution for communicating with the main add-on from experiments (`nsIObserverService`) won't work for WebExtensions.  There is [an API](https://developer.chrome.com/extensions/runtime#method-connect) we could use instead, but it's [not implemented yet](https://bugzilla.mozilla.org/show_bug.cgi?id=1213473), and worse, isn't going to block the first release of WebExtensions.
![😢](http://replygif.net/i/127.gif)

But, if we can convince the WebExtensions team that we really, _really_ need this, they will probably bump it up the priority queue.  Also, maybe there are other solutions we can use in the meantime, like hitting the Test Pilot server manually…

bwinton:
It looks like from the WebExtension side, we can use [`chrome.runtime.sendMessage`](https://developer.chrome.com/extensions/runtime#method-sendMessage).  I’m not sure what that’ll translate to on the Test Pilot add-on side, though…

ckprice:
A quick note that product would like to include Tab Center with the go to market release of Test Pilot in April.

/edit thanks @bwinton :blush: Not required for Tab Center, but still important as noted.

bwinton:
Fortunately, Tab Center isn't a WebExtension, however the No-More-404s extension is… (And future extensions will be, if at all possible. :wink:)

lmorchard:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1252215

ckprice:
I'm moving this into the current milestone to make sure we're tracking this. Wayback Machine is blocked by this bug.

ckprice:
> See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1252215

Confirmed in #amo that this is something that will have to ride the trains. Once it gets resolved, I'll request an uplift.

ckprice:
~~Ref 904, may want to make sure Optimizely is okay.~~

/edit WRONG BUG, thanks Les.

lmorchard:
> Ref #904, may want to make sure Optimizely is okay.

FWIW, I think this might be more relevant to #715 & possible add-on install whitelisting

lmorchard:
Starting to try catching myself up on the concerns surrounding this issue. 

[...]

PR #952, opened 2016-06-06
Add /api/metrics/testpilottest POST resource

lmorchard:
Add /api/metrics/testpilottest POST resource
- /api/metrics/testpilottest accepts POST'd JSON body which gets
  included in the Fields property of a [mozlog log event](https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md)
- Add django-cors-headers dependency to allow CORS requests to the API
- Tiny WebExtension example in docs/examples/webextension to show how to
  POST to the metrics ping resource
- Tests

Issue #433


lmorchard:
Rebased against master. Made further tweaks to ensure a WebExtension can POST to the metrics API.

docs/examples/webextension/background.js
```diff
@@ -0,0 +1,14 @@
...
+  body: JSON.stringify({
+    "some": "random",
+    "metrics": "ping",
+    "payload": "here"
+  })
+}).catch(e => {
+  console.log('problem sending metrics ping', e);
+}).then(resp => {
+  console.log('metrics ping success', resp);
+});
```
pdehaan, line 14: How does `.catch().then()` work?
I can't remember if that means if you did catch an error it'd fall through and, you'd get both messages:
- _"problems sending metrics ping"_ AND
- _"metrics ping success"_

Or should we reverse the order so it's `.then().catch()`?

lmorchard, line 14: I think it bails on the catch, but I swapped the order anyway

ghost:
maybe a section at the bottom of readme-metrics.md about how to use this (format of the JSON if nothing else?)

lmorchard:
I'm totally fuzzy on what the format of the JSON should be - that depends entirely on what gets done in the processing pipeline with the Fields section of the log events, and what the experiment wants to measure

lmorchard:
Probably the closest thing is [this section of README-METRICS](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md#data-collection-details) that talks about what happens with `request.summary` server log events. 

I'm thinking we'll probably have to see what a WebExtension experiment wants to measure and tweak the RedShift schema for a new `testpilottest` server event and the associated Fields properties. Maybe some will be similar to that generic `request.summary` schema, but needs some discussion

docs/examples/webextension/background.js
pdehaan, line 14: Turns out the `catch()` in the middle lets you handle the error and continue, so it'd have done both `.catch()` and `.then()`:

``` js
function promiseTest(data = new Error("uh oh!")) {
    if (data instanceof Error) {
      throw data;
    }
    resolve(data);
  });
}

promiseTest( /* error */ )
  .then((data) => {
    console.log(`first then(): ${data}`);
    return data;
  }).catch((err) => {
    console.error(`An error? ${err}`);
    return err;
  }).then((data) => {
    console.info(`after catch()? ${data}`);
  });
```

``` sh
$ node promise.test.js

An error? Error: uh oh!
after catch()? Error: uh oh!
```

ghost:
Hmm, yeah.  I thought we had an undefined payload{} section like we have for the client side...

I have a draft of metrics for firefoxnomore404s at https://github.com/internetarchive/FirefoxNoMore404s/pull/10 .  If that draft gets approved, this is only recording a very small set of data.  We _could_ just overload the context field here to log the action of the user.  (this makes sense if you look at the JSON at the bottom of my doc).

It's hacky, but if this is only temporary until webextensions can do their own logging.....

...talk me out of doing this the hacky way... :loud_sound:

testpilot/metrics/views.py
```diff
@@ -0,0 +1,30 @@
...
+    'testpilottest'
+]
+
+
+@csrf_exempt
+@api_view(['POST'])
+@permission_classes([])
+def log_ping(request, logger_name):
+    """Accept and log metrics pings"""
+    whitelist = getattr(settings, 'LOG_PING_WHITELIST', DEFAULT_LOG_PING_WHITELIST)
```
groovecoder, line 20: So we will need to add logger names for each experiment that uses this? I.e., for 404s and my "blok" add-on/experiment, we would need to have:
`settings.py`:

```
LOG_PING_WHITELIST = [
    'blok',
    'nomore404s'
]
```

?

lmorchard, line 20: No, this should be like [the general `testpilottest` ping in Telemetry](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md#data-collection-details). One log event type (and URL) for all experiments. You should include the experiment's add-on ID in the JSON payload. Don't really have a schema for that payload, but i figured it could look like the telemetry ping.

lmorchard, line 20: FWIW, the whitelist is kind of a leftover from when we thought we'd be doing more of our own metrics with events like `newuser` and `newinstall`

groovecoder, line 20: Can we update the `docs/examples/webextension/background.js`:`fetch` to include a more realistic example of the POST body with add-on ID?

```
  body: JSON.stringify({
    service: "blok@mozilla",
    msg: { domain: "economist.com", blockedTrackers: ["googleadservices.com", "facebook.net", "doubleclick.net"], reason: "broken"}
  })
```

Or whatever the add-ons need to POST that isn't auto-filled by the server?

lmorchard:
Going to close this, for now. Hopefully we can get #1008 into usable shape and get full Telemetry pings for webextensions

PR #954, opened 2016-06-07
Add dev/stage/prod S3 bucket URLs to allowed origins for images

lmorchard:
Issue #882


lmorchard:
Just saw PR #930 get merged, and then realized it broke images on dev. This should fix it, and I'll probably self-merge after Circle CI is green

PR #957, opened 2016-06-08
Add QA deployment verification doc

pdehaan:
r? @clouserw @ckprice

### DIFF
--- a/docs/DEPLOYMENT-VERIFICATION.md
+++ b/docs/DEPLOYMENT-VERIFICATION.md
@@ -1,0 +1,170 @@
+# Test Pilot Deployment Verification Test Plan
+
+## ENVIRONMENTS:
+
+| ENVIRONMENT | URL | FXA |
+|:------------|:----|:----|
+| Development | http://testpilot.dev.mozaws.net/    | https://stable.dev.lcip.org/
+| Stage       | https://testpilot.stage.mozaws.net/ | https://accounts.firefox.com/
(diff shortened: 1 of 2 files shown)


docs/DEPLOYMENT-VERIFICATION.md
```diff
@@ -0,0 +1,168 @@
+# Test Pilot Test Plan
+
+## ENVIRONMENTS:
```
ckprice, line 3: Do you think it's worth mentioning the `Preferences` drop-down in this section? Like, "this _should_ match with the drop-down, buuuut..."

ckprice:
Thanks for this @pdehaan! Wondering if we should add a link to [the deployment documentation](https://github.com/mozilla/testpilot/blob/master/docs/DEPLOYMENT.md#test-stage) as well? Right now it just links to the Google Doc, but this is way better.

ghost:
love it. r+

docs/DEPLOYMENT-VERIFICATION.md
pdehaan, line 3: I do, added to the wall of verification steps below, so people remember to confirm after installing the add-on.

pdehaan:
> @ckprice: Wondering if we should add a link to the deployment documentation as well?

Added a note in the DEPLOYMENT.md file below the Google Docs link:

> _Follow the steps in the ["Test Pilot Deployment Verification Test Plan" doc](DEPLOYMENT-VERIFICATION.md) to verify that the site works as expected._

PR #959, opened 2016-06-08
Changes bash script to sh script.

### LINKED ISSUE #958: Fix issue with run-common.sh in Dockerfile

kaganjd:
Fixes #958

### DIFF
--- a/bin/run-common.sh
+++ b/bin/run-common.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 ./manage.py migrate --noinput
 ./manage.py createinitialsuperuser


lmorchard:
:shipit:

pdehaan:
Do we need to remove the bash references from the other scripts as well?
https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=bash&type=Code

``` sh
$ grep -irn "/bin/bash" . --exclude-dir=node_modules

./bin/deploy-dev.sh:1:#!/bin/bash
./bin/docker-build.sh:1:#!/bin/bash
./bin/update-ip.sh:1:#!/bin/bash
```

meandavejustice:
@pdehaan Each of those bash scripts are run from outside the docker container, so they should be fine to stay as is.

PR #982, opened 2016-06-14
Fix unclosed tag attribute in landing-page

pdehaan:
Removed locale-specific download pages (since Mozilla.org does that for us, and is better for l10n), and fixed an unclosed `class` attribute in landing-page.js.

r? anybody with a thumb and a desire to press a green button.


meandavejustice:
u don't even need a thumb to merge m8

PR #1004, opened 2016-06-20
Use 127.0.0.1 instead of localhost in jpm watchpost.

chuckharmston:
I struggled getting autoinstaller working; it was trying to use IPv6 and failing silently. Changing the `watchpost` command to use the `127.0.0.1` IP address directly forces it to use IPv4 and fixes that issue.

Ref: https://github.com/palant/autoinstaller/issues/8#issuecomment-215949994


lmorchard:
Works for me! :+1:

Issue #477, opened 2016-03-14
Provide variant testing framework to experiments.


lmorchard:
Issue #366 introduced data models and server-side APIs for feature flipping flags & conditions. This needs to be passed along to experiments, so let's do that.

Probably the easiest way is to do this via messaging using nsIObserver. But, investigating that should be a part of this issue.

lmorchard:
Getting add-ons to cooperate close to startup time is proving a bit of a challenge. Going to swap this one to 4/8 and see if I can't get #479 done instead

lmorchard:
Just as a note about where I left off with this: As of #422, feature flags should be appearing in the data from `/api/me`, found in the `features` property in each entry of the `installed` set. (Though, FWIW, there's an issue with the format of this data - #742 - it should probably be an object indexed by add-on ID instead of an array.)

Anyway, the goal of this issue is to get that `features` data from `/api/me` into the main Test PIlot add-on. Then, relay it to individual experiments. I was trying to set up nsIObserver messaging to make that happen, but ran into a chicken-and-egg problem: 

Add-ons load in random order. Sometimes experiments come in before the main add-on, sometimes after. So, that means, if an experiment wants to ask the main add-on for those feature flags at start up - sometimes the main add-on isn't available yet. If the main add-on wants to push the data to experiments, sometimes _they're_ not available yet.

So, this issue probably needs a clever way to handle that and ensure all experiments and the main add-on are initialized and loaded before messages go around. Maybe experiments can check whether the main add-on is there, somehow, before pinging? Also need to be careful not to break existing experiments, e.g. by requiring some kind of messaging on startup

chuckharmston:
Current proposal: https://gist.github.com/chuckharmston/b8b728844c44fa931ace881e92286b4d

ianb:
Some feedback on the [proposal](https://gist.github.com/chuckharmston/b8b728844c44fa931ace881e92286b4d):
- In many cases we want a statistically significant number of people to get B, rather than an A/B ratio.  I think SHIELD tries to be clever about this, there might be something to take from that.
- The proposal seems to have the add-on giving information to Test Pilot, including weights.  This means that changing weights would require an add-on update.  Remote control of this would seem better.  But then that means people might move between buckets if you twiddle the knobs... which should be accounted for anyway, I suppose, as part of starting and stopping experiments.
[...]

PR #1008, opened 2016-06-21
(DO NOT MERGE) Issue #433: Dirty comm channel between main add-on and webextension

lmorchard:
- New /postmessage-proxy page served by Django
- Main add-on loads /postmessage-proxy into a background page worker in
  an iframe
- Example webextension loads /postmessage-proxy into a background page
  worker in an iframe
- proxy.js on /postmessage-proxy listens for postMessage events:
  - queueTelemetryPing to enqueue a telemetry ping to be sent by the
    main add-on
  - fetchTelemetryPings to fetch all queued telemetry pings from the
    main add-on and submit them all
- GIANT HACK: proxy.js maintains the queue as a JSON-serialized array in
  a cookie, because that's the only shared persistent data for the
  iframe I could get working. (LocalStorage fails with a security error)
- add vendor gulpfile task to copy over vendor scripts
- add vendor/js.cookie.js as a lazy way to manage cookies

Made a quick drawing of the messaging flow:
flow diagram

### DIFF
--- a/addon/data/postmessageProxyFrame.js
+++ b/addon/data/postmessageProxyFrame.js
@@ -1,0 +1,25 @@
+const TELEMETRY_QUEUE_POLL_INTERVAL = 5000;
+const proxyFrame = document.getElementById('tp-proxy');
+
+// Periodically poll for queued telemetry pings
+setInterval(function() {
+  proxyFrame.contentWindow.postMessage({
+    op: 'fetchTelemetryPings'
+  }, '*');
+}, TELEMETRY_QUEUE_POLL_INTERVAL);
+
+// Listen for postMessage
+window.addEventListener('message', function(ev) {
+  if (ev.data.op === 'telemetryPings') {
+    // Relay telemetryPings on to the add-on
+    self.port.emit('telemetryPings', ev.data.data);
+    return;
+  }
+});
+
+// Listen for message from add-on to update iframe src
+self.port.on('updateIFrameSrc', src => {
+  proxyFrame.src = src;
+});
+
+self.port.emit('ready');
--- a/docs/examples/webextension/background.html
+++ b/docs/examples/webextension/background.html
@@ -1,0 +1,6 @@
+<!doctype html>
+<html>
+    <body>
+        <iframe id="tp-proxy" src="http://testpilot.dev:8000/postmessage-proxy"></iframe>
+    </body>
+</html>
--- a/docs/examples/webextension/background.js
+++ b/docs/examples/webextension/background.js
@@ -1,0 +1,18 @@
+var EXPERIMENT_ID = 'webextension-example-1';
+
+function sendTelemetryPing(data) {
+  document.getElementById('tp-proxy').contentWindow.postMessage({
+    op: 'queueTelemetryPing',
+    data: {
+      subject: EXPERIMENT_ID,
+      data: data
+    }
+  }, '*');
+}
+
+// Start sending a ping with fake data every second.
+setInterval(function () {
+  sendTelemetryPing({
+    timesThingClicked: parseInt(Math.random() * 100)
+  });
+}, 1000);
--- a/testpilot/frontend/static-src/scripts/proxy.js
+++ b/testpilot/frontend/static-src/scripts/proxy.js
@@ -1,0 +1,41 @@
+var TELEMETRY_PINGS_KEY = 'telemetryPings';
+var TELEMETRY_PINGS_MAX_COUNT = 10;
+
+// Parse queued telemetry pings from a cookie
+function parseTelemetryPings() {
+  var pings;
+  try { pings = JSON.parse(Cookies.get(TELEMETRY_PINGS_KEY)); }
+  catch (e) { pings = []; }
+  return pings;
+}
+
+window.addEventListener('message', function(ev) {
+  // Only listen for messages from add-ons and webextensions
+  if (ev.origin.indexOf('moz-extension://') !== 0 &&
+      ev.origin.indexOf('resource://testpilot-addon') !== 0) {
+    return;
+  }
+
+  if (ev.data.op === 'queueTelemetryPing') {
+    // Parse the current list of pings, add this new one.
+    var pings = parseTelemetryPings();
+    pings.push(ev.data.data);
+    // Drop some older pings, if we have too many.
+    while (pings.length > TELEMETRY_PINGS_MAX_COUNT) { pings.shift(); }
+    // Serialize the queue back into a cookie.
+    Cookies.set(TELEMETRY_PINGS_KEY, JSON.stringify(pings));
+    return;
+  }
+
+  if (ev.data.op === 'fetchTelemetryPings') {
+    // Parse the current list of pings, clear the queue.
+    var pings = parseTelemetryPings();
+    Cookies.set(TELEMETRY_PINGS_KEY, JSON.stringify([]));
+    // Send the list of pings back to the requester
+    ev.source.postMessage({
+      op: 'telemetryPings',
+      data: JSON.stringify(pings)
+    }, '*');
+    return;
+  }
+});
--- a/testpilot/frontend/views.py
+++ b/testpilot/frontend/views.py
@@ -1,6 +1,12 @@
 from django.shortcuts import render
+from django.views.decorators.clickjacking import xframe_options_exempt
 
 
 def index(request, url=''):
     return render(request, 'testpilot/frontend/index.html',
                   {'user_id': request.user.email if not request.user.is_anonymous() else ''})
+
+
+@xframe_options_exempt
+def postmessage_proxy(request):
+    return render(request, 'testpilot/frontend/proxy.html')
--- a/addon/data/postmessageProxyFrame.html
+++ b/addon/data/postmessageProxyFrame.html
@@ -1,0 +1,9 @@
+<!doctype html>
+<html>
+    <head>
+        <meta charset="utf-8"/>
+    </head>
+    <body>
+        <iframe id="tp-proxy" src="http://testpilot.dev:8000/postmessage-proxy"></iframe>
+    </body>
+</html>
--- a/addon/index.js
+++ b/addon/index.js
@@ -37,6 +37,7 @@ Mustache.parse(templates.installed);
 Mustache.parse(templates.experimentList);
 
 const Metrics = require('./lib/metrics');
+const PostmessageProxy = require('./lib/postmessage-proxy');
 const survey = require('./lib/survey');
 
 const PANEL_WIDTH = 300;
@@ -571,6 +572,7 @@ exports.main = function(options) {
 
   initServerEnvironmentPreference();
   Metrics.init();
+  PostmessageProxy.init();
 };
 
 exports.onUnload = function(reason) {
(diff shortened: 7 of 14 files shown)


lmorchard:
/cc @chuckharmston @meandavejustice @groovecoder - all you folks probably have an interest in this.

This is probably a terrible idea, because I came up with it while sleep-deprived and flying back from London. But, what this does is implement a message-passing channel between webextensions and our main bootstrap add-on.

The mechanism is a new page at `/postmessage-proxy` on the Test Pilot site. The webextension and the main add-on each load this page into an iframe inside an invisible document. That iframe accepts `postMessage()` messages. 

The webextension sends a `queueTelemetryPing` message to the iframe. The main add-on periodically sends a `fetchTelemetryPings` message to the iframe. The iframe replies to `fetchTelemetryPings` with `telemetryPings` so the add-on can pick up telemetry payloads and send them with `submitExternalPing`

Here's the nasty hacky part: In the middle, the `/postmessage-proxy` page stores the telemetry pings queue as JSON serialized in a cookie. The special magic that makes everything work is that the cookie is shared between the separate instances of the page loaded in the main add-on and the webextension.

I wanted to use LocalStorage in the `/postmessage-proxy` page. But, whenever I try to get or set anything, I get a SecurityError in the console. Hopefully this has a fix, because the cookie method is not something I'm proud of.

Future expansion of this idea, if it can be polished up, can also include the messaging required for sending variant testing configuration around (for #477).

lmorchard:
Oh, and also this commit has the webextension and main add-on both hard-coded to load `/postmessage-proxy` from local dev. That needs some work, too.

lmorchard:
Oh yeah, and this results in real normal Telemetry pings, as opposed to the server-side solution in #952

groovecoder:
What if we use IndexedDB (via localForage)? Less hacky?

lmorchard:
> What if we use IndexedDB (via localForage)? Less hacky?

Tried both the localStorage & IndexedDB drivers in localForage. Same error: `SecurityError: The operation is insecure.` I'm not quite sure if this is a legitimate security complaint, or an upstream Firefox bug. I imagine an iframe in a background page in a webextension or add-on is a fairly edgy case.

docs/examples/webextension/background.js
```diff
@@ -0,0 +1,18 @@
+var EXPERIMENT_ID = 'webextension-example-1';
+
+function sendTelemetryPing(data) {
+  document.getElementById('tp-proxy').contentWindow.postMessage({
+    op: 'queueTelemetryPing',
+    data: {
+      subject: EXPERIMENT_ID,
+      data: data
+    }
+  }, '*');
```
groovecoder, line 10: In the "real world" implementation, I assume we would tell devs to only post the message to `targetOrigin` of `testpilot.firefox.com` ?

addon/data/postmessageProxyFrame.js
```diff
@@ -0,0 +1,17 @@
...
+
+// Periodically poll for queued telemetry pings
+setInterval(function() {
+  document.getElementById('tp-proxy').contentWindow.postMessage({
+    op: 'fetchTelemetryPings'
+  }, '*');
+}, TELEMETRY_QUEUE_POLL_INTERVAL);
+
+// Listen for postMessage
+window.addEventListener('message', function(ev) {
```
groovecoder, line 12: And would we want to check for `ev.origin == ?` here? (Though I'm not sure what the origin will be when the message comes from the add-on and/or web extension.)

testpilot/frontend/static-src/scripts/proxy.js
```diff
@@ -0,0 +1,41 @@
...
+    while (pings.length > TELEMETRY_PINGS_MAX_COUNT) { pings.shift(); }
+    // Serialize the queue back into a cookie.
+    Cookies.set(TELEMETRY_PINGS_KEY, JSON.stringify(pings));
+    return;
+  }
+
+  if (ev.data.op === 'fetchTelemetryPings') {
+    // Parse the current list of pings, clear the queue.
+    var pings = parseTelemetryPings();
+    Cookies.set(TELEMETRY_PINGS_KEY, JSON.stringify([]));
```
groovecoder, line 33: Should we wait until we've (successfully) posted the message before we empty the queue?

addon/data/postmessageProxyFrame.js
groovecoder, line 12: Ah, so this addon event listener will be getting the messages from testpilot.firefox.com ? So maybe that's the one we want to verify?

groovecoder:
I'm still trying to get my local testpilot install to work, but then I'll be eager to try this out with my extension. I agree the cookie thing is a bit hacky, but it does seem to be the best we can do for now? Do you have a bug # or issue # for the security error thrown by localStorage and IndexedDB?

groovecoder:
Would you want to do this instead of #952 ? Or in addition to it?

lmorchard:
> Would you want to do this instead of #952 ? Or in addition to it?

Instead of. This would give us normal full Telemetry pings for all experiment types.

lmorchard:
> Do you have a bug # or issue # for the security error thrown by localStorage and IndexedDB?

No, I have no idea yet whether it's a bug or working as intended. Still researching that / hoping someone else knows what's up.

testpilot/frontend/static-src/scripts/proxy.js
lmorchard, line 33: Problem with that, is there's no way to keep more queue items from rolling in while we're waiting for an acknowledgement. Also would require a significantly more complex system of messaging that we probably don't need for Telemetry pings.

docs/examples/webextension/background.js
lmorchard, line 10: Yeah, but until/unless I figure out a way to configure this for dev/stage/prod environment it's `*` for now

addon/data/postmessageProxyFrame.js
lmorchard, line 12: Same as above - the messages can come from one of prod / stage / dev / local and I don't have a way to configure that yet

groovecoder:
FWIW I've added both methods of boilerplate/example code into my extension: https://github.com/groovecoder/blok/compare/disable-and-ask-why-4?expand=1#diff-d8c11c63dfd474a75f05fbbdba0889f8R27

If I can get testpilot running locally I'll try them both out.

lmorchard:
FWIW, I also just tried setting the `postMessage` origin to `http://testpilot.dev:8000` rather than `*` - and now I get this error:

```
Failed to execute ‘postMessage’ on ‘DOMWindow’: The target origin provided
(‘http://testpilot.dev:8000’) does not match the recipient window’s origin 
(‘http://testpilot.dev:8000’).
```

😕 The hell? I feel like I've strayed into a minefield of corner cases here.

testpilot/frontend/views.py
```diff
@@ -1,6 +1,12 @@
 from django.shortcuts import render
+from django.views.decorators.clickjacking import xframe_options_exempt
 
 
 def index(request, url=''):
     return render(request, 'testpilot/frontend/index.html',
                   {'user_id': request.user.email if not request.user.is_anonymous() else ''})
+
+
+@xframe_options_exempt
```
groovecoder, line 10: Any chance we could use `parent-src` CSP around this to protect it from click-jacking? (IIRC `ALLOW-FROM` doesn't allow multiple origin values?)

docs/examples/webextension/background.html
```diff
@@ -0,0 +1,6 @@
+<!doctype html>
+<html>
+    <body>
+        <iframe id="tp-proxy" src="http://testpilot.dev:8000/postmessage-proxy">
```
groovecoder: needs `</iframe>`

groovecoder:
Wow, so [this actually works in my web extension](https://github.com/groovecoder/blok/commit/13cd284bcce344e60f8df521fd7259185a5d6124). :tada:

**Except** I couldn't put the `<iframe>` directly into my pageAction popup 😞 I had to:
1. `postMessage` from my pageAction popup to my background page
2. `postMessage` from my background page to the testpilot postmessage-proxy `<iframe>`

Still, I:
1. Disabled Test Pilot
2. Clicked the `postMessage`-firing button in my pageAction pup-up
3. Loaded up http://testpilot.dev:8000/postmessage-proxy
4. Saw my data in the `telemetryPings` cookie
5. Re-enabled Test Pilot
6. Waited a few seconds
7. Refreshed the postmessage-proxy page
8. Saw that Test Pilot had process my telemetry pings

I see `addon/dev-prefs.json` has `"toolkit.telemetry.server": "http://127.0.0.1"` ... any chance I can point that at a staging/sandbox telemetry IP to see this actually hit the telemetry pipeline?

Oh, and here's a fun video reward for making such an amazingly-creative solution:

https://www.youtube.com/watch?v=qybUFnY7Y8w

lmorchard:
> I see addon/dev-prefs.json has "toolkit.telemetry.server": "http://127.0.0.1" ... any chance I can point that at a staging/sandbox telemetry IP to see this actually hit the telemetry pipeline?

There's no sandbox for Telemetry as far as I know. The 127.0.0.1 server is intended to be an instance of [vdjeric/gzipServer](https://github.com/vdjeric/gzipServer), but it could be any HTTP server to see the POST requests. There's an about:config setting to turn on telemetry console messages when requests go out, too, but I can't remember what it is :/

lmorchard:
> Except I couldn't put the `<iframe>` directly into my pageAction popup

Hmm, is that not a spot where [`chrome.runtime.sendMessage()`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts#Communicating_with_background_scripts) could work? Not sure because I haven't actually gotten that far with building a webextension myself, though.

groovecoder:
Probably could use that too. Not sure if it gets me anything over postMessage right now. But my add-on is growing large enough that some actual architecture and design is in order. I'll keep `sendMessage` in mind for that.

lmorchard:
Not seeing any other feedback or ideas on this... @chuckharmston? @meandavejustice?

Might just see what I can do to clean up the last loose ends and merge this Monday if I don't get any more eyes on it before end of day. Mainly:
- [ ] <strike>Find a simpler / smaller cookie library?</strike>
- [x] Point the proxy frames at testpilot.firefox.com - or find some way to configure them based on main addon environment selection (might only be possible for the main addon itself)
- [ ] <strike>Tests? (not sure how feasible without an integration test solution)</strike>

Just going to go with the cookie hack until/unless a better working solution reveals itself.

groovecoder:
Not sure if it's any better, but ...

For another add-on we've coded it such that it uses whichever site was last visited in the browser. I.e., when the browser goes to testpilot.firefox.com, a content script configures the add-on to use testpilot.firefox.com, when the user goes to testpilot.dev:8000, the content script configures the add-on to use testpilot.dev:8000.

But that just shifts the issue - still hard to know which site you're using when doing development. Hmm ... It also means each web extension experiment would have to implement a similar mechanism, so that's double-bad.

groovecoder:
So yeah - 👍 to hard-code testpilot.firefox.com into the add-ons and the extension experiments by default. And make devs go in and change the code to use other environments.

lmorchard:
Aaaaand, of course, just as I'm about to merge this stuff... I've been told about [BroadcastChannel](https://hacks.mozilla.org/2015/02/broadcastchannel-api-in-firefox-38/), which [looks like it might be a way cleaner comm channel](https://github.com/rpl/testpilot-webextension-experiment/commit/3a2cadeb2f955926887d462419c894df4ebe1c77) for all this. Going to look into that really quick before merging this.

lmorchard:
Aaand BroadcastChannel is _way_ simpler & cleaner than this. So, closing in favor of PR #1022

Issue #610, opened 2016-04-19
Simplify addon/index.js


meandavejustice:
At the time of creating this issue `addon/index.js` is 527 lines long. There are some things that could easily be pulled out into separate modules in `lib/`. Comments should also be audited as a part of this effort.

lmorchard:
Been thinking about this a little bit, too, but haven't had time to chase things down. 

One big thing I'd like to see happen is to move most of the initialization stuff scattered across random spots throughout `index.js` and move it to `init()` functions in modules - basically like what I did with [`Metrics.init()`](https://github.com/mozilla/testpilot/blob/master/addon/lib/metrics.js#L29) that gets [called in `exports.main`](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L505). Then, `exports.main` can be a clearer table-of-contents of what happens on startup.

meandavejustice:
I like what you did in `Metrics.init` and `exports.main`. My dream would be to have the entire file just pull things into `exports.main` and kicked off from there. But that dream may be unattainable.

johngruen:
@dannycoates @meandavejustice can we close this given recent changes?

dannycoates:
I think this is an ongoing effort. I don't know if it makes sense to keep this bug open or not though. I'll defer to @meandavejustice since he opened it :)

meandavejustice:
Let's keep it open, We've gotten it a lot nicer, but `addon/index.js` is still pretty big.

lmorchard:
Yeah, no, this is not done. A few things off the top of my head yet to do:
- [ ] Move AddonManager & add-on management stuff into a submodule.
- [ ] Move environment management stuff into a submodule.
- [ ] Unit tests for everything for bonus points

Really, IMO, index.js should boil down to a screen or so of initializing everything at startup and tearing everything down on shutdown.

We could file separate issues for these things, but maybe keep this one open to track the overall effort. There might be more things we can clarify / split out as we go

chuckharmston:
Proposal for an end state: a very, very thin version of Django's `apps` functionality:

```
const modules = [
        Metrics,
        WebExtensionChannels,
        SharePrompt,
        ToolbarButton,
        ExperimentNotifications,
        SharePrompt
];

exports.main = options => {
        modules.forEach(module => {
                module.setup(options.reason, settings);
        });
};

exports.onUnload = reason => {
        modules.reverse().forEach(module => {
                module.teardown(reason, settings);
        });
}
```

[...]

PR #1009, opened 2016-06-22
Add variant testing framework for experiments.

chuckharmston:
So, here's a very rough sketch at what a fix for #477 could look like. It works like this:
1. Upon initialization, an experiment spams Test Pilot with a [test and variant definition](https://gist.github.com/chuckharmston/b8b728844c44fa931ace881e92286b4d#experiment---test-pilot) using [`nslObserverService`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIObserverService) and the `testpilot::register-variants` topic. This is done in an interval, as there is no guarantee that experiments are loaded after Test Pilot is.
2. The first time Test Pilot receives that message, it [does some magic](https://gist.github.com/chuckharmston/b8b728844c44fa931ace881e92286b4d) and deterministically (seeded by test name and `clientUUID`) places users into variants for each test.
3. Test Pilot stores and ships a payload containing those results back to the experiment, using `testpilot::receive-variants`. When received, the experiment clears the interval and can do whatever it would like with the payload.
4. In any future `testpilot::send-metric` messages sent by experiments, the variants are appended to the payload before `submitExternalPing` is called. This will let us segment any metric based on variant placement.

In addition to the code in the PR, we'd include something approximating the following in some sort of experiment developer SDK:

``` javascript
const {classes: Cc, interfaces: Ci, utils: Cu} = Components;

function TestPilotObserver(topic, resolve) {
  this.topic = topic;
  this.resolve = resolve;
  this.register();
}

TestPilotObserver.prototype = {
  observe: function(subject, topic, data) {
    this.resolve({
      subject: subject,
      topic: topic,
      data: data
    });
    this.unregister();
  },

  register: function() {
    let observerService = Cc["@mozilla.org/observer-service;1"]
                          .getService(Ci.nsIObserverService);
    observerService.addObserver(this, this.topic, false);
  },

  unregister: function() {
    let observerService = Cc["@mozilla.org/observer-service;1"]
                          .getService(Ci.nsIObserverService);
    observerService.removeObserver(this, this.topic);
  }
};

const registerVariants = function(addonId, payload) {
  let observerService = Cc["@mozilla.org/observer-service;1"]
                        .getService(Ci.nsIObserverService);

  let receiveTopic = 'testpilot::receive-variants';
  let registerTopic = 'testpilot::register-variants';

    let intervalPing = win.setInterval(() => {
      observerService.notifyObservers(
        {
          wrappedJSObject: {
            observersModuleSubjectWrapper: true,
            object: addonId
          }
        },
        registerTopic,
        JSON.stringify(payload)
      );
    }, 100);
    let observer = new TestPilotObserver(receiveTopic, data => {
      win.clearInterval(intervalPing);
      resolve(data);
    });
    observerService.addObserver(observer, receiveTopic, false);
  });
};
```

Which the experiments themselves would use like this:

``` javascript
registerVariants('universal-search@mozilla.com', payload).then(payload => {
  app.console.log(JSON.parse(payload.data));
});
```

Using the above code and [this sample JSON blob as the payload](https://gist.github.com/chuckharmston/b8b728844c44fa931ace881e92286b4d#experiment---test-pilot), Universal Search receives the following object:

screen shot 2016-06-21 at 6 27 56 pm

And the telemetry pings sent by Universal Search look like this:

screen shot 2016-06-21 at 6 23 07 pm

This is all very messy, but I wanted to get eyes on it early. r? @lmorchard 

Potentially-interested parties: @k88hudson @rjweiss

### DIFF
--- a/addon/lib/metrics.js
+++ b/addon/lib/metrics.js
@@ -12,11 +12,16 @@ Cu.import('resource://gre/modules/TelemetryController.jsm');
 
 const { setTimeout, clearTimeout } = require('sdk/timers');
 const Events = require('sdk/system/events');
-const store = require('sdk/simple-storage').storage;
 const PrefsService = require('sdk/preferences/service');
+const self = require('sdk/self');
+const store = require('sdk/simple-storage').storage;
+
+const seedrandom = require('seedrandom');
 
 // Event type for receiving pings from experiments
 const EVENT_SEND_METRIC = 'testpilot::send-metric';
+const EVENT_RECEIVE_VARIANT_DEFS = 'testpilot::register-variants';
+const EVENT_SEND_VARIANTS = 'testpilot::receive-variants';
 
 // Minimum interval of time between main Telemetry pings in ms
 const PING_INTERVAL = 24 * 60 * 60 * 1000;
@@ -33,6 +38,30 @@ const PREFERENCE_OVERRIDES = {
 
 let pingTimer = null;
 
+const variantMaker = {
+  makeTest: function(test) {
+    let summedWeight = 0;
+    const variants = [];
+    test.variants.forEach(variant => {
+      summedWeight += variant.weight;
+      for (let i = 0; i < variant.weight; i++) {
+        variants.push(variant.value);
+      }
+    });
+    const seed = `${test.name}_${store.clientUUID}`;
+    return variants[Math.floor(seedrandom(seed)() * summedWeight)];
+  },
+
+  parseTests: function(tests) {
+    const results = {};
+    Object.keys(tests).forEach(key => {
+      results[key] = this.makeTest(tests[key]);
+    });
+    return results;
+  }
+};
+
+
 module.exports = {
 
   init: function() {
@@ -47,6 +76,7 @@ module.exports = {
     this.maybePingTelemetry();
 
     Events.on(EVENT_SEND_METRIC, this.onExperimentPing);
+    Events.on(EVENT_RECEIVE_VARIANT_DEFS, this.onReceiveVariantDefs);
   },
 
   onEnable: function() {
@@ -103,8 +133,10 @@ module.exports = {
   },
 
   updateExperiment: function(addonId, data) {
+    const features = (addonId in store.experimentVariants ?
+                      store.experimentVariants[addonId] : {});
     store.telemetryPingPayload.tests[addonId] = Object.assign(
-      store.telemetryPingPayload.tests[addonId] || {features: {}},
+      store.telemetryPingPayload.tests[addonId] || {features: features},
       data
     );
   },
@@ -121,12 +153,31 @@ module.exports = {
     this.updateExperiment(addonId, {features: features});
   },
 
+  onReceiveVariantDefs: function(ev) {
+    if (!store.experimentVariants) {
+      store.experimentVariants = {};
+    }
+
+    const { subject, data } = ev;
+    const dataParsed = variantMaker.parseTests(JSON.parse(data));
+
+    store.experimentVariants[subject] = dataParsed;
+    Events.emit(EVENT_SEND_VARIANTS, {
+      data: JSON.stringify(dataParsed),
+      subject: self.id
+    });
+  },
+
   onExperimentPing: function(ev) {
     const { subject, data } = ev;
     const dataParsed = JSON.parse(data);
 
     // TODO: Map add-on ID (subject) to other pingTypes as necessary
     const pingType = 'testpilottest';
+
+    if (subject in store.experimentVariants) {
+      dataParsed.variants = store.experimentVariants[subject];
+    }
 
     const payload = {
       version: 1,
--- a/docs/examples/addon-restartless/bootstrap.js
+++ b/docs/examples/addon-restartless/bootstrap.js
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
+
+Cu.import('resource://gre/modules/XPCOMUtils.jsm');
+let console = Cu.import('resource://gre/modules/Console.jsm').console;
+let {setTimeout} = Cu.import('resource://gre/modules/Timer.jsm', {});
+
+XPCOMUtils.defineLazyModuleGetter(this, 'Services',
+  'resource://gre/modules/Services.jsm');
+XPCOMUtils.defineLazyModuleGetter(this, 'registerVariants',
+  'chrome://restartless-experiment/content/variants.js');
+
+
+const addonId = 'restartless-experiment@mozilla.com';
+let variants;
+
+// Define your tests...
+const variantTests = {
+  someFeature: {
+    name: 'Some Feature',
+    description: 'An example A/B test.',
+    variants: [
+      { description: 'A', value: 'A', weight: 1 },
+      { description: 'B', value: 'B', weight: 1 }
+    ]
+  }
+};
+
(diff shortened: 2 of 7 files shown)


MattGrimes:
@gregglind will also be interested here and probably have some good advice.

lmorchard:
Starting to take a look at this today (I hope). 

But, please don't assign me to things I haven't gotten a look at yet. We're using that as a signal to mean "I'm actively on this" - not as a request to look at a thing. If I'm busy on other things, my name on something means other folks might skip it when it could use some help sooner than I can get to it.

lmorchard:
Can you roll the experiment-side code that uses this into [the example add-on in `docs/examples/addon-sdk`](https://github.com/mozilla/testpilot/tree/master/docs/examples/addon-sdk)? Would be easier to see how this all fits together in a working demo.

chuckharmston:
> Can you roll the experiment-side code that uses this into [the example add-on in `docs/examples/addon-sdk`](https://github.com/mozilla/testpilot/tree/master/docs/examples/addon-sdk)? Would be easier to see how this all fits together in a working demo.

I can work towards that, but it's not copy/paste; the example code I wrote was for a restartless add-on, not an SDK one.

lmorchard:
> I can work towards that, but it's not copy/paste; the example code I wrote was for a restartless add-on, not an SDK one.

Or even start a new minimal restartless add-on next to the SDK one in examples - anything that's working and shows how it connects together

chuckharmston:
@lmorchard Did so in 2149802. Checking out this PR and adding a [proxy file](https://developer.mozilla.org/en-US/Add-ons/Setting_up_extension_development_environment#Firefox_extension_proxy_file) would be the easiest way to test it out.

addon/lib/metrics.js
```diff
@@ -33,6 +38,30 @@ const PREFERENCE_OVERRIDES = {
 
 let pingTimer = null;
 
+const variantMaker = {
```
lmorchard, line 41: Kind of a nit: Should this stuff really be a part of `metrics.js`? We do want to send the variant info in pings, but it seems like the rest might live better in its own module. We could defer to #610 as part of other refactorings

lmorchard:
Giving the example add-on a spin...

addon/lib/metrics.js
```diff
@@ -121,13 +151,32 @@ module.exports = {
...
+  },
+
   onExperimentPing: function(ev) {
     const { subject, data } = ev;
     const dataParsed = JSON.parse(data);
 
     // TODO: Map add-on ID (subject) to other pingTypes as necessary
     const pingType = 'testpilottest';
 
+    if (subject in store.experimentVariants) {
```
lmorchard, line 178: We didn't specify this data in the `testpilottest` ping - though that's a good place for it. 

But, we _did_ (loosely) specify it [in the main `testpilot` ping under the `features` key for each listed experiment](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md#data-collection-details). We should add this data to the main ping - either when [an experiment is updated in the ping](https://github.com/mozilla/testpilot/blob/master/addon/lib/metrics.js#L105) or at the [time of submission](https://github.com/mozilla/testpilot/blob/master/addon/lib/metrics.js#L97).

docs/examples/addon-restartless/bootstrap.js
```diff
@@ -0,0 +1,42 @@
...
+      { description: 'B', value: 'B', weight: 1 }
+    ]
+  }
+};
+
+function startup(data, reason) {
+  // Then, on startup, ask Test Pilot what to do.
+  registerVariants('restartless-experiment@mozilla.com', variantTests).then(payload => {
+    // When Test Pilot responds, finish initialization with the payload:
+    console.log('Finish initialization with:', JSON.parse(payload.data));
```
lmorchard: Should also probably add some code here that fires off a telemetry ping, to exercise the bit that adds variant info to the metrics payload. Took me a few minutes to realize it wasn't actually showing up in the main ping

lmorchard:
Okay, I think I've beat on this a bit - works & looks good, though I would like to see the variant data added to the main ping. r+ from me after that.

Other stuff is nitpicky and we can follow up with docs & further developed examples

addon/lib/metrics.js
chuckharmston, line 178: > We didn't specify this data in the testpilottest ping - though that's a good place for it. 

Why is that a good place for it? I feel like Test Pilot should be agnostic to experiments, while the experiments themselves should be able to segment on these values.

chuckharmston, line 178: Also, quoting the metrics readme:

> The second payload's schema will be defined by each experiment and needs to remain flexible for rapid changes.

I read this to say that, as long as this happens:

> Please note that each experiment will need to define its payload schema before being able to record data.

We're able to go freeform with the payload as far as the Test Pilot addon is concerned, yeah?

gregglind:
(hey peeps, related  (OR IS IT????) work at https://github.com/gregglind/shield-studies-addon-utils.

I have done tons of testing of maintaining branches across restarts, and other horrible things)

addon/lib/metrics.js
lmorchard, line 178: > Why is that a good place for it?

Actually, I misread: I thought this was in the payload envelope (down on line 180) and not in the `dataParsed`... the envelope is actually a better place for that variant data, if we're adding it to the `testpilottest` ping. Still, we'll probably need to make some downstream schema changes to pick it up in a useful way. (and the `version` probably needs a bump, as well)

But yeah, we _shouldn't_ be adding this directly to the data sent from the experiment in the ping.

lmorchard, line 178: That said, we should still have this stuff in the main `testpilot` ping

chuckharmston, line 178: I still don't understand why you want this on the `testpilot` ping, but not the `testpilottest` ping. Isn't the entire point to be able to segment on metrics that the experiments themselves are collecting?

chuckharmston:
Added a metrics ping in the sample add-on.

addon/lib/metrics.js
```diff
@@ -103,8 +133,10 @@ module.exports = {
   },
 
   updateExperiment: function(addonId, data) {
+    const features = (addonId in store.experimentVariants ?
+                      store.experimentVariants[addonId] : {};
```
lmorchard: Missing a `)` before the `;`, here

chuckharmston: Yeah, I think I accidentally deleted it while saving. Pushing a fix!

chuckharmston:
@lmorchard added the payload to the main ping, I believe this is ready to merge.

chuckharmston:
~~Errr, don't merge it yet. I had a rebase issue I'm working through.~~

chuckharmston:
All good, should be ready to merge for real.

addon/lib/metrics.js
```diff
@@ -103,8 +133,10 @@ module.exports = {
   },
 
   updateExperiment: function(addonId, data) {
+    const features = (addonId in store.experimentVariants ?
+                      store.experimentVariants[addonId] : {});
     store.telemetryPingPayload.tests[addonId] = Object.assign(
-      store.telemetryPingPayload.tests[addonId] || {features: {}},
+      store.telemetryPingPayload.tests[addonId] || {features: features},
```
lmorchard, line 139: Thinking through this & playing with it on my machine, it occurs to me that this won't work quite right.

The `|| {features: features}` part only works on the first initialization, where there's no ping data for an experiment yet. So, if an experiment is already enabled and it later adds or changes variants, this info never gets updated.

lmorchard:
This also could use a bump in the version for the add-on.

Think I'll do that myself for expediency and merge this, then submit a followup PR to tweak the main ping update.

lmorchard:
https://github.com/mozilla/testpilot/compare/8b26ffc1baa7...85e595ad907f

PR #1012, opened 2016-06-23
Add a fancy bleu background for legal Terms of Use

### LINKED ISSUE #1010: Highlight bold portions of the TOU
This is a follow on from #1005 and https://bugzilla.mozilla.org/show_bug.cgi?id=1281242

We should eventually restyle the bold text in the ToU to match the following: 

highlight-copy

pdehaan:
Fixes #1010


pdehaan:
firefox_test_pilot

pdehaan:
Counter proposal to embedding HTML in the Markdown templates is some Sass which [ab]uses `nth-child()`, like-a-so:

``` html
<style>
  .legal-container li:nth-child(8) p {
    background-color: #d3e7fc;
    padding: 1em;
  }
</style>
```

Obvious "con" here is that if we tweak the legal doc templates (ie: remove any FxA sections) or reorder sections, we'll have to adjust the child number accordingly.

PR #1013, opened 2016-06-23
Remove the bold from the legal Terms of Use page

pdehaan:
Totally looks like

firefox_test_pilot

r? @johngruen


johngruen:
R++

PR #1022, opened 2016-06-28
Use BroadcastChannel to allow WebExtensions to communicate with Test Pilot

### LINKED ISSUE #433: Need a way to connect WebExtension tests to the Test Pilot add-on.

lmorchard:
- New lib/webextension-channels module to manage channels for every
  enabled experiment
- Each "channel" creates a background page in the WebExtension's origin
  and listens for BroadcastChannel messages from the WebExtension
- Example WebExtension that uses the BroadcastChannel

Fixes #433

### DIFF
--- a/addon/lib/webextension-channels.js
+++ b/addon/lib/webextension-channels.js
@@ -1,0 +1,171 @@
+/*
+ * This Source Code is subject to the terms of the Mozilla Public License
+ * version 2.0 (the 'License'). You can obtain a copy of the License at
+ * http://mozilla.org/MPL/2.0/.
+ */
+
+/* global XPCOMUtils, Services */
+
+const { Ci, Cu } = require('chrome');
+const { Class } = require('sdk/core/heritage');
+const { Disposable } = require('sdk/core/disposable');
+
+const store = require('sdk/simple-storage').storage;
+
+const Metrics = require('./metrics');
+
+const TESTPILOT_TELEMETRY_CHANNEL = 'testpilot-telemetry';
+
+Cu.import('resource://gre/modules/XPCOMUtils.jsm');
+
+XPCOMUtils.defineLazyModuleGetter(this, 'Services',
+                                  'resource://gre/modules/Services.jsm');
+
+const {getExtensionUUID} = Cu.import('resource://gre/modules/Extension.jsm', {});
+
+function createChannelForAddonId(name, addonId) {
+  // The BroadcastChannel API allows messaging between different windows that
+  // share the same origin. Bug 1186732 extended this to WebExtensions (which
+  // may not have an origin) by adding a special URL that loads an about:blank
+  // page at the (generalized) "origin" of the extension.
+  //
+  // Load that about:blank page, and use its `window` to get a BroadcastChannel
+  // that allows two-way communication between the main Test Pilot extension and
+  // individual experiment extensions.
+
+  // Note: the `targetExtensionUUID` is different for each copy of Firefox,
+  // so that malicious websites can't guess the special URL associated with
+  // an extension.
+  const targetExtensionUUID = getExtensionUUID(addonId);
+
+  // Create the special about:blank URL for the extension.
+  const baseURI = Services.io
+        .newURI(`moz-extension://${targetExtensionUUID}/_blank.html`, null, null);
+
+  // Create a principal (security context) for the generalized origin given
+  // by the extension's special URL and its `addonId`.
+  const principal = Services.scriptSecurityManager
+        .createCodebasePrincipal(baseURI, { addonId });
+
+  // Create a hidden window and open the special about:blank page for the
+  // extension.
+  const addonChromeWebNav = Services.appShell.createWindowlessBrowser(true);
+  const docShell = addonChromeWebNav.QueryInterface(Ci.nsIInterfaceRequestor) // eslint-disable-line new-cap
+        .getInterface(Ci.nsIDocShell);
+  docShell.createAboutBlankContentViewer(principal);
+  const window = docShell.contentViewer.DOMDocument.defaultView;
+
+  // Finally, get the BroadcastChannel associated with the extension.
+  const addonBroadcastChannel = new window.BroadcastChannel(name);
+
+  // Callers need to keep the pointer to the window, otherwise the window's
+  // BroadcastChannel will get garbage collected.
+  return {
+    addonChromeWebNav,
+    addonBroadcastChannel
+  };
+}
+
+const WebExtensionChannel = Class({ // eslint-disable-line new-cap
+  implements: [Disposable],
+
+  initialize(targetAddonId) {
+    this.pingListeners = new Set();
+
+    this.targetAddonId = targetAddonId;
+
+    const {
+      addonChromeWebNav,
+      addonBroadcastChannel
+    } = createChannelForAddonId(TESTPILOT_TELEMETRY_CHANNEL, targetAddonId);
+
+    // NOTE: Keep a ref to prevent it from going away during garbage collection
+    // (or the BroadcastChannel will stop working).
+    this.addonChromeWebNav = addonChromeWebNav;
+    this.addonBroadcastChannel = addonBroadcastChannel;
+
+    this.handleEventBound = ev => this.handleEvent(ev);
+  },
+
+  dispose() {
+    this.addonBroadcastChannel.removeEventListener('message', this.handleEventBound);
+    this.addonBroadcastChannel.close();
+    this.addonChromeWebNav.close();
+    this.pingListeners.clear();
+
+    this.addonBroadcastChannel = null;
+    this.addonChromeWebNav = null;
+  },
+
+  registerPingListener(callback) {
+    this.pingListeners.add(callback);
+
+    if (this.pingListeners.size >= 0) {
+      this.addonBroadcastChannel.addEventListener('message', this.handleEventBound);
+    }
+  },
+
+  unregisterPingListener(callback) {
+    this.pingListeners.delete(callback);
+
+    if (this.pingListeners.size === 0) {
+      this.addonBroadcastChannel.removeEventListener('message', this.handleEventBound);
+    }
+  },
+
+  handleEvent(event) {
+    if (event.data) {
+      this.notifyPing(event.data, {addonId: this.targetAddonId});
+    }
+  },
+
+  notifyPing(data, sender) {
+    for (let pingListener of this.pingListeners) { // eslint-disable-line prefer-const
+      try {
+        pingListener({
+          senderAddonId: sender.addonId,
+          testpilotPingData: data
+        });
+      } catch (err) {
+        console.error('Error executing pingListener', err); // eslint-disable-line no-console
+      }
+    }
+  }
+});
+
+let channels = {};
+
+module.exports = {
+  WebExtensionChannel,
+
+  // Update all the channels on init.
+  init() {
+    this.updateExperimentChannels();
+  },
+
+  // Drop refs to channels for garbage collection
+  destroy() {
+    channels = {};
+  },
+
+  // Rebuild channels for all known experiments
+  updateExperimentChannels() {
+    channels = {};
+    if (store.installedAddons) {
+      Object.keys(store.installedAddons).forEach(id => {
+        const channel = new WebExtensionChannel(id);
+        channels[id] = channel;
+        channel.registerPingListener(data =>
+          this.handleWebExtensionPing(id, data));
+      });
+    }
+  },
+
+  // Pass a ping message along to Telemetry via Metrics
+  handleWebExtensionPing(id, data) {
+    Metrics.onExperimentPing({
+      subject: id,
+      data: JSON.stringify(data)
+    });
+  }
+};
--- a/docs/examples/webextension/background.js
+++ b/docs/examples/webextension/background.js
@@ -1,0 +1,12 @@
+var TESTPILOT_TELEMETRY_CHANNEL = 'testpilot-telemetry';
+var testpilotPingChannel = new BroadcastChannel(TESTPILOT_TELEMETRY_CHANNEL);
+setInterval(function () {
+  testpilotPingChannel.postMessage({
+    boolData: true,
+    arrayOfData: ["one", "two", "three"],
+    nestedData: {
+      intData: 10,
+    },
+  });
+  console.log("TEST PILOT PING SENT", Date.now());
+}, 5000);
--- a/addon/index.js
+++ b/addon/index.js
@@ -38,6 +38,7 @@ Mustache.parse(templates.experimentList);
 
 const Metrics = require('./lib/metrics');
 const survey = require('./lib/survey');
+const WebExtensionChannels = require('./lib/webextension-channels');
 
 const PANEL_WIDTH = 300;
 const FOOTER_HEIGHT = 50;
@@ -233,8 +234,8 @@ panel.on('show', showExperimentList);
 panel.port.on('back', showExperimentList);
 
 function showExperimentList() {
-  panel.port.emit('show', getExperimentList(store.availableExperiments,
-                                            store.installedAddons));
+  panel.port.emit('show', getExperimentList(store.availableExperiments || {},
+                                            store.installedAddons || {}));
 }
 
 panel.port.on('link', url => {
@@ -475,6 +476,7 @@ const addonListener = {
         version: addon.version
       });
       Metrics.experimentEnabled(addon.id);
+      WebExtensionChannels.updateExperimentChannels();
     }
   },
   onDisabled: function(addon) {
@@ -486,6 +488,7 @@ const addonListener = {
         version: addon.version
       });
       Metrics.experimentDisabled(addon.id);
+      WebExtensionChannels.updateExperimentChannels();
     }
   },
   onUninstalling: function(addon) {
@@ -511,6 +514,7 @@ const addonListener = {
       }
 
       Metrics.experimentDisabled(addon.id);
+      WebExtensionChannels.updateExperimentChannels();
     }
   }
 };
@@ -525,6 +529,7 @@ const installListener = {
                formatInstallData(install, addon), addon);
     });
     Metrics.experimentEnabled(addon.id);
+    WebExtensionChannels.updateExperimentChannels();
   },
   onInstallFailed: function(install) {
     app.send('addon-install:install-failed', formatInstallData(install));
@@ -571,6 +576,7 @@ exports.main = function(options) {
 
   initServerEnvironmentPreference();
   Metrics.init();
+  WebExtensionChannels.init();
 };
 
 exports.onUnload = function(reason) {
@@ -579,6 +585,7 @@ exports.onUnload = function(reason) {
   panel.destroy();
   button.destroy();
   Metrics.destroy();
+  WebExtensionChannels.destroy();
 
   if (reason === 'uninstall' || reason === 'disable') {
     Metrics.onDisable();
--- a/docs/examples/webextension/manifest.json
+++ b/docs/examples/webextension/manifest.json
@@ -1,0 +1,19 @@
+{
+  "manifest_version": 2,
+  "name": "Test Pilot WebExtension Example",
+  "version": "1.0",
+  "description": "This is a WebExtension built as an example Test Pilot experiment",
+  "icons": {
+    "32": "icons/icon-32.png"
+  },
+  "permissions": ["background"],
+  "applications": {
+    "gecko": {
+      "id": "testpilotexample1@mozilla.org",
+      "strict_min_version": "45.0"
+    }
+  },
+  "background": {
+    "scripts": ["background.js"]
+  }
+}


groovecoder:
I've got my stuff working with the old iframe postMessage branch for today's meeting/demo. But I'll look at switching to this tomorrow.

lmorchard:
@groovecoder Yeah, I think this is what we're going to end up going with. [It's _way_ simpler, and doesn't require an iframe page](https://github.com/mozilla/testpilot/pull/1022/commits/506f83888d80c83e5bbebe805cccbc93caa314e7#diff-471217898021b66d8e609e881600b0f3R1) - or even necessarily having a local Test Pilot server working.

docs/examples/webextension/background.js
```diff
@@ -0,0 +1,12 @@
+var TESTPILOT_TELEMETRY_CHANNEL = 'testpilot-telemetry';
+var testpilotPingChannel = new BroadcastChannel(TESTPILOT_TELEMETRY_CHANNEL);
+setInterval(function () {
+  testpilotPingChannel.postMessage({
+    boolData: true,
+    arrayOfData: ["one", "two", "three"],
+    nestedData: {
+      intData: 10,
+    },
+  });
```
groovecoder, line 10: Could use some elaboration on what the `postMessage` payload should contain. Just a JSON blob of data? Any required properties?

lmorchard:
FWIW, this is largely based on @rpl's work [experimenting with BroadcastChannel](https://github.com/rpl/testpilot-webextension-experiment/commit/3a2cadeb2f955926887d462419c894df4ebe1c77)

docs/examples/webextension/background.js
lmorchard, line 10: Same as the other mechanism - just a JS object with what you want included in the payload of the Telemetry ping. Test Pilot itself is agnostic on this, so it's more a concern between you & the metrics team to spec out.

lmorchard, line 10: That is to say, what you send here is the `...` in the [example ping shown in README-METRICS.md for the `testpilottest` ping](https://github.com/mozilla/testpilot/blob/master/docs/README-METRICS.md#per-experiment-testpilottest-ping) - Test Pilot wraps that in the other envelope info and the full client environment data.

groovecoder:
I ran `npm run package` and tried to install the built `.xpi` file, but I got:

```
testpilot-addon:TypeError: can't convert undefined to object
Stack trace:
updateExperimentChannels@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-addon/lib/webextension-channels.js:130:5
init@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-addon/lib/webextension-channels.js:119:5
exports.main@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-addon/index.js:579:3
run@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:153:7
startup/</<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:87:9
Handler.prototype.process@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:937:23
this.PromiseWalker.walkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:816:7
Promise*this.PromiseWalker.scheduleWalkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:747:11
this.PromiseWalker.schedulePromise@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:779:7
this.PromiseWalker.completePromise@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:714:7
handler@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/window.js:56:3
EventListener.handleEvent*EventTargetInterposition.methods.addEventListener@resource://gre/modules/RemoteAddonsParent.jsm:652:5
AddonInterpositionService.prototype.interposeProperty/desc.value@resource://gre/components/multiprocessShims.js:160:52
@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/window.js:54:1
evaluate@resource://gre/modules/commonjs/toolkit/loader.js:279:19
load@resource://gre/modules/commonjs/toolkit/loader.js:331:5
_require@resource://gre/modules/commonjs/toolkit/loader.js:653:16
require@resource://gre/modules/commonjs/toolkit/loader.js:607:12
startup/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:72:21
Handler.prototype.process@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:937:23
this.PromiseWalker.walkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:816:7
Promise*this.PromiseWalker.scheduleWalkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:747:11
this.PromiseWalker.schedulePromise@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:779:7
this.PromiseWalker.completePromise@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:714:7
readURI/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/net/url.js:50:9
NetUtil_asyncFetch/<.onStopRequest@resource://gre/modules/NetUtil.jsm:128:17
```

Probably because I have/had no other installed add-ons when I installed this Test Pilot one?

groovecoder:
Got the same error when I disabled and re-enabled this Test Pilot build while I had my add-on installed. 😞 

I also tried adding a breakpoint at `Metrics.onExperimentPing({` but it was never reached.

My new code is: https://github.com/mozilla/blok/compare/telemetry-ping-6?expand=1

lmorchard:
Ugh, yeah, I have a general bad habit of not trying things in the blank slate state with a fresh profile. Just wrapped that code in a check whether the list of installed experiments is even defined yet, hopefully that improves things.

groovecoder:
I don't get that error anymore, but I get a different error (twice) when loading the Test Pilot add-on now:

```
[Exception... "Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIURI.hostPort]"  nsresult: "0x80004005 (NS_ERROR_FAILURE)"  location: "JS frame :: resource://gre/modules/PopupNotifications.jsm :: PopupNotifications_refreshPanel/< :: line 667"  data: no]
PopupNotifications_refreshPanel/<()
PopupNotifications.jsm:667
forEach()
self-hosted
PopupNotifications_refreshPanel()
PopupNotifications.jsm:616
PopupNotifications_showPanel()
PopupNotifications.jsm:728
PopupNotifications_update()
PopupNotifications.jsm:844
PopupNotifications_show()
PopupNotifications.jsm:394
gXPInstallObserver.showInstallConfirmation()
browser-addons.js:206
gXPInstallObserver.observe/showNotification()
browser-addons.js:383
gXPInstallObserver.observe()
browser-addons.js:399
notifyObservers()
amWebInstallListener.js:51
Installer.prototype.checkAllDownloaded()
amWebInstallListener.js:171
Installer()
amWebInstallListener.js:81
extWebInstallListener.prototype.onWebInstallRequested()
amWebInstallListener.js:335
buildNextInstall()
extensions.js:1293
gViewController.commands.cmd_installFromFile.doCommand/buildNextInstall/<()
extensions.js:1303
safeCall()
AddonManager.jsm:179
AddonManagerInternal.getInstallForFile/<.nextObject/<()
AddonManager.jsm:1955
this.XPIProvider.getInstallForFile/<()
XPIProvider.jsm:3892
makeSafe/<()
XPIProvider.jsm:2001
AddonInstall.prototype.initLocalInstall/</<()
XPIProvider.jsm:5317
makeSafe/<()
XPIProviderUtils.js:168
completeAddon()
XPIProviderUtils.js:157
this.AddonRepository.getCachedAddonByID<()
AddonRepository.jsm:578
next()
self-hosted
TaskImpl_run()
Task.jsm:319
TaskImpl()
Task.jsm:280
createAsyncFunction/asyncFunction()
Task.jsm:254
getRepositoryAddon()
XPIProviderUtils.js:159
this.XPIDatabase.getAddon/<()
XPIProviderUtils.js:1119
Handler.prototype.process()
Promise-backend.js:937
this.PromiseWalker.walkerLoop()
Promise-backend.js:816
bound walkerLoop()
self-hosted
bound bound walkerLoop()
```

I tried adding `console.log` statements ... 

```
   // Rebuild channels for all known experiments
   updateExperimentChannels() {
     channels = {};
+    console.log('store: ' + store);
+    console.log('store.installedAddons: ' + store.installedAddons);
     if (store.installedAddons) {
       Object.keys(store.installedAddons).forEach(id => {
         const channel = new WebExtensionChannel(id);
```

But I don't see this messages in the browser toolbox console when I install the `addon.xpi` When should I expect the `updateExperimentChannels` "method" to be called?

lmorchard:
Hmm, that error is confusing - doesn't mention any Test Pilot code anywhere, and I can't repro on a fresh Firefox profile. But it _does_ look like it happens during the doorhanger to warn of an add-on installation, maybe?

As for `updateExperimentChannels`, it should get called at startup and whenever an experiment is enabled / disabled.

lmorchard:
I tossed those `console.log` statements into my local copy. I see the messages on startup and on experiment enable - both in a fresh profile and in my daily dev browser.

lmorchard:
@groovecoder all that said, your WebExtension code looks correct and a WebExtension shouldn't be able to cause these errors... so there's still probably something weird happening on the Test Pilot side 😞

groovecoder:
I didn't get the error in a fresh profile.

Do I need to do something with my add-on code to make sure it's registered as a testpilot experiment? So I can make sure it's causing `updateExperimentChannels` to fire?

lmorchard:
> Do I need to do something with my add-on code to make sure it's registered as a testpilot experiment?

Oh, right! Yeah, you do: In your local `testpilot.dev` instance - which you _do_ need, now that I'm thinking this through - you have to add your webextension as an experiment. And then, enable it from the local site. The add-on only cycles through known & installed experiments to set up BroadcastChannels, and it fetches that list via the API from the site.

lmorchard:
This is also getting me thinking about coming up with a way to make this process simpler - because you _shouldn't_ have to have a full local instance of the Test Pilot site just to develop a WebExtension. 

Will probably take me some time, but we might be able to come up with just a stub JSON file to feed the Test Pilot add-on in the future.

Edit: https://github.com/mozilla/testpilot/issues/1029

groovecoder:
Are there docs to add & enable my web extension as an experiment in `testpilot.dev` ?

lmorchard:
No docs. You go to /admin/, login as admin with password admin, and add an Experiment record with an add-on id that matches your add-on

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,147 @@
...
+  const baseURI = Services.io
+        .newURI(`moz-extension://${targetExtensionUUID}/_blank.html`, null, null);
+  const principal = Services.scriptSecurityManager
+        .createCodebasePrincipal(baseURI, { addonId });
+
+  const addonChromeWebNav = Services.appShell.createWindowlessBrowser(true);
+  const docShell = addonChromeWebNav.QueryInterface(Ci.nsIInterfaceRequestor) // eslint-disable-line new-cap
+        .getInterface(Ci.nsIDocShell);
+  docShell.createAboutBlankContentViewer(principal);
+  const window = docShell.contentViewer.DOMDocument.defaultView;
```
jaredhirsch, line 56: @lmorchard Is this (lines 28-38) the best way to get a DOM window pointer for the given moz-extension URL? I guess I'm wondering if you found this snippet somewhere, or if it was suggested by someone on the FF team.

I asked about safely getting a DOM Window in `#fx-team` but got no response. I'm wondering if there might be a simpler or safer way--usually Gecko has N ways to do something, but only 1-3 ways that are considered good ways. I'll look further into it.

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,147 @@
...
+
+const store = require('sdk/simple-storage').storage;
+
+const Metrics = require('./metrics');
+
+const TESTPILOT_TELEMETRY_CHANNEL = 'testpilot-telemetry';
+
+Cu.import('resource://gre/modules/XPCOMUtils.jsm');
+
+// Lazy imports.
```
jaredhirsch: This comment is probably not needed

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,147 @@
...
+    } = createChannelForAddonId(TESTPILOT_TELEMETRY_CHANNEL, targetAddonId);
+
+    // NOTE: Keep a ref to prevent it from going away during garbage collection
+    // (or the BroadcastChannel will stop working).
+    this.addonChromeWebNav = addonChromeWebNav;
+    this.addonBroadcastChannel = addonBroadcastChannel;
+  },
+
+  dispose() {
+    this.addonBroadcastChannel.removeEventListener();
```
jaredhirsch: The docs suggest calling `close()` to free the channel before deleting it: https://mdn.io/Broadcast_Channel_API#Disconnecting_a_channel

I tried looking at [the c++ `Close` implementation](https://dxr.mozilla.org/mozilla-central/source/dom/broadcastchannel/BroadcastChannel.cpp#600) to see how reference freeing works, but it's too Gecko-y for me. I'll ask about this as well (unless you already did).

jaredhirsch: Also, did you want to pass arguments to this function, matching the unregisterPingListener function?

`this.addonBroadcastChannel.removeEventListener('message', this);`

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,147 @@
...
+  WebExtensionChannel,
+
+  // Update all the channels on init.
+  init() {
+    this.updateExperimentChannels();
+  },
+
+  // Drop refs to channels for garbage collection
+  destroy() {
+    channels = {};
```
jaredhirsch, line 148: Continuing with the theme of this review, I'm not sure if this will free the channels for GC, how sure are you feeling about this / have you asked anybody on the DOM team about it?

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,147 @@
...
+  },
+
+  // Rebuild channels for all known experiments
+  updateExperimentChannels() {
+    channels = {};
```
jaredhirsch, line 153: Same question as line 124

jaredhirsch:
@lmorchard I'm still trying to get docker sorted, but I read through the code and added some comments. I'll verify that stuff works tomorrow.

addon/lib/webextension-channels.js
lmorchard, line 56: @6a68 This is mostly not my code, it comes from <a href="https://github.com/rpl/testpilot-webextension-experiment/commit/3a2cadeb2f955926887d462419c894df4ebe1c77">experiments with BroadcastChannel over here</a>. Honestly, I didn't really know you could get a DOM window pointer for another origin from add-on code :)

lmorchard:
@6a68 I didn't write most of this stuff dealing with BroadcastChannel, so I mostly wanted to see if it works for our purposes. I don't really have my head totally wrapped around what all it does, so some patches / advice are welcome - because you probably know way more than I do :)

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,147 @@
...
+
+    this.addonBroadcastChannel = null;
+    this.addonChromeWebNav = null;
+  },
+
+  registerPingListener(callback) {
+    this.pingListeners.add(callback);
+
+    if (this.pingListeners.size >= 0) {
+      this.addonBroadcastChannel.addEventListener('message', this);
```
groovecoder: Confused by this ...

When the `WebExtensionChannel` has `pingListener`, the code here adds `this` as the `message` event listener. Does something automatically send/forward/bubble the message event to `this.handleEvent`, which actually calls `notifyPing`?

lmorchard:
Updated to close BroadcastChannel while disposing the WebExtensionChannel

lmorchard:
Could have sworn that this all worked when I tried it locally. But, looking at the event handling, I'm not sure how it did... hmm. Updated with a new commit to add/remove event handler with a bound function. Haven't tried this locally yet because meetings, but pushing my changes to the PR so other folks can look.

addon/lib/webextension-channels.js
lmorchard, line 148: If I understand it right, the `WebExtensionChannel` is also a `Disposable`, so `dispose` gets called on GC. It's not actually a `BroadcastChannel` but it does _contain_ one. I [just added a `.close()` call in `dispose`](https://github.com/mozilla/testpilot/pull/1022/files#diff-8205e50d3a6376cdb23f71fe68c0e2c4R69) - so I think that should cover it?

groovecoder:
YES! 🎉 `this.handleEventBound` makes it work! When I click a disable reason button in my web extension, it sends a message to the broadcast channel, and test pilot calls `Metrics.onExperimentPing` !! Thanks for all the time on this, I think we have it working finally!

addon/lib/webextension-channels.js
jaredhirsch: @groovecoder `handleEvent` is indeed a less-known, but standard, DOM API that I've seen used in Gecko code quite a bit:

https://developer.mozilla.org/en-US/docs/Web/API/EventListener

https://dxr.mozilla.org/mozilla-central/search?q=handleEvent&redirect=false

lmorchard:
I think the last open issue might be what @6a68 raised about creating the DOM window pointer in the WebExtension's origin. If that's more a nit than a gotcha, we can merge this for next push

Good thing about this approach is we should be able to extend it to support @chuckharmston's new variant testing framework in WebExtensions, next.

addon/lib/webextension-channels.js
jaredhirsch, line 56: > Honestly, I didn't really know you could get a DOM window pointer for another origin from add-on code :)

@lmorchard Ah yeah, once you `require('chrome')` you effectively have rootly powers, and a lot of weird stuff becomes possible ^_^

After looking at this snippet for a while, and reading through https://bugzil.la/1186732, I'm much less scared of it. But it might be nice to document it for the sake of our future selves. Here's an attempt, take whatever seems useful, or skip it if it seems to make things more confusing overall:

``` js
function createChannelForAddonId(name, addonId) {
  // The BroadcastChannel API allows messaging between different windows that
  // share the same origin. Bug 1186732 extended this to WebExtensions (which
  // may not have an origin) by adding a special URL that loads an about:blank
  // page at the (generalized) "origin" of the extension.
  //
  // Load that about:blank page, and use its `window` to get a BroadcastChannel
  // that allows two-way communication between the main Test Pilot extension and
  // individual experiment extensions. 

  // Note: the `targetExtensionUUID` is different for each copy of Firefox,
  // so that malicious websites can't guess the special URL associated with
  // an extension. 
  const targetExtensionUUID = getExtensionUUID(addonId);

  // Create the special about:blank URL for the extension.
  const baseURI = Services.io
        .newURI(`moz-extension://${targetExtensionUUID}/_blank.html`, null, null);

  // Create a principal (security context) for the generalized origin given
  // by the extension's special URL and its `addonId`.
  const principal = Services.scriptSecurityManager
        .createCodebasePrincipal(baseURI, { addonId });

  // Create a hidden window and open the special about:blank page for the
  // extension.
  const addonChromeWebNav = Services.appShell.createWindowlessBrowser(true);
  const docShell = addonChromeWebNav.QueryInterface(Ci.nsIInterfaceRequestor) // eslint-disable-line new-cap
        .getInterface(Ci.nsIDocShell);
  docShell.createAboutBlankContentViewer(principal);
  const window = docShell.contentViewer.DOMDocument.defaultView;

  // Finally, get the BroadcastChannel associated with the extension.
  const addonBroadcastChannel = new window.BroadcastChannel(name);

  // Callers need to keep the pointer to the window, otherwise the window's 
  // BroadcastChannel will get garbage collected.
  return {
    addonChromeWebNav,
    addonBroadcastChannel
  };
} 
```

addon/lib/webextension-channels.js
```diff
@@ -0,0 +1,149 @@
...
+      addonChromeWebNav,
+      addonBroadcastChannel
+    } = createChannelForAddonId(TESTPILOT_TELEMETRY_CHANNEL, targetAddonId);
+
+    // NOTE: Keep a ref to prevent it from going away during garbage collection
+    // (or the BroadcastChannel will stop working).
+    this.addonChromeWebNav = addonChromeWebNav;
+    this.addonBroadcastChannel = addonBroadcastChannel;
+
+    this.handleEventBound = ev => this.handleEvent(ev);
```
jaredhirsch, line 87: This seems weird to me. I'll look into it this evening.

addon/lib/webextension-channels.js
lmorchard, line 56: Think I'll just plonk that on in - verbose commenting of this magic is a great addition, thanks!

addon/lib/webextension-channels.js
lmorchard, line 87: I could probably just do `this.handleEvent.bind(this)`, but thought this looked clearer. I think the add/remove event methods need a consistent bound function ref to deal with. Though I _did_ think `handleEvent` and passing `this` was enough.

addon/lib/webextension-channels.js
jaredhirsch, line 148: Oh, nice! I skimmed right past the mention of `Disposable`, which I wasn't familiar with. :+1:

addon/lib/webextension-channels.js
jaredhirsch, line 87: I agree, there's probably something fishy here: the BroadcastChannel implementation of add/removeEventListener just wraps the default implementation. That said, it does work and it's readable, so I think we're fine to just leave it and move on.

jaredhirsch:
@lmorchard I think this is good to land.

I ran out of time doing server setup and wasn't able to get the addons communicating, but it sounds like you and @groovecoder both saw it working, so it's probably fine. The channel creation snippet seems sane, and the rest of the code LGTM. :beers:

groovecoder:
giphy

lmorchard:
Seems like we've beaten this enough to merge it, so I'll go ahead & do that... now

Issue #1006, opened 2016-06-21
Heartbeat interval responses are still low


MattGrimes:
I just checked again today. We are still seeing very low numbers for the day 7 interval and almost nothing for later intervals. According to Marina this isn't due to high attrition before the 7 day mark. I suspect something else.

johngruen:
@meandavejustice this is our bread and butter. could you take a look at this ASAP. Should be prioritized higher than any work on min-vid

meandavejustice:
@johngruen I'll take a look today

meandavejustice:
@MattGrimes This should have been fixed in https://github.com/mozilla/testpilot/commit/041aa535ca5be8ca9952f11f7129bb877a3e21a7 I'm looking into when it was actually deployed right now.

UPDATE: It was included in this release https://github.com/mozilla/testpilot/releases/tag/2016-06-06 I'm having trouble finding proof of when it was actually deployed.. We should fix that :/ ~~I'm guessing there is a bugzilla bug somewhere~~ https://bugzilla.mozilla.org/show_bug.cgi?id=1279287

johngruen:
@ckprice should know when it was deployed...I think he's back in sunny California?

pdehaan:
> I think he's back in sunny California?

FWIW, I haven't seen him all week :(
I think he went to Paris after the work week, but am not 💯 on when he's back.

---

https://testpilot.firefox.com/__version__ shows what SHA is currently deployed:

``` json
{ "commit": "ac480a9bea2be7fb0c7e866a4274cdd9945c719a",
  "build": "https://circleci.com/gh/mozilla/testpilot/506",
  "version": "2016-06-09",
  "source": "https://github.com/mozilla/testpilot" }
```

So the current version was deployed on [June 9th](https://github.com/mozilla/testpilot/releases/tag/2016-06-09) (tagged by @wresuolc), and does include the https://github.com/mozilla/testpilot/commit/041aa535ca5be8ca9952f11f7129bb877a3e21a7 commit in question https://github.com/mozilla/testpilot/commits/2016-06-09

pdehaan:
Any updates on this?

jaredhirsch:
@pdehaan Dave just fixed one bug (#1023), I'm helping him figure out if there are still others lurking.

johngruen:
#1024

meandavejustice:
[...]

PR #1023, opened 2016-06-28
Don't remove surveyChecks unless addon is uninstalled

meandavejustice:
- refs #1006
- only call `survey.destroy()` if the addon is being `uninstalled`
  we have been calling on `unload`.


lmorchard:
LGTM

PR #1025, opened 2016-06-30
Adds Discourse link to experiment details page

kaganjd:
- Fixes #1019
- updates frontend view, template, and test for experiment page
- updates experiment model, serializer, and test
- adds migration for model update
- adds discourse key to app.l20n
- updates `fixtures/demo_data.json` with `discourse_url`

PR #1030, opened 2016-07-01
Changes to upgrade l20n.js

### LINKED ISSUE #497: Upgrade to official l20n.js release
[I left this TODO back when we first incorporated l20n.js](https://github.com/mozilla/testpilot/blob/334d776393425df52564d9f1cad99c2b882ed736/testpilot/frontend/static-src/app/main.js#L6):

> // TODO: Switch back to L20N official release when Bug 1240192 is resolved
> // https://bugzilla.mozilla.org/show_bug.cgi?id=1240192

It looks like bug 1240192 is resolved now, so let's get rid of our patched build and switch to the official release.

lmorchard:
- Convert .l20n format en-US locale to .ftl
- Add babel-polyfills and l20n 4.0.0-alpha.3 to package.json
- Include the correct modules in vendor bundle
- Introduce the new modules as dependencies in the app bundle
- Exit gulp if gulpfile.js changes - should restart in frontend watcher

Fixes #497


lmorchard:
This doesn't _quite_ close #497, because this doesn't use an official release. Instead it uses today's current master [per advice from :stas](https://bugzilla.mozilla.org/show_bug.cgi?id=1280410#c13). It does get us up to date with the `.ftl` string format, though - so hopefully an official release will be a drop-in upgrade.

lmorchard:
I _am_ a little unhappy about adding another `<script>` tag, but whenever I'd try adding `l20n.js` to the app or vendor bundle, I'd get nothing but errors in the console. I think something in the build process doesn't agree with the already-preprocessed JS.

jotes:
@lmorchard thx that you've taken this issue -> I was totally lost in the time during last few weeks.

lmorchard:
Also thanks for looking into it!

locales/en-US/app.l20n
```diff
@@ -1,122 +0,0 @@
...
-
-<measurements "Your privacy">
-<experimentPrivacyNotice "You can learn more about the data collection for {{title}} here.">
-<contributorsHeading "Brought to you by">
-<version "Version">
-<changelog "changelog">
-<lastUpdate "Last Update">
-<contribute "Contribute">
-<bugReports "Bug Reports">
-<discourse "Discourse">
```
pdehaan, line 106: I think we lost this string which was added recently in #1025

lmorchard, line 106: Good catch!

lmorchard:
Tweaked a few more strings that I hadn't quite converted correctly.

stasm:
@lmorchard, I installed testpilot locally but I don't know how to try to add l20n to the app or the vendor bundle.  What were the errors that you saw?  Perhaps related to minification?

Also, v4.0.0-a2 is now on npm.

lmorchard:
> I don't know how to try to add l20n to the app or the vendor bundle.

The previous version got added to the app bundle with [this `import` line](https://github.com/mozilla/testpilot/pull/1030/files#diff-0830938c0d34346320870a2fe4743068L8) removed in this PR. That was the built release, not the source modules.

> What were the errors that you saw? Perhaps related to minification?

The main error I get is: `ReferenceError: regeneratorRuntime is not defined` around this line from l20n:

```
var mapValues = regeneratorRuntime.mark(function mapValues(arr) {
```

Not great context, I know, and I need to dig into this some more to see what's up. Also going to try installing the new npm release and see if I can roll building it into our overall build process.

stasm:
I know this error! :)  It looks like adding the script to the bundle also automatically babel-transpiles it?  L20n uses generators (`yield` etc.) and Babel transpiles them to state machines using http://facebook.github.io/regenerator/.  This requires a small runtime library which exposes the `regeneratorRuntime` entry point.

You can add the runtime with:

```
$ npm install --save regenerator-runtime

require('regenerator-runtime/runtime');
```

Or, if you're using Babel anyways, you could use their polyfills which also adds support for things like `Map`, `Promise` etc.:

```
$ npm install --save babel-polyfill

require('babel-polyfill');
```

lmorchard:
Tried adding either / both of those modules along with `require`s in the main app bundle - and neither fixed the error. It _does_ look like `regenerator-runtime` is in the vendor bundle when I grep for it, though.

The weird thing, though, is that I'm trying to bundle in `l20n/dist/compat/web/l20n` which works when I include it directly on the page with a `<script>` tag. So, I wonder what I'm missing?

lmorchard:
Either way, I'm not too sad to bump that to a future PR, because l20n _does_ work as a standalone library included in the page. So, it would be nice to get the PR merged with that caveat, as long as all the localizations work

stasm:
I'm a bit puzzled right now because `dist/compat/web/l20n` is transpiled by Babel and uses `regeneratorRuntime` as well: https://github.com/l20n/l20n.js/blob/58644b2fff88fc0fb43c802a77be5b9ffc2df01f/dist/compat/web/l20n.js#L90

Perhaps use `dist/bundle/web/l20n` (es6) for now and I'll try to help with the compat version in a new PR?

lmorchard:
I can't get either bundle or compat to work in our build process. Going to just go with including it as a standalone script and work further on it in another PR. Mainly looking to get the thumbs up on whether the PR works as-is

stasm:
Isn't it weird that it works?  Does testpilot already include babel-polyfill or regenerator-runtime?  My suggestion was to use `dist/bundle/web/l20n` as a standalone lib for now and figure out how to use the compat version (standalone or baked into one of your static bundles) in a new PR.

lmorchard:
> Isn't it weird that it works? Does testpilot already include babel-polyfill or regenerator-runtime? My suggestion was to use dist/bundle/web/l20n as a standalone lib for now

I think we're getting turned around... [a copy of `dist/bundle/web/l20n.js` _is_ what this PR uses right now as `testpilot/frontend/static-src/vendor/l20n.js`](https://github.com/mozilla/testpilot/pull/1030/commits/d82927c37a344f36dbcace8aba6f3ba11742400d) and includes it as a standalone script. That seems to work - should it not?

Hmm, actually maybe it shouldn't, since it looks like it's not the transpiled version.

stasm:
Ah, great, thanks for clearing that up.  I misunderstood :)

lmorchard:
Ooooookay, I think I just untangled some things. Just using `require('babel-polyfill')` didn't fix things for me. But, doing _this_ appears to have fixed things:

```
import 'babel-polyfill/browser';
import 'l20n/dist/compat/web/l20n';
```

That gets the correct polyfill module into the build and uses the transpiled version of l20n, because I think Browserify only runs Babel transforms on local code - and not on dependencies in `node_modules`.

So, [I just added a commit](https://github.com/mozilla/testpilot/pull/1030/commits/56b0cc93ee92e459ce9022be149a0849c0552c67) to the PR that moves l20n.js into our build process from the npm alpha release and stops using my own local build. Think I might squash some things and update the PR again

locales/en-US/app.ftl
```diff
@@ -0,0 +1,125 @@
...
+bugReports = Bug Reports
+discourse = Discourse
+nowActive = Active
+userCount =
+    [html/title] Active Users of this Experiment
+tourOnboardingTitle = {$title} enabled!
+tourStartButton = Take the Tour
+tourCancelButton = Skip
+tourDoneButton = Done
+userCountContainer = There are <span class="bold">{$installation_count}</span> people trying {$title} right now!
```
stasm: The `class` attribute will be filtered out by l20n.js so it can be removed here. It should be defined on the source `<span>` in the HTML—l20n.js will reuse that original `<span>` and the `class` will be preserved.

stasm:
That's great to hear, @lmorchard! The FTL part looks good to me, thanks!

lmorchard:
@chuck @meandavejustice - Either of you folks want to give this a spin for an r+?

stasm:
@lmorchard I published l20n.js 4.0.0-alpha.3 today which is a small update with better compatibility with Safari and Safari for iOS.

lmorchard:
Updated to 4.0.0-alpha.3, things still look good to me

gulpfile.js
```diff
@@ -26,7 +26,20 @@ const md = new Remarkable({
...
+  'l20n'
+];
+const includeVendorModules = [
+  'babel-polyfill/browser',
+  'l20n/dist/compat/web/l20n'
+];
+const vendorModules = Object.keys(packageJSON.dependencies)
+  .filter(name => {
+    if (excludedVendorModules.indexOf(name) !== -1) { return false; }
+    return true;
```
jaredhirsch: nit: you could just `return excludedVendorModules.indexOf(name) > -1`

jaredhirsch:
@lmorchard R+, works for me locally. I edited some of the strings, saw gulp pick up the changes, and saw the changes reflected in the page after a refresh. I managed to find one nit, happy to merge once it's addressed, or feel free to merge yourself.

gulpfile.js
meandavejustice: could go even further and put `excludedVendorModules.indexOf(name) > -1` on the same line as the arrow function and drop `return`

meandavejustice:
~~@lmorchard So l20n is distributed with generators instead of browser ready code?~~

I see, this seems reasonable https://github.com/l20n/l20n.js/blob/master/docs/compat.md

:+1: from me.

lmorchard:
> So l20n is distributed with generators instead of browser ready code? Sounds like we should address that upstream instead of processsing all of our code through babel

FWIW, we _already_ process all our code through Babel by way of Browserify - that's how we get to use ES6 features in our source. Adding babel-polyfill is actually something we probably should have done a long time ago. The l20n code is from the `l20n/dist/compat/web/l20n` module, already passed through Babel in their own build process.

gulpfile.js
lmorchard: Looks like `excludedVendorModules.indexOf(name) < 0` is actually what I need here, but that does work as a one-liner. I think I had it more complicated thinking I'd add more conditions, but YAGNI

lmorchard:
Updated to simplify the `excludeVendorModules` logic, going to let CircleCI pass and then self-merge

gulpfile.js
pdehaan: FWIW, there may also be an `.includes()` API, depending on our Node+Babel env:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

> _The `includes()` method determines whether an array includes a certain element, returning `true` or `false` as appropriate._

PR #1033, opened 2016-07-07
Update DEPLOYMENT.md

ckprice:
We have a new QA team! Adding their steps to docs.

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -1,5 +1,17 @@
 This document details the process we use to deploy Test Pilot to our stage and production environments.
 
+## Overview of schedule ##
+
+(you can read more below)
+
+- Friday at 2-4pm PST we tag and push stage, and send email out.
+- Monday at regular stand-up we review anything that Softvision found and fix it.
+- Tuesday at 8am PST we push and Softvision verifies.
+
+## Softvision ##
+
+Softvision is our embedded QA team. Their main functions are to write test plans and verify deployments.
+
 ## Team Notification ##
 
 During the checkin before the end of the [current milestone](https://github.com/mozilla/testpilot/milestones), an email will be sent to our distribution list (`testpilot-dev@mozilla.com`) informing the team that we will be building a release against `master`.
@@ -40,7 +52,7 @@ Ops will inform us once we are deployed to our *staging environment*: [http://te
 
 This will happen on Friday at the end of sprint.
 
-Send out an email notification to `testpilot-dev@mozilla.com` to please test the staging environment.
+Send out an email notification to `testpilot-dev@mozilla.com` to please test the staging environment. [Example](https://mail.mozilla.org/pipermail/testpilot-dev/2016-July/000198.html).
 
 Include Softvision and the bug link in the email notification.
 
@@ -60,7 +72,17 @@ On the following Monday, during our checkin, Softvision will give us an update o
 
 Once we are comfortable that the site has been tested, update the bug with approval to move to our *production environment*: [http://testpilot.firefox.com](http://testpilot.firefox.com).
 
+We'll target Tuesday 8AM PST for deployment.
+
 ## Verify Production ##
 
-Softvision will verify production for us, and report any bugs.
+Softvision will verify production for us, and report any bugs on Tuesday.
+
+## Rolling Back/Forward Process ##
+
+If something is found that requires an immediate change to production (or staging) code, please follow the below guidelines.
+
+In all cases, engineering will have the final approval on the roll back.
+
+Ideally, we should "roll forward" with `git revert` to undo problematic commits and push a new release. Of course, that's also problematic if we don't know what commits caused a problem. Or if the problematic commits also included the database migrations - in which case, we also need to pair the `git revert` with a commit that undoes the migration.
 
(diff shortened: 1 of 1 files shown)


ckprice:
@clouserw - need some clarification on the `Verify Production` section.

docs/DEPLOYMENT.md
```diff
@@ -8,34 +8,59 @@ Note: we auto deploy the master branch to our *development environment*: [http:/
 
 ## Tag Release ##
 
-On the Monday after the milestone closes, Tag the release.
+This will happen on Friday at the end of sprint.
```
jaredhirsch, line 23: 'This' is a bit vague. Maybe 'Releases should normally happen on Friday...' ?

docs/DEPLOYMENT.md
```diff
@@ -8,34 +8,59 @@ Note: we auto deploy the master branch to our *development environment*: [http:/
...
 
 1. https://github.com/mozilla/testpilot/releases/new
 2. Tag Version: YYYY-MM-DD (append -N if more than one release is tagged on a given day: 2016-04-08-1)
 3. Release Title: YYYY-MM-DD
 4. Click `Publish`
 5. Wait for CircleCI to confirm that the release has passed ([tags uploaded to container](https://hub.docker.com/r/mozilla/testpilot/tags/)).
 
+Please be as detailed as possible in the release notes.
```
jaredhirsch: It's probably good to either point to an example or list the details that should be included. The more explicit the docs are, the easier it'll be for people to take vacations and hand off chores ^_^

docs/DEPLOYMENT.md
jaredhirsch, line 23: Never mind, I see that sentence has been sprinkled throughout, it's fine

docs/DEPLOYMENT.md
```diff
@@ -8,34 +8,59 @@ Note: we auto deploy the master branch to our *development environment*: [http:/
...
 
+This will happen on Friday at the end of sprint.
+
 Once we've confirmed our release passed CircleCI, and we've confirmed that it is uploaded into the container, we need to request for Ops to deploy to stage.
 
 This is done via a deployment bug.
 
 Example: https://bugzilla.mozilla.org/show_bug.cgi?id=1263682
 
+Include Softvision in the CC:
```
jaredhirsch, line 43: What bugzilla accounts specifically should be included? The example doesn't CC anybody outside the dev team.

docs/DEPLOYMENT.md
```diff
@@ -8,34 +8,59 @@ Note: we auto deploy the master branch to our *development environment*: [http:/
...
 
+This will happen on Friday at the end of sprint.
+
 Ops will inform us once we are deployed to our *staging environment*: [http://testpilot.stage.mozaws.net](http://testpilot.stage.mozaws.net).
 
 ## Test Stage ##
 
+This will happen on Friday at the end of sprint.
+
 Send out an email notification to `testpilot-dev@mozilla.com` to please test the staging environment.
```
jaredhirsch: Is there a good example email that we could link to? Archives are here: https://mail.mozilla.org/pipermail/testpilot-dev/

docs/DEPLOYMENT.md
```diff
@@ -8,34 +8,59 @@ Note: we auto deploy the master branch to our *development environment*: [http:/
...
 
-We have a [list of steps for basic functionality](https://docs.google.com/document/d/1cYelgp7dPwHD1eQIs4POgFqF2NOzxmVpKP4rnPR-RDA/edit), but new features should also be tested.
+Include Softvision and the bug link in the email notification.
```
jaredhirsch, line 57: Again, I don't know what 'include softvision' means. Are there specific email addresses?

docs/DEPLOYMENT.md
```diff
@@ -8,34 +8,59 @@ Note: we auto deploy the master branch to our *development environment*: [http:/
...
 ## Approve Production ##
 
-Once we are comfortable that the site has been tested, update the bug with approval to move to our *production environment*: [http://testpilot.firefox.com](http://testpilot.firefox.com).
\ No newline at end of file
+Once we are comfortable that the site has been tested, update the bug with approval to move to our *production environment*: [http://testpilot.firefox.com](http://testpilot.firefox.com).
+
+## Verify Production ##
+
+Softvision will verify production for us, and report any bugs.
+
```
jaredhirsch, line 88: This might be a good opportunity to document our rollback process?

jaredhirsch:
@ckprice Added a few comments. Thanks for keeping the docs maintained :beers:

docs/DEPLOYMENT.md
lmorchard, line 88: Yeah, we don't really have one & we should come up with one. We've done a simple release rollback at least once and got lucky. But we shouldn't reflexively reach for that option as a panic button when things break.

Rolling back to the previous release is dangerous / impossible when we have database migrations - meaning the site is likely to break on a rollback and require careful manual ops intervention to repair. In other words, compound the problem and require hours to get back to working order.

Ideally, we should "roll forward" with `git revert` to undo problematic commits and push a new release. Of course, that's also problematic if we don't know what commits caused a problem. Or if the problematic commits also included the database migrations - in which case, we also need to pair the `git revert` with a commit that undoes the migration.

docs/DEPLOYMENT.md
ckprice, line 43: For now, the list is Paul and Cosmin. I think we'll just keep it as Softvision team as we may not have the same team all time?

/edit: added a Softvision section at the top.

ckprice:
Any other comments? I'll probably just merge this in a bit.

PR #1051, opened 2016-07-08
Make "enabled" indicators more clear

### LINKED ISSUE #968: Make 'Active' / Experiment started indicator more clear.
User testing shows that users want to know when experiments start collecting data. We should do a better job indicating when an experiment has started as well as the correlation btwn experiment activity and metrics collection

### LINKED ISSUE #1067: Remove button-related code from experiment-row-view.js
#1051 replaces "Manage" and "Get Started" buttons with a participant count. But there's still lingering button-related code [here](https://github.com/mozilla/testpilot/blob/67ad38a84bd36be94de501e9d6b6af2d5ea70cb2/testpilot/frontend/static-src/app/views/experiment-row-view.js#L72) that needs to be updated or deleted.

kaganjd:
-fixes #968 
-fixes #1067 
-changes "active" indicator circle to "enabled" indicator bar on experiment detail page
-adds "enabled" tab to experiment list view
-replaces "manage" and "get started" buttons with participant count
-adds localization for enabled tab and enabled indicator bar
-creates color variables for the greens in the design spec

### DIFF
--- a/testpilot/frontend/static-src/app/templates/experiment-page.js
+++ b/testpilot/frontend/static-src/app/templates/experiment-page.js
@@ -5,11 +5,10 @@ export default `
       <header data-hook="header-view"></header>
     </div>
     <div class="default-background">
-      <div class="spacer"></div>
       <div class="details-header-wrapper">
+        <div class="status-bar is-enabled" data-hook="is-enabled" data-l10n-id="isEnabled"><span data-hook="title"></span> is enabled.</div>
         <div class="details-header content-wrapper">
           <h1 data-hook="title"></h1>
-          <span class="now-active" data-hook="now-active" data-l10n-id="nowActive">Active</span>
           <div class="experiment-controls">
             <a data-hook="highlight-privacy" class="highlight-privacy" data-l10n-id=highlightPrivacy>Your privacy</a>
             <a data-l10n-id="giveFeedback" data-hook="feedback" id="feedback-button" class="button default" target="_blank">Give Feedback</a>
--- a/testpilot/frontend/static-src/app/views/experiment-row-view.js
+++ b/testpilot/frontend/static-src/app/views/experiment-row-view.js
@@ -4,6 +4,9 @@ import BaseView from './base-view';
 
 export default BaseView.extend({
   template: `<div data-hook="show-detail" class="experiment-summary">
+                <div class="experiment-actions">
+                  <div data-l10n-id="experimentListEnabledTab" data-hook="enabled-tab" class="tab show-when-enabled"></div>
+                </div>
               <div class="experiment-icon-wrapper" data-hook="bg">
                 <div class="experiment-icon" data-hook="thumbnail"></div>
               </div>
@@ -12,11 +15,8 @@ export default BaseView.extend({
                   <h3 data-hook="title"></h3>
                 </header>
                 <p data-hook="description"></p>
+                <span class="participant-count" data-l10n-id="participantCount"</span>
               </div>
-               <div class="experiment-actions">
-                  <button data-l10n-id="experimentListInactiveHover" class="button default show-when-inactive">Get Started</button>
-                  <button data-l10n-id="experimentListActiveHover" class="button secondary show-when-active">Manage</button>
-               </div>
              </div>`,
 
   props: {
@@ -49,11 +49,15 @@ export default BaseView.extend({
       },
       hook: 'thumbnail'
     },
-    'model.enabled': {
+    'model.enabled': [{
       type: 'booleanClass',
       hook: 'show-detail',
-      name: 'active'
+      name: 'enabled'
     },
+    {
+      type: 'toggle',
+      hook: 'enabled-tab'
+    }],
     'loggedIn': {
       type: 'booleanClass',
       hook: 'show-detail',
@@ -71,16 +75,9 @@ export default BaseView.extend({
 
   openDetailPage(evt) {
     evt.preventDefault();
-    if (this.model.enabled) {
       app.sendToGA('event', {
         eventCategory: 'ExperimentsPage Interactions',
-        eventAction: 'Manage experiment button',
-        eventLabel: this.model.title
-      });
-    } else {
-      app.sendToGA('event', {
-        eventCategory: 'ExperimentsPage Interactions',
-        eventAction: 'Get Started experiment button',
+        eventAction: 'Open detail page',
         eventLabel: this.model.title
       });
     }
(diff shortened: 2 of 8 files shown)


jaredhirsch:
@johngruen Hey, could you take a look at this PR?

meandavejustice:
changes to `locales/en-US/app.l20n` should actually be in `locales/en-US/app.ftl` since the `.l20n` file was recently removed.

johngruen:
hey, got pulled off on other tangents today. will look in the morning.

johngruen:
@kaganjd looks good, i think you need to add these lines in '`locales/en-us/app.ftl`

```
experimentListEnabled = Enabled
isEnabled = {$title} is enabled.
```

johngruen:
@kaganjd err, it looks like the whole l20n file has been removed and replaced with the .ftl file, so maybe back that thing out entirely.

testpilot/frontend/static-src/styles/modules/_experiment-details.scss
```diff
@@ -1,5 +1,18 @@
-.details-header-wrapper {
+.is-enabled {
```
johngruen: let's stick with the flex-box convention we're doing elsewhere here...

nix the `line-height` and `text-align` rules
and add the following rules:

```
@include flex-container(row, center, center);
height: $grid-unit * 1.8;
padding-top: 1px;
```

testpilot/frontend/static-src/app/views/experiment-row-view.js
```diff
@@ -13,10 +16,6 @@ export default BaseView.extend({
                 </header>
                 <p data-hook="description"></p>
               </div>
-               <div class="experiment-actions">
```
johngruen, line 16: Hey, let's keep these in for now, actually, i don't want to back them out until we've replaced them with user counts

testpilot/frontend/static-src/styles/modules/_experiment-details.scss
kaganjd: For some reason, when I use those rules, my text aligns the text to the left and top. Is that what you get?

Code:

```
.is-enabled {
  @include flex-container(row, center, center);
  background: $green;
  border-bottom: 1px solid $dark-green;
  color: $white;
  display: block;
  font-weight: 300;
  height: $grid-unit * 1.8;
  left: 0;
  padding-top: 1px;
  position: absolute;
  top: 0;
  width: 100%;
}
```

screen shot 2016-07-12 at 5 50 41 pm

johngruen: huh, whelp, what you had works, so go with it!

johngruen:
Hey @kaganjd just noticed that the scroll behavior is a little wonky on the experiment detail page 

screen shot 2016-07-13 at 12 18 11 pm

When an experiment isn't installed, there's way too much white space on the top here. Looking at your code, you might want to bind is `is-enabled` to the `.details-header-wrapper` then use CSS to show and hide the `.is-enabled` span based on a class change. I might also lose the `.spacer` div there and just add padding-top equivalent to the height of the `.is-enabled` span when the span is not visible and the page is not scrolled.

Oh, also, 
I don't believe the span needs to be absolutely positioned. Might make it easier.

testpilot/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -5,11 +5,11 @@ export default `
       <header data-hook="header-view"></header>
     </div>
     <div class="default-background">
-      <div class="spacer"></div>
       <div class="details-header-wrapper">
+      <span class="is-enabled" data-hook="is-enabled" data-l10n-id="isEnabled"><span data-hook="title"></span> is enabled.</span>
+      <div class="spacer"></div>
```
johngruen: nit: these lines should be tabbed over one.

jaredhirsch:
Thanks for taking a look, @johngruen.

@kaganjd I think you can simplify things on the experiment detail page by making `.is-enabled` a div, not a span, and removing absolute positioning. If it's a div and not absolutely positioned, then it should take up the right amount of vertical space when it's shown, and when it's hidden, there won't be any empty space to worry about.

So, summarizing, I think this will work:
- remove spacer div
- change .is-enabled from a span to a div
- remove position: absolute, left: 0, top: 0 from .is-enabled, so that when it's shown, it pushes the rest of the header downward
- remove display: block from .is-enabled, since display: block is the default display style for divs
- remove width: 100% from .is-enabled, since block elements fill width by default

Ping me on IRC if you run into any surprises :-)

_Edited to clarify that I'm referring to jg's feedback on the experiment detail page_

johngruen:
@kaganjd @6a68: There needs to be more space top of the title when the experiment isn't active && the page is not scrolled. Sorry to be a pain about this. 

Because we're showing and hiding the notification element you need something to make up for the height lost when it isn't there. I think I have a strategy for this if you need a bit of an assist. Ping me in the morning and we can hack on it together for a bit.

jaredhirsch:
@johngruen No worries, let's get it right :-)

kaganjd:
@johngruen will do. Just so we have it tomorrow, here are dimensions of the spacer div I got rid of:

```
.spacer {
  height: $grid-unit * 2.5;
}
```

meandavejustice:
looks like there is a linting issue

trailing comma on line 60 in `testpilot/frontend/static-src/app/views/experiment-page.js`

testpilot/frontend/static-src/styles/modules/_experiment-list.scss
```diff
@@ -90,19 +84,38 @@
...
+      line-height: $grid-unit * 1.5;
+      outline: none;
+      padding: 0 20px;
+      position: absolute;
+      text-align: center;
+      text-decoration: none;
+      top: -3px;
+      transition: background 150ms;
+      white-space: nowrap;
+      z-index: 99999;
```
meandavejustice, line 118: woah

jaredhirsch:
@johngruen You're taking this bug, is that right? Do you want to close and reopen the PR?

testpilot/frontend/static-src/app/views/experiment-row-view.js
```diff
@@ -71,16 +75,9 @@ export default BaseView.extend({
...
-        eventLabel: this.model.title
-      });
-    } else {
       app.sendToGA('event', {
         eventCategory: 'ExperimentsPage Interactions',
-        eventAction: 'Get Started experiment button',
+        eventAction: 'Open detail page',
         eventLabel: this.model.title
       });
     }
```
meandavejustice, line 83: this also needs to be removed

kaganjd:
I removed

johngruen:
Re-PRing

testpilot/frontend/static-src/app/views/experiment-row-view.js
johngruen, line 83: @meandavejustice so neither of these should be in there?

PR #1054, opened 2016-07-09
Remove user dependency from installation tracking API

### LINKED ISSUE #1040: Remove user account dependency in tracking experiment installations
The method we use to track experiment installations on the server (mainly for counting users) has some FxA dependencies that need changing:
- [x] [UserInstallation model](https://github.com/mozilla/testpilot/blob/master/testpilot/experiments/models.py#L112) includes a pointer to user account along with browser client ID. 
- [x] [The API](https://github.com/mozilla/testpilot/blob/master/testpilot/experiments/views.py#L47) for managing installation status requires authentication
- [x] [The add-on](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L427) uses authentication cookies as part of the request to that API.
- [x] [The `installation_count` property of the Experiment model](https://github.com/mozilla/testpilot/blob/master/testpilot/experiments/models.py#L64) counts distinct users in installation records - this should just be a straight installation count after the user column is removed.

lmorchard:
- Migration to drop user field from UserInstallation model, users
  relation in Experiment model.
- Remove `distinct('user')` filter from `installation_count` property in
  Experiment model.
- Drop auth requirement from installation tracking API resources.
- Remove unused `experiment-installation-list` API resource.
- Remove list of installed experiments from `/api/me` resource.
- Stop doing a full sync of experiments with the server, list of enabled
  experiments from the add-on is authoritative
- Separate call to fetch installed experiments and request to fetch
  `/api/me` resource in frontend Me model.
- Remove deletion of user installations from user retire API resource.
- Bump add-on to v0.8.0
- Updated tests.

Fixes #1040

### DIFF
--- a/testpilot/frontend/static-src/app/models/me.js
+++ b/testpilot/frontend/static-src/app/models/me.js
@@ -32,40 +32,35 @@ export default Model.extend({
   },
 
   fetch() {
+    return Promise.all([
+      this.fetchInstalledExperiments(),
+      this.fetchMeResource()
+    ]);
+  },
+
+  fetchInstalledExperiments() {
+    this.hasAddon = Boolean(window.navigator.testpilotAddon);
+    if (!this.hasAddon) { return false; }
+
+    // HACK: Previous versions of the add-on expected installation data from
+    // the server, but that's no longer available. So, let's fake it, with an
+    // empty data structure as appropriate to the addon version (or lack of one
+    // exposed)
+    const installedData = window.navigator.testpilotAddonVersion ? {} : [];
+
+    return app.waitForMessage('sync-installed', installedData).then(result => {
+      this.clientUUID = result.clientUUID;
+      this.installed = result.installed;
+    });
+  },
+
+  fetchMeResource() {
     return fetch(this.url, {
       headers: { 'Accept': 'application/json' },
       credentials: 'same-origin'
     }).then(response => response.json()).then(userData => {
       this.user = userData;
       if (!this.user.profile) { return false; }
-
-      this.hasAddon = Boolean(window.navigator.testpilotAddon);
-      if (!this.hasAddon) { return false; }
-
-      let installedData;
-      if (!window.navigator.testpilotAddonVersion) {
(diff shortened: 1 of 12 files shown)


lmorchard:
Updated commit with a rebase against master

chuckharmston:
@lmorchard Looks like there's a conflict, unfortunately.

lmorchard:
Yup, looks like a test that 2 of my PRs touched. Just rebased from master and resolved conflicts

testpilot/frontend/static-src/app/models/me.js
```diff
@@ -32,40 +32,35 @@ export default Model.extend({
...
+  },
+
+  fetchInstalledExperiments() {
+    this.hasAddon = Boolean(window.navigator.testpilotAddon);
+    if (!this.hasAddon) { return false; }
+
+    // HACK: Previous versions of the add-on expected installation data from
+    // the server, but that's no longer available. So, let's fake it, with an
+    // empty data structure as appropriate to the addon version (or lack of one
+    // exposed)
```
chuckharmston, line 48: How will the add-on handle this?

testpilot/users/views.py
```diff
@@ -32,13 +30,6 @@ def list(self, request):
                 "name": "Test Pilot",
                 "url": settings.ADDON_URL
             },
```
chuckharmston, line 32: Unnecessary trailing comma here, if you care about that. `¯\_(ツ)_/¯`

chuckharmston:
This is great. Shipping when tests pass.

testpilot/frontend/static-src/app/models/me.js
lmorchard, line 48: It'll look to the older add-ons like the server thinks no experiments are installed. So, on a page load of the Test Pilot site, the add-on will send a PUT to the server for each installed experiment as part of the "sync" process. A few extra HTTP requests until those add-ons get updated. But, if the server already has UserInstallation records for those experiments, the requests will be idempotent no-ops.

PR #1053, opened 2016-07-09
Remove account deletion step of "Uninstall" flow

lmorchard:
- Remove frontend request to delete user during uninstall
- Remove `/users/retire/` API resource
- Updated tests

Fixes #1045

PR #1064, opened 2016-07-12
WIP no longer unsubscribe users from email lists when they unistall

johngruen:
@lmorchard Here's how i would handle the UI changes.


lmorchard:
The unsubscribe call [was removed in another PR](https://github.com/mozilla/testpilot/pull/1053/files#diff-80eb70da53930bdf6ead5c585c81c484L52), so I'll just go ahead and merge this as the UI side of it

Issue #1046, opened 2016-07-07
Ensure content for "Contributors" list on experiment details page is retained after user account deletion


lmorchard:
The "contributors" list for each experiment is made up of a list of user accounts.

This should probably be reworked to be a simple list of name, title, and avatar totally separate from user accounts.

ckprice:
I wonder if it makes sense to remove this from TXP-7, and add it to our flat files milestone when it's created.

There may be a more interesting way to manage this list once we figure out the files architecture, and I think we can continue to edit the contributors list in the DB after we remove Firefox Accounts.

lmorchard:
Problem is that the data behind this column possibly goes away when TXP-7 is finished. The contributors list is made of users. That said, we could possibly retain those specific users - since those are all Mozillians and possibly all admins.

pdehaan:
If we move to flat files (backed by Markdown+JSON/Yaml) we could probably just use front matter.
For example:
- https://gohugo.io/content/front-matter/
- https://jekyllrb.com/docs/frontmatter/

So in the case of an experiment, the front-matter could just specify a sorted array of contributors and have no dependencies on a database or anything else.

---

```
---
title: Tab Center
version: "1.21"
slug: tab-center
changelog_url: https://github.com/bwinton/VerticalTabs/releases/latest
contribute_url: https://github.com/bwinton/VerticalTabs#readme
bug_report_url: https://github.com/bwinton/VerticalTabs/issues
discourse_url: https://discourse.mozilla-community.org/c/test-pilot/tab-center
description: "What would it be like to move tabs from the top of the browser to the side? We wanted to find out!"
tags: [ "Just Updated" ]
lastmod: 2016-07-05
date: "2012-04-06"
contributors:
  - {name: "John G", bio: "Firefox UX"}
  - {name: "Philipp S", bio: "Firefox UX"}
  - {name: "Blake W", bio: "Firefox UX"}

---

<p class="lead">Take your tabs from in the way to tucked away. Tab Center moves your tabs to the side of your browser window so they're out of sight when you don't need them, and easy to expand and grab when you do.</p>

![Some image]()
Tuck away those tabs until you need them.

![Second image]()
Pin Tab Center to see everything you've got open.

![Third image]()
Redecorate and move the furniture! Tab Center works with your favorite Firefox themes.

```

lmorchard:
Flat files issue over here: https://github.com/mozilla/testpilot/issues/988 (also a meeting this Weds)

[...]

Issue #1041, opened 2016-07-07
Update landing page to accomodate removal of FxA auth


lmorchard:
- [x] Update landing page to link directly to Install.
- [ ] Updated Landing page UI to reflect new onboarding steps (at the bottom of the landing page).

Mockup: https://mozilla.invisionapp.com/share/Q77S1H6ZC#/screens/168769316

(Do we need assets for this?)

lmorchard:
The first TODO up there will be covered by #1065

lmorchard:
Looks like the only asset we might need are the chat bubbles image for step number 3

lmorchard:
@johngruen Do we have that chat bubbles image somewhere already? Or does that still need slicing from a mockup?

johngruen:
@lmorchard ah, yeah: https://www.dropbox.com/s/8zsjpvhf9jsz6xe/chat-icon.svg?dl=0

Thanks, i didn't realize this was covering both issues.

Issue #1034, opened 2016-07-07
Remove Test Pilot user database.


ckprice:
Drop email addresses from the database.  This data is in exact target now and we don't need a copy of it.

lmorchard:
Something that occurred to me: We probably want to keep the actual DB tables, but delete _most_ of the records. That is, all users _without_ the `is_staff` and/or `is_superuser` flags. Until we get to using flat files (#988), we still need admin logins to access the admin pages.

lmorchard:
See also: #1056

lmorchard:
Collapsing #1056 & #1046 into this issue as things to keep in mind when we delete data:
- [ ] Retain contributor accounts & profiles (i.e. make sure `is_staff=True`)
- [ ] Retain admin accounts (i.e. `is_superuser=True`)
- [ ] Ensure admin accounts can be logged into at `/admin/` by setting passwords

We need to keep the actual user database around, but we can delete all records where `is_superuser=false AND is_staff=false` and clean up all data not related to core staff

ghost:
Marking as blocked just to keep us from landing this yet.  Next sprint please.

ckprice:
I'm going to take this off this milestone as it doesn't block TXP-7.

lmorchard:
So, I think this can be unblocked now, and taken care of with some SQL run on stage / prod with the next deployment. That, or I can write a Django DB migration that takes care of it. I'm going to poke around a bit on my local machine and see if I can't come up with something solid before tomorrow EoD.

Basically, I think we want to wipe out all accounts & profiles & email address records - _except_ for users with /admin/ access and users included as contributors on experiments.

PR #1065, opened 2016-07-12
Remove requirement to sign-in with FxA

### LINKED ISSUE #1047: Remove sign-in requirement in frontend app
- [ ] [The `me` API resource on the server](https://github.com/mozilla/testpilot/blob/master/testpilot/users/views.py#L20) lists installed experiments and profile data for the logged in user - we'll no longer have this data or a login
- [ ] [The `me` model on the frontend](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/models/me.js) fetches some of its data from the auth'd user's profile on the server. Remove these parts and rely only on what the add-on knows - i.e. experiment installation state, client UUID, etc.

We should also treat presence of the add-on now as basically the "logged in" state.

lmorchard:
This is a big one. I did it in a few steps, so I left those as separate commits for now. But, here's what's happening:
- Change landing page call-to-action to add-on install button
- No more sign-in button on landing page
- Only require presence of add-on to see /experiments
- Add-on presence now enables "get started" / "manage" buttons in
  experiment listings
- Add-on presence indicates active user for header settings button
- Don't attempt to fetch server /api/me on frontend app startup
- Increase time to wait for response from add-on to messages
- Add more details about timeout failure while waiting for message
  responses from add-on
- Hardcode add-on installation URL to `/static/addon/addon.xpi` because
  /api/me is no longer available to configure that
- Remove django-allauth URL routes for sign in / out
- Remove the /api/me viewset
- Clean up related tests

Fixes #1047


lmorchard:
Have to run for the day because I'm starving & eyes have glazed over. But I _think_ this might be in good shape? Needs a serious beating by someone other than me to find what I might have missed. I'll probably scour through it again tomorrow to see if I missed anything after I've slept on it.

lmorchard:
Also note that the landing page messaging is still a little wonky with this PR. We've got issue #1041 to clean up the calls-to-action and whatnot.

testpilot/frontend/static-src/app/main.js
```diff
@@ -65,9 +65,11 @@ app.extend({
   },
 
   // Send webChannel message to addon, use a Promise to wait for the answer.
-  waitForMessage(type, data, timeout = 1000) {
+  waitForMessage(type, data, timeout = 5000) {
```
ghost, line 68: why the increase?

lmorchard, line 68: I suspect, but can't be sure, that this was contributing to some of the "never finishes installing" errors on some machines. If the add-on took more than a second to finish installing & set up, the "Installing" barber poll button would never stop. Just thought of that while I was in here poking at things, so figured I'd give it a bump. An alternative might be to not have a timeout here at all and just have it continue to wait and never error with a timeout.

lmorchard:
Added another commit to also cover issue #1041 in this PR

testpilot/frontend/static-src/app/views/experiment-row-view.js
```diff
@@ -54,7 +54,7 @@ export default BaseView.extend({
       hook: 'show-detail',
       name: 'active'
     },
-    'loggedIn': {
+    'hasAddon': {
       type: 'booleanClass',
       hook: 'show-detail',
       name: 'logged-in'
```
meandavejustice: should this prop be updated to `has-addon`?

testpilot/frontend/static-src/test/views/landing-page.js
```diff
@@ -57,7 +57,7 @@ test('Expected sendToGA call made when logged out', t => {
 
 test('Expected sendToGA call made with no experiments', t => {
   t.plan(3);
-  app.me.user.id = 'gary@busey.net';
+  app.me.hasAddon = true;
   app.sendToGA = (name, data) => {
     t.ok(data.dimension1 === true, 'is logged in');
```
meandavejustice: message should read 'addon is installed'

testpilot/frontend/static-src/test/views/landing-page.js
```diff
@@ -30,21 +30,21 @@ test('Landing Page view renders', t => {
...
 
-  t.ok(view.loggedIn);
+  t.ok(view.hasAddon);
 });
 
-test('Expected sendToGA call made when logged out', t => {
+test('Expected sendToGA call made with no add-on', t => {
   t.plan(3);
   app.sendToGA = (name, data) => {
     t.ok(data.dimension1 === false, 'is not logged in');
```
meandavejustice: message needs updated

testpilot/frontend/static-src/test/views/landing-page.js
```diff
@@ -69,11 +69,11 @@ test('Expected sendToGA call made with no experiments', t => {
...
-    {'addon_id': '@experiment1'},
-    {'addon_id': '@experiment2'}
-  ];
+  app.me.hasAddon = true;
+  app.me.installed = {
+    '@experiment1': {},
+    '@experiment2': {}
+  };
   app.sendToGA = (name, data) => {
     t.ok(data.dimension1 === true, 'is logged in');
```
meandavejustice: message needs updated

testpilot/frontend/static-src/test/views/landing-page.js
```diff
@@ -57,7 +57,7 @@ test('Expected sendToGA call made when logged out', t => {
 
 test('Expected sendToGA call made with no experiments', t => {
   t.plan(3);
-  app.me.user.id = 'gary@busey.net';
```
meandavejustice, line 60: always sad to lose a reference to gary busey

testpilot/settings.py
```diff
@@ -509,7 +509,8 @@ def lazy_langs():
     "https://code.cdn.mozilla.net/fonts/fira.css":
         "sha384-APhs/OUouhH+ZbBANL3+7a5J1sVSYyfUhGxTWiDMkTuEIO5fXWZonzNEj+CagDB7",
 }
-STATICINTEGRITY_CACHE_TTL = 30
+STATICINTEGRITY_CACHE_TTL = config('STATICINTEGRITY_CACHE_TTL',
```
meandavejustice, line 512: could you explain this change somewhere? is this just removing the hardcoded value?

testpilot/urls.py
```diff
@@ -4,21 +4,16 @@
...
-from .users import views as users_views
 
 # Allow apps to contribute API parts
 router = routers.DefaultRouter(trailing_slash=False)
 experiment_views.register_views(router)
-users_views.register_views(router)
 
 urlpatterns = patterns(
     '',
     # users app is special, because it handles /accounts and /users
```
meandavejustice: comment needs to be updated now that we've removed `/users`

meandavejustice:
I see that we aren't just removing the User model entirely, why is this?

meandavejustice:
Although it's a little nitpicky, I think we should update the `.logged-in` classname in `testpilot/frontend/static-src/styles/modules/_experiment-list.scss` and corresponding templates in this pr, otherwise it may never be addressed again.

meandavejustice:
We still have copy about being logged-in to fxa in `docs/FAQs.md` and `docs/DEPLOYMENT-VERIFICATION.md`

meandavejustice:
This works and looks great! thanks for doing such a thorough job on this @lmorchard 
I'm happy to merge after the above comments are addressed

lmorchard:
> I see that we aren't just removing the User model entirely, why is this?

We're still using the UserProfile model for the contributors list (#1046 / #1034)

testpilot/settings.py
lmorchard, line 512: Yeah, just making the value configurable, so I can set it to a very short duration on dev where the static files change very often during development

lmorchard:
I think I caught all the comments in this new commit, and squashed some things together

meandavejustice:
Looks great to me, once tests pass I'll go ahead and merge

PR #1077, opened 2016-07-15
Wait longer for add-on setup, reload install page if it times out

### LINKED ISSUE #1075: "Installing..." keeps spinning after add-on installation attempt
Occasionally on a fresh Firefox profile, I try to install the add-on but the "Installing..." barberpole never stops spinning. 

There will be an error in the Web Console like "Uncaught promise rejection: waitForMessage timed out". Having trouble reproducing this to capture the exact error message. 

[waitForMessage is here](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/main.js#L68) and it has a timeout for when the add-on takes too long to respond. It [gets called here in the frontend](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/models/me.js#L41) after the install button [detects that the add-on has been installed](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/views/landing-page.js#L55).

My guess is that between installation and when the add-on finally starts responding to messages, there's a gap of sometimes more than 5 seconds and that's why the error happens. Still investigating, though.

Every time this has happened to me, a page refresh seems to fix everything and the add-on is actually working.

lmorchard:
When the add-on is first installed from the "Install the Test Pilot Add-on" button, we start an "Installing" animation. Once the add-on is installed & set up, we try to detect that and move on to the next step. If the add-on takes too long, we time out with an unhandled error.

This PR attempts to handle that error by reloading the page. If the add-on was installed but just took awhile to set up, this should result in the user seeing the experiments page. Otherwise, they'll see the "Install the Test Pilot Add-on" button again

Fixes #1075

### DIFF
--- a/testpilot/frontend/static-src/app/views/landing-page.js
+++ b/testpilot/frontend/static-src/app/views/landing-page.js
@@ -61,6 +61,10 @@ export default PageView.extend({
       app.me.fetch().then(() => {
         app.webChannel.sendMessage('show-installed-panel', {});
         app.router.redirectTo('experiments');
+      }).catch(() => {
+        // HACK (for Issue #1075): Timed out while waiting for the initial sync
+        // message, so just reload the page to stop waiting.
+        window.location.reload();
       });
     }, 1000);
   },
--- a/testpilot/frontend/static-src/app/main.js
+++ b/testpilot/frontend/static-src/app/main.js
@@ -65,7 +65,7 @@ app.extend({
   },
 
   // Send webChannel message to addon, use a Promise to wait for the answer.
-  waitForMessage(type, data, timeout = 5000) {
+  waitForMessage(type, data, timeout = 10000) {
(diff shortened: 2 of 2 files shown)


testpilot/frontend/static-src/app/views/landing-page.js
```diff
@@ -61,6 +61,10 @@ export default PageView.extend({
       app.me.fetch().then(() => {
         app.webChannel.sendMessage('show-installed-panel', {});
         app.router.redirectTo('experiments');
+      }).catch(() => {
+        // HACK: Timed out while waiting for the initial sync message, so just
+        // reload the page to stop waiting.
```
jaredhirsch: Nit: reference issue #1075 in the HACK notice

jaredhirsch:
:beers:

PR #1102, opened 2016-07-19
Attempt to make frontend-to-addon comms more robust

### LINKED ISSUE #1101: Blank page is randomly displayed when redirected after add-on instalation or experiment page opened in a new tab
[Affected versions]:
- Nightly 50.0a1
- Add-on v0.8.0-tag-2016-07-15-1

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Steps to reproduce]:
1. Go to https://testpilot.firefox.com/ and install the Test Pilot add-on.
2. Observe where you are redirected after installation.
3. Click the "Test Pilot" icon from toolbar and click on any of the available experiments.

[Expected result]:
2. The user is redirected to the experiments page.
3. Experiment description page is opened in a new tab. 

[Actual result]:
2. The user is randomly redirected to a blank page even if the link in address bar seems correct.
3. New tab opened and user is randomly redirected to a blank page even if the link in address bar seems correct.

[Notes]:
- Added a screenshot with the browser console errors. The web console does not display almost any errors except "waitForMessage timeout: sync-installed".
- Tried with a new profile and cache cleared but same issue occurs.
- The page only loads after multiple refreshes
  console errors
  .

lmorchard:
- Perform initial experiments API fetch and add-on installed experiment
  message happen in series, rather than in parallel
- `app.waitForMessage` will now retry several times before finally
  timing out with a rejection (default 3 times)
- If all else fails on app init, bounce to error page instead of blank
  white page

Fixes #1101

### DIFF
--- a/testpilot/frontend/static-src/app/main.js
+++ b/testpilot/frontend/static-src/app/main.js
@@ -42,39 +42,62 @@ app.extend({
     app.me = new Me();
     app.experiments = new ExperimentsCollection();
 
-    Promise.all([
-      app.me.fetch(),
-      app.experiments.fetch()
-    ]).then(() => {
+    app.pageManager = new PageManager({
+      pageContainer: document.querySelector('[data-hook=page-container]')
+    });
+
+    if (!app.router.history.started()) {
+      app.router.history.start();
+      // HACK for Issue #124 - sometimes popstate doesn't fire on navigation,
+      // but pageshow does. But, we just want to know if the URL changed.
+      addEventListener('pageshow', app.router.history.checkUrl, false);
+    }
+
+    app.experiments.fetch().then(() => app.me.fetch()).then(() => {
       app.me.updateEnabledExperiments(app.experiments);
-
-      app.pageManager = new PageManager({
-        pageContainer: document.querySelector('[data-hook=page-container]')
-      });
-
-      if (!app.router.history.started()) {
-        app.router.history.start();
-        // HACK for Issue #124 - sometimes popstate doesn't fire on navigation,
-        // but pageshow does. But, we just want to know if the URL changed.
-        addEventListener('pageshow', app.router.history.checkUrl, false);
-      }
     }).catch((err) => {
-      // for now, log the error in the console & do nothing in the UI
       console && console.error(err); // eslint-disable-line no-console
+      app.router.redirectTo('error');
     });
   },
 
   // Send webChannel message to addon, use a Promise to wait for the answer.
-  waitForMessage(type, data, timeout = 10000) {
+  waitForMessage(type, data, timeout = 5000, retries = 3) {
     return new Promise((resolve, reject) => {
-      const rejectTimer = setTimeout(() => {
-        reject('waitForMessage timeout: ' + type);
-      }, timeout);
-      this.once('webChannel:' + type + '-result', (result) => {
-        clearTimeout(rejectTimer);
+      let attempts = 0;
+      let retryTimer;
+
+      // On timeout, try again or reject with an error if out of retries.
+      const timeoutHandler = () => {
+        if (attempts++ < retries) {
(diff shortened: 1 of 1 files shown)


lmorchard:
I _still_ can't reliably reproduce this issue - and I have never been able to do so on local development. It happens for me like 1 out of 20 attempts on staging. My strong hunch is that this is a race condition, but I don't know where the race is happening. 

So, this PR attempts to stop the number of things happening at once and retry things a few times before reporting an error. And then, on error, show the error page <strike>reload the page rather than just show a blank screen with an error in the console</strike>.

<strike>I'm a little worried that that reload might turn into an infinite loop if there's another more persistent issue here that depends on environmental context I don't have on my machine.</strike>

testpilot/frontend/static-src/app/main.js
```diff
@@ -59,22 +56,49 @@ app.extend({
...
-      // for now, log the error in the console & do nothing in the UI
       console && console.error(err); // eslint-disable-line no-console
+      // HACK (for Issue #1101): Last ditch - reload the page and try again
+      window.location.reload();
     });
   },
 
   // Send webChannel message to addon, use a Promise to wait for the answer.
-  waitForMessage(type, data, timeout = 10000) {
+  waitForMessage(type, data, timeout = 5000, retries = 3) {
```
lmorchard, line 65: FWIW, reduced timeout to 5 seconds from 10 - because the retries should actually lead to a total timeout of _15 seconds_ before rejection.

lmorchard:
@meandavejustice @6a68 @chuckharmston - this could use some additional eyes and sanity checking, because I'm kind of handwaving at the problem here 😿

lmorchard:
@ckprice - oh and hey, if you can try this out, it could be helpful since you reported being able to reliably reproduce this!

chuckharmston:
> If all else fails on app init, just try reloading the page.

I'm not a fan of this; it could lead to a loop of endless refreshes if the user is like @ckprice and can reproduce reliably. Maybe ship them to the error page, instead?

chuckharmston:
Otherwise, code looks good and logic is sound. Unfortunately, like you I can't reproduce the original issue, so I'm reticent to merge anything. I'll let someone else check it out.

lmorchard:
> I'm not a fan of this; it could lead to a loop of endless refreshes if the user is like @ckprice and can reproduce reliably. Maybe ship them to the error page, instead? 

Yeah, the more I think about that, the less I like it. Just pushed a tweak to bounce to the error view, rather than reload. Better than a blank white page, at least.

testpilot/frontend/static-src/app/main.js
```diff
@@ -42,39 +42,62 @@ app.extend({
...
-    ]).then(() => {
-      app.me.updateEnabledExperiments(app.experiments);
+    app.pageManager = new PageManager({
+      pageContainer: document.querySelector('[data-hook=page-container]')
+    });
 
-      app.pageManager = new PageManager({
-        pageContainer: document.querySelector('[data-hook=page-container]')
-      });
+    if (!app.router.history.started()) {
```
lmorchard, line 49: Also, moving page manager & router init outside of the experiments fetch means we can show the error page. 

Side effect will be that the page will start to render before the API actually returns results, so typically a sub-second lag until filling in with experiments. Previously, the whole page would just stay white until the experiments came in.

ckprice:
> @ckprice - oh and hey, if you can try this out, it could be helpful since you reported being able to reliably reproduce this!

Pulled this down locally and was unable to trigger the error.

My "pretty reliable" STR's on production have become less reliable in the past couple of hours FYI. One in every 10-12 tries or so.

lmorchard:
Unless there are objections, I may end up self-merging this tomorrow just to get it in there for next deploy.

meandavejustice:
@lmorchard pulling this down this morning for a quick check then I'm fine to merge for you

PR #1111, opened 2016-07-20
[fixes #555] Refresh demo_data.json

### LINKED ISSUE #555: Refresh demo_data.json with up to date info

ckprice:
Fixes #555.

Can be tested with ([ref](https://github.com/mozilla/testpilot/blob/master/docs/README-DOCKER.md)).

```
mkdir -p media
cp -r fixtures/media/demo media
docker exec testpilot_server_1 ./manage.py loaddata fixtures/demo_data.json
```

Once we confirm this works for someone else locally, I'm happy to try to apply to stage (with instruction).


lmorchard:
FWIW, the only way I can think of to apply this to stage is to file a bug with ops to run those commands for us - i.e. like we do for deployment. We don't have a way to self-service on this beyond making manual changes through the admin pages.

lmorchard:
This also doesn't necessarily _fix_ issue #555, because this demo data will go out of date every time we launch a new experiment or make changes to content or site features. Issue #555 should become more of a going concern like unit tests whenever we work on the site.

[Like I wrote earlier](https://github.com/mozilla/testpilot/issues/555#issuecomment-207581824), the demo data is not intended to reflect current experiments, so much as it should exercise features on the site.

If we want to keep staging data somewhat current with production data, we probably want a different import/export mechanism than `demo_data.json`. Maybe an issue to set up periodic or on-demand experiment data dumps from prod to stage to dev.

ckprice:
> Like I wrote earlier, the demo data is not intended to reflect current experiments, so much as it should exercise features on the site.

Yes. Things like installing experiments and taking tours weren't happening because stuff was pretty out of date. #555 describes a refresh of `demo_data.json`, it doesn't imply any mechanism for sync'ing environments or anything else.

The recent QA concerns on stage may have been conflated with this bug a bit, but I think this closes #555. If we think we need some kind of mechanism to keep production updated with staging, let's file another issue.

lmorchard:
Running through the stuff in the commit, it just looks like a lot of content - but not an exact duplicate of production. So, maybe it's just kind of overkill (IMO) in exercising features? That makes my comments more nits than real problems.

In that case, this works for me as-is. We should just try to keep this stuff updated and maybe look into ways to get better content into staging as a separate issue

lmorchard:
Just filed #1113 to carry over the content-in-staging part of things here.

ckprice:
> FWIW, the only way I can think of to apply this to stage is to file a bug with ops to run those commands for us - i.e. like we do for deployment. We don't have a way to self-service on this beyond making manual changes through the admin pages.

Thanks. I'll take a todo to manually update stage.

PR #1120, opened 2016-07-22
Update SRI hash for pontoon dependency

lmorchard:
This should fix the hash mismatch error reported here:

https://bugzilla.mozilla.org/show_bug.cgi?id=1288286


lmorchard:
Just going to merge, because I've already updated to this hash on prod admin

PR #1122, opened 2016-07-22
Use setInterval in add-on for periodic experiment updates

lmorchard:
Realized I only used a `setTimeout` that only fired once. We need this to happen periodically & repeatedly.


lmorchard:
Small change, tests pass, will self-merge to make sure it gets into this release

PR #1132, opened 2016-07-26
Update SRI hash for Pontoon dependency

lmorchard:


lmorchard:
Passes tests, already made the change in /admin/... I'll just merge

PR #1136, opened 2016-07-26
Use module-global `Metrics` instead of `this` in metrics.js

### LINKED ISSUE #1135: variant error: this.experimentFeaturesChanged is not a function
I'm unable to send variant data from universal search to test pilot.

[Here](https://github.com/mozilla/universal-search/pull/261/commits/89f536f346a24dd7f1d17143c7714f89a3ee858e) is a universal search commit that mostly contains variant-related code.

Each time I send the data over, I see this error in the browser toolbox console:

```
TypeError: this.experimentFeaturesChanged is not a function
Stack trace:
module.exports.onReceiveVariantDefs@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-addon/lib/metrics.js:163:5
Observer<.observe@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/system/events.js:81:7
Search.prototype.fetchVariantData/sendPing@chrome://universalsearch-lib/content/universal-search.js:109:7
setTimeout_timer@resource://gre/modules/Timer.jsm:30:5
```

Looks like the value of `this` isn't properly bound at line 163 of lib/metrics.js.

lmorchard:
Fixes #1135

### DIFF
--- a/addon/lib/metrics.js
+++ b/addon/lib/metrics.js
@@ -62,7 +62,7 @@ const variantMaker = {
 };
 
 
-module.exports = {
+const Metrics = module.exports = {
 
   init: function() {
     if (!store.telemetryPingPayload) {
@@ -73,10 +73,10 @@ module.exports = {
       };
     }
 
-    this.maybePingTelemetry();
+    Metrics.maybePingTelemetry();
 
-    Events.on(EVENT_SEND_METRIC, this.onExperimentPing);
-    Events.on(EVENT_RECEIVE_VARIANT_DEFS, this.onReceiveVariantDefs);
+    Events.on(EVENT_SEND_METRIC, Metrics.onExperimentPing);
+    Events.on(EVENT_RECEIVE_VARIANT_DEFS, Metrics.onReceiveVariantDefs);
   },
 
   onEnable: function() {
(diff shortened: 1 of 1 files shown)


addon/lib/metrics.js
```diff
@@ -75,8 +75,8 @@ module.exports = {
 
     this.maybePingTelemetry();
 
-    Events.on(EVENT_SEND_METRIC, this.onExperimentPing);
-    Events.on(EVENT_RECEIVE_VARIANT_DEFS, this.onReceiveVariantDefs);
+    Events.on(EVENT_SEND_METRIC, this.onExperimentPing.bind(this));
+    Events.on(EVENT_RECEIVE_VARIANT_DEFS, this.onReceiveVariantDefs.bind(this));
```
jaredhirsch: This won't work because we use `Events.off` to disconnect the functions further down

jaredhirsch: You'll need to assign the bound functions somewhere and pass 'em to `on` and `off`

lmorchard:
Updated to something more drastic: Switched away from `this` to using a `Metrics` global reference to `module.exports` - [we use that pattern elsewhere](https://github.com/mozilla/testpilot/blob/master/addon/lib/toolbar-button.js#L89) and `module.exports` is a weird object anyway. (e.g. it's frozen such that any assignments to `this.foo = 'bar'` is a runtime error, so we shouldn't treat it like a typical object)

jaredhirsch:
lgtm but I don't have time right now to verify locally, I can look later this afternoon

jaredhirsch:
looks good, merging

PR #1139, opened 2016-07-26
Indicator for just launched & updated experiments

### LINKED ISSUE #1038: Update UI on website with new experiment UI.
UI added when a new experiment is added to Test Pilot. Removed once a user has enabled the experiment or clicked through to the details page.

UX: See the New Experiment Notification card here: https://invis.io/Q77S1H6ZC

lmorchard:
- Add `lastSeen` property to `experiment` model persisted in
  `localStorage`
- Update `lastSeen` on experiment detail page view
- When experiment created & never seen, show a "Just Launched" indicator
  for 2 weeks in experiment card view.
- When experiment modified since last seen, show a "Just Updated"
  indicator for 2 weeks in experiment card view.
- Style tweaks, strings, and tests

Fixes #1038

### DIFF
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -51,6 +51,8 @@ notFoundHeader = Four Oh Four!
 experimentListPageHeader = Ready for Takeoff!
 experimentListPageSubHeader  = Pick the features you want to try. <br> Check back soon for more experiments.
 experimentListEnabledTab = Enabled
+experimentListJustLaunchedTab = Just Launched
+experimentListJustUpdatedTab = Just Updated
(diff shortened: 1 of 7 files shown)


lmorchard:
@johngruen Improvised here a little bit, by putting a 2-week limit on Just Launched / Just Modified indicators. 

We can change that value, but we might want to keep _some_ limit there - because otherwise all our existing experiments will appear as "Just Launched" when we deploy this.

johngruen:
@lmorchard looks good in dev, the blue/white borders don't show up if the user doesn't have the add-on installed. I filed #1145 as a follow up

PR #1143, opened 2016-07-27
Stop logging request.summary for unimportant paths

### LINKED ISSUE #646: Let's not log heartbeat hits
I think we should have a list somewhere of URLs which we don't log.  We had 41k hits yesterday, of which 80 were to URLs that weren't:

```
/favicon.ico
/__heartbeat__
/__lbheartbeat__
/nginx_status
/robots.txt
```

Let's not bother sending logs for those  URLs

dannycoates:
Howdy!

I think this fixes #646. This is my first patch to testpilot, I've never used django, and fairly noob to python on the whole, so I'd ❤️ some feedback. Is there a good way to test this?

### DIFF
--- a/testpilot/settings.py
+++ b/testpilot/settings.py
@@ -379,6 +379,19 @@ CSP_STYLE_SRC = (
     'https://*.mozilla.net',
 )
 
+LOG_PATH_BLACKLIST = (
+    '/favicon.ico',
+    '/__heartbeat__',
+    '/__lbheartbeat__',
+    '/nginx_status',
+    '/robots.txt'
+)
+
+
+def skip_logging_blacklisted_paths(record):
+    return not (record.path and record.path in LOG_PATH_BLACKLIST)
+
+
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
@@ -388,6 +401,10 @@ LOGGING = {
         },
         'require_debug_true': {
             '()': 'django.utils.log.RequireDebugTrue',
+        },
+        'log_path_blacklist': {
+            '()': 'django.utils.log.CallbackFilter',
+            'callback': skip_logging_blacklisted_paths
         },
     },
     'formatters': {
@@ -423,6 +440,7 @@ LOGGING = {
         'request.summary': {
             'handlers': ['console'],
             'level': 'INFO',
+            'filters': ['log_path_blacklist'],
         },
         'django': {
             'handlers': ['console'],
--- a/testpilot/base/tests.py
+++ b/testpilot/base/tests.py
@@ -9,11 +9,38 @@ from django.db import OperationalError
 
 from constance.test import override_config
 
+from testfixtures import LogCapture
+
 from ..utils import TestCase
 from .jinja import _hash_content, waffleintegrity, TestPilotExtension
 
 import logging
 logger = logging.getLogger(__name__)
+
+
+class RequestSummaryLoggingTests(TestCase):
+
+    def setUp(self):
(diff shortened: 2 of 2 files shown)


testpilot/settings.py
```diff
@@ -379,6 +379,22 @@ def lazy_langs():
...
+    '/favicon.ico',
+    '/__heartbeat__',
+    '/__lbheartbeat__',
+    '/nginx_status',
+    '/robots.txt'
+]
+
+
+def skip_logging_blacklisted_paths(record):
+    if record.path:
```
lmorchard: I think you could actually reduce this whole function to something like this:

```
def skip_logging_blacklisted_paths(record):
    return not (record.path and record.path in LOG_PATH_BLACKLIST)
```

testpilot/settings.py
```diff
@@ -389,6 +405,10 @@ def lazy_langs():
         'require_debug_true': {
             '()': 'django.utils.log.RequireDebugTrue',
         },
+        'log_path_blacklist': {
+            '()': 'django.utils.log.CallbackFilter',
```
lmorchard, line 406: Huh, I didn't even know this existed - nice find for a Django & Python noob 😄

lmorchard:
Works for me, the blacklisted paths are filtered from logging!

> Is there a good way to test this?

If you really want to dig into writing a test for this, here are some pointers off the top of my head:
- [`testpilot/base/tests.py`](https://github.com/mozilla/testpilot/blob/master/testpilot/base/tests.py) would be a good generic place for it. 
- We have some tests in [`testpilot/experiments/tests.py`](https://github.com/mozilla/testpilot/blob/master/testpilot/experiments/tests.py) that [capture log messages using LogCapture](https://github.com/mozilla/testpilot/blob/master/testpilot/experiments/tests.py#L210) to ensure certain events are emitted after requests.
- There's an [`override_settings`](https://github.com/mozilla/testpilot/blob/master/testpilot/base/tests.py#L131) decorator that lets you temporarily change settings for a test - i.e. set `LOG_PATH_BLACKLIST` to some test list to make sure it works

If you put all the above together, you can probably get a nice test that makes a few [mock requests](https://github.com/mozilla/testpilot/blob/master/testpilot/base/tests.py#L25) and asserts whether or not log events should be emitted as appropriate with respect to the `LOG_PATH_BLACKLIST`

All that said, though, I think I'd be happy to merge this as-is and circle back to the tests after - they'll probably be a bit more involved than the work you did to get the blacklist working and we appreciate the contribution 😀

dannycoates:
Thanks @lmorchard! 

I added a couple test cases but I wasn't able to get `@override_settings` to work. I originally tried this...

``` py
@override_settings(LOG_PATH_BLACKLIST=('/dummy'))
def test_unblackisted_are_logged(self):
        self.handler.records = []
        url = '/__heartbeat__'
        resp = self.client.get(url)
        self.assertEqual(200, resp.status_code)
        record = self.handler.records[0]
        self.assertEqual(url, record.path)
```

but it failed because `/__heartbeat__` was blocked (by `settings.py`).

Anyway, I think this is ready for r?

lmorchard:
Hmm, not quite sure why `@override_settings` didn't work. Unless, maybe, the `skip_logging_blacklisted_paths` function in `settings.py` just doesn't see the override and continues using the local variable.

Meh, either way, these tests seem to work and that's fine by me

lmorchard:
And looks like all the tests ran & passed on CircleCI, so r+ from me!

PR #1144, opened 2016-07-27
Add notifications to Test Pilot for new/updated experiments

### LINKED ISSUE #753: Add push notifications to Test Pilot for New Experiments
So that I can see when a new experiment is released, I would like to receive push notifications in Firefox from Test Pilot.

### LINKED ISSUE #1126: Measure click through & install rates on variant push notification messages
- [ ] It should measure clicks on the push notification that lead to experiment installs
- [ ] It should measure the different click through rates of notifications with different copy

lmorchard:
- New `addon/lib/experiment-notifications.js` module for the feature
- Check for new & updated experiments on every refresh of data. If a
  notification hasn't already been sent, choose one of the items of
  notification content and display a notification.
- Add ExperimentNotification content model to Experiment, with admin &
  API support.
- Add `html_url` field to ExperimentSerializer to point to
  human-readable HTML page
- Router bugfix - since router was started before experiment data was
  loaded, links directly to experiment subpages would 404
- Tests & migrations

Fixes #753, Fixes #1126


lmorchard:
I _think_ this is ready, or at least could use some eyes on it. But there are a few things that could use tweaking / second opinions:
- [ ] The add-on only refreshes experiment data once daily - and that's when the data is checked for new & updated experiments to fire notifications. Do we want to make this more frequent?
- [ ] [`addon/data/notification-icon.png`](https://github.com/lmorchard/testpilot/blob/8b4a5992a26b904f821a261b1a67f54b2a249609/addon/data/notification-icon.png) copter image is huge - probably should be scaled down to icon size for the notification. Though, I don't know what the size is yet
- [ ] Add-on notifications don't look exactly like web push notifications, but [I added a hardcoded 'via testpilot.firefox.com' line in the notification text](https://github.com/mozilla/testpilot/pull/1144/files#diff-2747efc5ad9bb8131eee041a83458b30R79) which seems hackish but looks more like the mockup
- [ ] Different notification messages can be specified for experiments when they're new vs when they're updated - not asked for in the issue, but seemed handy.
- [ ] "Cohorts" for receiving messages is just [a random selection](https://github.com/mozilla/testpilot/pull/1144/files#diff-2747efc5ad9bb8131eee041a83458b30R73) right now. Hope that's ok?
- [ ] `utm_content` param on notification links is just the database ID of the ExperimentNotification - do we want something more descriptive than that? (e.g. a manually-entered short slug that's readable in GA?)

lmorchard:
Oh yeah, and some tips to exercise this locally:
- On install, you probably shouldn't see a notification because existing experiments are old
- Modify one of those existing experiments to add some notification content in /admin/ and you should see an update notification shortly after saving.
- Add a new experiment in /admin/ and you should see an update shortly after saving.

meandavejustice:
Alright, this is working when updating activity stream, and adding a new experiment.

#### here is a screenshot

notification-tp

### your checklist

> The add-on only refreshes experiment data once daily - and that's when the data is checked for new & updated experiments to fire notifications. Do we want to make this more frequent?

I'm good to merge with this number, although I think twice a day would still be reasonable. Probably can address it in a follow up issue.

> addon/data/notification-icon.png copter image is huge - probably should be scaled down to icon size for the notification. Though, I don't know what the size is yet

created https://github.com/mozilla/testpilot/issues/1149 

> Add-on notifications don't look exactly like web push notifications, but I added a hardcoded 'via testpilot.firefox.com' line in the notification text which seems hackish but looks more like the mockup

would like to get @johngruen 's opinion on this before merging

> Different notification messages can be specified for experiments when they're new vs when they're updated - not asked for in the issue, but seemed handy.

I think this is great!

johngruen:
@meandavejustice looking

johngruen:
@lmorchard everything looks great, i think the 'update' notifications are a little to easy to trigger. making an adjustment to the changelog URL, for example, probably shouldn't kick off a push notification, nor should every dot release of the add-on. Could you add a field in the admin maybe... i'm seeing just numeric input that we could increment to signal that an update push notification should go out?

I'm picturing something like this: 
launch an experiment we set this field to zero
do a bunch of dot releases that just fix nits, field stays at zero
do a big release with major changes, bump the number to 1, change triggers push
do a bunch of releases with minor changes, number stays at 1
do one more big release, number goes to two, change triggers push

johngruen:
@lmorchard one more thing, i tried and tried to find a spec for the right sized icon, but it doesn't exist :)... i'll toss a small version your way in the morning!

lmorchard:
> Could you add a field in the admin maybe... i'm seeing just numeric input that we could increment to signal that an update push notification should go out?

Easier to add another date field that we can manually set. Maybe call it `notifyAfter` and the code can use that for comparisons. I'd drop the new vs updated message flags and we can just change the set of messages when an experiment launches vs when it updates.

johngruen:
@lmorchard one more nit, when a user installs the test pilot add on they get hammered with push notifications (at least that's what i'm seeing on my local version). Can we make is so existing notifications don't get sent to new users?

johngruen:
notification-icon

smaller version of the icon!

lmorchard:
> one more nit, when a user installs the test pilot add on they get hammered with push notifications (at least that's what i'm seeing on my local version).

@johngruen [That shouldn't be happening for any experiments created / modified more than 2 weeks ago.](https://github.com/mozilla/testpilot/pull/1144/files#diff-2747efc5ad9bb8131eee041a83458b30R44) How new are the dates on your local data?

lmorchard:
Actually, I did just realize a bug in the logic that I'll fix: It never keeps track of notifications as handled, so it would just keep notifying on every experiment data update until 2 weeks had passed. That's bad.

lmorchard:
Updated the commit:
- Replaced the notification icon
- Changed the experiment update period to 12 hours
- Switched to a `notify_after` date field in notifications to schedule them manually
- Removed `is_new` & `is_updated` filter flags - the `notify_after` schedule should work better
- Tracking processed notifications so they don't get repeatedly displayed

lmorchard:
@johngruen @meandavejustice This could use another round of review after these changes.

lmorchard:
Oh yeah, and one more update to bump the add-on version

johngruen:
looking

johngruen:
> one more nit, when a user installs the test pilot add on they get hammered with push notifications (at least that's what i'm seeing on my local version).
> 
> @johngruen That shouldn't be happening for any experiments created / modified more than 2 weeks ago. How new are the dates on your local data?

yeah, the date system makes this issue much better, but it's still confusing to get popped with notifications when you first install the add-on....

i'm happy to have this fixed in a follow up issue though. This all looks really great @lmorchard thank you so much!

lmorchard:
@johngruen I guess I don't understand what you're seeing, because the issue I saw should be fixed... how new are the dates on your local data? Need more info before filing a follow up issue

addon/lib/experiment-notifications.js
```diff
@@ -0,0 +1,93 @@
...
+    // Initialize notifcations last sent timestamp to a truthy value that's
+    // also less than whatever the current time is.
+    if (!('notificationsLastChecked' in store)) {
+      store.notificationsLastChecked = 1;
+    }
+
+    // Ensure we have an initial empty object for processed notifications
+    if (!('notificationsProcessed' in store)) {
+      store.notificationsProcessed = {};
+    }
```
meandavejustice, line 21: I think we should be cleaning these up on `uninstall` in our unload handler

johngruen:
@lmorchard STR
1. Uninstall Test Pilot add-on
2. in your local admin, set a notification on an experiment, to happen right now
3. change something in the experiment copy
4. save
5. install the test pilot add-on

Expected
Test Pilot Installed

Actual
Test Installed with a confusing push notification

lmorchard:
Just pushed another commit with further tweaks:
- <strike>`notificationsLastChecked` gets initialized with the current date on install, which should prevent a bunch of notifications on first start (I think that covers @johngruen's issue?)</strike> Nope, that didn't fix it. Still thinking about this, ugh.
- Added an `uninstall()` method called on uninstall that deletes the notification module data from `store`

lmorchard:
Okay, let's try this: On the first run, the list of "read" notifications is empty. So, if it's empty, let's just "mark all as read" on that first run and only start displaying notifications after that. Just updated the commit to do that.

PR #1146, opened 2016-07-28
Add border to Just Launched cards when user does not have add-on

### LINKED ISSUE #1145: Add border to Just Launched cards when user does not have add-on installed
# E Current:

screen shot 2016-07-28 at 2 07 46 pm
## Expected:

Just Launched cards should have the same blue border whether or not users have the Test Pilot add-on installed

lmorchard:
Fixes #1145


johngruen:
thanks @lmorchard

PR #1147, opened 2016-07-28
hack to add no more 404s subtitle

johngruen:
@lmorchard, care to take a look at this one: 

all the little hacks are noted as TODOs


lmorchard:
And it even has tests! Works for me.

PR #1155, opened 2016-07-29
Replace post-install doorhanger with new tab

### LINKED ISSUE #1074: Addon: Replace post-install doorhanger with new tab
This issue should have been filed with the rest of TXP-7, but there accounts-backout UX does away with the installed-panel UI. Instead it opens a second tab in the background on install with the following:

add-on notice 2x

Flow should be: 
1. user installs add-on
2. browser redirects to the experiments home page
3. the above page opens in a background tab.

There is no panel UI in the updated flow

jaredhirsch:
To avoid duplicating styles and images, the new tab opens a page on the
Test Pilot website, not a page stored inside the add-on.

Fixes #1074.


johngruen:
@6a68 ~~looking~~ looking in a bit

jaredhirsch:
@johngruen thanks! I've had occasional problems where the addon couldn't figure out the environment, so it wouldn't open the new tab (added a TODO about that in the code). If you surf to testpilot.dev:8000/onboarding you should see the page, independent of any addon tab opening errors

johngruen:
@6a68, sounds good...do you think the env problem is an artifact of you switching around envs a lot before hand? ie something our users would see?

meandavejustice:
After a fresh install, just had it try and open `resource://testpilot-addon/data/undefined/onboarding` in a background tab

johngruen:
style nits on the onboarding page:
- use this image https://cloud.githubusercontent.com/assets/3323249/16886039/cd9ca366-4a9f-11e6-8832-7aac43158f37.png
- the background size of the image (and width, height) should be `386px 182px`
- add `$grid-unit * 2` bottom-margin to the `toolbar-button-onboarding` div

johngruen:
also cannot get the page to trigger on a clean install

addon/index.js
```diff
@@ -45,6 +44,8 @@ const WebExtensionChannels = require('./lib/webextension-channels');
 const ToolbarButton = require('./lib/toolbar-button');
 const ExperimentNotifications = require('./lib/experiment-notifications');
 
+const EXPERIMENT_UPDATE_INTERVAL = 24 * 60 * 60 * 1000; // 1 day
```
lmorchard: I think this is going to need a rebase against master / small manual tweak, [because we changed this to 12 hours in the notifications PR](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L10). But, the `const` got moved to the top, so this won't technically look like a conflict.

lmorchard: Yeah, [looks like linting on CircleCI caught it](https://circleci.com/gh/mozilla/testpilot/675)

jaredhirsch: derp, on it

addon/index.js
```diff
@@ -157,6 +158,15 @@ function initServerEnvironmentPreference() {
   });
 }
 
+function openOnboardingTab() {
+  // TODO: Settings is populated async, so we need a global afterSettingsReady
+  // promise that this function can wait on.
+  tabs.open({
+    url: SERVER_ENVIRONMENTS.production.BASE_URL + '/onboarding',
```
lmorchard, line 160: Hmm, a little leary of hardcoding production here. But, the only idea I have to fix it is already in the TODO so...

lmorchard, line 160: FWIW, we basically have an `afterSettingsReady` in [`updatePrefs()`](https://github.com/mozilla/testpilot/blob/dc5f374b956394682c373c9bd890aa152436dd1f/addon/index.js#L77) - that's where we do all the post-environment stuff. Would have to maybe set a global var like `doOpenOnboardingTab` in the startup code and then resolve it in `updatePrefs()` by showing the tab.

jaredhirsch, line 160: @lmorchard I was actually thinking of passing a callback to updatePrefs to call when it was totally done, simpler than a promise and does what's needed. THoughts?

jaredhirsch, line 160: this seems like a good candidate for a followup bug, actually. getting pretty late on friday.

lmorchard:
Looks okay to me. One gotcha for testing - have to close `about:addons` after uninstalling before re-installing to get the `installed` state that triggers the new tab. That, or exit & restart Firefox after "removing" the add-on

addon/index.js
jaredhirsch, line 160: opened issue #1159 to track this todo

PR #1168, opened 2016-08-01
Disable SRI enforcement for 3rd party resources

### LINKED ISSUE #1134: Remove SRI enforcement from the site
SRI turns out to be trouble for 3rd party services that update their JS-include-based API hooks often - e.g. Pontoon, Google Analytics, Optimizely. The hashes change without notice, and SRI's rejection of resources has the potential for major site breakage.

Thought we needed SRI for security requirements, but can't find an email or bug mentioning it. So, since it seems more trouble than it's worth without better tooling (e.g. Issue #1121), let's remove it for now.

lmorchard:
- No more SRI for fira.css
- No more SRI for Pontoon JS

Fixes #1134

### DIFF
--- a/testpilot/frontend/templates/testpilot/frontend/index.html
+++ b/testpilot/frontend/templates/testpilot/frontend/index.html
@@ -3,9 +3,7 @@
 <head>
     <meta charset="utf-8" />
     <link rel="shortcut icon" href="{{ static('images/favicon.ico') }}">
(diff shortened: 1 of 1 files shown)


lmorchard:
This is a bit of a compromise tweak. SRI is still on the site, but only for the resources we control and generate hashes for from our own server. Those bits haven't been problematic, just the 3rd party resources.

PR #1177, opened 2016-08-03
update docs for new Docker for Mac

clouserw:


README.md
```diff
@@ -51,44 +51,23 @@ See some of our [Tips and Tricks](docs/README-DOCKER.md).
 
 #### First Thing's First
```
jaredhirsch: @clouserw While we're in here, could this be corrected to `First Things First`?

jaredhirsch:
I'll take a look. Docker on Mac is really exciting

lmorchard:
Might also want to toss a reference in there to [the docs to transition from Docker Toolkit to Docker for Mac](https://docs.docker.com/docker-for-mac/docker-toolbox/), if only at the end for us folks who already have the old-style setup running

README.md
ghost: will fix

ghost:
New patch w/ comments addressed

jaredhirsch:
Finally got Docker Toolbox uninstalled, new instructions work for me :tada:

jaredhirsch:
Merging. Followup question: should we get rid of `docs/README-DOCKER.md`?

lmorchard:
> Followup question: should we get rid of docs/README-DOCKER.md

I don't think so... though it could be better written or folded in elsewhere. It has a lot of useful info, all still relevant to Docker for Mac.

PR #1179, opened 2016-08-03
Docker local development improvements

lmorchard:
- Reformat docs/README-DOCKER.md to make tips more easily linkable
- Add a 15 second delay to server bin/run-dev.sh to try to compensate
  for postgres warmup and frontend_watcher initial build time
- Move frontend_watcher to much smaller Alpine linux, upgrade node to
  v6.2.0 from Alpine package
- Speed up server build by caching Alpine package installation
  separately from Python dependency installation
- Add comments to docker-compose-base.yml for optional base images to
  kickstart the process for local dev

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,16 @@ RUN addgroup -g 10001 app && \
 # https://github.com/python-pillow/Pillow/issues/1763#issuecomment-204252397
 ENV LIBRARY_PATH=/lib:/usr/lib
 
+# Install & cache packages for alpine
+RUN apk --no-cache add ca-certificates postgresql-dev build-base libjpeg-turbo-dev zlib-dev
+
 # Install & cache dependencies for Django/Python using pip8
 COPY requirements.txt /app/requirements.txt
-RUN apk --no-cache add ca-certificates postgresql-dev build-base libjpeg-turbo-dev zlib-dev && \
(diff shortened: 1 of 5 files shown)


lmorchard:
Two things:
- There's a V8 deprecation warning on starting up `frontend_watcher` - I think that's fine for now, and the gulp build works fine. We probably just need to upgrade Node dependencies. (Worth a follow-up issue)
- Not sure the commented-out lines in `docker-compose-base.yml` are the best way to pass this along. But, if you really don't want to build fresh images, those lines will pull down ready-built images from Docker Hub and get things started faster.
  
  Those images might be stale, though, if a) what you're working on has new dependencies and b) I haven't gotten around to updating `lmorchard/testpilot_frontend_watcher` (very likely). But it could be interesting to experiment with as a faster kickstart. Could also be handy to rework our Circle CI runs to also build & use a frontend_watcher image and push to Docker Hub regularly.

lmorchard:
Oh also, if your `frontend_watcher` build fails, swapping those lines in docker-compose-base can get you set up with a build that succeeded. So, that could be worth playing with

jaredhirsch:
Oh, this is awesome. I'll have a look

jaredhirsch:
@lmorchard frontend_watcher actually seems to be working for me, and the docs look much nicer. merging :beers:

PR #1180, opened 2016-08-03
Initialize add-on toolbar last clicked timestamp to now

### LINKED ISSUE #1170: `new` badge on doorhanger, but no new experiments.
From [TXP-3](https://trello.com/c/4PNAwpwQ/3-new-experiment-add-on-ui)

> It should trigger a toolbar badge whenever a new experiment is launched

I see the "New" badge on the toolbar when I bring up Firefox after this morning's deployment. We did not launch a new experiment, so we should not be showing the toolbar badge.

/cc @johngruen 

References from chat:

> `lorchard`> If that's not what we want, we can change this 1 to a Date.now() so it doesn't do that https://github.com/mozilla/testpilot/blob/master/addon/lib/toolbar-button.js#L150
> `lorchard`> It currently should show up as new whenever there's a created/modified date newer than the last click on the toolbar button but not older than 2 weeks
> `lorchard`> Oh wait, no, just created time not modified
> `lorchard`> That's what it does now, but it also sets the last-clicked to 1 on initial install meaning you'll see "new" on first update

What's "first update" in this context?

/edit And first update is update of test pilot since the badge feature was added

lmorchard:
Should result in no "New" badge on initial install and only show the badge when things are created after then

Fixes #1170

### DIFF
--- a/addon/lib/toolbar-button.js
+++ b/addon/lib/toolbar-button.js
@@ -145,9 +145,7 @@ const ToolbarButton = module.exports = {
 
     // Initialize the last button click timestamp if necessary.
     if (!('toolbarButtonLastClicked' in store)) {
-      // HACK: Set this to 1 so there's a value, but one that will initially be
-      // less than the current time.
-      store.toolbarButtonLastClicked = 1;
+      store.toolbarButtonLastClicked = Date.now();
     }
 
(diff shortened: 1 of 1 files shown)


dannycoates:
lgtm 👍

PR #1181, opened 2016-08-03
Add a legacy user migration page. Fixes #1079

### LINKED ISSUE #1079: Build the legacy user migration page
As part of our migration from the legacy Test Pilot we need to build a simple landing page which we show before we remove the old add-on.

You can see the spec at https://mozilla.invisionapp.com/share/Q77S1H6ZC#/screens/169034030
- The "Check out Test Pilot" button should just link to the home page
- The "Check out Test Pilot" button should be instrumented to determine how many people hit the download funnel.  I think this can be done simply by making the download button a new referral name in our stats (similar to a snippet, or whatever).  /cc @mattgrimes to say if that will work, and if so, to give the name of it

clouserw:
Not ready to land, but r?

This is my first time in these internals so making an early PR for review.  Two things are left:

1) The header only works when I have the add-on installed (which people looking at this page won't have).  Loading the page in safari or private browsing mode shows the page with no header.  The other pages all seem to be a series of special cases so they are hard to copy directly. :)

2) The spacing of the paragraphs are wrong.  @johngruen said he wanted to come up with some fancy CSS.

The spec is at https://mozilla.invisionapp.com/share/Q77S1H6ZC#/screens/169034030

### DIFF
--- a/testpilot/frontend/static-src/app/lib/page-manager.js
+++ b/testpilot/frontend/static-src/app/lib/page-manager.js
@@ -5,6 +5,7 @@ import ErrorPage from '../views/error-page';
 import ExperimentPage from '../views/experiment-page';
 import ExperimentListPage from '../views/experiment-list-page';
 import LandingPage from '../views/landing-page';
+import LegacyPage from '../views/legacy-page';
 import NotFoundPage from '../views/not-found-page';
 import OnboardingPage from '../views/onboarding-page';
 import RetirePage from '../views/retire-page';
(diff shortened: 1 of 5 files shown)


johngruen:
@clouserw I made a style PR into this branch

ghost:
Thanks.  This looks great and is ready for review

lmorchard:
Looks good to me. Could use a frontend test to assert existence of the page, but meh

PR #1182, opened 2016-08-03
Change the toolbar button icon based on theme color

### LINKED ISSUE #719: Test pilot icon appears white

dannycoates:
this should fix #719 

p.s. I also snuck in changing it to a ToggleButton 🍌

### DIFF
--- a/addon/lib/toolbar-button.js
+++ b/addon/lib/toolbar-button.js
@@ -28,30 +28,31 @@ let panel;
 let collapsed;
 let prefs;
 
-function themeChanged() {
-  setTimeout(setButton);
+function backgroundChanged() {
+  setTimeout(setButtonIcon);
 }
 
-function isLightTheme() {
-  const chromeWindow = viewFor(windows.activeWindow);
-  return !chromeWindow.document.getElementById('nav-bar').matches('[brighttext]');
+function isLightBackground() {
+  const w = viewFor(windows.activeWindow);
+  let el = w.document.getElementById('toggle-button--testpilot-addon-testpilot-link');
+  if (!el) { return true; }
+
+  while (el.parentNode) {
+    if (el.matches('[brighttext]')) {
+      return false;
+    }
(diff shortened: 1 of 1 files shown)


addon/lib/toolbar-button.js
```diff
@@ -23,11 +26,22 @@ let settings;
 let button;
 let panel;
 let collapsed;
+let prefs;
 
-function setActionButton(dark) {
-  const iconPrefix = dark ? './icon-inverted' : './icon';
+function themeChanged() {
+  setTimeout(setButton);
```
dannycoates: weirdly, without the `setTimeout` it sometimes (always?) picks up the style _before_ the change is applied

addon/lib/toolbar-button.js
```diff
@@ -23,11 +26,22 @@ let settings;
...
+
+function isLightTheme() {
+  var chromeWindow = viewFor(windows.activeWindow);
+  return !chromeWindow.document.getElementById('nav-bar').matches('[brighttext]');
+}
 
-  button = ActionButton({ // eslint-disable-line new-cap
+function setButton() {
+  const iconPrefix = isLightTheme() ? './icon' : './icon-inverted';
+  if (button) { button.destroy(); }
```
dannycoates: the console was complaining about something already existing when `setButton` ran a second time. This `destroy` made it go away.

dannycoates:
This doesn't work yet when the button is moved into the hamburger menu. I'm looking into it now

dannycoates:
> This doesn't work yet when the button is moved into the hamburger menu

It sort of works now. It ends up with the right shade, but while in customize mode its wrong for dark themes. IDK if there's much else I can do without using an overlay...

lmorchard:
Seems to work for me

PR #1186, opened 2016-08-04
Implements new metrics ping format.

### LINKED ISSUE #1058: reformat our metrics ping 🚵🏾
We have a new metrics ping defined which will include more useful information in a more precise format and also give us more knobs to turn when we're scaling (the format supports batching).

I patched our documentation with the new definition at https://github.com/clouserw/testpilot/compare/master...metricsv2 but am struggling to write the actual patch so I'm filing this.

chuckharmston:
r? @lmorchard @clouserw 

Of note:
- We're no longer pinging on any intervals, so I got to remove big chunks of code.
- ~~This does not change the experiment ping format, which is more complicated. We need experiments to report additional data that they don't currently, so we may need to coordinate efforts.~~
- It's currently reporting timestamps as floats to maintain millisecond precision while still reporting seconds. Do they need to be integers?

Closes #1058.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -348,11 +348,11 @@ exports.main = function(options) {
     store.clientUUID = require('sdk/util/uuid').uuid().toString().slice(1, -1);
   }
 
+  Metrics.init();
   if (reason === 'install' || reason === 'enable') {
     Metrics.onEnable();
   }
 
-  Metrics.init();
   WebExtensionChannels.init();
   ToolbarButton.init(settings);
   ExperimentNotifications.init();
--- a/addon/lib/metrics.js
+++ b/addon/lib/metrics.js
@@ -10,7 +10,7 @@ const {Cu} = require('chrome');
 
 Cu.import('resource://gre/modules/TelemetryController.jsm');
 
-const { setTimeout, clearTimeout } = require('sdk/timers');
+const { AddonManager } = require('resource://gre/modules/AddonManager.jsm');
 const Events = require('sdk/system/events');
 const PrefsService = require('sdk/preferences/service');
 const self = require('sdk/self');
@@ -23,20 +23,15 @@ const EVENT_SEND_METRIC = 'testpilot::send-metric';
 const EVENT_RECEIVE_VARIANT_DEFS = 'testpilot::register-variants';
 const EVENT_SEND_VARIANTS = 'testpilot::receive-variants';
 
-// Minimum interval of time between main Telemetry pings in ms
-const PING_INTERVAL = 24 * 60 * 60 * 1000;
-
-// Interval between Telemetry ping time checks, more frequent than the actual
-// ping so we can play catch-up if the computer was asleep at the ping time.
-const PING_CHECK_INTERVAL = 10 * 60 * 1000;
-
 // List of preferences we'll override on install & restore on uninstall
 const PREFERENCE_OVERRIDES = {
   'toolkit.telemetry.enabled': true,
   'datareporting.healthreport.uploadEnabled': true
 };
 
-let pingTimer = null;
+// nsIObserver message subjects.
+const TELEMETRY_TESTPILOT = 'testpilot';
+const TELEMETRY_EXPERIMENT = 'testpilottest';
 
 const variantMaker = {
   makeTest: function(test) {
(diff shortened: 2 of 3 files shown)


addon/lib/metrics.js
```diff
@@ -65,90 +59,70 @@ const variantMaker = {
...
-      store.lastTelemetryPingTimestamp = false;
-      store.telemetryPingPayload = {
-        version: 1,
-        tests: {}
-      };
-    }
-
-    Metrics.maybePingTelemetry();
-
+    store.browserLoadedTimestamp = Date.now();
```
chuckharmston, line 69: So, we're supposed to record times based on when the browser was loaded. I'm not sure that's possible from an addon; the best I was able to figure out is the amount of time since the add-on was loaded, which will be roughly equivalent in most cases (but if the user installed or enabled the testpilot addon since they launched the browser, it will be the amount of time since that event).

Thoughts? @ilanasegall @rjweiss

addon/index.js
```diff
@@ -439,12 +439,12 @@ exports.main = function(options) {
     store.clientUUID = require('sdk/util/uuid').uuid().toString().slice(1, -1);
   }
 
+  Metrics.init();
```
chuckharmston, line 351: Moved this up to make sure we logged `browserLoadedTimestamp` before running `onEnable`. Possibly want to revert this, pending the outcome of [this comment](https://github.com/mozilla/testpilot/pull/1186/files#r73460377).

addon/lib/metrics.js
```diff
@@ -65,90 +59,71 @@ const variantMaker = {
...
     }
   },
 
   destroy: function() {
-    // Stop the ping timer, if any
-    if (pingTimer) {
-      clearTimeout(pingTimer);
-    }
     Events.off(EVENT_SEND_METRIC, Metrics.onExperimentPing);
+    Events.off(EVENT_RECEIVE_VARIANT_DEFS, Metrics.onReceiveVariantDefs);
```
chuckharmston, line 106: Drive-by fix in a separate commit.

addon/lib/metrics.js
```diff
@@ -65,90 +59,71 @@ const variantMaker = {
...
+    Metrics.pingTelemetry(self.id, 'disabled', Date.now());
+    Metrics.prefs.restore();
+  },
+
+  prefs: {
+    // Backup existing preference settings and then override.
+    backup: function() {
+      store.metricsPrefsBackup = {};
       Object.keys(PREFERENCE_OVERRIDES).forEach(name => {
-        PrefsService.set(name, store.metricsPrefsBackup[name]);
```
chuckharmston, line 95: This is a weird diffing artifact; in the pre-refactor version of this, the order was the same.

chuckharmston:
@clouserw Who on the Insights side would be best to review this?

ghost:
This looks really clean.  I like.

addon/lib/metrics.js
```diff
@@ -185,7 +156,7 @@ const Metrics = module.exports = {
     };
 
     TelemetryController.submitExternalPing(
-      pingType, payload,
```
ghost, line 188: It won't let me comment on the line, but right above here you have "version: 1" -- it's supposed to be the version of the add-on.  Is that something we can get from our end, or does the add-on have to pass that up to us?

chuckharmston, line 188: This is for the experiment ping, which I didn't really touch in this PR, noted in the pull request description. The only changes here were moving the `pingType` into a constant to match the style of the main add-on ping.

~~When we change the experiment ping, we could possibly keep the version here, but it may make more sense for add-ons to ship the version of the payload that they are using.~~

Edit: dur, version of the _experiment_, not the _schema_. I think experiments would need to ship that as part of the message, but I don't have my head around all the various APIs available to SDK add-ons. I know that `AddonManager` provides interfaces for introspecting installed add-ons, but that might not be worth the overhead introduced.

addon/lib/metrics.js
```diff
@@ -65,90 +59,71 @@ const variantMaker = {
...
-      Metrics.maybePingTelemetry,
-      PING_CHECK_INTERVAL
-    );
-  },
-
-  pingTelemetry: function() {
+  pingTelemetry: function(object, eventName, eventTimestamp) {
+    const payload = {
+      timestamp: (Date.now() - store.browserLoadedTimestamp) / 1000,
+      test: self.id,
```
ghost, line 112: I'm missing it in my docs, but similar to the experiment ping, this is supposed to have a `version` value too.  Please can you add:

```
`version: self.version`
```

sorry for missing that in the docs

chuckharmston, line 112: ☑️ 

The docs aren't merged yet so I can't update it there; you'll have to do that.

ghost, line 112: roger that

addon/lib/metrics.js
ghost, line 69: I & R's opinions are better than mine for this, but what you have here seems reasonable to me.

addon/lib/metrics.js
```diff
@@ -160,34 +137,33 @@ const Metrics = module.exports = {
...
-    if (store.experimentVariants && subject in store.experimentVariants) {
-      dataParsed.variants = store.experimentVariants[subject];
-    }
-
-    const payload = {
-      version: 1,
-      test: subject,
-      payload: dataParsed
-    };
+    AddonManager.getAddonByID(subject, addon => {
```
chuckharmston, line 156: There's a problem here: if this ping happens in an experiment's `unload` handler, this function will return `null` in the lookup. If add-ons are using the event-driven model appropriately this shouldn't be a problem, but Tab Center [works on a schedule](https://github.com/bwinton/TabCenter/blob/1e95d534c4bf37c5a66cb8aa85ea1459cfee6982/main.js#L98) right now, and [forces a flush](https://github.com/bwinton/TabCenter/blob/1e95d534c4bf37c5a66cb8aa85ea1459cfee6982/main.js#L102-L104) on the `unload` handler. This will cause `payload.version` to be null in the latter case here.

CC @clouserw, @bwinton

chuckharmston, line 156: /me spent way too long figuring out why it was always `null` for Tab Center.

bwinton, line 156: I would be up to switching Tab Center to an event-driven model if that makes people's lives easier.  (I started off batching because we thought it would be nicer on the Test Pilot side, but now that Test Pilot has to handle events, perhaps using them in Tab Center wouldn't be so bad.)

ghost, line 156: @bwinton - I was talking to @ilanasegall the other day and she was not a fan of our batching in tab center either.  She says she's working around it on the server side, but if we can change it, I imagine it'd be good.  If she'd rather not mess with it, we could skip it also and just not have a version for tab center.  

@chuckharmston - so, worst case scenario if someone implements the metrics poorly is that version is blank?  The pings still happen other than that?  If that's the case, I think we're fine with this.  We're working with developers so, just like any other API of ours, if they mis-code it we can adjust the experiment.  We should write that in the docs somewhere though.

ghost:
I missed the top-level question about floats vs integers.  I think integers are fine (unsigned if that's a thing).  We've agreed on no milliseconds

lmorchard:
This is looking pretty good to me, but don't want to merge if we're waiting on telemetry folks' input. Will take a look again on Monday though!

chuckharmston:
> I missed the top-level question about floats vs integers. I think integers are fine (unsigned if that's a thing). We've agreed on no milliseconds

Updated to use integers.

addon/lib/metrics.js
```diff
@@ -62,93 +57,81 @@ const variantMaker = {
...
   },
 
   onDisable: function() {
-    // Restore previous preference settings before override.
-    if (store.metricsPrefsBackup) {
+    Metrics.pingTelemetry(self.id, 'disabled', Date.now());
+    Metrics.prefs.restore();
+  },
+
+  prefs: {
```
lmorchard, line 84: Are either of these `Metrics.prefs.{backup,restore}` methods called outside this module? If so, they could just be module-private functions like `makeTimestamp`

addon/index.js
lmorchard, line 351: "this comment" doesn't seem to go anywhere, now... What was the feedback on this? Seems weird to start from time zero on every browser restart and upgrade of the main add-on

lmorchard:
Had an ignorable refactoring nitpick and a question about the timestamp resetting to zero on every restart / upgrade. Beyond that, looks like this needs a rebase but otherwise looks okay to me

ghost:
The timestamp resetting on every restart is how the rest of telemetry works.  We're just being consistent (and also confirmed with Ilana that this won't affect her analysis negatively).

lmorchard:
> The timestamp resetting on every restart is how the rest of telemetry works

Ahh, okay. Then @chuckharmston, feel free to self-merge after a rebase against master! r+ from me

PR #1187, opened 2016-08-04
Recommends DevPrefs Add-on for local development.

chuckharmston:
This prevents the need for manual prefs-setting, and importantly also enables `extensions.logging.enabled` both `dom.report_all_js_exceptions`. Without that, unfortunate engineers like me are sent into a tailspin trying to figure out why exceptions and console calls are vanishing.


chuckharmston:
As always, thanks to @Noitidart for the awesome add-on! `^.^`

Noitidart:
Thanks @chuckharmston for the note! :)

PR #1190, opened 2016-08-04
Update list of installed experiments after adding or removing an experiment

### LINKED ISSUE #723: Update doorhanger `Active` state on enablement without a refresh.
STR:
- Install Experiment.
- Click on Test Pilot icon while staying in the same tab (don't refresh).

Expected:
- Doorhanger should indicate that the newly enabled experiment is "Active".

Actual Result:
- Experiment is not indicated as "Active".
- Refreshing web page fixes the issue.

not active

jaredhirsch:
Fixes #723.


lmorchard:
Seems to work for me

Issue #1159, opened 2016-07-29
Onboarding new tab page should use correct env URL


jaredhirsch:
This is a [TODO](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L157) from [PR 1155](https://github.com/mozilla/testpilot/pull/1155#discussion_r72857565):

Rather than hard-code the production test pilot URL in the `openOnboardingTab` method, we should wait until the settings data is ready, then open the onboarding tab using the correct server environment base URL.

lmorchard:
Fixed by https://github.com/mozilla/testpilot/commit/06f2605126949f942384fca5d6f8765441cc2cb4

Issue #1021, opened 2016-06-27
Test Pilot logging errors when other extensions are installed


bobsilverberg:
This is happening in Firefox 48, but does not seem to be happening in Firefox 49 and 50. When installing a web extension addon, such as the sample attached (change the extension from `zip` to `xpi` to install), errors are logged to the console, such as:

```
1467061502401   addons.manager  WARN    InstallListener threw exception when calling onNewInstall: TypeError: app is undefined (resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-addon/index.js:536:5) JS Stack trace: installListener.onNewInstall@index.js:536:5 < AddonManagerInternal.callInstallListeners@AddonManager.jsm:1716:15 < this.AddonManagerPrivate.callInstallListeners@AddonManager.jsm:2976:12 < AddonInstall.prototype.initLocalInstall/</<@XPIProvider.jsm:5319:11 < makeSafe/<@XPIProviderUtils.js:168:17 < getRepositoryAddon@XPIProviderUtils.js:149:5 < this.XPIDatabase.getAddon/<@XPIProviderUtils.js:1119:9 < Handler.prototype.process@Promise-backend.js:937:23 < this.PromiseWalker.walkerLoop@Promise-backend.js:816:7 < this.PromiseWalker.scheduleWalkerLoop/<@Promise-backend.js:750:11
1467061519962   addons.manager  WARN    InstallListener threw exception when calling onInstallStarted: TypeError: app is undefined (resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-addon/index.js:533:5) JS Stack trace: installListener.onInstallStarted@index.js:533:5 < AddonManagerInternal.callInstallListeners@AddonManager.jsm:1716:15 < this.AddonManagerPrivate.callInstallListeners@AddonManager.jsm:2976:12 < AddonInstall.prototype.startInstall@XPIProvider.jsm:6056:10 < AddonInstall.prototype.install@XPIProvider.jsm:5380:7 < AddonInstallWrapper.prototype.install@XPIProvider.jsm:6439:5 < gXPInstallObserver.showInstallConfirmation/options.eventCallback/this.acceptInstallation@browser-addons.js:146:15 < oncommand@browser.xul:1:1
```

[bug1281030test-1.0.zip](https://github.com/mozilla/testpilot/files/335837/bug1281030test-1.0.zip)

dannycoates:
[...]

PR #1195, opened 2016-08-05
Extract PageMod bits of addon into a new module

dannycoates:
This is a heavyweight fix for #1021. I apologize for the huge diff. 

The main problem in #1021 is that the `installListener` functions could run either before `app` was initialized [1] or the worker port for the pagemod had been detached [2]. The quick solution would be to check if `app` is defined and that there's still a worker on the pagemod. Since I felt like I had time I opted to try a less hacky approach. Hopefully I didn't get _too_ carried away.

The gist of this patch is
- extract the pagemod bits into its own module `app.js` to make changing envs cleaner
- initialize `app` to the most likely env on startup
- set the `active` flags correctly without refreshing from the server
- trigger the onboarding tab after the final env is known

Which is _way_ more stuff than required for #1021 so I'm ok with undoing anything that is too far.

(questions and comments coming inline)

[1] - Example: a fresh browser start then install an addon without ever visiting testpilot
[2] - Example: visited testpilot, closed the tab, then install an addon

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -6,41 +6,27 @@
 
 const EXPERIMENT_UPDATE_INTERVAL = 12 * 60 * 60 * 1000; // 12 hours
 
-const settings = {};
-
-let setInstalledFlagPageMod;
-let messageBridgePageMod;
-let app;
-let experimentsUpdateTimer = null;
-
-const {Cc, Ci, Cu} = require('chrome');
-
+const { Cc, Ci, Cu } = require('chrome');
 const AddonManager = Cu.import('resource://gre/modules/AddonManager.jsm').AddonManager;
-
 const cookieManager2 = Cc['@mozilla.org/cookiemanager;1']
                        .getService(Ci.nsICookieManager2);
 
 const self = require('sdk/self');
 const store = require('sdk/simple-storage').storage;
-const {PageMod} = require('sdk/page-mod');
 const tabs = require('sdk/tabs');
 const request = require('sdk/request').Request;
 const simplePrefs = require('sdk/simple-prefs');
 const URL = require('sdk/url').URL;
 const history = require('sdk/places/history');
-const {setInterval, clearInterval} = require('sdk/timers');
-
-const Mustache = require('mustache');
-const templates = require('./lib/templates');
-Mustache.parse(templates.installed);
-Mustache.parse(templates.experimentList);
-
 const Metrics = require('./lib/metrics');
 const survey = require('./lib/survey');
 const WebExtensionChannels = require('./lib/webextension-channels');
 const ToolbarButton = require('./lib/toolbar-button');
 const ExperimentNotifications = require('./lib/experiment-notifications');
+const { App } = require('./lib/app');
 
+const settings = {};
+let app;
 // Canned selectable server environment configs
 const SERVER_ENVIRONMENTS = {
   local: {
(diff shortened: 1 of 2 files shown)


addon/lib/app.js
```diff
@@ -0,0 +1,91 @@
...
+      contentScriptWhen: 'start',
+      attachTo: ['top', 'existing'],
+      contentScriptOptions: {
+        version: options.addonVersion
+      }
+    });
+    this.loadExperimentList();
+  },
+  loadExperimentList: function() {
+    const r = new Request({
```
dannycoates, line 55: `requestAPI` in `index.js` does stuff for CSRF that I'm not doing here because I tried it and it worked without it. Is this ok?

addon/index.js
```diff
@@ -145,106 +141,50 @@ function initServerEnvironmentPreference() {
...
-            }, store.availableExperiments[addon.id]);
-          });
-    return store.installedAddons;
+          .forEach(setAddonActiveState);
   });
 }
 
+function setAddonActiveState(addon) {
+  const xp = store.availableExperiments[addon.id];
+  if (xp) { delete xp.active; }
```
dannycoates, line 177: I should comment this as a HACK. Because of the way `toolbar-button.js` works the active flag gets "remembered" too long. This corrects it.

addon/index.js
```diff
@@ -145,106 +141,50 @@ function initServerEnvironmentPreference() {
     if (lastVisitedName && lastVisitedName !== currName) {
       // Switch to the last visited environment.
       simplePrefs.prefs.SERVER_ENVIRONMENT = lastVisitedName;
+      updatePrefs();
```
dannycoates, line 144: Moved this call to here instead of always running at the end of this function because `updatePrefs` has already ran with the `currName` env in `main`.

addon/index.js
```diff
@@ -145,106 +141,50 @@ function initServerEnvironmentPreference() {
...
-    });
-
-    // Update UI since we just updated experiments.
-    ExperimentNotifications.maybeSendNotifications();
-    ToolbarButton.updateButtonBadge();
-
-    // Query all installed addons
-    return new Promise(resolve => AddonManager.getAllAddons(resolve));
-  }).then(addons => {
+function experimentsLoaded(experiments) {
```
dannycoates, line 163: This replaces the second part of `updateExperiments` and is triggered by the `app` 'loaded' event.

lmorchard:
I'll take a look through this stuff today, too.

FWIW, we've also got issue #610 for refactoring all the cluttered mess in `index.js` out into separate modules. So, this looks like it should be a great step along the way to closing that!

addon/lib/app.js
lmorchard, line 55: You might try seeing what happens when you enable / disable an experiment. We might need the CSRF stuff to enable the [`DELETE` / `PUT` requests we send to the API for counting experiment installs](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L299). 

But, I can't remember if we need CSRF for those. This might be a throwback to when we had FxA and needed to POST to an `/auth/logout` URL to log the user out.

dannycoates, line 55: So, yeah, those work without CSRF headers against prod as well.

addon/index.js
```diff
@@ -145,106 +141,50 @@ function initServerEnvironmentPreference() {
...
-    updatePrefs();
   });
 }
 
 function openOnboardingTab() {
-  // TODO: Settings is populated async, so we need a global afterSettingsReady
-  // promise that this function can wait on.
   tabs.open({
-    url: SERVER_ENVIRONMENTS.production.BASE_URL + '/onboarding',
+    url: settings.BASE_URL + '/onboarding',
```
lmorchard, line 158: This also fixes #1159, FWIW

addon/lib/app.js
lmorchard, line 55: Oh wait, this is just doing the GET for /api/experiments and has nothing to do with the DELETE / PUT stuff - derp, never mind, this doesn't need CSRF headers at all.

lmorchard:
Finally got around to building & trying this out after skimming the code. Everything seems to work

addon/lib/app.js
dannycoates, line 55: Just FYI, the installation requests also seem to work without CSRF 😄

dannycoates:
Thanks for the quick review @lmorchard 🍻

PR #1199, opened 2016-08-05
move stars out of templates and onto server

### LINKED ISSUE #1194: Let's move all of the blue/stars into the server side page that loads ampersand
We should also add a loader in there that gets nuked when the ampersand app fires up. This will do wonders for perceived performance.

johngruen:
Fixes #1194


lmorchard:
Works for me! The background & new loading spinner look way better than just the default white page

Issue #741, opened 2016-05-09
Should the production add-on be modifiable to point to non-prod endpoints


johngruen:
Since we're likely to break other endpoints and do experiments with them, we might not want to make them discoverable from the add-on. Maybe we could add some kind of flag in the build process on prod to disable the endpoint select list.

lmorchard:
FWIW, there is no build process specifically for prod - the same server image & add-on build gets deployed to all environments.

But, we've got #581 that tries to set the env to the last-visited site after installation. I was thinking as a follow up to that, we might be able to get the add-on to detect & switch automatically to whatever environment you're visiting right then & there - rather than only at startup & installation. Then, the preference drop-down could go away. Maybe have a hidden setting to switch in about:config if you really, really need to lock it to a server for dev purposes

ghost:
Let's make this an about:config pref

lmorchard:
https://github.com/mozilla/testpilot/pull/1201/commits/929dceac5c63f148e1f35dc8016b8fb40ee428dc

PR #1201, opened 2016-08-05
switch env pref to about:config

### LINKED ISSUE #1196: Use a manually-set preference to change Test Pilot add-on environment
The main add-on has two things for dev & testing convenience that are most likely causing problems:
- [The environment selector in `about:addons` settings](https://github.com/mozilla/testpilot/blob/master/addon/package.json#L44) - general users have found it and gotten into unexpected situations.
- [The code that inspects browser history on installation and auto-selects an environment based on the last Test Pilot environment you visited](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L118) - this seems to be changing the environment at undesirable times.

We should replace both of these with a single manually-set pref - either in `about:config` or in a `user.js` file. It could be set before add-on installation and would allow the "Install" button to work properly on respective environments.

dannycoates:
fixes #1196

Uses the `testpilot.env` from about:config to set which server environment to use, defaults to `production`

options:
- local
- dev
- stage
- production


dannycoates:
now with fast env switching!

foo

dannycoates:
also fixes #741

lmorchard:
Works for me! Could use a follow up PR with some docs on how to build / install / use the new Ground Control add-on. Maybe we should also rename it to something like "Test Pilot Ground Control" but that's just a nit

PR #1203, opened 2016-08-07
style button badge with css

dannycoates:
fixes #1165 

Thanks @bwinton for the css tip and technique 🍻

PR #1210, opened 2016-08-08
style button icon with css instead of js

dannycoates:
Follow-up to #1203 

Remove all the js hackery in favor of smooth css


lmorchard:
Works for me!

bwinton:
So, there's a small problem with the CSS when the browser window doesn't have focus…
screen shot 2016-08-05 at 18 39 24
Maybe we can make the selector a little more restrictive in a followup PR?

(Found by @phlsa, reported via Slack.)

lmorchard:
Yeah, let's file a new issue for that. This PR is already merged

SoftVision-PaulOiegas:
Yep we found it too in #1215 . But it's only visible for a few minutes after installation.

PR #1214, opened 2016-08-09
document about:config

dannycoates:
Steps to change the config and links from the READMEs

### DIFF
--- a/docs/ADDON-ENVIRONMENT.md
+++ b/docs/ADDON-ENVIRONMENT.md
@@ -9,6 +9,7 @@ Below are the environments for the website:
 
 | ENVIRONMENT | URL |
 |:------------|:----|
+| Local       | http://testpilot.dev:8000/
 | Development | http://testpilot.dev.mozaws.net/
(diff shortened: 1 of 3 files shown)


docs/ADDON-ENVIRONMENT.md
```diff
@@ -17,12 +18,24 @@ Since the website talks to the add-on, it's important that the two environments
...
+  - `production`
+  - `stage`
+  - `dev`
+  - `local`
+
+## Fast Switcher Addon (Ground Control) Method
+
+1. Ensure you've completed the installation and development steps from [`addon/README.md`](../addon/README.md)
+2. In a terminal go to the `groundcontrol/` directory
+3. Run `npm install` and `npm start`
```
lmorchard: Worth mentioning that this requires use of the [Extension Auto-installer add-on](https://addons.mozilla.org/en-US/firefox/addon/autoinstaller/) that we use with the main add-on. `npm start` won't do anything without it

jaredhirsch:
looks great! thanks, @dannycoates :beers:

PR #1218, opened 2016-08-09
Initial implementation of integration tests

### LINKED ISSUE #352: Add tests for addon / discussion
Would be interested to see some prior art here.
- [ ] Unit tests
- [ ] integration tests
  cc/ @6a68 @lmorchard

lmorchard:
- Add integration testing framework using Marionette
- Test to exercise add-on installation flow
- Refactor circle.yml into small shell scripts to make the main file
  simpler to read
- Upgrade to Node v6.2.0 to match frontend watcher
- Update jpm to 1.1.3 for add-on

Fixes #352

### DIFF
--- a/addon/bin/sign
+++ b/addon/bin/sign
@@ -58,7 +58,7 @@ function distAddon() {
     if (err) return console.error(err);
 
     // sign our add-on
-    var generatedXpi = '@testpilot-addon-' + version + '.xpi';
+    var generatedXpi = 'testpilot-addon.xpi';
 
     signAddon(generatedXpi, function(err, signedXpiPath) {
       if (err) return console.error(err);
--- a/addon/package.json
+++ b/addon/package.json
@@ -1,7 +1,7 @@
 {
   "title": "Test Pilot",
   "name": "testpilot-addon",
-  "version": "0.8.3",
+  "version": "0.8.5",
   "private": true,
   "description": "Test Pilot is a privacy-sensitive user research program focused on getting new features into Firefox faster.",
   "repository": "mozilla/testpilot",
@@ -25,7 +25,7 @@
     "watch-ui": "watchify-server ui-test.js --index ui-test.html",
     "lint": "eslint .",
     "sign": "./bin/update-version && ./bin/sign",
-    "package": "jpm xpi && mv @testpilot-addon-$npm_package_version.xpi ../testpilot/frontend/static-src/addon/addon.xpi && mv @testpilot-addon-$npm_package_version.update.rdf ../testpilot/frontend/static-src/addon/update.rdf",
+    "package": "jpm xpi && mv testpilot-addon.xpi ../testpilot/frontend/static-src/addon/addon.xpi && mv @testpilot-addon-$npm_package_version.update.rdf ../testpilot/frontend/static-src/addon/update.rdf",
     "lint-addon": "addons-linter ../testpilot/frontend/static-src/addon/addon.xpi -o text"
   },
   "license": "MPL-2.0",
@@ -34,7 +34,7 @@
     "babel-eslint": "4.1.8",
     "eslint": "^1.10.3",
     "eslint-config-airbnb": "^0.0.8",
-    "jpm": "1.0.6",
+    "jpm": "1.1.3",
     "jsonwebtoken": "5.7.0",
     "request": "2.69.0",
     "watchify-server": "1.0.2"
--- a/bin/circleci/setup-integration-tests.sh
+++ b/bin/circleci/setup-integration-tests.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -ex
+
+# TODO: Update this occasionally from here:
+# https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds
+FIREFOX_URL=http://archive.mozilla.org/pub/firefox/tinderbox-builds/mozilla-release-linux64-add-on-devel/1470649852/firefox-48.0.1.en-US.linux-x86_64-add-on-devel.tar.bz2
+
+wget -qO $HOME/fx-release.tar.bz2 $FIREFOX_URL
+tar -C $HOME -jxf $HOME/fx-release.tar.bz2
+$HOME/firefox/firefox-bin --version
+
+cd integration
+pip install -r requirements.txt
--- a/circle.yml
+++ b/circle.yml
@@ -8,78 +8,42 @@
 # DOCKER_PASS
 
 machine:
+  hosts:
+    testpilot.dev: 127.0.0.1
   services:
     - docker
 
   node:
-    version: 4.4.0
+    version: 6.2.0
 
 dependencies:
 
   # Keep docker dir and addon dependencies
   cache_directories:
     - "~/docker"
+    - "~/virtualenvs/venv-system"
     - "addon/node_modules"
+    - "node_modules"
 
   pre:
     # https://discuss.circleci.com/t/nvm-commands-are-not-evaluated/943
-    - nvm use 4.4.0 && nvm alias default 4.4.0
+    - nvm use 6.2.0 && nvm alias default 6.2.0
 
   override:
     - docker info
     - node --version
     - npm --version
-
-    # used by npm test for frontend testing
     - google-chrome --version
-
-    # build the container, use circleci's docker cache workaround
-    # only use 1 image per day to keep the cache size from getting
-    # too big and slowing down the build
-    - I="image-$(date +%j).tgz"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; pigz -d -c ~/docker/$I | docker load; fi
-
-    # create a version.json
-    - >
-        printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n'
-        "$CIRCLE_SHA1"
-        "$CIRCLE_TAG"
-        "$CIRCLE_PROJECT_USERNAME"
-        "$CIRCLE_PROJECT_REPONAME"
-        "$CIRCLE_BUILD_URL"
-        > version.json
-
-    # build addon - only sign when on master branch or a tag
-    - cd addon/ &&
-      npm config set spin false &&
-      npm install &&
-      npm run lint &&
-      if [[ $CIRCLE_BRANCH == 'master' || $CIRCLE_TAG != '' ]]; then
-        npm run sign;
-      else
-        npm run package;
-      fi &&
-      npm run lint-addon
-
-    # build static assets
-    - npm install
-    - npm run build
-    - npm test
-
-    # build the actual deployment container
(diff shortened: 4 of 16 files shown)


addon/bin/sign
```diff
@@ -58,7 +58,7 @@ function distAddon() {
     if (err) return console.error(err);
 
     // sign our add-on
-    var generatedXpi = '@testpilot-addon-' + version + '.xpi';
+    var generatedXpi = 'testpilot-addon.xpi';
```
lmorchard, line 61: Seems the newest jpm changes the output .xpi name

addon/package.json
```diff
@@ -25,7 +25,7 @@
     "watch-ui": "watchify-server ui-test.js --index ui-test.html",
     "lint": "eslint .",
     "sign": "./bin/update-version && ./bin/sign",
-    "package": "jpm xpi && mv @testpilot-addon-$npm_package_version.xpi ../testpilot/frontend/static-src/addon/addon.xpi && mv @testpilot-addon-$npm_package_version.update.rdf ../testpilot/frontend/static-src/addon/update.rdf",
+    "package": "jpm xpi && mv testpilot-addon.xpi ../testpilot/frontend/static-src/addon/addon.xpi && mv @testpilot-addon-$npm_package_version.update.rdf ../testpilot/frontend/static-src/addon/update.rdf",
```
lmorchard, line 28: jpm strikes again with filename change here

addon/package.json
```diff
@@ -34,7 +34,7 @@
     "babel-eslint": "4.1.8",
     "eslint": "^1.10.3",
     "eslint-config-airbnb": "^0.0.8",
-    "jpm": "1.0.6",
+    "jpm": "1.1.3",
```
lmorchard, line 37: When I ran these tests on a fresh CircleCI cache, there were issues with the add-on output filename. So, I just went ahead and upgraded jpm and dealt with it

lmorchard:
Only one integration test so far, [exercising initial installation of the add-on through onboarding](https://github.com/mozilla/testpilot/pull/1218/files#diff-e2601bf2d555ee9647801210a15bf0ecR35). Should be an ongoing effort to add some more tests like enabling / disabling experiments, but I wanted to get this stuff into a PR sooner than later and follow up with further work.

lmorchard:
Should hopefully not be too hard to run this test suite locally, if you have a full Docker dev env going. [Comments / critique / etc welcome on the docs I threw together!](https://github.com/lmorchard/testpilot/blob/ce013fae002f9abc805e6f6cbfbe057c0ee203bd/docs/TESTING.md#integration-tests)

lmorchard:
Should also be able to see [how I restructured the CircleCI test runs](https://circleci.com/gh/lmorchard/testpilot/344), which now include a run of these integration tests.

lmorchard:
Rebased against master for good measure

dannycoates:
> Should hopefully not be too hard to run this test suite locally, if you have a full Docker dev env going.

worked for me 💯

jaredhirsch:
Taking a look

addon/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "Test Pilot",
   "name": "testpilot-addon",
-  "version": "0.8.3",
+  "version": "0.8.5",
```
jaredhirsch, line 4: just curious, what happened to 0.8.4?

lmorchard, line 4: Kind of arbitrary... I think we made a few changes to the add-on this sprint, but never got around to bumping the version. We should probably just establish a habit of bumping the add-on version every sprint end

bin/circleci/setup-integration-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -ex
+
+# TODO: Update this occasionally from here:
```
jaredhirsch, line 4: Should we set up a recurring task somewhere to ensure this stays fresh? Maybe as part of an end-of-sprint checklist?

circle.yml
```diff
@@ -8,78 +8,42 @@
...
-    # build the actual deployment container
-    - docker build -t app:build .
-
-    # Clean up any old images and save the new one
-    - I="image-$(date +%j).tgz"; mkdir -p ~/docker; rm ~/docker/*; docker save app:build | pigz --fast -c > ~/docker/$I; ls -l ~/docker
+    - ./bin/circleci/build-version-json.sh
+    - ./bin/circleci/build-addon.sh
+    - ./bin/circleci/build-frontend.sh
+    - ./bin/circleci/build-server.sh
+    - ./bin/circleci/setup-integration-tests.sh
```
jaredhirsch, line 41: :+1: so nice to see these broken out

bin/circleci/setup-integration-tests.sh
jaredhirsch, line 4: Might also be nice to do boring stuff like update pip/npm dependencies on a twice-monthly schedule. Thoughts?

lmorchard, line 4: Might be worth doing that... I didn't want to automate it, so we could maybe pin it to a known version. Might also be interesting someday to expand this into a set of versions to try against, because at some point we'll be spread across different versions that yield different onboarding experiences (e.g. when we start showing up in add-on install whitelist)

integration/__init__.py
```diff
@@ -0,0 +1,23 @@
+from marionette import Marionette, MarionetteTestCase
+from firefox_puppeteer.testcases import BaseFirefoxTestCase
+
+
+class FirefoxTestCase(BaseFirefoxTestCase, MarionetteTestCase):
+
+    # HACK: Replicate setUp from superclass, but just without the navigation to
+    # browser.newtab.url
+    def setUp(self, *args, **kwargs):
```
jaredhirsch, line 9: It would be good to set the element search timeout to a generous number inside this setUp method, so that you can call find_element without having to use as many explicit waits. I think this would do it: `self.marionette.set_search_timeout(10000)`.

I tried looking in the [BaseFirefoxTestCase](https://dxr.mozilla.org/mozilla-central/source/testing/puppeteer/firefox/firefox_puppeteer/testcases/base.py), and it doesn't seem to set the search timeout either. I could be missing it in another base class, or a config file, though.

integration/test_installation.py
```diff
@@ -0,0 +1,79 @@
...
+        FirefoxTestCase.setUp(self)
+
+    def test_initial_installation(self):
+        m = self.marionette
+        b = self.browser
+
+        # Navigate to the site and click the install button
+        with m.using_context(m.CONTEXT_CONTENT):
+            m.navigate(SITE_URL)
+            m.find_element(By.CSS_SELECTOR,
```
jaredhirsch, line 42: This line is the reason I mentioned the search timeout above: it's possible for find_element to give up before the front-end JS has a chance to template the page.

integration/test_installation.py
```diff
@@ -0,0 +1,79 @@
...
+                           'button[data-hook=install]').click()
+            w = Wait(m, ignored_exceptions=NoSuchElementException)
+            w.until(lambda m: m.find_element(By.CSS_SELECTOR,
+                                             'button[data-hook=install].state-change'))
+
+        # Click through the blocked notification
+        b.wait_for_notification(AddOnInstallBlockedNotification)
```
jaredhirsch, line 49: To avoid breakage a few months from now, it would be nice to figure out which version of FF will have the non-blocked notification, then conditionally wait for either the HappyAddOnNotification or the current AddOnInstallBlockedNotification. It should be super easy to check the FF version from marionette.

integration/test_installation.py
```diff
@@ -0,0 +1,79 @@
...
+        # HACK: Things seem to fail intermittently here if we don't wait a tick
+        time.sleep(0.5)
```
jaredhirsch, line 51: Rather than using sleep calls, it's better to either wait for a specific element to appear, become activated, or disappear, or just set a longer implicit wait time. I bet setting a long implicit wait time in the base class would allow this hack to be thrown away.

jaredhirsch:
@lmorchard Looks great! Added a couple of comments about implicit waits vs sleep(), this stuff was drummed into my head by the webQA folks. The thing about sleeps is that, if the VM gets really sluggish, tests will randomly fail; if, on the other hand, we wait a super long time for each element to appear, it can smooth over some of those problems.

R+ from me once the explicit sleep is gone

jaredhirsch:
@meandavejustice We've had two sets of eyes on this, ejecting you from the review ;-)

bin/circleci/setup-integration-tests.sh
lmorchard, line 4: We've also got an issue open to update dependencies, since we're pretty far behind now and a mass upgrade will inevitably break things :(

integration/__init__.py
lmorchard, line 9: Hmm, yeah, I think I'm just using whatever the default is. I'll add that to this base test case though, for our tests

integration/test_installation.py
lmorchard, line 51: Well, that's the weird hack here... The blocked notification _does_ show up, and the allow_button is non-null and actually found. And the `click()` method fires without error. It's just that the confirmation notification never shows up & _that_ search times out. The elements are all there - Firefox just doesn't seem respond to a click on it for at least 0.5 sec.

integration/test_installation.py
lmorchard, line 49: We've got a month or so at least... but I'm not sure what version it will show up in. Would rather get this test in as-is, working with the current version, and iterate further in future PRs. We should probably improve this whole thing to run integration tests against a matrix of Firefox versions & release lines

lmorchard:
> R+ from me once the explicit sleep is gone

I can't get this to pass reliably without the sleep, though. It fails maybe every 3rd or 4th time. The search timeout doesn't seem to help. It seems like all the elements are there, it's just that the "Allow" button isn't responding to clicks immediately after it's found

jaredhirsch:
> I can't get this to pass without the sleep, though. The search timeout doesn't seem to help. It seems like all the elements are there, they're just not yet responding to clicks.

@lmorchard Cool, just land it as-is and file a bug. I'll try to get it into TXP-0 and bang on it myself next sprint.

bin/circleci/setup-integration-tests.sh
jaredhirsch, line 4: :+1:

integration/test_installation.py
jaredhirsch, line 49: > We should probably improve this whole thing to run integration tests against a matrix of Firefox versions & release lines

:100: that sounds awesome

integration/test_installation.py
jaredhirsch, line 51: Oh, interesting. Definitely add that info to the followup bug. I wonder if marionette is clicking the button before the XBL code is ready.

jaredhirsch:
looks like all my comments were addressed, merging

PR #1219, opened 2016-08-09
Update sample data documentation.

chuckharmston:
I was getting odd tracebacks about 50% of the time after running `loaddata`. I get them 0% of the time if restarting after running `loaddata`.

### DIFF
--- a/docs/README-DOCKER.md
+++ b/docs/README-DOCKER.md
@@ -13,11 +13,14 @@ lists some hints & tips we've accumulated through the course of daily work.
 This is necessary for running Django commands, among other things.
 
 ### To load some sample demo content into your Docker dev environment:
+
 ```
 mkdir -p media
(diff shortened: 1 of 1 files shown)


lmorchard:
What were the tracebacks? Might be something else worth fixing.

chuckharmston:
Seems like it was a recurrence of the timing issues we've had previously; the frontend code hadn't finished building yet.

lmorchard:
Hmm, this could be something we want to fix in either Gulp or Django or both. Gulp runs "clean" on every file change, which deletes everything & starts from scratch. Django throws errors when static files linked to using `static()` are missing. So, Django tracebacks while Gulp is building. :(

chuckharmston:
Yeah, exactly.

I'm not sure what a fix would look like, but it certainly isn't this PR.

Issue #1216, opened 2016-08-09
Small problem with the CSS when the browser window doesn't have focus.


bwinton:
Introduced in #1210.

Maybe we can make the selector a little more restrictive?
<img width="103" alt="screen shot 2016-08-05 at 18 39 24" src="https://cloud.githubusercontent.com/assets/159571/17519591/9ec398d6-5e1a-11e6-86f6-91fab0ee3b9a.png">
(Found by @phlsa, reported via Slack.)

johngruen:
@bwinton is this a dupe of #1215

bwinton:
I don't _think_ so, although depending on how you choose to fix that bug, this might no longer be a problem…

(On Mac, the toolbar icons should be greyer when the window isn't focused, so that's expected behaviour for this platform at least…)

dannycoates:
@bwinton are you talking about the blurred text? FWIW badge text has always looked that way when the window is unfocus with the dev edition theme for me. A `text-shadow: none;` seems to fix it.

bwinton:
Yeah, well, partially the blurred text, partially the bold background, mostly the way they interact to make it really hard to read.  😉  Could you post a screenshot of what it looked like before your patch (or on a different button)?

dannycoates:
<img width="73" alt="screen shot 2016-08-09 at 11 08 43 am" src="https://cloud.githubusercontent.com/assets/87619/17527807/b5307c72-5e21-11e6-9920-fd14defaa72b.png">

On the left is with adding `text-shadow: none;` and the right is no custom style with the color set by `button.badgeColor = '#1996F7'` with the dev edition theme.

dannycoates:
i think #1221 fixes this

PR #1221, opened 2016-08-09
Fix button css issues

### LINKED ISSUE #1215: After fresh install, the "Test Pilot" add-on toolbar icon is grayed out if browser window is not focused
[Affected versions]:
- Nightly 51.0a1
- Test Pilot 0.8.3 -tag-2016-07-29

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- Use a clean profile without Test Pilot add-on installed.

[Steps to reproduce]:
1. Go to http://testpilot.dev.mozaws.net and install Test Pilot add-on.
2. Observe the Test Pilot icon from the main toolbar.
3. Click outside the browser window to remove the focus from it.
4. Observe again the Test Pilot icon from the main toolbar.

[Expected result]:
- The Test Pilot icon is displayed in black color even if browser window is focused or not.

[Actual result]:
- The Test Pilot icon is displayed in black color but becomes grayed out when browser window is not in focus.

[Notes]:
- This issue is visible only for a few minutes after installation. 
- I have observed a similar behaviour on Windows 7 x64. After installation, hovering the Test Pilot icon changes its color to gray instead of displaying the hover selection overlay.
- Added a screen recording with the issue https://goo.gl/7aFvHz.

dannycoates:
I hope this fixes #1215 and #1216 

@johngruen I shouldn't be allowed to css without adult supervision

### DIFF
--- a/addon/data/button.css
+++ b/addon/data/button.css
@@ -4,7 +4,7 @@
  * http://mozilla.org/MPL/2.0/.
  */
 
-#toggle-button--testpilot-addon-testpilot-link {
+#toggle-button--testpilot-addon-testpilot-link > .toolbarbutton-badge-stack > .toolbarbutton-icon {
   list-style-image: none !important;
   background-image: url('resource://testpilot-addon/data/icon-16.png') !important;
   background-size: 16px 16px !important;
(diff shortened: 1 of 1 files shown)


lmorchard:
Seems to work for me, but I've also no experience with CSS as applied to XUL. I think we need adults from another species to grok that. So, eh, r+ from me

PR #1224, opened 2016-08-09
Use browser start time for metrics timestamps.

jaredhirsch:
@chuckharmston R? Uses the cleverly concealed `nsIAppStartup` service to get the app startup time.

### DIFF
--- a/addon/lib/metrics.js
+++ b/addon/lib/metrics.js
@@ -4,10 +4,11 @@
  * http://mozilla.org/MPL/2.0/.
  */
 
-/* global TelemetryController */
+/* global TelemetryController, Services */
 
 const {Cu} = require('chrome');
 
+Cu.import('resource://gre/modules/Services.jsm');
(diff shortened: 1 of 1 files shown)


addon/lib/metrics.js
```diff
@@ -66,7 +67,7 @@ function makeTimestamp(time) {
 const Metrics = module.exports = {
 
   init: function() {
-    store.browserLoadedTimestamp = Date.now();
+    store.browserLoadedTimestamp = Services.startup.getStartupInfo().process;
```
chuckharmston: We should probably just not set this, and instead replace `store.browserLoadedTimestamp`'s use in `makeTimestamp` a few lines above.

chuckharmston: For linting reasons, might want to throw a `// noqa` on that line, since `Services` won't be required.

jaredhirsch: sure, works for me

Issue #1131, opened 2016-07-25
Weird layout gap with even number of experiments in a row


pdehaan:
Found in http://testpilot.dev.mozaws.net/__version__ (e9e40d0052b61faa8017abf8f4716dd947d24039):

``` json
{ "commit": "e9e40d0052b61faa8017abf8f4716dd947d24039",
  "source": "https://github.com/mozilla/testpilot",
  "version": "",
  "build": "https://circleci.com/gh/mozilla/testpilot/647" }
```

---
### Actual results:

![firefox_test_pilot](https://cloud.githubusercontent.com/assets/557895/17120953/c54bee5a-5284-11e6-840b-b09b077226d6.png)

ckprice:
Fixed in #1226.

PR #1225, opened 2016-08-10
Add public experiment detail page (closes #419, #1237).

### LINKED ISSUE #419: Create a public details page for tests
This bug is about allowing users who have not signed in or installed the add-on to view experiment detail pages (e.g. https://testpilot.firefox.com/experiments/activity-stream).

/edit removing the original opening, as it was incorrect.

chuckharmston:
Throwing this out there for some early review, because I'm faking it 'til I make it with Ampersand and would appreciate any course-correction.
# Screenshots
## Firefox, with add-on installed

with
## Firefox, without add-on installed

without
## Chrome

chrome
# TODO
- [x] Dynamically determine the y-position at which to fix the header.
- [x] Add listing of additional experiments at the bottom of page for users without the add-on.
- [x] Handle non-ff cases
- [x] Get metrics changes defined, implement them.
- [x] Fix/remove old/broken tests.
- [x] Write new tests for all the introduced variations.

### DIFF
--- a/testpilot/frontend/static-src/app/lib/install-addon.js
+++ b/testpilot/frontend/static-src/app/lib/install-addon.js
@@ -1,0 +1,35 @@
+import app from 'ampersand-app';
+import cookies from 'js-cookie';
+
+
+export default function installAddon(view, eventCategory, callback) {
+  const downloadUrl = '/static/addon/addon.xpi';
+
+  view.query('[data-hook=install]').classList.add('state-change');
+  view.query('.default-btn-msg').classList.add('no-display');
+  view.query('.progress-btn-msg').classList.remove('no-display');
+
+  app.sendToGA('event', {
+    eventCategory: eventCategory,
+    eventAction: 'button click',
+    eventLabel: 'Install the Add-on',
+    outboundURL: downloadUrl
+  });
+
+  cookies.set('first-run', 'true');
+
+  // Wait for the add-on to be installed.
+  // TODO: Should we have a timeout here, give up after a few intervals? If
+  // user cancels add-on install, this will never stop spinning.
+  const interval = setInterval(() => {
+    if (!window.navigator.testpilotAddon) { return; }
+    clearInterval(interval);
+    app.me.fetch().then(() => {
+      callback();
+    }).catch(() => {
+      // HACK (for Issue #1075): Timed out while waiting for the initial sync
+      // message, so just reload the page to stop waiting.
+      window.location.reload();
+    });
+  }, 1000);
+}
--- a/testpilot/frontend/static-src/app/templates/experiment-page.js
+++ b/testpilot/frontend/static-src/app/templates/experiment-page.js
@@ -17,8 +17,8 @@ export default `
 
             <a data-hook="highlight-privacy" class="highlight-privacy" data-l10n-id=highlightPrivacy>Your privacy</a>
             <a data-l10n-id="giveFeedback" data-hook="feedback" id="feedback-button" class="button default" target="_blank">Give Feedback</a>
-            <button data-hook="uninstall" id="uninstall-button" class="button secondary"><span class="state-change-inner"></span><span data-l10n-id="disableExperimentTransition" class="transition-text">Disabling...</span><span data-l10n-id="disableExperiment" class="default-text">Disable <span data-hook="title"></span></span></button>
-            <button data-hook="install" id="install-button" class="button default"><span class="state-change-inner"></span><span data-l10n-id="enableExperimentTransition" class="transition-text">Enabling...</span><span data-l10n-id="enableExperiment" class="default-text">Enable <span data-hook="title"></span></span></button>
+            <button data-hook="uninstall-experiment" id="uninstall-button" class="button secondary"><span class="state-change-inner"></span><span data-l10n-id="disableExperimentTransition" class="transition-text">Disabling...</span><span data-l10n-id="disableExperiment" class="default-text">Disable <span data-hook="title"></span></span></button>
+            <button data-hook="install-experiment" id="install-button" class="button default"><span class="state-change-inner"></span><span data-l10n-id="enableExperimentTransition" class="transition-text">Enabling...</span><span data-l10n-id="enableExperiment" class="default-text">Enable <span data-hook="title"></span></span></button>
           </div>
         </div>
       </div>
--- a/testpilot/frontend/static-src/app/templates/experiment-page.js
+++ b/testpilot/frontend/static-src/app/templates/experiment-page.js
@@ -1,8 +1,7 @@
 export default `
   <section id="details" data-hook="experiment-page">
-    <div class="shifted-stars">
-      <header data-hook="header-view"></header>
-    </div>
+    <header data-hook="header-view"></header>
+    <section data-hook="testpilot-promo"></section>
     <div class="default-background">
       <div class="details-header-wrapper" data-hook="has-status">
         <div class="status-bar" data-hook="status-type">
@@ -14,7 +13,7 @@ export default `
             <h1 data-hook="title"></h1>
             <h4 data-hook="subtitle" class="subtitle"></h4>
           </header>
-            <div class="experiment-controls">
+            <div class="experiment-controls" data-hook="active-user">
 
             <a data-hook="highlight-privacy" class="highlight-privacy" data-l10n-id=highlightPrivacy>Your privacy</a>
             <a data-l10n-id="giveFeedback" data-hook="feedback" id="feedback-button" class="button default" target="_blank">Give Feedback</a>
@@ -23,6 +22,7 @@ export default `
           </div>
         </div>
       </div>
+      <div class="sticky-header-sibling"></div>
 
       <div data-hook="details">
           <div class="details-content content-wrapper">
@@ -35,7 +35,12 @@ export default `
                   <span data-l10n-id="userCountContainer">There are <span data-l10n-id="userCount" class="bold" data-hook="install-count"></span>
                   people trying <span data-hook="title"></span> right now!</span>
                 </section>
-                <section>
+                <div data-hook="inactive-user">
+                  <section data-hook="introduction-container">
+                    <div data-hook="introduction-html"></div>
+                  </section>
+                </div>
+                <section data-hook="active-user">
                   <table class="stats">
                     <tr data-hook="version-container">
                       <td data-l10n-id="version">Version</td>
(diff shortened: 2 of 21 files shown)


chuckharmston:
@johngruen Flagging that this promo is really tall, around 500px; tall enough that I'd be concerned that the user not realize that they're on an experiment page at all (blah blah below-the-fold-blah blah).

testpilot/frontend/static-src/styles/modules/_experiment-list.scss
```diff
@@ -185,3 +178,13 @@
   height: $grid-unit * 4.8;
   opacity: .9;
 }
+
+[data-hook='experiment-page'] {
```
chuckharmston: There don't seem to be any norms around per-usage overrides of default styles, so I wasn't really sure where to put this, or what the most appropriate selector would be. Does this warrant a new file?

testpilot/frontend/static-src/app/views/experiment-list-view.js
```diff
@@ -1,19 +1,32 @@
...
     BaseView.prototype.render.apply(this, arguments);
+    console.log('experiments', app.experiments);
+
+    const filteredExperiments = () => {
+      if (this.except) {
+        const filtered = app.experiments.models.filter(experiment => {
+          return experiment.slug !== this.except;
+        });
+        return new ExperimentsCollection(filtered);
+      }
```
chuckharmston, line 24: This is the only way I could manage to filter a collection: by creating a new collection from a filtered subset of the original one's models. I assume there's an easier way?

chuckharmston:
Updated the PR description with full-page screenshots. Missing tests, but definitely could use review.

@lmorchard @meandavejustice: r? I've annotated a few specific questions I have, but would appreciate an overall look at it.

@johngruen:
- Mind taking a look at visuals? The one obvious quirk is the spacing of additional experiments if there are only two, which is actually https://github.com/mozilla/testpilot/issues/1131. ~~I hope to close that separately before this lands.~~ There's a PR open for that: https://github.com/mozilla/testpilot/pull/1226
- How should we instrument these new variations? Should we be tweaking any existing funnels?

testpilot/frontend/static-src/app/views/experiment-list-view.js
lmorchard, line 24: There is [ampersand-filtered-subcollection](https://github.com/AmpersandJS/ampersand-filtered-subcollection), but I think the approach is generally the same thing just in a packaged module

lmorchard:
JS-wise, I think this looks pretty good. Any confusion or negative reaction you might be having is probably a natural allergy to how tangly our Ampersand stuff has gotten

testpilot/frontend/static-src/app/views/experiment-list-view.js
meandavejustice, line 24: Looking over ampersand-filtered-subcollection and it looks like overkill for what we need anyways. I think what you've got here is rather clear.

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -7,9 +7,16 @@ import DetailView from './detail-view';
...
 import template from '../templates/experiment-page';
 import DisableDialogView from './disable-dialog-view';
+import ExperimentListView from './experiment-list-view';
 import ExperimentTourDialogView from './experiment-tour-dialog-view';
+import TestpilotPromoView from './testpilot-promo-view';
 
-const changeHeaderOn = 127;
+function changeHeaderOn() {
+  const header = document.getElementsByClassName('details-header-wrapper')[0];
+  const content = document.getElementsByClassName('details-content')[0];
```
meandavejustice: There a reason you are using `getElementsByClassName[0]` over `querySelector`? Is it faster?

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -7,9 +7,16 @@ import DetailView from './detail-view';
 import ContributorView from './contributor-view';
 import template from '../templates/experiment-page';
 import DisableDialogView from './disable-dialog-view';
+import ExperimentListView from './experiment-list-view';
 import ExperimentTourDialogView from './experiment-tour-dialog-view';
+import TestpilotPromoView from './testpilot-promo-view';
 
-const changeHeaderOn = 127;
+function changeHeaderOn() {
```
meandavejustice: this might be easier to read as `onHeaderChange`

testpilot/frontend/static-src/styles/modules/_experiment-list.scss
meandavejustice: ccing  @johngruen as he has been owning styles

testpilot/frontend/static-src/app/views/experiment-page.js
chuckharmston: If you're using it to grab something that is ID-like as this is, then yeah, it's faster.

testpilot/frontend/static-src/styles/modules/_experiment-list.scss
johngruen: @chuckharmston I knocked the need for this out in my PR into your fork...didn't see your override here, you can kill it entirely :)

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -217,6 +248,11 @@ export default PageView.extend({
     return this;
   },
 
+  beforeRender() {
+    this.activeUser = app.me.hasAddon;
+    this.isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
```
jaredhirsch, line 266: @chuckharmston To ensure `isFirefox` only flags desktop FF, have a look at [this MDN page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox).

I'm not totally sure, but I think `Gecko/20100101` is unique to desktop foxes.

chuckharmston, line 266: I copy/pasted this from [the landing page's view](https://github.com/mozilla/testpilot/blob/35802081bee163996b6766d8d119de769284e321/testpilot/frontend/static-src/app/views/landing-page.js#L21); worth improving, but probably orthogonal to this PR. I'll file a bug.

chuckharmston, line 266: https://github.com/mozilla/testpilot/issues/1233

chuckharmston:
@johngruen How do we want to track this? It seems like there are at least two new funnels that we want to track for users without the add-on:
1. Homepage -> experiment detail page -> install Test Pilot
2. Homepage -> experiment detail page -> install Test Pilot -> install experiment

testpilot/frontend/static-src/app/views/experiment-page.js
jaredhirsch, line 266: :+1: sounds good

johngruen:
@chuckharmston sounds right.

ckprice:
Metrics will be defined in #1237.

testpilot/frontend/static-src/app/views/experiment-list-view.js
```diff
@@ -1,19 +1,32 @@
...
 
   props: {
-    hasAddon: 'boolean'
+    hasAddon: 'boolean',
+    except: { type: 'string', required: false }
   },
 
   render() {
     BaseView.prototype.render.apply(this, arguments);
+    console.log('experiments', app.experiments);
```
chuckharmston: Get rid of this.

testpilot/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -65,25 +68,33 @@ export default `
...
+                </div>
               </div>
             </div>
 
             <div class="details-description">
-              <section data-hook="introduction-container" class="introduction">
-                <div data-hook="introduction-html"></div>
-              </section>
+              <div data-hook="active-user">
+                <section data-hook="introduction-container active-user" class="introduction">
```
chuckharmston: Don't need `active-user` here.

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -187,27 +196,49 @@ export default PageView.extend({
...
     this.pageTitleL10nID = 'pageTitleExperiment';
+
+    app.me.on('change:hasAddon', this.render, this);
   },
 
   render() {
     PageView.prototype.render.apply(this, arguments);
+
+    if (!this.activeUser) {
+      console.log('model', this.model);
```
chuckharmston: Remove this.

chuckharmston:
Alright, tests have been added and are passing. Now waiting on @ckprice to confirm via #1237 that we don't need any new instrumentation for metrics.

chuckharmston:
Instrumentation added in a few separate commits that I'll squash before merging. r? @lmorchard @ckprice 

@ckprice: everything should be to your spec in #1237, with one addition, as noted [in this comment](https://github.com/mozilla/testpilot/issues/1237#issuecomment-240428888).

@lmorchard: of note, I refactored the add-on install, and couldn't work out how to test the full flow locally.

addon/lib/metrics.js
```diff
@@ -143,31 +161,34 @@ const Metrics = module.exports = {
...
     Events.emit(EVENT_SEND_VARIANTS, {
       data: JSON.stringify(dataParsed),
       subject: self.id
     });
   },
 
   onExperimentPing: function(ev) {
-    const timestamp = Date.now();
     const { subject, data } = ev;
+    const dataParsed = JSON.parse(data);
```
pdehaan: Do we need to wrap `JSON.parse()` in a `try..catch` in case it receives invalid `data`?

chuckharmston: This shouldn't be here, I must have had a rebase go wrong.

testpilot/frontend/static-src/app/lib/install-addon.js
```diff
@@ -0,0 +1,36 @@
...
+    outboundURL: downloadUrl
+  });
+
+  cookies.set('first-run', 'true');
+
+  // Wait for the add-on to be installed.
+  // TODO: Should we have a timeout here, give up after a few intervals? If
+  // user cancels add-on install, this will never stop spinning.
+  const interval = setInterval(() => {
+    console.log(2);
```
pdehaan: We need this valuable `console.log()`?

chuckharmston: ☑️

addon/lib/metrics.js
chuckharmston: Yeah, I did. Should be fixed. ☑️

ckprice:
Pulled this down. Looked at pings being sent through the network console, and everything that was in #1237 seems to be covered. 👍

testpilot/frontend/static-src/app/views/testpilot-promo-view.js
```diff
@@ -0,0 +1,56 @@
...
+                <div class="button-icon-badge"></div>
+              </div>
+              <div class="button-copy">
+                <div data-l10n-id="landingDownloadFirefoxTitle" class="button-title">Firefox</div>
+                <div data-l10n-id="landingDownloadFirefoxSubTitle" class="button-description">Free Download</div>
+              </div>
+            </a>
+          {{/isFirefox}}
+          {{#isFirefox}}
+          <button data-hook="install" class="button extra-large primary">
```
lmorchard, line 30: Since we added another install button, we should add an integration test for it. Add-on installation flow is one of our sensitive spots that's ended up broken a lot in the past month or two. Don't want to block this PR, though, since it's been enough of a journey so far. Filed a follow up issue: https://github.com/mozilla/testpilot/issues/1257

lmorchard:
> @lmorchard: of note, I refactored the add-on install, and couldn't work out how to test the full flow locally.

Might have already covered this in IRC, but:
1. run dev edition with signature enforcement turned off per [addon readme](https://github.com/mozilla/testpilot/blob/master/addon/README.md#development) 
2. run `npm package` in the `addon/` directory to generate an unsigned `.xpi`
3. gulp frontend watcher should detect the new build & move it to the right spot
4. `testpilot.env` needs to be set to `local` in `about:config`

After all that, the local install button should work.

lmorchard:
Ran this through its paces again locally myself, and everything seems to work. :+1: from me - is this waiting on anything else? If not, I'd say it's good to merge!

chuckharmston:
Thanks, Les!

Do the honors? :shipit:

PR #1227, opened 2016-08-10
Add install error statusbar message

### LINKED ISSUE #660: Add a failure state for experiment install

dannycoates:
fixes #660 

This covers download and install errors emitted by the `installListener` from the add-on, which should cover the most common errors. It doesn't cover slow or hung downloads like I described [here](https://github.com/mozilla/testpilot/issues/660#issuecomment-238396305).

There's one deviation from the design, I didn't disable the button on an error. Its not harmful to try clicking it again right then and there, and it might work, so it saves a page refresh or whatever, but I'll add it if we really want it.

Testing it top to bottom is a bit convoluted. I don't know a good way to mock it up yet. So, I go to an experiment page, then turn off wifi, then click the enable button. After about 60s the download fails and displays the error. If you just want to see how the banner looks, you can go to the js console and run something like `app.experiments.models[0].error = true`

### DIFF
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -68,6 +68,7 @@ experimentListEnabledTab = Enabled
 experimentListJustLaunchedTab = Just Launched
 experimentListJustUpdatedTab = Just Updated
 isEnabledStatusMessage = {$title} is enabled.
+installErrorMessage = Uh oh. {$title} could not be enabled. Try again later.
 participantCount = <span>{$installation_count}</span> participants
 
 giveFeedback = Give Feedback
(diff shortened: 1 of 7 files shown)


dannycoates:
i think this is ready for r?

lmorchard:
Taking a look... skimming through the code, things look good. Tests pass locally. Giving it a quick manual spin - looks good 👍

SoftVision-PaulOiegas:
Seems to be working ok. At least I have verified it with bad or no connectivity and the errors are correctly appearing.
One thing only here, if the experiment was already enabled once, even with no internet connection the user can still disable and enable it. I'm not seeing it an issue, but do we want this ?

dannycoates:
> One thing only here, if the experiment was already enabled once, even with no internet connection the user can still disable and enable it. I'm not seeing it an issue, but do we want this ?

I think we do want this. If the experiment is still in the browser cache it will install from there even with no connection. Clearing the cache will force the browser to re-download it.

SoftVision-PaulOiegas:
Fine by me. Indeed it helps in this way.

Issue #1169, opened 2016-08-01
Berberpole animation along with "Installing" text are not displayed for bottom page installation button


SoftVision-CosminMuntean:
[Affected versions]:
- Nightly 50.0a1
- Test Pilot 0.8.3 -tag-2016-07-29

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Steps to reproduce]:
1. Go to https://testpilot.stage.mozaws.net page.
2. From the bottom of the page click on the big green "Install the Test Pilot Add-on" button.
3. Observe the button behavior.

[Expected result]:
- The button changes into "Installing..." and a "berberpole" animation is displayed under the text.

[Actual result]:
- "Berberpole" animation along with "Installing" text are not displayed for bottom page installation button.

[Notes]:
- Added a screen recording with the issue: https://goo.gl/UKYA0J

SoftVision-CosminMuntean:
I have verified this issue and is no longer reproducible.

SoftVision-PaulOiegas:
Seems we have a regression here. Probably because of #1225.
Reopening issue.

johngruen:
Can't repro on prod

PR #1228, opened 2016-08-11
disable install when installing

### LINKED ISSUE #1090: Disable the 'install' button when in the 'installing' state
Very minor polish bug: when the button is in the 'installing' barberpole state, it seems like it shouldn't be clickable.

johngruen:
Fixes #1090 and #1169

### DIFF
--- a/testpilot/frontend/static-src/app/views/landing-page.js
+++ b/testpilot/frontend/static-src/app/views/landing-page.js
@@ -42,10 +42,17 @@ export default PageView.extend({
 
   installClicked() {
     const downloadUrl = '/static/addon/addon.xpi';
+    const installButton = document.getElementsByClassName('install');
+    const defaultMessage = document.getElementsByClassName('default-btn-msg');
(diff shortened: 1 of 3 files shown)


testpilot/frontend/static-src/app/views/landing-page.js
```diff
@@ -42,10 +42,17 @@ export default PageView.extend({
 
   installClicked() {
     const downloadUrl = '/static/addon/addon.xpi';
+    const installButton = document.getElementsByClassName('install');
```
lmorchard, line 45: nit: Should make these lookups with ampersand's `this.query()`. Even better, implement these as view bindings on a boolean property. But, I'll probably end up just merging this as-is because ampersand is annoying

SoftVision-PaulOiegas:
This works as intended but we have a downside here. If you chose to allow the add-on download and when prompted to install you click "Cancel", the installation process cannot be restarted unless you refresh the page.

PR #1234, opened 2016-08-11
Fix error on removing add-on built after signing

lmorchard:
The add-on build in CircleCI completes with an error during clean up. Doesn't break the build, but this PR should fix the error. Not urgent at all for this sprint.


lmorchard:
Tests passed, error's gone, going to self-merge

PR #1238, opened 2016-08-11
fixed some linter warnings

dannycoates:
Make the linter happy


pdehaan:
We should check to see if Circle-CI is running linting on all the JS stuff, or if it only lints the /addon/ directory.

PR #1241, opened 2016-08-12
don't show exp as both new and active in doorhanger

### LINKED ISSUE #1239: The "View all experiments" button is not entirely visible if a new experiment is enabled
[Affected versions]:
- Nightly 51.0a1
- Test Pilot 0.8.5 -tag-2016-08-11

[Affected Platforms]:
- All Windows

[Prerequisites]:
- Test Pilot add-on is installed on a clean profile.
- No new experiments are enabled on the used profile.

[Steps to reproduce]:
1. Click the "Test Pilot" button from browser toolbar.
2. Click on a new experiment (eg: NoMore404s).
3. Click on the "Enable" button for the experiment.
4. After the experiment is enabled click the "Test Pilot" button from browser toolbar and observe the bottom part of the displayed door-hanger.

[Expected result]:
- The list with the available experiments is correctly displayed. 

[Actual result]:
- The "View all experiments" button is not entirely visible.  placed a few pixels downward  

[Notes]:
- This issue occurs because enabled new experiments have three lines of text instead of two which forces the button to move lower then the designed container space.
- Added a screenshot with the issue: 
  view all experiments button

dannycoates:
fixes #1239

### DIFF
--- a/addon/lib/toolbar-button.js
+++ b/addon/lib/toolbar-button.js
@@ -32,10 +32,9 @@ function getExperimentList(availableExperiments, installedAddons) {
     const experiment = availableExperiments[k];
     if (installedAddons[k]) {
       experiment.active = installedAddons[k].active;
-    } else {
-      const created = (new Date(experiment.created)).getTime();
-      experiment.isNew = (now - created) < NEW_EXPERIMENT_PERIOD;
     }
+    const created = (new Date(experiment.created)).getTime();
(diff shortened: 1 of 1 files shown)


SoftVision-PaulOiegas:
Only one status is now correctly displayed in doorhanger. 
However, the "New Experiment" status is re-displayed after disabling the experiment. The issue was addressed in #1242 and as Danny said we would need to keep track of what experiments are enabled/disabled in order to fix this.

PR #1243, opened 2016-08-12
added extensions.webapi.testing pref to env switcher

### LINKED ISSUE #1222: Toggle extensions.webapi.testing when we point somewhere that isn't production
If testpilot.env exists (and/or is not production) we should also set `extensions.webapi.testing` to `true` so that we can install add-ons from dev and stage.

I don't have a way to install add-ons from localhost.  https://dxr.mozilla.org/mozilla-central/source/toolkit/mozapps/extensions/AddonManager.jsm#2906 is the code we're going to run into.

dannycoates:
I don't know if switching the pref is better here or in the addon itself... Since its more of a dev thing I put it here first.

closes #1222

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -56,6 +56,20 @@ const SERVER_ENVIRONMENTS = {
   }
 };
 
+function loadEnvironment() {
+  const env = aboutConfig.get('testpilot.env', 'production');
+  if (!aboutConfig.has('testpilot.env')) {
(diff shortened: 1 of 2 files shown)


lmorchard:
Hmm, might be better to stick this in the main addon. We change the testpilot.env pref in tests and thing

dannycoates:
> Hmm, might be better to stick this in the main addon

sounds good, added.

PR #1244, opened 2016-08-12
Can we remove the CSRF header from addon requests?

dannycoates:
I'm sure this was required at some point, however testing this now against all environments works fine without it. (should it?)


meandavejustice:
seems fine to me. @lmorchard what do you think?

lmorchard:
Looks fine to me - I haven't had a chance to actually try it yet. My only worry would be ensuring the PUT & DELETE requests for counting experiments make it through. If you see that working, feel free to merge if I don't get around to it. Really, I think this stuff is just a leftover from when we used Firefox Accounts.

lmorchard:
Going to actually look at this in a few minutes - looks good, probably good to do unless something weird happens.

Issue #1220, opened 2016-08-09
Refactor add-on for unit testing


lmorchard:
In PR #1218, I claim to fix #352. That's true, insofar as integration tests will exercise the add-on. Still, it would also be nice to have some good unit tests for the add-on by itself. 

However, I found it hard to write unit tests with all the dependencies and Firefox APIs the add-on uses. We have some magical tools for injecting mocks in Python & Node, but I couldn't find something like that for Firefox add-ons.

So, to get good unit test coverage, I think significant clean up & refactoring is necessary:
1. Finish #610 and everything we can from `index.js` into submodules.
2. Rework submodules to use [a dependency injection pattern](http://jasonpolites.github.io/tao-of-testing/ch3-1.1.html): implement constructors / initializers to be called from `index.js` that supply the submodule with any `require()`'d modules or instances of Firefox services.
3. Write unit tests that swap in mock APIs & services for the submodules' constructors & initializers, so they work against test code.

This seems kind of drastic, so I'd really like it if someone came along and told me there was a better way :)

lmorchard:
A pointer from addons in IRC: 

https://developer.mozilla.org/en-US/Add-ons/SDK/Low-Level_APIs/_loader

tl;dr: Might be able to write a custom loader so that mock modules get loaded during testing. Need to investigate.

jaredhirsch:
I'm happy to review when is ready

lmorchard:
Hoping to have this in a PR on Monday. Two main things left:
- [ ] Finish 5 more tests to complete the suite
- [ ] Get this stuff running in Circle CI

PR #1254, opened 2016-08-15
Unit tests for addon/lib/metrics.js

lmorchard:
- Use the downloaded Firefox for both integration and add-on unit tests.
- Switch from Ci.import() to require() for some JS modules in
  addon/lib/metrics.js, making them replaceable with mock APIs
- New tests in addon/test/test-metrics.js
- Include add-on testing in bin/circleci/build-addon.sh
- Utilities for mocking out API functions and monitoring calls
- Custom module loader to substitute mock modules on require() calls in
  the test subject module

Issue #1220


lmorchard:
@6a68, @meandavejustice, @dannycoates, @chuckharmston, and anyone else I forgot - this is worth getting lots of eyes on. 

This implements tests for all the public methods of the Metrics submodule and includes a set of mocking utilities that I hope will be useful for backfilling unit tests for the rest of the add-on. 

Lots of critique is welcome, especially since we should maintain this unit test suite going forward for all add-on work.

addon/test/lib/mock-utils.js
```diff
@@ -0,0 +1,103 @@
...
+exports.resetCallbacks = function(callbacks) {
+  Object.keys(callbacks).forEach(prefix => {
+    const items = callbacks[prefix];
+    Object.keys(items).forEach(name => items[name].reset());
+  });
+};
+
+// Build a custom module loader that substitutes mock modules for the test
+// subject's require() calls.
+exports.loader = function(fromTest, subjectPath, modules) {
```
dannycoates: It took me a minute to figure out what `fromTest` is and why it's necessary. It might help if the name had "module" in it. :shrug:

addon/test/test-metrics.js
```diff
@@ -0,0 +1,348 @@
...
+  'resource://gre/modules/Services.jsm': {
+    Services: {
+      startup: {
+        getStartupInfo: () => (
+          {process: Date.now() - (BROWSER_DURATION * 1000)}
+        )
+      }
+    }
+  }
+});
```
dannycoates, line 55: This is nice!

addon/test/test-metrics.js
```diff
@@ -0,0 +1,348 @@
...
+  callbacks: MockUtils.callbacks({
+    seedrandom: ['seedrandom'],
+    Events: ['on', 'off', 'emit'],
+    TelemetryController: ['submitExternalPing'],
+    AddonManager: ['getAddonByID'],
+    PrefsService: ['set', 'get']
+  })
+};
+
+const mockLoader = MockUtils.loader(module, './lib/metrics.js', {
```
lmorchard, line 35: Oh yeah and the `./lib/metrics.js` here is weird, versus the `require('../lib/metrics');` path a few lines down. Turns out the loader path is always "relative" from the root of the add-on project. That is, _unless_ you're using some exceptions like modules from `sdk/`

dannycoates:
I see the main advantage to this over running the tests with something like [proxyquire](https://github.com/thlorenz/proxyquire) on node is that its running on firefox, which though similar to node, has a bunch of differences that we [might|definitely] care about. The disadvantage is that running them is harder. So although it'd be nice and convenient to unit test with node tools we care enough about the differences to make this choice. Is my assumption correct?

addon/test/test-metrics.js
dannycoates, line 35: what if the api was like 

``` js
const Metrics = MockUtils.require('./lib/metrics', thing, module)
```

would that work?

lmorchard:
> Is my assumption correct?

Yeah, different JS engine that could occasionally make a difference. And we may occasionally not be able or want to mock something out that's in the add-on SDK or the chrome guts of Firefox.

addon/test/test-metrics.js
lmorchard, line 35: I couldn't get something like that to work, since this is using the SDK's require() implementation and I can't really tweak that much. If I recall, it freezes the current "relative" path in a closure that I can't change or inspect from the test module

dannycoates, line 35: heh, i missed the subtle one dot difference in your previous comment

dannycoates:
This looks awesome, man! 👍

dannycoates:
> different JS engine that could occasionally make a difference

Also "environment" differences. Just in my limited experience with the SDK I've seen several common node modules fail because they assume things like `process` and `__dirname` etc exist.

bin/circleci/build-addon.sh
```diff
@@ -4,6 +4,8 @@ cd addon/
 npm config set spin false
 npm install
 npm run lint
+# Use jpm directly here so we can specify the Firefox binary
+./node_modules/.bin/jpm test --verbose --binary=$HOME/firefox/firefox-bin
```
meandavejustice: we should add this to the addon's package.json npm scripts as `test`

only difference is we should allow the ability to specify firefox binary when running.

meandavejustice:
We should add some info in the `addon/README.md` on how to run these

lmorchard:
Updated to add some notes in docs and move to an npm script

chuckharmston:
I spent some time poking here, and I was able to get these running (fast!) and made sense of all the code (comments were very helpful).

Probably not perfect, but I say we ship it and iterate. This sort of thing is better proved out by usage than review. r+ from me!

lmorchard:
All thumbs up, none down, I'll just merge this and we can wrestle with it from there. If anyone has some spare time to try backfilling more unit tests for modules in `addon/lib/`, that would be awesome. Tempted to file issues for each of them as followup

PR #1255, opened 2016-08-16
update ping definitions

clouserw:
r? @chuckharmston


chuckharmston:
Double-checking with Les that we can merge this despite the broken build, but it looks good otherwise.

chuckharmston:
I'm going to ship it anyway. Other builds are failing and this is all docs.

:shipit:

PR #1261, opened 2016-08-18
Show email modal after installing on experiment page (closes #1259).

### LINKED ISSUE #1259: Present email opt-in after installing from an experiment detail page
STR:
- Have Test Pilot uninstalled.
- (currently this is only on dev) Configure your add-on environment to dev.
- Visit an experiment detail page e.g. http://testpilot.dev.mozaws.net/experiments/activity-stream
- Install Test Pilot.
- Detail page refreshes except you now have Test Pilot installed.

Expected Result:
- Email opt-in modal is presented to the user.

Actual Result:
- No email opt-in modal is presented to the user.
- If a user browses away from Test Pilot, then hits the homepage later, the modal appears.

I _think_ this is the "Expected Result" we want. @chuckharmston @johngruen thoughts?

chuckharmston:

screen shot 2016-08-18 at 9 39 17 am

### DIFF
--- a/testpilot/frontend/static-src/app/views/testpilot-promo-view.js
+++ b/testpilot/frontend/static-src/app/views/testpilot-promo-view.js
@@ -49,8 +49,6 @@ export default BaseView.extend({
   },
 
   installClicked() {
-    installAddon(this, 'ExperimentDetailsPage Interactions', () => {
-      this.parentView.render();
-    });
+    installAddon(this, 'ExperimentDetailsPage Interactions', () => {});
   }
 });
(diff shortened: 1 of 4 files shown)


testpilot/frontend/static-src/app/views/testpilot-promo-view.js
```diff
@@ -49,8 +49,6 @@ export default BaseView.extend({
   },
 
   installClicked() {
-    installAddon(this, 'ExperimentDetailsPage Interactions', () => {
-      this.parentView.render();
-    });
+    installAddon(this, 'ExperimentDetailsPage Interactions', () => {});
```
chuckharmston, line 52: This actually caused it to render twice, since I already had this in the experiment page view:

``` javascript
app.me.on('change:hasAddon', this.render, this);
```

It took me an embarrassing amount of time to figure out why it would quickly dismiss the modal.

PR #1263, opened 2016-08-18
add social links to the footer

johngruen:
fixes #294

PR #1264, opened 2016-08-19
Update GA docs with social link click events

ckprice:
Following up on #1263.


ckprice:
Just adding a couple rows to docs. Merging.

Issue #1250, opened 2016-08-15
Replace most of the django-derived experiment data


johngruen:
I've started to rough out a [template](https://github.com/mozilla/testpilot-meta/blob/master/src/template/data.hjson) for experiment details using [hjson](http://hjson.org/) and gulp. Let's take this further by migrating it into the repo and replacing as much of the django API as we can for experiments.

johngruen:
@lmorchard just to be clear, my little experiment in the meta repo is not meant to be proscriptive. If it helps, please use it, but feel free to ignore it!

lmorchard:
Got some initial progress, playing with YAML instead of HJSON. We could possibly move things pretty easily to Jekyll someday if we start with YAML now, and it doesn't look much different than HJSON. 

Using a gulp task to import & massage content from the existing prod API, that results in experiment content files like this:

https://github.com/lmorchard/testpilot/blob/02395866a9681f3d3e6b045a4d1e2d9c26a512f1/content-src/experiments/activity-stream.yaml

Then, another gulp task builds a static version of `/api/experiments` that seems to work seamlessly with the frontend app and the add-on without changes.

Still need to do something about pulling images and things into the repo, so that will come next.

pdehaan:
+1 to YAML.

Also, you can probably remove a bunch of the optional single/double quotes. May also be easier to use multiline string format for some of the HTML blocks (like `measurements` which has escaped quotes-n-stuff):

http://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

lmorchard:
Well, this is just whatever the YAML serializer is doing, so i can get a quick import. But for manual edits, we can do whatever is valid yaml

lmorchard:
b56aaf3244e5c532fc6c66c095478d2706644bc4

PR #1268, opened 2016-08-19
Tools to import API content and build static experiments JSON

lmorchard:
- New Gulp tasks:
  - `import-api-content` - pulls down content from production API,
    generates .yaml files and downloads image resources
  - `experiments-json` - builds `static/api/experiments.json` and
    `static/experiments/{id}.json` from YAML files.
- New Waffle flag in Django:
  - `static_experiments_json` - switches to serving up
    `static/api/experiments.json` at `/api/experiments.json` and
    `static/api/experiments/{id}.json` at `/api/experiments/{id}.json`
    instead of the dynamic REST framework view
- Convenience script to run API import
- Frontend image lint should ignore imported experiment assets (for now)
- Tweaks to add JSON content-type
- Add-on makes URLs in experiment fields absolute to the current
  environment if they're not already
- Switch to using `/api/experiments.json` and
  `/api/experiments/{pk}.json` URLs with the `.json` extension,
  supported by Django and easier to support with static files

Issue #1250


lmorchard:
Hmm. This is a bit difficult to try out at the moment, and it's failing an image test in gulp. Will add a commit with some convenience scripts & docs, and see if I can fix that test.

In the meantime, it would be good to at least get some eyes skimming the code!

lmorchard:
Okay, some notes:
- On initial checkout of this PR, you should see a new Gulp task run in `frontend_watcher` called `experiments-json`. That should result in some data at `testpilot/frontend/static/api/experiments.json` 
  
  You should also see `experiments-json` run again whenever YAML files under `content-src` are manually edited or created.
- You can run another import from production like so:
  
  ```
  docker exec -ti testpilot_frontend_watcher_1 ./bin/run-import-api-content.sh
  ```
  
  That should update YAML files under `content-src/experiments` and download a bunch of images. If you're running Gulp on your host, you can accomplish the same thing with `gulp import-api-content`. 
- You can switch between the Django REST API and this new static file in your dev site by [adding & enabling a new Waffle flag](http://testpilot.dev:8000/admin/waffle/flag/) named `static_experiments_json`. By default, the old REST API shows up at [/api/experiments](http://testpilot.dev:8000/api/experiments). When the flag is enabled, the contents of `static/api/experiments.json` should appear instead.
- When the static JSON is enabled, you should see the frontend on your local dev server get populated with content from production and there should be no errors.

lmorchard:
Ugh. Falling down the CI tree and hitting every branch along the way

lmorchard:
Yay, the tests passed after revealing an actual bug! \o/

dannycoates:
I noticed while doing gemini that I had to make some of the urls absolute in order for the addon to work https://github.com/dannycoates/gemini/blob/master/gulpfile.js#L295-L301

dannycoates:
I think it might just be `thumbnail` that needs the full url
screen shot 2016-08-22 at 11 38 05 am

I also see 404s on `/api/experiments/5`

dannycoates:
> I also see 404s on /api/experiments/5

Oh yeah, my local db doesn't have no-more-404s. derp. 😜

dannycoates:
This looks good to me once the thumbnails work 👍

lmorchard:
> I also see 404s on /api/experiments/5

I forget where these individual experiment resources are fetched... Might have to have the script generate those, too. I'll poke around

> I noticed while doing gemini that I had to make some of the urls absolute in order for the addon to work

Hmm. Might be a better idea to do this absolutifying in the add-on, that way the JSON can be the same content regardless of host (e.g. local, stage, prod, branch, etc) and the add-on can resolve the URL because it knows which host it fetched the data from.

lmorchard:
Updated the PR to include:
- Tweaks to add-on turn site-relative URLs to absolute URLs in experiment records, using the base URL of the current environment. That seems to fix the thumbnails in the add-on  doorhanger.
- Now generating and serving JSON resources for individual experiments at `/api/experiments/{id}.json`. Changed the generated experiment data to point at those, so hopefully that fixes the 404.

dannycoates:
thumbnails look good 👍 but the installation url returns 403 now

lmorchard:
> thumbnails look good 👍 but the installation url returns 403 now

Which installation URL? Does it do it with the old API turned back on?

dannycoates:
> Which installation URL?

http://testpilot.dev:8000/api/experiments/2/installations/aa47cb02-f9b2-9942-ba89-b699d16d8de1

on both PUT and DELETE I get 403

> Does it do it with the old API turned back on?

It does it on both

lmorchard:
> on both PUT and DELETE I get 403

Oh, belated mention since I got distracted with lunch: I can reproduce this, but not sure what's happening yet. Weird because this doesn't 403 for me:

```
➜  testpilot git:(1250-static-experiment-json) ✗ curl -X PUT 'http://testpilot.dev:8000/api/experiments/2/installations/551f0c01-ccfe-fa46-aece-873e43ef5f55'
{"url":"http://testpilot.dev:8000/api/experiments/2/installations/551f0c01-ccfe-fa46-aece-873e43ef5f55","experiment":"http://testpilot.dev:8000/api/experiments/2","client_id":"551f0c01-ccfe-fa46-aece-873e43ef5f55","addon_id":"tabcentertest1@mozilla.com","created":"2016-08-23T19:48:43.600352Z","modified":"2016-08-23T19:49:14.794612Z"}
```

Definitely no auth or CSRF cookies in that plain old curl request.

lmorchard:
Hmm, did `git revert 5b0884a8345f0023de37e3b1b5f2fbb3ebe8d1a1` (i.e. [the CSRF header removal commit](https://github.com/mozilla/testpilot/commit/5b0884a8345f0023de37e3b1b5f2fbb3ebe8d1a1)) and the 403s go away. Roll back the revert, and the 403s come back. Need to see what in there might be the culprit.

lmorchard:
@dannycoates: Okay, I think I might have figured it out. Just stripped away more Django REST framework stuff from the PUT/DELETE handler and made it CSRF exempt. That seems to resolve the 403 errors for me without reverting the change to remove the cookie code in the add-on.

dannycoates:
this lgtm, r+

jaredhirsch:
:tada:

PR #1274, opened 2016-08-24
'Ending Soon' UI

### LINKED ISSUE #1271: Graduation "Ending Soon" UI
Implement the first part of #1247 

https://mozilla.invisionapp.com/share/Q77S1H6ZC#/181298063_One_Week_Out

dannycoates:
closes #1271 

This includes the "Ending Soon" messaging in the addon, experiment page, and experiment card.

It depends on a new `completed` date field in the experiment data. I haven't tried to add it to the database here since I'm hoping we'll be db-free before we need it. If not, I'll have to figure that out :)

Screenshots:

screen shot 2016-08-24 at 11 37 31 am

screen shot 2016-08-24 at 11 39 47 am

screen shot 2016-08-24 at 11 40 40 am

### DIFF
--- a/addon/lib/templates.js
+++ b/addon/lib/templates.js
@@ -13,8 +13,9 @@ module.exports.experimentList = `
         <div class="icon" style="background-image:url('{{thumbnail}}');"></div>
       </div>
       <div class="experiment-title">{{title}}
-        <span class="active-span {{#active}}visible{{/active}}">Enabled</span>
-        <span class="is-new-span {{#isNew}}visible{{/isNew}}">New Experiment</span>
+        <span class="active-span" style="{{#hideActive}}display:none;{{/hideActive}}">Enabled</span>
+        <span class="is-new-span">New Experiment</span>
+        <span class="eol-span {{#showEol}}visible{{/showEol}}">{{eolMessage}}</span>
       </div>
     </a>
   {{/experiments}}
--- a/testpilot/frontend/static-src/app/views/eol-view.js
+++ b/testpilot/frontend/static-src/app/views/eol-view.js
@@ -1,0 +1,22 @@
+import BaseView from './base-view';
+
+export default BaseView.extend({
+  template: `<div class="eol-block"><div data-hook="ending-soon" data-l10n-id="eolMessage">
+<strong>This experiment is ending on <span data-hook="completedDate"></span></strong>.<br/><br/>
+After then you will still be able to use <span data-hook="title"></span> but we will no longer be providing updates or support.</div>
+</div>`,
+  props: {
+    completedDate: 'string',
+    title: 'string'
+  },
+  bindings: {
+    completedDate: '[data-hook=completedDate]',
+    title: '[data-hook=title]'
+  },
+  getL10nArgs: function() {
+    return {
+      title: this.title,
+      completedDate: this.completedDate
+    };
+  }
+});
--- a/testpilot/frontend/static-src/app/views/experiment-page.js
+++ b/testpilot/frontend/static-src/app/views/experiment-page.js
@@ -12,6 +12,7 @@ import DisableDialogView from './disable-dialog-view';
 import ExperimentListView from './experiment-list-view';
 import ExperimentTourDialogView from './experiment-tour-dialog-view';
 import TestpilotPromoView from './testpilot-promo-view';
+import EolView from './eol-view';
 
 function changeHeaderOn() {
   const mainHeader = document.getElementById('main-header');
@@ -251,6 +252,13 @@ export default PageView.extend({
         except: this.model.slug,
         eventCategory: 'ExperimentsDetailPage Interactions'
       }), '[data-hook="experiment-list"]');
+    }
+
+    if (this.model.completed) {
+      this.renderSubview(new EolView({
+        completedDate: new Date(this.model.completed).toLocaleDateString(),
+        title: this.model.title
+      }), '[data-hook="eol-message"]');
     }
 
     postInstallModal(this);
--- a/testpilot/frontend/static-src/styles/modules/_experiment-list.scss
+++ b/testpilot/frontend/static-src/styles/modules/_experiment-list.scss
@@ -103,6 +103,10 @@
   }
 }
 
+.eol-message {
+  color: $eol-dark-orange;
+}
+
 .experiment-actions {
   .has-addon:not(.active) &.show-when-inactive {
     display: block;
--- a/addon/data/panel.css
+++ b/addon/data/panel.css
@@ -169,6 +169,12 @@ a {
     font-size: 14px;
 }
 
+.eol-span {
+    display: block;
+    color: #f46323;
+    font-size: 14px;
+}
+
 .installed-message {
     background: #fff;
     display: flex;
@@ -239,4 +245,3 @@ a {
     transform: scale(1);
   }
 }
-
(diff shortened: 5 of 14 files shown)


testpilot/frontend/static-src/app/views/eol-view.js
```diff
@@ -0,0 +1,18 @@
+import BaseView from './base-view';
+
+export default BaseView.extend({
+  template: `<div class="eol-block"><div data-hook="ending-soon">
+<strong>This experiment is ending on <span data-hook="completedDate"></span></strong>.<br/><br/>
+After then you will still be able to use <span data-hook="title"></span> but we will no longer be providing updates or support.</div>
+<!--<div data-hook="experiment-completed">
+This experiment was retired on <span data-hook="completedDate"></span> and is no longer supported.</br>
+</div>--></div>`,
```
dannycoates: I couldn't figure out how to get `data-l10n-id` to work here. The bindings wouldn't fill in like they do with, for example `{$title} is enabled`... Maybe we'll be on to react before we need translations???

dannycoates:
Hey, I just branched you

and this is craaaazy

but here's my PR, so

r? me maybe

chuckharmston:
https://youtu.be/-qTIGg3I5y8

testpilot/frontend/static-src/styles/modules/_experiment-list.scss
```diff
@@ -103,6 +103,10 @@
   }
 }
 
+.eol-message {
+  color: #f46323;
```
chuckharmston: We've been putting colors values in `testpilot/frontend/static-src/styles/_variables.scss`

chuckharmston: /g

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -253,6 +254,13 @@ export default PageView.extend({
       }), '[data-hook="experiment-list"]');
     }
 
+    if (this.model.completed) {
+      this.renderSubview(new EolView({
+        completedDate: new Date(this.model.completed).toLocaleDateString(),
+        title: this.model.title
+      }), '[data-hook="eol-message"');
```
chuckharmston: Missing a `]` here.

testpilot/frontend/static-src/app/views/eol-view.js
```diff
@@ -0,0 +1,18 @@
...
+  props: {
+    completedDate: 'string',
+    title: 'string'
+  },
+  bindings: {
+    'completedDate': '[data-hook=completedDate]',
```
chuckharmston: No single quotes around keys for consistency?

testpilot/frontend/static-src/app/views/eol-view.js
```diff
@@ -0,0 +1,18 @@
+import BaseView from './base-view';
+
+export default BaseView.extend({
+  template: `<div class="eol-block"><div data-hook="ending-soon">
+<strong>This experiment is ending on <span data-hook="completedDate"></span></strong>.<br/><br/>
+After then you will still be able to use <span data-hook="title"></span> but we will no longer be providing updates or support.</div>
+<!--<div data-hook="experiment-completed">
```
chuckharmston: Did you mean to ship the commented code?

addon/lib/templates.js
```diff
@@ -13,8 +13,9 @@ module.exports.experimentList = `
         <div class="icon" style="background-image:url('{{thumbnail}}');"></div>
       </div>
       <div class="experiment-title">{{title}}
-        <span class="active-span {{#active}}visible{{/active}}">Enabled</span>
-        <span class="is-new-span {{#isNew}}visible{{/isNew}}">New Experiment</span>
+        <span class="active-span" style="{{#hideActive}}display:none;{{/hideActive}}">Enabled</span>
+        <span class="is-new-span">New Experiment</span>
+        <span class="eol-span {{#showEol}}visible{{/showEol}}">{{eolMessage}}</span>
```
chuckharmston, line 18: I still don't really know what I'm doing with ampersand, but I think we want to be using hooks for this, right?

chuckharmston, line 18: Although I do see that this is what already exists.

testpilot/frontend/static-src/app/views/eol-view.js
chuckharmston: I have no idea how this magic works. @lmorchard?

chuckharmston:
Looks good to me, although I'm not sure what's going on with the l20n bit you called out.

testpilot/frontend/static-src/app/views/eol-view.js
lmorchard: Yeah, l10n is not sane in our current environment. L20N comes around and plonks string replacements down over the work that Ampersand already did with hooks & bindings - both systems compete to alter the DOM.

For example, consider [the place where we display the usage count](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/templates/experiment-page.js#L35) on the experiment page:
- It has an `data-l10n-id="userCountContainer"`
- [There's a localized string for that.](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl#L136)
- Ampersand does it's thing with hooks & bindings there when data changes.
- But, then we have [some code that populates `data-l10n-args` with `model` properties on every DOM node with a `data-l10n-id` attribute whenever model data changes](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/views/base-view.js#L30). 
  
  (Part of the reason for this is because L20N didn't have any concept of global variables for the whole page, just per-DOM-node data. So, we splatter the variables over every relevant node. Ew.) 
- L20N sees a `data-l10n-data` change on a node with a `data-l10n-id`. So, whatever Ampersand did with hooks & bindings is replaced by what L20N does with localized strings and their placeholders

TL;DR: You can override [`getL10nArgs`](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/views/base-view.js#L24) from the base page view to apply something other than model attributes. I think you'd need to do that here, because this view isn't using a single `model` and instead has its own props. Something like this:

``` javascript
  getL10nArgs() {
    return {
      title: this.title,
      completedDate: this.completedDate
    };
  }
```

Does all that make sense? I mean, insofar as it's terrible and nonsensical and shames me?

addon/lib/templates.js
dannycoates: this is in the addon panel, so handlebars. 🍌

dannycoates:
thanks @chuckharmston! updated.

testpilot/frontend/static-src/app/views/eol-view.js
dannycoates: Thanks for the thorough explanation @lmorchard. Adding `getL10nArgs` made it work :)

addon/lib/templates.js
chuckharmston: (－‸ლ)

lmorchard:
> I'm hoping we'll be db-free before we need it. If not, I'll have to figure that out

A very optimistic hope, but failing that we could do a follow up to add this to the Django machinery if/when we need to graduate an experiment in the next few sprints.

lmorchard:
Looks good to me, though could use some frontend tests to put it through its paces since it's a bit tricky to test manually without the field in Django

dannycoates:
> though could use some frontend tests to put it through its paces

im looking at that now

lmorchard:
Tests pass, works for me! We can follow up with adding fields in Django or (hopefully) the static side of things when we get there

PR #1276, opened 2016-08-24
Reorganize documentation (closes #1266).

### LINKED ISSUE #1266: Reorganize documentation
We have lots of docs, and they're scattered and can be difficult to find. I'm going to spend a few hours organizing them and cleaning them up.

chuckharmston:
In this PR, I:
- Moved all substantial documentation into its own files inside of `docs`
  - Exceptions: `addon/README.md`, `LICENSE.md`, and `CONTRIBUTING.md`, all of which have magic names for special treatment.
- Changed all filenames to lowercase.
  - Exceptions: the aforementioned magic names.
- Added a full table of contents in the main README, with some categorization.
- Made sure all documentation files were linked from there.
- Moved all documentation out of the main README.
- Used [shields](http://shields.io/) for our badges for a consistent visual style.
- Fixed some document hierarchies.
- Added links to the main README from all other subpages.
- Moved images into their own directory inside of `docs`.
- Changed all links and images to use relative URLs, which GitHub likes.
- Did a bit of renaming as made sense when looking at docs holistically.

Browsable here: https://github.com/chuckharmston/testpilot/blob/1266-docs-reorg/README.md

r? Would love lots of eyes on this.


docs/code_of_conduct.md
```diff
@@ -1,3 +1,5 @@
+[👈 Back to README](../README.md)
```
lmorchard, line 1: My only nit here is that we might want to leave this at the top level along with things like CONTRIBUTING.md, because I've noticed a few other projects doing it as an ad-hoc practice. But, it's not something that gets special treatment from GitHub, so the case for organizing this into the docs dir is just as valid IMO.

lmorchard:
Everything's neater, nothing's left out - looks 💯 to me. Could only come up with one picky nit that we might as well ignore. I might wait until a bit later this morning to merge in case anyone else has comments, but r+ from me to merge before the freeze today

ckprice:
Looks great to me too, thanks chuck!

lmorchard:
On that note, I'll just hit the button!

PR #1277, opened 2016-08-24
Enable multi-process add-on flag. Fixes #1246

### LINKED ISSUE #1246: Test pilot should be marked as multiprocess compatible
Currently Test Pilot isn't marked as multiprocess compatible. Issue #457 says that test pilot works with e10s, however, there may still be some incompatibilities that are masked by Firefox's shims.

The shims may lead to performance issues, although I doubt Test Pilot would really be hitting those that hard.: https://developer.mozilla.org/Add-ons/SDK/Guides/Multiprocess_Firefox_and_the_SDK#Testing_with_shims_enabled

Setting the flag for Test Pilot should be as simple as adding the following to `addon/package.json`:

```
"permissions": {
  "multiprocess": true
}
```

Obviously, it should be tested to make sure it does work.

clouserw:
This is untested, but we talked about it and don't think it should be a problem...

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -43,5 +43,8 @@
   "dependencies": {
     "mustache": "2.2.1",
     "seedrandom": "^2.4.2"
+  },
+  "permissions": {
+    "multiprocess": true
   }
 }


ghost:
Thanks @chuckharmston .  I built this and installed it on OS X with nightly.  Stuff seemed to work.  I made the "untested" comment above because I think this needs to be tested broadly across firefox versions and platforms, but I don't know all the details of e10s.  Maybe there are just a few we need to hit.

chuckharmston:
That's good enough to quell my concerns. Let's ship it.

PR #1278, opened 2016-08-24
Delete old user accounts from the database

### LINKED ISSUE #1034: Remove Test Pilot user database.

lmorchard:
- Ensure contributors on experiments are counted as staff
- Delete all user non-staff user accounts
- Drop old tables used by the old Firefox Accounts integration

Fixes #1034

### DIFF
--- a/testpilot/users/migrations/0003_delete_nonstaff_users.py
+++ b/testpilot/users/migrations/0003_delete_nonstaff_users.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+from testpilot.utils import cleanup_nonstaff_users
+
+
+def delete_nonstaff_users_forwards(apps, schema_editor):
+    Experiment = apps.get_model('experiments', 'Experiment')
+    User = apps.get_model('auth', 'User')
+    cleanup_nonstaff_users(User, Experiment.objects.all())
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0002_userprofile_invite_pending'),
+    ]
+
+    operations = [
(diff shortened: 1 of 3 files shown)


lmorchard:
I _think_ this covers all the old user data still hanging around. Can anyone think of things I missed?

chuckharmston:
I have no holes to poke in the code, but it would be very nice to [write a unit test](https://www.caktusgroup.com/blog/2016/02/02/writing-unit-tests-django-migrations/) for this, since it's so...destructive.

lmorchard:
Hmm, not sure how to unit test the DROP TABLE bits without writing another migration to add the tables for tests. Might just look into the bit that only deletes non-staff users

lmorchard:
FWIW, working on a test, but having some difficult wrestling with the TestMigrations class in the linked blog article. Might end up missing the freeze with that, today. 

Seems fiddly with regards to requiring reversible migrations (which this decidedly isn't) and <strike>names of installed apps it can't find</strike> models from other apps within the testpilot project (e.g. need Experiment model within the user app migration tests to deal with contributors, but testpilot.experiments isn't found as installed). 

Still poking though

lmorchard:
Added a test exercising the account deletion code that I moved to the utils library. 

A _little_ leary of it: The Experiment model during migration has the default Manager (`Experiment.objects.all()`). But, thanks to django-hvad, Experiment gets a localization-aware Manager swapped in with different queryset APIs during tests & normal app execution (`Experiment.objects.language().fallbacks('en').all()`). But, I think the two lookups are basically equivalent.

chuckharmston:
This looks great to me. The only ask I'd have is that we verify on stage that the correct users are kept and deleted once we get this pushed out there.

:shipit:❗

PR #1279, opened 2016-08-25
added a static site build

dannycoates:
This adds the `static` gulp task which builds a static version of the site into the `/dist` directory. The build for the django assets remains the same.

`npm start` will build the static site and host a server on port 8000.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -424,16 +424,27 @@ gulp.task('watch', function watchTask() {
   gulp.watch('gulpfile.js', () => process.exit());
 });
 
+let serverPort = 8000;
+
 // Set up a webserver for the static assets
 gulp.task('connect', function connectTask() {
   connect.server({
     root: DIST_PATH,
     livereload: false,
-    port: 8000
+    port: serverPort
   });
 });
 
 gulp.task('server', function() {
+  serverPort = 9988;
+  return runSequence(
+    'static',
+    'connect',
+    'watch'
+  );
+});
+
+gulp.task('static-only-server', function() {
   return runSequence(
     'static',
     'connect',
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,8 +13,12 @@ const gulp = require('gulp');
 const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
 // const imagemin = require('gulp-imagemin');
+const merge = require('merge-stream');
 const minifycss = require('gulp-cssnano');
+const multiDest = require('gulp-multi-dest');
 const normalize = require('node-normalize-scss');
+const rename = require('gulp-rename');
+const RevAll = require('gulp-rev-all');
 const runSequence = require('run-sequence');
 const sass = require('gulp-sass');
 const sassLint = require('gulp-sass-lint');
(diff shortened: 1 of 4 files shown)


testpilot/frontend/static-src/index.html
```diff
@@ -0,0 +1,65 @@
...
+    <title>Firefox Test Pilot</title>
+
+    <link rel="canonical" href="http://testpilot.firefox.com/">
+    <meta name="description" content="Test new Features. Give us feedback. Help build Firefox." />
+    <meta property="og:description" content="Test new Features. Give us feedback. Help build Firefox." />
+    <meta property="og:image" content="http://testpilot.firefox.com/static/images/thumbnail-facebook.1877727a8303.png" />
+    <meta property="og:locale" content="en_US" />
+    <meta property="og:title" content="Firefox Test Pilot" />
+    <meta property="og:type" content="website" />
+    <meta property="og:url" content="http://testpilot.dev.mozaws.net/" />
```
pdehaan: Should this be hard-coded to dev?

gulpfile.js
```diff
@@ -312,8 +320,82 @@ function buildExperimentsJSON(path) {
...
+  return merge(
+    gulp.src(SRC_PATH + 'index.html')
+      .pipe(multiDest(paths)),
+    gulp.src('./legal-copy/privacy-notice.html')
+      .pipe(rename('index.html'))
+      .pipe(gulp.dest(STAGE_PATH + '/privacy')),
+    gulp.src('./legal-copy/terms-of-use.html')
+      .pipe(rename('index.html'))
+      .pipe(gulp.dest(STAGE_PATH + '/terms'))
+  );
```
chuckharmston, line 360: This is kind of a bummer. I wonder if there's a way we can do some sort of discovery to avoid having to [define everything twice](https://github.com/mozilla/testpilot/blob/master/testpilot/frontend/static-src/app/lib/router.js#L10-L21).

lmorchard: Hmm, I want to say I used this [express-history-api-fallback](https://www.npmjs.com/package/express-history-api-fallback) module at some point in a gulp debug server. Lets you use a single index.html as the ultimate fallback for any URL that would get routed in the client-side app via history API (like we do)

testpilot/frontend/static-src/app/collections/experiments.js
```diff
@@ -6,7 +6,7 @@ import Experiment from '../models/experiment';
 export default Collection.extend({
   model: Experiment,
   indexes: ['slug'],
-  url: '/api/experiments',
+  url: '/api/experiments.json',
```
lmorchard, line 9: Oh, good catch, I missed that in my PR!

gulpfile.js
```diff
@@ -344,12 +426,18 @@ gulp.task('watch', ['build'], function watchTask() {
 // Set up a webserver for the static assets
 gulp.task('connect', function connectTask() {
   connect.server({
-    root: DEST_PATH,
+    root: DIST_PATH,
     livereload: false,
-    port: 9988
+    port: 8000
```
lmorchard: Might want to put this at a different port than 8000, to avoid conflicting with Django in case you want to run both at the same time for comparison's sake. I think that's why I had it at a weird port

gulpfile.js
lmorchard: Though, I guess that only solves things for us in dev, and doesn't really help in prod/stage/etc where we'll be running under nginx or Amazon S3 without necessarily having the routing smarts

gulpfile.js
dannycoates: hmm, yeah. I was thinking about it from the angle of making sure it works with the addon. I guess either way something isn't going to work. I guess we could make a separate task for each...

gulpfile.js
dannycoates: I imagine this to be a transitional approach. (It was the easiest thing I could think of)

I'm sure we'll find a better way.

lmorchard:
Going to finally take a good dive into this today, since I think building on this (or something like it) will be a focus for me this week

lmorchard:
Finally got the node version sorted out on my host machine and got this running. Works for me - think I'll merge this and work with it from there

PR #1284, opened 2016-08-25
fix margins and privacy on small details pages

### LINKED ISSUE #1282: Add padding back to middle breakpoint on details page
screen shot 2016-08-25 at 12 54 41 pm

@clouserw this needs to land before we tag

johngruen:
Fixes #1282 

@chuckharmston i also added the ability to show the install and show privacy on the installed details page

### DIFF
--- a/testpilot/frontend/static-src/app/templates/experiment-page.js
+++ b/testpilot/frontend/static-src/app/templates/experiment-page.js
@@ -91,6 +91,13 @@ export default `
                 </section>
               </div>
               <div class="details-list"></div>
+              <div data-hook="active-user">
+                <section data-hook="measurements-container" class="measurements">
+                  <h3 data-l10n-id="measurements">Your privacy</h3>
+                  <div data-hook="measurements-html" class="measurement"></div>
+                  <a class="privacy-policy" data-l10n-id="experimentPrivacyNotice" data-hook="privacy-notice-url">You can learn more about the data collection for <span data-hook="title"></span> here.</a>
+                </section>
+              </div>
             </div>
           </div>
         </div>
--- a/testpilot/frontend/static-src/app/views/experiment-page.js
+++ b/testpilot/frontend/static-src/app/views/experiment-page.js
@@ -317,6 +317,7 @@ export default PageView.extend({
     function cleanup() {
       evt.target.disabled = false;
       evt.target.classList.remove('state-change');
+      evt.target.style.width = '';
       app.off('webChannel:addon-install:download-failed');
       app.off('webChannel:addon-install:install-failed');
       app.off('webChannel:addon-install:install-ended');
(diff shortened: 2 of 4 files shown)


testpilot/frontend/static-src/app/templates/experiment-page.js
```diff
@@ -91,6 +91,13 @@ export default `
                 </section>
               </div>
               <div class="details-list"></div>
+              <div data-hook="active-user">
```
johngruen, line 94: I realize duping this hook is a hack, but it will appease legal and handle the edge case

testpilot/frontend/static-src/app/views/experiment-page.js
```diff
@@ -317,6 +317,7 @@ export default PageView.extend({
     function cleanup() {
       evt.target.disabled = false;
       evt.target.classList.remove('state-change');
+      evt.target.style.width = '';
```
johngruen, line 320: added these to clear the button width after install/uninstall otherwise the calculated width for the state change persists and you get funky button sizes

testpilot/frontend/static-src/styles/modules/_experiment-details.scss
```diff
@@ -165,11 +232,11 @@
 }
 
 .eol-block {
-  padding: 10px !important;
-  border-radius: 3px;
   background: $eol-light-orange;
   border: 1px solid $eol-orange;
+  border-radius: 3px;
   margin-bottom: 20px;
+  padding: 10px !important;
```
johngruen, line 239: @pdehaan this is left over from @dannycoates...if you want to get on somebody's case about using important, take it up with him :)

testpilot/frontend/static-src/styles/_layout.scss
```diff
@@ -85,3 +85,13 @@
 .overflow-hidden {
   overflow: hidden;
 }
+
+.show-on-small {
+  @include respond-to('not-small') {
+    display: none;
+  }
+
+  @include respond-to('small') {
```
chuckharmston: No need for this media query. Just have it be block by default, and override when not-small.

chuckharmston: Careful where you use this, though. We have a bunch of elements that want to be `flex` and not `block`.

johngruen: yeah, i might change this

chuckharmston: It might make more sense as a mixin, now that I think about it. Single param, default `block`, can be overriden when appropriate.

chuckharmston:
Seems fine. I'll hold off on merging myself in case you want to address the nit above; feel free to ship whenever, @johngruen.

johngruen:
@chuckharmston good call on the mixins! be good to go now

PR #1292, opened 2016-08-25
Ask user to share after having an experiment installed for 3 days (closes #1097, closes #1099).

### LINKED ISSUE #1097: Test Pilot should ask user to share after having an experiment installed for 3 days
- Should launch the page from #1096 
- Should only launch on a new browser session

### LINKED ISSUE #1099: Measure conversion of social media shares to new installs
The share URLs from #1096 should ship with `utm_` parameters indicating the link's origin. These will be used to create a dashboard to track the conversion rate from these links to Test Pilot installs.

chuckharmston:
r?

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -14,12 +14,14 @@ const tabs = require('sdk/tabs');
 const request = require('sdk/request').Request;
 const { PrefsTarget } = require('sdk/preferences/event-target');
 const URL = require('sdk/url').URL;
-const Metrics = require('./lib/metrics');
-const survey = require('./lib/survey');
-const WebExtensionChannels = require('./lib/webextension-channels');
-const ToolbarButton = require('./lib/toolbar-button');
-const ExperimentNotifications = require('./lib/experiment-notifications');
+
 const { App } = require('./lib/app');
+const ExperimentNotifications = require('./lib/experiment-notifications');
+const Metrics = require('./lib/metrics');
+const SharePrompt = require('./lib/share-prompt');
+const survey = require('./lib/survey');
+const ToolbarButton = require('./lib/toolbar-button');
+const WebExtensionChannels = require('./lib/webextension-channels');
 
 const settings = {};
 let prefs;
@@ -355,6 +357,7 @@ exports.main = function(options) {
   WebExtensionChannels.init();
   ToolbarButton.init(settings);
   ExperimentNotifications.init();
+  SharePrompt.init(settings);
 
   if (reason === 'install') {
     openOnboardingTab();
@@ -369,6 +372,7 @@ exports.onUnload = function(reason) {
   WebExtensionChannels.destroy();
   ToolbarButton.destroy();
   ExperimentNotifications.destroy();
+  SharePrompt.destroy(reason);
 
   if (reason === 'uninstall' || reason === 'disable') {
     Metrics.onDisable();
--- a/addon/lib/share-prompt.js
+++ b/addon/lib/share-prompt.js
@@ -1,0 +1,75 @@
+ /*
+  * This Source Code is subject to the terms of the Mozilla Public License
+  * version 2.0 (the 'License'). You can obtain a copy of the License at
+  * http://mozilla.org/MPL/2.0/.
+  */
+
+const store = require('sdk/simple-storage').storage;
+const tabs = require('sdk/tabs');
+
+let SETTINGS;
+
+
+module.exports = {
+  DELAY: 3 * 24 * 60 * 60 * 1000,  // 3 days
+  LOCAL_DELAY: 1000,  // 1 second
+  PATH: 'share',
+  QS: 'utm_source=testpilot-addon&utm_medium=firefox-browser&utm_campaign=share-page',
+
+  makeUrl: function() {
+    return `${SETTINGS.BASE_URL}/${this.path}?${this.qs}`;
+  },
+
+  hasExperimentInstalled: function() {
+    try {
+      return Object.keys(store.installedAddons).length > 0;
+    } catch (err) {
+      return false;
+    }
+  },
+
+  openShareTab: function() {
+    tabs.open({
+      url: this.makeUrl()
+    });
+  },
+
+  calculateDelay: function() {
+    return SETTINGS.env === 'local' ? this.LOCAL_DELAY : this.DELAY;
+  },
+
+  shouldOpenTab: function() {
+    return (
+      store.sharePrompt.hasBeenShown === false &&
+      store.sharePrompt.showAt &&
+      store.sharePrompt.showAt <= Date.now()
+    );
+  },
+
+  setup: function() {
+    if (typeof store.sharePrompt === 'undefined') {
+      store.sharePrompt = {
+        showAt: 0,
+        hasBeenShown: false
+      };
+    }
+    if (this.hasExperimentInstalled() && !store.sharePrompt.showAt) {
+      store.sharePrompt.showAt = Date.now() + this.calculateDelay();
+    }
+  },
+
+  destroy: function(reason) {
+    if (reason === 'uninstall') {
+      delete store.sharePrompt;
+    }
+  },
+
+  init: function(settings) {
+    SETTINGS = settings;
+    this.setup();
+    if (this.shouldOpenTab()) {
+      this.openShareTab();
+      store.sharePrompt.hasBeenShown = true;
+    }
+  }
+};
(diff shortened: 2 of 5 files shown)


chuckharmston:
NB: for users who already have Test Pilot + experiments installed, this will only trigger 3 days after the _next_ experiment is installed.

It wasn't specced, but as a one-time measure to bridge that gap, we could toss this tab to any users who have experiments installed when they receive this update. I'm going to build this and we can back it out if it's not desired.

chuckharmston:
The above is now reflected in the code. Additionally, I set the timeout to 1 second, rather than 3 days, in all non-production environments to make for easier testing.

chuckharmston:
Also updated so that the onboarding tab and share tab never open simultaneously, giving priority to the former.

lmorchard:
Will take a look at this today, too

lmorchard:
Would be very handy to have unit tests for this stuff, now that we have the capability

addon/index.js
```diff
@@ -342,6 +358,11 @@ exports.main = function(options) {
 
   if (reason === 'install') {
     openOnboardingTab();
+  } else if (
+    store.showSharePromptAt && Date.now() > store.showSharePromptAt ||
+    typeof store.showSharePromptAt === 'undefined' && hasExperimentInstalled()
```
lmorchard: Trying to work though the logic here to make sure I've got it right... Looks like there are 2 cases when the share-solicitation tab gets shown:
1. Existing Test Pilot users with at least one experiment installed will see a sharing-solicitation immediately on their next browser restart.
2. Future Test Pilot users on their first experiment enable will get a sharing-solicitation scheduled for the next browser restart after at least 3 days have passed.

In either of these cases, `store.showSharePromptAt` gets set to `false` after showing the tab. That, in turn, records that the tab has been shown and no future showings will be scheduled on experiment enables.

As long as I've got all that right, I'd say case 1 sounds okay if we assume existing Test Pilot users have had an experiment around for more than 3 days, so this is basically playing catch-up ([like you wrote above](https://github.com/mozilla/testpilot/pull/1292#issuecomment-242577504)).

lmorchard:
Okay, basically three nits:
1. This stuff would be great if it lived in its own submodule in `lib/` with associated unit tests. (So we don't add to #610)
2. We should probably delete `store.showSharePromptAt` on uninstall. We're not consistent about it, but we've started adding [cleanup-on-uninstall code](https://github.com/mozilla/testpilot/blob/master/addon/lib/experiment-notifications.js#L28) in spots to make uninstall / reinstall behavior more consistent. 
3. Might be clearer if `store.showSharePromptAt` got split into 2 properties instead of doing double duty - i.e. as a datestamp to schedule the next solicitation showing, while also serving as a marker that the tab was shown via being undefined / being assigned `false`

meandavejustice:
I am unable to test this because of a [seemingly unrelated issue](https://github.com/mozilla/testpilot/issues/1324) But this looks good to me. It has my 👍 so long as @lmorchard 's comments are addressed

chuckharmston:
@lmorchard Your comments should all be addressed. I refactored this enough that it probably warrants a re-review.

chuckharmston:
@ckprice mind reviewing the instrumentation? Everything in #1099 should be here except for the actual logging of the pageview, which I've omitted due to #1362.

addon/lib/share-prompt.js
```diff
@@ -0,0 +1,73 @@
+ /*
```
lmorchard, line 1: :+1: yay modularization!

addon/test/test-share-prompt.js
```diff
@@ -0,0 +1,168 @@
+/*
```
lmorchard, line 1: :+1: yay tests!

docs/metrics/ga.md
```diff
@@ -73,6 +73,7 @@ Here are the current events on the website as of this writing
 | Click take survey after Leave                            | RetirePage Interactions            | button click      | take survey                  |
 | Click on Twitter link in footer                          | FooterView Interactions            | social link clicked      | Twitter       |
 | Click on GitHub link in footer                           | FooterView Interactions            | social link clicked      | GitHub        |
+| Click on a button in the Share section                   | ShareView Interactions             | button click             | {facebook,twitter,email,copy} |
```
lmorchard, line 76: :+1: yay doco!

docs/metrics/ga.md
```diff
@@ -104,12 +105,12 @@ Here is a list of dimensions we are currently using
 
 | Page                              | Description                                       | dimension | values |
 |-----------------------------------|---------------------------------------------------|-----------|--------|
-| Home Page, Experiment Detail Page | Does the user have the add-on installed           | 1         | {0,1}  |
-| Home Page                         | Does the user have any experiments installed      | 2         | {0,1}  |
-| Home Page                         | How many experiments does the user have installed | 3         | {n}    |
-| Experiment Detail Page            | Is the experiment enabled                         | 4         | {0,1}  |
-| Experiment Detail Page            | Experiment title                                  | 5         | "xyz"  |
-| Experiment Detail Page            | Installation count                                | 6         | {n}    |
+| Home Page, Experiment Detail Page, Share Page | Does the user have the add-on installed           | 1         | {0,1}  |
```
lmorchard, line 108: :+1: yay more doco!

addon/test/test-share-prompt.js
lmorchard: :+1: and tests pass with `jpm test --verbose --binary /Applications/FirefoxDeveloperEdition.app/Contents/MacOS/firefox-bin`

addon/lib/share-prompt.js
```diff
@@ -0,0 +1,73 @@
...
+        hasBeenShown: false
+      };
+    }
+    if (this.hasExperimentInstalled() && !store.sharePrompt.showAt) {
+      store.sharePrompt.showAt = Date.now() + this.calculateDelay();
+    }
+  },
+
+  destroy: function() {
+    delete store.sharePrompt;
```
lmorchard: Hmm, you probably want to move this into an `uninstall` method and only call in `onUnload` when [`reason` is `uninstall`](https://github.com/chuckharmston/testpilot/blob/e41ddb458c7b16edb37988ef3ce35311c9d9da04/addon/index.js#L381). 

Doing this on all `onUnload` cases [includes `shutdown`, `upgrade`, and `downgrade`](https://developer.mozilla.org/en-US/Add-ons/SDK/Tutorials/Listening_for_load_and_unload#exports.onUnload%28%29). That means you'll delete your tracking state on every browser shutdown or add-on upgrade and forget you already showed the prompt.

In retrospect, I think `destroy` is a terrible name for this pattern in the add-on modules :/ `teardown` might be better.

lmorchard (changes requested):
Looks good except for the one kink I saw with `onUnload` and `destroy`.

ckprice (approved):
Approving the collections part of this. Everything in #1099 (minus the `pageview` calls as you mentioned) LGTM 👍  Events are firing, and shared URL's have the proper campaign tags.

addon/lib/share-prompt.js
dannycoates: All the data in `store` gets deleted anyway on uninstall, right? I'd suggest just removing `destroy` altogether.

lmorchard: No, it doesn't, and that's one of our ongoing issues. The SDK doesn't auto-delete the data in store. So, if you uninstall our add-on, it sticks around. If you re-install, you don't get a clean slate.

dannycoates: > The SDK doesn't auto-delete the data in store

Oh interesting... That's not the behavior I've observed on another addon. 🤔

chuckharmston: That's a really good catch, @lmorchard.

chuckharmston:
@ckprice It may be worth trying to get https://mzl.la URLs for each different share URL, since they're long and sort of ugly with the `utm_`s. That domain is administered as a custom bit.ly domain for `mozilla.org`. Since we're on `firefox.com`, we'd need an administrator for the bit.ly account to create them manually. Maybe you know who to talk to about that?

chuckharmston:
Ready for a re-review.

lmorchard:
Will re-review today

addon/lib/share-prompt.js
```diff
@@ -0,0 +1,75 @@
...
+        hasBeenShown: false
+      };
+    }
+    if (this.hasExperimentInstalled() && !store.sharePrompt.showAt) {
+      store.sharePrompt.showAt = Date.now() + this.calculateDelay();
+    }
+  },
+
+  destroy: function(reason) {
+    if (reason === 'uninstall') {
```
lmorchard, line 62: Ah, cool, better to move this reason handling here than doing it in index.js

lmorchard (approved):

PR #1297, opened 2016-08-26
Backfill unit tests for add-on experiment notifications

lmorchard:
Just to be sure this stuff does what I've been saying it does, here are some unit tests!

### DIFF
--- a/addon/test/test-experiment-notifications.js
+++ b/addon/test/test-experiment-notifications.js
@@ -1,0 +1,224 @@
+/*
+ * This Source Code is subject to the terms of the Mozilla Public License
+ * version 2.0 (the "License"). You can obtain a copy of the License at
+ * http://mozilla.org/MPL/2.0/.
+ */
+// const self = require('sdk/self');
+const { before } = require('sdk/test/utils');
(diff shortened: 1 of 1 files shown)


dannycoates:
Reading through these did raise some questions in my head about the design, but as far as confirming the existing behavior it looks good to me. I'll raise those questions in another issue.

lmorchard:
> Reading through these did raise some questions in my head about the design, but as far as confirming the existing behavior it looks good to me. I'll raise those questions in another issue.

Cool! That's actually a super great outcome of writing these tests. Not sure I'm happy with how this came together, so improvements are totally welcome

jaredhirsch:
This all looks good to me as well.

Couple very minor annoyances:
- I do wish the test runner were a little less chatty by default, or maybe used a junit-style dot reporter, but it's a minor thing. Probably better to have more detail vs. less when reading unit test failure logs on CI anyways.
- Initially, I got a pretty bizarre-looking xvfb error on Linux, because xvfb wasn't installed. Might be good to add that line to the addon's README development section.

Anyhow, nice work @lmorchard :beers:, merging

PR #1313, opened 2016-08-30
Split usage counts out of experiments JSON API

### LINKED ISSUE #1312: Split usage counts out of experiments JSON API
While we work toward #1039, an interim step would be to move the experiment usage counts out of the experiments content JSON and into its own resource. That way, we could publish the site & JSON statically and refer to just the dynamic data at a separate URL.

That JSON could look something like this:

``` json
{
  "wayback_machine@mozilla.org": 18713,
  "@activity-streams": 43608,
  "tabcentertest1@mozilla.com": 29117,
  "universal-search@mozilla.com": 35062
}
```

That is, an object with properties named for experiment add-on IDs pointing at the counts.

This needs backend changes to create the API resource and frontend changes to fetch & use it. Hopefully this format can be carried over to the solution for #1039 - in which case we'd only need to change what URL gets fetched!

lmorchard:
- Create /api/experiments/usage_count.json server resource
- Switch to using usage counts resource in frontend
- Produce static api/experiments/usage_counts.json

Fixes #1312

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -282,12 +282,13 @@ function writeExperimentYAML(experiment) {
 
 gulp.task('experiments-json', function generateStaticAPITask() {
   return gulp.src(CONTENT_SRC_PATH + 'experiments/*.yaml')
-    .pipe(buildExperimentsJSON('experiments'))
+    .pipe(buildExperimentsJSON())
     .pipe(gulp.dest(DEST_PATH + 'api'));
 });
 
-function buildExperimentsJSON(path) {
+function buildExperimentsJSON() {
   const index = {results: []};
+  const counts = {};
 
   function collectEntry(file, enc, cb) {
     const yamlData = file.contents.toString();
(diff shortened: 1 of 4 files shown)


lmorchard:
Added a tweak to gulpfile.js to build a static api/experiments/usage_counts.json resource

lmorchard:
Hmm, seems like the install integration test is failing, but not sure why :(

jaredhirsch:
@lmorchard Yeah, this failure is weird. The broken test is passing on my ubuntu machine.

Anyway, if this error is to be believed:

``` sh
TEST-UNEXPECTED-ERROR | test_installation.py 
TestAddonInstallation.test_initial_installation | NoSuchElementException: 
NoSuchElementException: Unable to locate element: button[data-hook=install]
```

then the first thing to try might be setting a super long implicit wait (via `set_search_timeout`), or a `sleep` if you'd rather, to see if it's a simple timing / waiting error or something deeper. If it's just a timing issue, then adding an explicit wait at the previous line is probably the way to go.

You could also try grabbing all the buttons out of the DOM via `find_elements`, and console.logging their data-hook attributes, just above the busted line. That might give you a clue into what's in the DOM when the error occurs.

Another kind of desperation move might be taking screenshots just before the error's thrown--dunno if this is supported by circle-ci, but [marionette does provide it](http://marionette-client.readthedocs.io/en/latest/reference.html?highlight=screenshot#marionette_driver.marionette.Marionette.screenshot). I used to resort to this with webdriver + sauce labs, it can sometimes immediately tell you what's up.

lmorchard:
It might actually be a real test failure - I'm seeing JS errors reproduced locally when I turn off NODE_ENV=development when I build the frontend with this PR. When that happens, there's really no install button for the integration test to find

lmorchard:
Alright, rewrote the bit that fetches usage counts in the frontend, and that seems to have fixed the local error and made the integration test pass. Hooray!

lmorchard:
Hot off the IRC presses - we also have a JSON resource from telemetry with counts!

https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json

I think we should still roll with this PR as-is. But, when we get closer to the point of deploying static files, switching to this Telemetry resource is just a matter of changing `usageCountsUrl`

lmorchard:
closing in favor of #1321

PR #1318, opened 2016-09-01
1305 static deploy

lmorchard:
This piggybacks on PR #1313, so review / merge that one first. But, this could use some eyes now. 
- circle.yml changes to support both Docker and static build branches
- bugfix to static build to exclude rewriting references to addon
- integration test tweak to accept trailing slash on /onboarding URL

Basically, this supports the usual Docker build for most branches. But, it adds static-only build & deploy to S3 for pre-defined static branches. 

So far, the static branches are `1305-static-deploy` (for this PR) and `development-static`. Results of deployment are found here:
- http://testpilot-static.lmorchard.mozaws.net.s3-website-us-east-1.amazonaws.com/
- http://testpilot-static.dev.mozaws.net.s3-website-us-east-1.amazonaws.com/

Still have some work to do here, including setting up CloudFront for SSL & a better hostname on development and also adding new environments to the main add-on so it supports these new hosts.

### DIFF
--- a/bin/circleci/build-frontend.sh
+++ b/bin/circleci/build-frontend.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 set -ex
 npm install
-npm run build
+if [[ " $TESTPILOT_STATIC_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
+    npm run static
+else
+    npm run build
+fi
 npm test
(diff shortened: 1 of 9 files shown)


lmorchard:
Oh and also: This should be okay to merge on master, since the circle.yml should do the normal thing on master and all non-static branches. If that breaks, then there are bugs to be fixed.

lmorchard:
closing in favor of #1321

Issue #525, opened 2016-03-24
Refine dev server auto-deployment to support additional servers mapped to branches


lmorchard:
Using Circle CI variables and rules in the .yml, it's could be possible to spin up additional dev servers on Elastic Beanstalk and map them to different branches. 

For example, if we had a `ux-research` branch, changes committed there could be auto-deployed to a `testpilot-ux-research` server.

lmorchard:
This will be super-easy to do with static deployment on circleci and doesn't need special work or research

PR #1320, opened 2016-09-01
Add static-dev and static-stage envs for add-on

### LINKED ISSUE #1319: Add new static dev environments to add-on
I propose two new hosts for our static deployment track:
- https://testpilot-static.dev.mozaws.net
- https://testpilot-static.stage.mozaws.net

The add-on should support both of these in the `testpilot.env` preference

lmorchard:
Fixes #1319


lmorchard:
I'm not very creative, so I just turned the dev & stage critters upside down for the static counterparts in groundcontrol

dannycoates:
👍

Side note: I'm trying out an 'any' [option](https://github.com/dannycoates/x16/commit/e3e9a97df05cf1113294b7d3b89d4023a8f57117) in x16. It makes the addon work with any page that uses the `message-bridge`. It isn't meant to replace the presets, but allows you to try the addon with "unofficial" txp deployments, like say for a deployment per PR branch.

lmorchard:
Think I'll merge this. We should circle back and think about adding an 'any' option, because that would more easily let us deploy many environments - e.g. for UX testing for issue #525

lmorchard:
Filed #1325 as a followup

Issue #1307, opened 2016-08-29
Convert to React views & Redux state management from Ampersand views & models


lmorchard:
This work in progress has been merged into the [development-static](https://github.com/mozilla/testpilot/tree/development-static) branch and [is auto-deploying](https://circleci.com/gh/mozilla/testpilot/tree/development-static) to https://testpilot-static.dev.mozaws.net

if you set `testpilot.env` to `static-dev` in about:config, [this environment](https://testpilot-static.dev.mozaws.net) should fully work. Anything not listed as a known issue in the TODOs below is something new and needs to be fixed - so feel free to edit this & add to the list!

To run the [development-static](https://github.com/mozilla/testpilot/tree/development-static) branch locally, just check it out and then run this:

```
cd addon
npm install && npm run package
cd ..
npm install
while [ 1 ]; do NODE_ENV=development gulp; sleep 1; done
```

Docker is no longer needed. Just build the add-on and start up gulp. The while loop is to re-start gulp when it exits, which it will do if it sees edits to gulpfile.js or build tasks.

**Note**: If you get an ["Unexpected token" error](https://pastebin.mozilla.org/8909837) when trying to run gulp, make sure you're running at least Node v6.2.0.

Remaining TODOs:
- [x] Auto-deployment does not invalidate CloudFront distribution, so things stay stale after deployment
- [x] Links in Settings menu do the wrong things, retire flow is broken.
- [x] Port over [the new mozAddonManager code](https://github.com/mozilla/testpilot/commit/2c570b23840c1afbf06a298c39f3cdd8375c99f6) for add-on installation
- #1349 <strike>[ ] Try building with React in production mode, to be sure everything still works</strike>
- #1350 <strike>[ ] Update docs to describe new static generation & React architecture. No more Docker.</strike>
- #1351 <strike>[ ] Frontend tests - port over existing? start fresh with [Mocha](https://mochajs.org/) + [Enzyme](http://airbnb.io/enzyme/)?</strike>

johngruen:
@lmorchard: settings menu should only appear if the user has the add-on installed

lmorchard:
> settings menu should only appear if the user has the add-on installed

That's what I'm seeing... are you seeing something different?

dannycoates:
I also see settings without the addon

ckprice:
Can't see a couple GA events/pageviews firing
[...]

Issue #1306, opened 2016-08-29
Implement static page rendering with progressive enhancement


lmorchard:
Probably way too big an issue to file as such, but here goes:

We should build a static HTML rendering of all pages - experiment pages, legal pages, everything - in a way that works for most browsers.

Then, when the JS kicks in and we detect Firefox, we can flip on/off relevant bits. Same goes for whether we do or don't detect the Test Pilot add-on as installed.

Ideally, it would be great if we could statically render React templates and dynamically render the JS-dependent bits also within React. Not-as-ideally, we could render the base cases with something like Jekyll and then sprinkle in React magic after load.

lmorchard:
Also, possibly blocked by #1307 (React conversion) unless we want to jump straight at Jekyll / Liquid Template rendering.

Prior art: https://github.com/chuckharmston/testpilot-flatfiles

fzzzy:
I'm actually most of the way done with this at https://github.com/fzzzy/testpilot/tree/remove-react-router , so I'll close this as a dup of https://github.com/mozilla/testpilot/issues/2260

PR #1321, opened 2016-09-03
Changes for static build & deployment

lmorchard:
Okay, this one is pretty big, because I'm rolling up my PRs so far this week and then some. Main highlights:
- Split usage counts out of main /api/experiments.json so we can switch URLs to [the Telemetry-based JSON](https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json) and solve #1312 
- Reorganizing gulpfile.js in advance of replacing the frontend with React for #1307
- Moved all the frontend source files out of the server app, anticipating getting get rid of Django
- Static generation of landing pages with social sharing metadata, not quite solving #1306 but also not regressing from the current site
- Dual-purpose Circle CI build - we can specify branches for static-only build / test / deploy without Docker server build / test / deploy

Just to make sure this stuff is working in CircleCI, this branch is publishing here:
- http://testpilot-static.lorchard.mozaws.net.s3-website-us-east-1.amazonaws.com/

Can't do a full installation because there is no `testpilot.env` corresponding to that host in the add-on, but you can at least see the experiment detail pages.


dannycoates:
This looks really nice @lmorchard. Breaking up the gulp file makes each task very clear.

johngruen:
Maybe gulp-king @pdehaan could give this a glance?

circle.yml
```diff
@@ -12,9 +12,10 @@ machine:
     testpilot.dev: 127.0.0.1
   services:
     - docker
-
   node:
     version: 6.2.0
+  environment:
+    TESTPILOT_STATIC_BRANCHES: development-static 1305-revamped-static
```
pdehaan: what is "1305-revamped-static"? seems oddly specific.

frontend/config.js
```diff
@@ -0,0 +1,24 @@
...
+  PRODUCTION_EXPERIMENTS_URL: 'https://testpilot.firefox.com/api/experiments',
+  IMAGE_NEW_BASE_PATH: 'images/experiments/',
+  IMAGE_NEW_BASE_URL: '/static/images/experiments/',
+
+  'sass-lint': true,
+  'js-lint': true
+};
+
+// Pull in local debug-config.json overrides
+const tryRequire = require('try-require');
```
pdehaan: this looks... malformed.

circle.yml
lmorchard: It's the specific branch that this PR is on in my repo, deploying to my throwaway S3 bucket for a demo of this PR:

http://testpilot-static.lorchard.mozaws.net.s3-website-us-east-1.amazonaws.com/

frontend/config.js
lmorchard: I don't know what that means

frontend/src/vendor/babel-polyfill.js
```diff
@@ -0,0 +1,4 @@
...
...n i(t),o?o(t):!0}})},{32:32,7:7}],210:[function(t,n,r){var e=t(32);e(e.S,"Reflect",{ownKeys:t(80)})},{32:32,80:80}],211:[function(t,n,r){var e=t(32),i=t(7),o=Object.preventExtensions;e(e.S,"Reflect",{preventExtensions:function preventExtensions(t){i(t);try{return o&&o(t),!0}catch(n){return!1}}})},{32:32,7:7}],212:[function(t,n,r){var e=t(32),i=t(90);i&&e(e.S,"Reflect",{setPrototypeOf:function setPrototypeOf(t,n){i.check(t,n);try{return i.set(t,n),!0}catch(r){return!1}}})},{32:32,90:90}],213:[function(t,n,r){function set(t,n,r){var c,l,h=arguments.length<4?t:arguments[3],v=i.f(a(t),n);if(!v){if(s(l=o(t)))return set(l,n,r,h);v=f(0)}return u(v,"value")?v.writable!==!1&&s(h)?(c=i.f(h,n)||f(0),c.value=r,e.f(h,n,c),!0):!1:void 0===v.set?!1:(v.set.call(h,r),!0)}var e=t(67),i=t(70),o=t(74),u=t(39),c=t(32),f=t(85),a=t(7),s=t(49);c(c.S,"Reflect",{set:set})},{32:32,39:39,49:49,67:67,7:7,70:70,74:74,85:85}],214:[function(t,n,r){var e=t(38),i=t(43),o=t(67).f,u=t(72).f,c=t(50),f=t(36),a=e.RegExp,s=a,l=a.prototype,h=/a/g,v=/a/g,p=new a(h)!==h;if(t(28)&&(!p||t(34)(function(){return v[t(117)("match")]=!1,a(h)!=h||a(v)==v||"/a/i"!=a(h,"i")}))){a=function RegExp(t,n){var r=this instanceof a,e=c(t),o=
```
pdehaan: Where is this file from? Is this something we can install via npm and copy over at build time instead of including in source control?

frontend/tasks/content.js
```diff
@@ -0,0 +1,150 @@
...
+  Object.keys(imageFields).forEach(key => {
+    const items = (key === '') ? [experiment] : experiment[key];
+    const fields = imageFields[key];
+    items.forEach(item => fields.forEach(field => {
+      // Grab the original image URL, bail if it's not available
+      const origURL = item[field];
+      if (!origURL) { return; }
+
+      // Chop off the protocol & domain, convert gravatar param to .jpg
+      const path = origURL.split('/').slice(3).join('/').replace('?s=64', '.jpg');
```
pdehaan, line 105: Probably well out of scope of this PR, but I had to read that `.split().slice(3).join()` a couple times to figure it out. Not sure if it's easier to use `url.parse().path` instead (only difference seems to be the leading "/"):

``` js
const url = require('url');

const origURL = "https://msn.com/foo/bar/baz/image?s=64";

const path = origURL.split('/').slice(3).join('/').replace('?s=64', '.jpg');
const path2 = url.parse(origURL).path.replace('?s=64', '.jpg');

console.log(path);  // "foo/bar/baz/image.jpg"
console.log(path2); // "/foo/bar/baz/image.jpg"
```

gulpfile.js
```diff
@@ -1,455 +1,63 @@
...
 
-gulp.task('default', ['build', 'watch']);
+require('./frontend/tasks/content');
+require('./frontend/tasks/scripts');
+require('./frontend/tasks/styles');
+require('./frontend/tasks/images');
+require('./frontend/tasks/assets');
+require('./frontend/tasks/pages');
+require('./frontend/tasks/server');
+require('./frontend/tasks/dist');
```
pdehaan, line 19: :clap: Much cleaner!

frontend/src/vendor/babel-polyfill.js
pdehaan: Ah, I think this may be _./node_modules/babel-polyfill/dist/polyfill.min.js_.

Not sure if it's worth using some Gulp task or npm `postinstall` hook, to copy the file from _./node_modules/_ into _./frontend/src/vendor/*_ and then maybe we .gitignore that directory.

frontend/config.js
pdehaan: Sorry, for some reason my computer was missing the closing `)`, or else my eyes are starting to go. But it looks fine now, so I'll delete these useless and now confusing comments.

frontend/src/vendor/babel-polyfill.js
lmorchard: Hmm, actually, we might not need this at all because [we import `babel-polyfill/browser` in `main.js`](https://github.com/lmorchard/testpilot/blob/6c81c08e63117f3e27500275ef548ab2f54480a4/frontend/src/app/main.js#L1). This might be an old artifact from when I was wrestling with getting l20n to work.

Updated commit to delete this!

frontend/tasks/content.js
lmorchard: Will probably leave that alone... It's a dirty dirty hack from a few PRs ago, but will go away once we do the last import from the Django-based site.

circle.yml
lmorchard: And just updated the commit to remove this, since I want to merge soon and this doesn't belong in the actual project

lmorchard:
Alright, let's merge this

Issue #1329, opened 2016-09-08
Use SSL for local development.


chuckharmston:
In #1223 I am introducing some functionality that only works over HTTPS. I'm working around it with a proxy for now, but it'd be nice if we shipped with some dummy certs or LE configuration that let us use HTTPS for local work without configuration.

lmorchard:
See also, for dev on elastic beanstalk (not a dupe, but also important): https://github.com/mozilla/testpilot/issues/548

lmorchard:
I remember we did this with MDN, but via a full Apache instance in Vagrant. Let's not do that here. Maybe we can use something like [django-sslserver](https://github.com/teddziuba/django-sslserver)?

lmorchard:
Going to say this is closed since #1394 was merged. Could use some docs though

PR #1330, opened 2016-09-09
Use mozAddonManager to install Test Pilot extension when available.

### LINKED ISSUE #1223: Use the amazing new mozAddonManager API
Ok, using something newer than Firefox 48 (I think) go to about:addons, click "Get Add-ons" in the menu on the left and then enable/disable some add-ons that you see there.  See how amazing and easy it is to enable add-ons?  We want that!

Good news!  [Bug 1288178](https://bugzilla.mozilla.org/show_bug.cgi?id=1288178) gave testpilot.firefox.com access to that API.  We're going to try to get it uplifted -- if we get it into beta, it will go live on **Sep 13** for general release users.  That means we'll need to do some detection to support the old method of installing as well as the new method.  That can be done with `if (navigator.mozAddonManager)`.

Before this will work on our site, we'll need to ensure that the add-on .xpi URLs live on testpilot.firefox.com/files/.  [Bug 1291357](https://bugzilla.mozilla.org/show_bug.cgi?id=1291357) is taking care of proxying that to our CDN, but we'll still have to get the other files up there.

I'm filing this bug a bit early to get it on the radar and mark it as needs:ux in case we want to adjust our install buttons at all to be more similar to the disco pane sliders.

/cc @johngruen @sevaan

chuckharmston:
This works great ~~, except for one blocking problem, which is filed here: https://bugzilla.mozilla.org/show_bug.cgi?id=1301609~~. I didn't have `testpilot.env` set appropriately. Works great, no longer WIP.

Testing this locally is...complex. I'll follow up with a comment with full instructions.

Closes #1223

### DIFF
--- a/frontend/src/app/lib/install-addon.js
+++ b/frontend/src/app/lib/install-addon.js
@@ -3,19 +3,53 @@ import cookies from 'js-cookie';
 import EmailDialogView from '../views/email-opt-in-dialog-view';
 
 
+const MOZADDONMANAGER_ALLOWED_HOSTNAMES = [
+  'testpilot.firefox.com',
+  'testpilot.stage.mozaws.net',
+  'testpilot.dev.mozaws.net'
+];
+
+
+function mozAddonManagerInstall(url) {
+  return navigator.mozAddonManager.createInstall({ url }).then(install => {
+    return new Promise((resolve, reject) => {
+      install.addEventListener('onInstallEnded', () => resolve());
+      install.addEventListener('onInstallFailed', () => reject());
+      install.install();
+    });
+  });
+}
+
 function installAddon(view, eventCategory, callback) {
-  const downloadUrl = '/static/addon/addon.xpi';
+  const { protocol, hostname, port } = window.location;
+  const path = '/static/addon/addon.xpi';
+  const downloadUrl = `${protocol}//${hostname}${port ? ':' + port : ''}${path}`;
 
   view.query('[data-hook=install]').classList.add('state-change');
   view.query('.default-btn-msg').classList.add('no-display');
   view.query('.progress-btn-msg').classList.remove('no-display');
 
-  app.sendToGA('event', {
+  const useMozAddonManager = (
+    navigator.mozAddonManager &&
+    protocol === 'https:' &&
+    MOZADDONMANAGER_ALLOWED_HOSTNAMES.includes(hostname)
+  );
+
+  const gaEvent = {
     eventCategory: eventCategory,
     eventAction: 'button click',
-    eventLabel: 'Install the Add-on',
-    outboundURL: downloadUrl
-  });
+    eventLabel: 'Install the Add-on'
+  };
+
+  if (useMozAddonManager) {
+    app.sendToGA('event', gaEvent);
+    mozAddonManagerInstall(downloadUrl).then(() => {
+      console.log('Installed extension via mozAddonManager');
+    });
+  } else {
+    gaEvent.outboundURL = downloadUrl;
+    app.sendToGA('event', gaEvent);
+  }
 
   // Wait for the add-on to be installed.
   // TODO: Should we have a timeout here, give up after a few intervals? If


chuckharmston:
To test:
1. Use Firefox >= 49.
2. Create a new profile.
3. Install [DevPrefs](https://addons.mozilla.org/en-US/firefox/addon/devprefs/).
4. Set `testpilot.env` to `production`.

`mozAddonManager` only works for a set of whitelisted hostnames. It also requires SSL, which our dev environment does not currently handle (#1329). So to test this PR, you need to use a proxy to map requests from `https://testpilot.firefox.com` to `http://testpilot.dev:8000`.
1. Install [Charles](https://www.charlesproxy.com/).
2. Install the [Charles proxy configuration extenstion](https://addons.mozilla.org/en-US/firefox/addon/charles-proxy/).
3. Install the [Charles certificate](https://www.charlesproxy.com/documentation/using-charles/ssl-certificates/) in Firefox.
4. Set `extensions.install.requireBuiltInCerts`, to allow Firefox to install extensions from sites with SSL certs that are not built-in.
5. In Charles, enable SSL proxying for Test Pilot. `Proxy > SSL Proxying Settings`, then [add a record for Test Pilot](https://cloud.githubusercontent.com/assets/23885/18376028/810b372a-7618-11e6-8284-58ad85d91c8b.png).
6. In Charles, compose a `GET` request for Test Pilot by hitting the pen icon in the toolbar. I don't know why it's worded like this, but the record should [look like this](https://cloud.githubusercontent.com/assets/23885/18376131/47e5a93e-7619-11e6-9bd1-504e2e6a5aed.png).
7. Right click on the new record in the sidebar and hit "Map Remote". In the new form, [configure it like this](https://cloud.githubusercontent.com/assets/23885/18376230/674cb88e-761a-11e6-8b6a-5ef17fe4640d.png).

From here, requests to `testpilot.firefox.com` should actually hit your local dev environment while Charles is running. To ensure that this is working properly, I recommend replacing the logo/wordmark in `templates/landingpage-js` with ``.

Some caveats: make sure your dev site is running, updated, and is not cached. It may well not be, due to recent changes that Les is working on for TXP-30. To be sure, I recommend updating master and fully rebuilding your Docker containers. Also make sure that you've built the addon (`npm run package` in `/addon`) and that the frontend watcher has picked it up and moved it to the frontend `src` directory.

From here, you should be able to visit testpilot.firefox.com (with fancy unicorn indicating that it's actually your local dev site) and hit "install". ~~The install button will spin forever due to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1301609), but the addon should otherwise install and initialize correctly without prompts.~~ Check for the toolbar button! Attempts to install the addon from `testpilot.dev:8000` should give you the regular prompts, but still work.

Infinite good vibes to and at least one 🍹 for @mstriemer in Hawaii for helping me work through this.

chuckharmston:
I'm confused at the build failure. Wouldn't [these lines](https://github.com/mozilla/testpilot/blob/6bd599eb1c3396883a7a98acaf6671dd49e03132/frontend/src/app/main.js#L4-L5) make `Promise` generically available anywhere in the built file?

chuckharmston:
Also: best I can tell, we don't currently have tests for this bit of code, so they've been omitted here.

lmorchard:
> Also: best I can tell, we don't currently have tests for this bit of code, so they've been omitted here.

FWIW, [we _do_ have a test for add-on installation](https://github.com/mozilla/testpilot/blob/master/integration/test_installation.py#L35) - our only [integration test](https://github.com/mozilla/testpilot/blob/master/docs/development/testing.md#integration-tests) so far. Would be nice to find a way to rig this new installation path up for testing, too.

Will give a try at the rest of this locally probably sometime this afternoon when meetings have cleared

lmorchard:
> I'm confused at the build failure. Wouldn't these lines make Promise generically available anywhere in the built file?

I think occasionally, the test harness fails with that. Not sure why. I just hit the Rebuild button on CircleCI to give it another try. When in doubt, if a failure looks weird, try that

chuckharmston:
> FWIW, we do have a test for add-on installation. Would be nice to find a way to rig this new installation path up for testing, too.

I don't think integration tests are the right place for this; we need to mock browser APIs, observe calls, and fake responses. Client unit tests seem more appropriate to me; filed #1334 for that.

lmorchard:
And it looks like on a re-run, [it's the integration test that fails](https://circleci.com/gh/mozilla/testpilot/913)

lmorchard:
> I don't think integration tests are the right place for this; we need to mock browser APIs, observe calls, and fake responses.

The integration test ensures that the add-on install & onboarding process happens as expected, at least from an outside-in user's testing perspective. Using this new mozAddonManager API changes that flow. We should be able to at least assert what that new flow looks like from a user perspective

chuckharmston:
Happy to be pointed to methods for mocking browser APIs or running multiple versions of Firefox to check both code paths. I'm not personally familiar with anything.

lmorchard:
> Happy to be pointed to methods for mocking browser APIs

I'd need to get pointed to that first to point you to it :)

> or running multiple versions of Firefox to check both code paths. I'm not personally familiar with anything.

For CircleCI, at least, I think this would be a matter of [downloading multiple unbranded versions](https://github.com/mozilla/testpilot/blob/master/bin/circleci/setup-test-dependencies.sh#L6) and then [running the integration suite](https://github.com/mozilla/testpilot/blob/master/bin/circleci/run-integration-tests.sh#L27) against each downloaded binary.

lmorchard:
To be fair, though, those tweaks to CI are substantial. On the other hand, changing how our main add-on gets installed is a pretty big deal and we've been suffering from bugs in untested code paths. So, I'm kind of ¯\(°_o)/¯ about being a jerk about tests in this particular PR. 

I haven't had a chance to review this locally yet, so if that works out we can probably defer CI tests to a follow up issue in the interest of getting this merged

chuckharmston:
The other problem is that we'd need to somehow get CI to fake out the hostname to one of the whitelisted ones, get SSL working in tests, and change Firefox prefs to allow checks to non-built in CAs. I'm not sure how that'd even work here.

lmorchard:
> The other problem is that we'd need to somehow get CI to fake out the hostname to one of the whitelisted ones. I'm not sure how that'd even work here.

That part doesn't seem too bad - [we can set /etc/hosts entries in circle.yml](https://github.com/mozilla/testpilot/blob/master/circle.yml#L12)

chuckharmston:
Yeah, I submitted that comment too early. Added two more bits:

> get SSL working in tests, and change Firefox prefs to allow checks to non-built in CAs

frontend/src/app/lib/install-addon.js
```diff
@@ -3,19 +3,52 @@ import cookies from 'js-cookie';
...
+  return navigator.mozAddonManager.createInstall({ url }).then(install => {
+    return new Promise((resolve, reject) => {
+      install.addEventListener('onInstallEnded', () => resolve());
+      install.addEventListener('onInstallFailed', () => reject());
+      install.install();
+    });
+  });
+}
+
 function installAddon(view, eventCategory, callback) {
```
jaredhirsch, line 23: @chuckharmston I think if you pass in `navigator` here, you'll easily be able to mock out `mozAddonManager` for testing.

It [looks like](https://github.com/mozilla/testpilot/blob/master/frontend/src/test/browser.js#L8) the existing frontend tests expect a test/lib directory to exist, so you should be able to just copy an existing test into a new lib directory and add the happy path test.

Happy to help further if you'd like; let me know

lmorchard:
Finally clear of some meetings and trying to run this locally. Not having much luck with [an unbranded Nightly 49 build](http://archive.mozilla.org/pub/firefox/tinderbox-builds/mozilla-beta-macosx64-add-on-devel/1472838649/) and Charles 4.0 yet. I just get [a weird blank window](https://www.dropbox.com/s/c39139kfnx5n1yx/Screenshot%202016-09-09%2015.25.58.png?dl=0) when I hit Tools > Charles Proxy > Install Charles Root Certificate. Anyone else trying to work through this? Going to try another build of Firefox and see if this repeats there.

Also couldn't get Charles working in Firefox Beta 49. Firefox Dev Edition 50.0a2 seems finally to have worked with Charles, though. Let's see if the rest of things work...

lmorchard:
Okay, after finally getting through all the steps using Firefox Dev Edition 50.0a2 - and having @chuckharmston remind me that `extensions.install.requireBuiltInCerts` should be **`false`** - it appears that everything works locally for me. 

Also tried with Firefox 47.0 and a signed add-on, and install worked with the old code path. Also, I see the integration test on CircleCI with unbranded Firefox 48.0.1 just passed.

So, that seems pretty good. I'll say r+ and we can save boiling the ocean for automated tests in follow up issues already filed.

PR #1338, opened 2016-09-12
Fix paths to privacy & terms pages in Django

lmorchard:
These things moved with the changes to static build & deployment. Derp.


dannycoates:
LGTM

PR #1341, opened 2016-09-13
close settings menu on item click

dannycoates:


lmorchard:
r+ will merge after circleci is green

PR #1348, opened 2016-09-14
Sort experiment cards by id, extract experiment list into component

lmorchard:
No issue, but following up @johngruen's [last comment on 1307](https://github.com/mozilla/testpilot/issues/1307#issuecomment-247060129)

### DIFF
--- a/frontend/src/app/components/ExperimentCardList.js
+++ b/frontend/src/app/components/ExperimentCardList.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import ExperimentRowCard from './ExperimentRowCard';
+import Loading from './Loading';
+
+export default function ExperimentCardList({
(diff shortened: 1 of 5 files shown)


dannycoates (approved):
This LGTM. I find the whole `isExperimentEnabled` thing a bit awkward, but not a blocker.

lmorchard:
> This LGTM. I find the whole isExperimentEnabled thing a bit awkward, but not a blocker.

Yeah, I'm trying to find a better way to do that. What we had before (`experiment.enabled`) is a little harder here, because actually that info is represented in the installed list we get from the add-on. That list and the list of experiments each come in async from each other and they can change, so it's a little hard to mash them together & keep them in sync. That `isExperimentEnabled` thing is a helper I made as a compromise

lmorchard:
Hmm, this reviewer thing is kind of neat

PR #1353, opened 2016-09-14
make survey optional after rating

### LINKED ISSUE #1327: Make feedback survey optional after heartbeat trigger
Currently we pop a survey straight from the heartbeat rating.
Let's see what happens if we issue a 'thank you' then give users the choice to take a quick survey.

dannycoates:
fixes #1327 

Testing this is currently as hard as it ever was. What I did was comment out the restrictive bits of  `init` to:

``` js
function init() {
  // wait about 10 minutes before prompting user for survey
  // setTimeout(() => {
    // Only check/ask for survey if the user has addons installed.
    if (store.installedAddons && Object.keys(store.installedAddons).length) {
      const experiment = getRandomExperiment();
      const interval = checkInstallDate(experiment.installDate, experiment.addon_id);
      // if (!interval) return;
      launchSurvey(experiment, interval);
    }
  // }, TEN_MINUTES);
}
```

Maybe an about:config flag would be a good way to force it to show...

### DIFF
--- a/addon/lib/survey.js
+++ b/addon/lib/survey.js
@@ -9,8 +9,9 @@
  *   https://github.com/lmorchard/fireriver/blob/master/lib/main.js#L369
  *   & https://dxr.mozilla.org/mozilla-central/source/browser/components/uitour/UITour.jsm
  */
-const {Cc, Ci} = require('chrome');
-const {setTimeout} = require('sdk/timers');
+
+const { Services } = require('resource://gre/modules/Services.jsm');
+const { setTimeout, clearTimeout } = require('sdk/timers');
 const tabs = require('sdk/tabs');
 const querystring = require('sdk/querystring');
 const store = require('sdk/simple-storage').storage;
@@ -18,7 +19,7 @@ const store = require('sdk/simple-storage').storage;
 const NUM_STARS = 5; // Number of survey stars
 const TEN_MINUTES = 60 * 1000 * 10;
 
-module.exports = {init: init, destroy: destroy};
+module.exports = {init, destroy, launchSurvey};
 
 // set timecheck flags for initial run.
 if (store.surveyChecks === undefined) store.surveyChecks = {};
@@ -31,7 +32,10 @@ function init() {
   setTimeout(() => {
     // Only check/ask for survey if the user has addons installed.
     if (store.installedAddons && Object.keys(store.installedAddons).length) {
-      showRandomSurvey(getRandomExperiment());
+      const experiment = getRandomExperiment();
+      const interval = checkInstallDate(experiment.installDate, experiment.addon_id);
+      if (!interval) return;
+      launchSurvey(experiment, interval);
     }
   }, TEN_MINUTES);
 }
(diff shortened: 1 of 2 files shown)


lmorchard:
> Testing this is currently as hard as it ever was.

Haven't skimmed the code yet, but is there a way you could wrap this stuff in unit tests?

dannycoates:
@rayborn I added the `surveyClicked` parameter which will be `true` if the user clicked the button and `false` if they didn't. A complete url will look something:

https://qsurvey.-------.com/s3/no-more-404s?id=wayback_machine%40mozilla.org&installed%5B0%5D=wayback_machine%40mozilla.org&rating=5&interval=2&surveyClicked=false

dannycoates:
> is there a way you could wrap this stuff in unit tests?

Probably, but it's mostly ui code. After mocking all that there isn't much unit left to test.

ghost:
Great, I'll take a look tomorrow morning (Ping me if you feel that this should be prioritized)

dannycoates:
@lmorchard I added some tests for `init`. I think the rest will require integration tests.

lmorchard (approved):
Did the quick tweaks to init to force display of the survey bar, tried rating without survey, tried with survey - everything works for me!

addon/lib/survey.js
```diff
@@ -81,61 +75,34 @@ function checkInstallDate(installDate, addonId) {
...
+    .then(
+      rating => {
+        const urlParams = {
+          id: experiment.addon_id,
+          installed: Object.keys(store.installedAddons),
+          surveyClicked: false,
+          rating,
+          interval
+        };
+        return showSurveyButton({ experiment })
```
lmorchard, line 84: Only nit I'd have here is rather than use resolve / reject to detect whether the button was clicked or not, resolve a boolean and switch on that. Promise rejection, to me, usually means we should bail out of the whole process because there was an error

dannycoates: Yeah, I generally agree, and I thought about doing it that way too. The trivial reason I went this way was so that the `rating` from `showRating` wouldn't need to be null checked. Then to be consistent with that I went the same way with `showSurveyButton`... A _VERY_ trivial difference that I'm not strongly attached to at all.

dannycoates: Actually you're right... rejecting a promise for expected control flow is bad form. I've changed it.

ghost:
I was confused on why this wasn't working, but then I read more of SG's documentation on partial survey responses and it turns out [partials are only recorded if users complete at least one page.](https://help.surveygizmo.com/help/response-status)

So the bad news is that we're not going to get these users, but the good news is that this is at parity with our current approach.   We can remove the GET since it's not going to be recorded.

Long term I'd still suggest Telemetry as the end-goal, but this is a step up from what we have, so I'm fine from a data perspective.

dannycoates:
@rayborn that's a bummer.

@chuckharmston @lmorchard Is it as easy as something like this to add the rating to telemetry?

``` js
Metrics.pingTelemetry({ addon_id, rating, interval }, 'rating', Date.now());
```

ghost:
Since we have parity, let's land this and file a separate issue to add telemetry.  I think it is as easy as your comment, but I'd like approval from ilana or sunah

chuckharmston:
> @chuckharmston @lmorchard Is it as easy as something like this to add the rating to telemetry?

Not quite, but close: https://github.com/mozilla/testpilot/blob/master/addon/lib/metrics.js#L117-L121

PR #1354, opened 2016-09-14
Turn off enable button animation when download ends in error

### LINKED ISSUE #1352: When a download fails the button remains in the enabling state.
[As reported](https://github.com/mozilla/testpilot/issues/1307#issuecomment-246845942) by @dannycoates in #1307:

> When a download fails the button remains in the enabling state.
> 
> screen shot 2016-09-13 at 2 54 14 pm

lmorchard:
Fixes #1352


lmorchard:
@dannycoates - I think this fixes the download button thing

lmorchard:
Updated commit to fix some linting goofs

dannycoates (approved):
The logic is sort of confusing, but it works. Onward and upward!

lmorchard:
Yeah, I kind of think most of what I'm doing here with looking up the experiment from slug should be handled somehow in the router, but my React-fu isn't strong enough to figure that out yet

PR #1357, opened 2016-09-15
Don't merge - Revert 2c570, 'Use mozAddonManager to install Test Pilot extension when available.'

chuckharmston:
Please don't merge this; I'm putting it up in the event that we are unable to get whitelisted.


chuckharmston:
Closing in favor of #1360

Issue #1350, opened 2016-09-14
Update docs to describe new static generation & React architecture


lmorchard:
It's all still kind of a mystery. And not the good kind.

PR #1361, opened 2016-09-15
Require ENABLE_PONTOON=1 during frontend build to include Pontoon JS in pages

### LINKED ISSUE #1356: Remove pontoon.js from production instance
The [pontoon.js](https://developer.mozilla.org/en-US/docs/Implementing_Pontoon_Mozilla) script is required to allow in-page localization of Test Pilot with [Pontoon](https://pontoon.mozilla.org/).

It only has to be loaded in the environment that is used for in-page localization, e.g. staging server. You can remove it from the production instance.

lmorchard:
Fixes #1356

### DIFF
--- a/bin/circleci/build-frontend.sh
+++ b/bin/circleci/build-frontend.sh
@@ -2,7 +2,11 @@
 set -ex
 npm install
 if [[ " $TESTPILOT_STATIC_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
-    npm run static
+    if [[ " $TESTPILOT_ENABLE_PONTOON_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
+        NODE_ENV=production ENABLE_PONTOON=1 npm run static
+    else
+        NODE_ENV=production ENABLE_PONTOON=0 npm run static
+    fi
 else
     npm run build
 fi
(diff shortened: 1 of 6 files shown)


lmorchard:
This should exclude the Pontoon script tag by default on all pages, _unless_ ENABLE_PONTOON=1 is set during frontend build. That's enabled in this PR for develpoment-static branch, but it should be off for any other deployments in the future.

You can test this locally by running gulp and verifying that there's no Pontoon script tag on the page. Restart with `ENABLE_PONTOON=1 gulp` and the script tag should appear.

chuckharmston (changes requested):
This looks great!

Only nit: this seems like the sort of thing we should be documenting.

ghost:
Thanks for cooking this up, Les and for reviewing it, Chuck

lmorchard:
> Only nit: this seems like the sort of thing we should be documenting.

Definitely a good point. One caveat is that the static build system isn't documented at all yet, so there's not really a good place to put just the mention of this flag right now. Hoping to work on those docs today for #1350

lmorchard:
May collide with PR #1366, but I added a section to deployment docs describing how ENABLE_PONTOON is used.

chuckharmston (approved):

Issue #1351, opened 2016-09-14
Update & re-enable frontend unit tests for React


lmorchard:
Everything's different. The old frontend tests don't work. We should either port them or write new ones. Maybe with [Mocha](https://mochajs.org/) + [Enzyme](http://airbnb.io/enzyme/)?

lmorchard:
Going to say this is closed, though frontend tests need to be backfilled now that we have a system to support them

PR #1365, opened 2016-09-16
Switch to Mocha / Chai / Enzyme for testing under React

lmorchard:
Issue #1351 

This isn't a full test suite, more a working proposal for how to do it going forward with Mocha, Chai, and Enzyme. Rather than go much further down the road and try to hero-code my way to a full suite, I think I'll just submit this as a PR now

[The second commit](https://github.com/mozilla/testpilot/pull/1365/commits/a9904e32ce078fe49cbe5cc7971dc0466d3fe6ab) is where the new stuff lives, if you want to see it without the noise of cleaning up the old test suite

These tests run in node with `npm test` or `npm run test:watch` and don't require a browser instance. I think that should be good enough for asserting things about the DOM and UI logic. It's also a lot easier to get running, IMO. If we have things we want to specifically test against Firefox, that could be more a job for integration tests & marionette


chuckharmston (approved):
This is awesome. :shipit:

Issue #1367, opened 2016-09-16
Move component state to Redux store


chuckharmston:
In most cases, things you'd consider storing as component-level state may be helpful for other components to know. For example:

https://github.com/mozilla/testpilot/blob/622e5285dae57a3952817aa29ea2854f884db5ec/frontend/src/app/components/MainInstallButton.js#L11

We currently store `isInstalling` on the install button's component. If this is stored in the Redux store, the email modal component would be able to monitor its value and pop up whenever it changes from `false` to `true`, rather than needing to be manually-triggered.

There's a lot of good discussion about this out there:

https://github.com/reactjs/redux/issues/1287
http://redux.js.org/docs/FAQ.html#do-i-have-to-put-all-my-state-into-redux-should-i-ever-use-reacts-setstate

There is one important benefit to everything going into the Redux store: it turns your app into a finite state machine. The Redux store becomes the input, and the output is the rendered DOM. This makes some good things easy: testing complex interactions between components, time travel, bug reproduction, and testing. 

For the sake of being consistent and explicit, I'd suggest implementing a guideline to not use component state, instead putting everything into the Redux store and passing values down as props.

johngruen:
we should do this at some point.

chuckharmston:
For a less contrived example of how this might be useful, see the `addon.restart` store bits in https://github.com/mozilla/testpilot/pull/1369.

fzzzy:
I think this is mostly done now: as you can see here https://github.com/mozilla/testpilot/blob/master/frontend/src/app/reducers/addon.js RestartState.isRequired is part of the AddonState in redux now, and here https://github.com/mozilla/testpilot/blob/master/frontend/src/app/reducers/experiments.js you can see that inProgress is now a part of the Experiment redux state.

However, an anti-example is https://github.com/mozilla/testpilot/blob/master/frontend/src/app/reducers/newsletter-form.js where I think possibly all that state should be local to the component with setState.

PR #1369, opened 2016-09-18
If necessary, prompt user to restart after add-on installation when using mozAddonManager (closes #1360).

### LINKED ISSUE #1360: Implement in product restart Prompt for e10s users in 49
Per: https://bugzilla.mozilla.org/show_bug.cgi?id=1302846

The issue is that we need to get all Test Pilot experiments on the e10s whitelist for FF49 if we want e10s users to avoid browser restart. If we are unable to make this list, we need a graceful fallback. Fortunately, mozAddonManager allows for [in-content restart UI](https://github.com/mozilla/addons-frontend/pull/568).

@chuckharmston here's a provisional mock up. I'm not 100% wedded to the design, but the idea here is to rive users inexorably toward that restart button.

@andymckay has offered to provide support in implementation here. 

e10s restart

chuckharmston:
~~This is WIP, getting it out there so it's easy to follow progress.~~
## Screencaps

As of ab498de, here's what the e10s experience looks like:
- [In 49 with e10s, where a restart is required](https://vid.me/rrPw)
- [In 50 with e10s, where a restart is not required](https://vid.me/LLC8)

As of d1cced6, showing the two reboot flows:
- [When installing from the landing page](https://vid.me/4Mq9)
- [When installing from an experiment page](https://vid.me/Wm3N)
## TODO
- [x] Detect necessary post-install restarts.
- [x] Prompt user for restart if necessary.
- [x] Make the prompt look [like this](https://cloud.githubusercontent.com/assets/3323249/18564740/3965abe2-7b8d-11e6-95fe-1e62d07c0930.png).
- ~~On add-on first run, if there are no open test Test Pilot tabs, open one.~~ (#1374)
- ~~Instrument this process (TBD @johngruen @ckprice)~~ (#1375)
## Testing this

To test this flow locally, you'll need a few things:
- Firefox === 49
- New profile
- No add-ons installed
- `browser.tabs.remote.force-enable` set to `false`
- `browser.tabs.remote.autostart` set to `true`
- You must use a signed copy of the add-on. Download the XPI from the production site, and move it to `/frontend/build/static/addon/addon.xpi`. You may need to do this multiple times as the gulp task rebuilds the site.
- The [full SSL proxy setup from the `mozAddonManager` PR](https://github.com/mozilla/testpilot/pull/1330#issuecomment-245821762).

You can verify that e10s is enabled by going to about:support and looking for the `Multiprocess Windows` value; you want it to be greater than 0.

### DIFF
--- a/frontend/src/app/actions/addon.js
+++ b/frontend/src/app/actions/addon.js
@@ -3,5 +3,6 @@ import { createActions } from 'redux-actions';
 export default createActions(
   {},
   'setHasAddon', 'setInstalled', 'setClientUuid',
-  'enableExperiment', 'disableExperiment'
+  'enableExperiment', 'disableExperiment',
+  'requireRestart'
 );
--- a/frontend/src/app/components/App.js
+++ b/frontend/src/app/components/App.js
@@ -1,0 +1,20 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+
+
+class App extends Component {
+  render() {
+    const { addon } = this.props;
+    if (addon.restartRequired) {
+      return <p>Restart Required</p>;
+    }
+    return this.props.children;
+  }
+}
+
+
+export default connect(
+  state => ({
+    addon: state.addon
+  })
+)(App);
--- a/frontend/src/app/components/App.js
+++ b/frontend/src/app/components/App.js
@@ -1,12 +1,14 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 
+import Restart from '../containers/Restart';
+
 
 class App extends Component {
   render() {
-    const { addon } = this.props;
-    if (addon.restartRequired) {
-      return <p>Restart Required</p>;
+    const { restart } = this.props.addon;
+    if (restart.isRequired) {
+      return <Restart experimentTitle={ restart.forExperiment }/>;
     }
     return this.props.children;
   }
--- a/frontend/src/app/components/MainInstallButton.js
+++ b/frontend/src/app/components/MainInstallButton.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { PropTypes } from 'react';
 import classnames from 'classnames';
 
 import { installAddon } from '../lib/addon';
@@ -10,6 +10,13 @@ export default class MainInstallButton extends React.Component {
     this.state = {
       isInstalling: false
     };
+  }
+
+  install() {
+    const { eventCategory, hasAddon } = this.props;
+    if (hasAddon) { return; }
+    this.setState({ isInstalling: true });
+    installAddon(this.context.store, eventCategory);
   }
 
   render() {
@@ -49,11 +56,8 @@ export default class MainInstallButton extends React.Component {
       </div>
     );
   }
-
-  install() {
-    const { eventCategory, hasAddon } = this.props;
-    if (hasAddon) { return; }
-    this.setState({ isInstalling: true });
-    installAddon(eventCategory);
-  }
 }
+
+MainInstallButton.contextTypes = {
+  store: PropTypes.object.isRequired
+};
--- a/frontend/src/app/components/MainInstallButton.js
+++ b/frontend/src/app/components/MainInstallButton.js
@@ -12,16 +12,18 @@ export default class MainInstallButton extends React.Component {
     };
   }
 
-  install() {
+  install(evt, experimentTitle) {
     const { eventCategory, hasAddon } = this.props;
     if (hasAddon) { return; }
     this.setState({ isInstalling: true });
-    installAddon(this.context.store, eventCategory);
+    installAddon(this.context.store, eventCategory, experimentTitle);
   }
 
   render() {
     const { isFirefox, hasAddon } = this.props;
     const isInstalling = this.state.isInstalling && !hasAddon;
+    const experimentTitle = ('experimentTitle' in this.props &&
+                             this.props.experimentTitle);
 
     return (
       <div>
@@ -40,7 +42,7 @@ export default class MainInstallButton extends React.Component {
           </div>
         ) : (
           <div>
-            <button onClick={e => this.install(e)} data-hook="install"
+            <button onClick={e => this.install(e, experimentTitle)} data-hook="install"
                     className={classnames('button extra-large primary install', { 'state-change': isInstalling })}>
               {hasAddon && <span className="progress-btn-msg" data-l10n-id="landingInstallingButton">Installed</span>}
               {!hasAddon && !isInstalling &&
--- a/frontend/src/app/containers/Restart.js
+++ b/frontend/src/app/containers/Restart.js
@@ -1,0 +1,38 @@
+import React from 'react';
+
+import Header from '../components/Header';
+import Footer from '../components/Footer';
+
+export default class Restart extends React.Component {
+  renderSubtitle() {
+    const { experimentTitle } = this.props;
+    if (experimentTitle) {
+      return (
+        <p data-l10n-id="restartRequiredFromExperiment">
+           data-l10n-args={JSON.stringify({ experimentTitle })}
+          Restart Firefox to enable { experimentTitle }.
+        </p>
+      );
+    }
+    return (
+      <p data-l10n-id="restartRequiredFromLanding">
+        Restart Firefox to choose your features.
+      </p>
+    );
+  }
+
+  render() {
+    return (
+      <div className="full-page-wrapper space-between">
+        <Header forceHideSettings={ false } />
+        <div className="centered-banner restart-message">
+          <h2 data-l10n-id="restartRequiredHeader">Test Pilot Installed!</h2>
+          { this.renderSubtitle() }
+        </div>
+        <footer id="main-footer" className="content-wrapper">
+          <Footer />
+        </footer>
+      </div>
+    );
+  }
+}
--- a/frontend/src/app/index.js
+++ b/frontend/src/app/index.js
@@ -10,7 +10,7 @@ import { createHistory } from 'history';
 import { useRouterHistory } from 'react-router';
 import { syncHistoryWithStore, routerReducer, routerMiddleware } from 'react-router-redux';
 import { Provider } from 'react-redux';
-import { combineReducers, createStore, applyMiddleware } from 'redux';
+import { compose, combineReducers, createStore, applyMiddleware } from 'redux';
 
 import createLogger from 'redux-logger';
 import thunk from 'redux-thunk';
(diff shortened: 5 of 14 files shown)


frontend/src/app/components/App.js
```diff
@@ -0,0 +1,20 @@
...
+    const { addon } = this.props;
+    if (addon.restartRequired) {
+      return <p>Restart Required</p>;
+    }
+    return this.props.children;
+  }
+}
+
+
+export default connect(
```
chuckharmston, line 18: Ideally this is the only connected component, and the entirety of the redux store is proliferated to children via props.

It's orthogonal to this PR so I didn't implement it here, but we would likely do it like this:

```
{React.cloneElement(this.props.children, ...this.props)}
```

frontend/src/app/components/MainInstallButton.js
```diff
@@ -12,6 +12,13 @@ export default class MainInstallButton extends React.Component {
     };
   }
 
+  install() {
+    const { eventCategory, hasAddon } = this.props;
+    if (hasAddon) { return; }
+    this.setState({ isInstalling: true });
```
chuckharmston, line 18: This is the sort of thing I'd like to get rid of in #1367; there's no reason this value shouldn't be on the Redux store and available for other components to access.

frontend/src/app/components/MainInstallButton.js
```diff
@@ -49,11 +56,8 @@ export default class MainInstallButton extends React.Component {
...
-
-  install() {
-    const { eventCategory, hasAddon } = this.props;
-    if (hasAddon) { return; }
-    this.setState({ isInstalling: true });
-    installAddon(eventCategory);
-  }
 }
+
+MainInstallButton.contextTypes = {
```
chuckharmston, line 63: This is how you get access to the store object. [More info](https://egghead.io/lessons/javascript-redux-passing-the-store-down-implicitly-via-context)

frontend/src/app/index.js
```diff
@@ -41,11 +41,14 @@ const store = createStore(
...
-    routerMiddleware(browserHistory),
-    createLogger()
+  compose(
+    applyMiddleware(
+      thunk,
+      promise,
+      routerMiddleware(browserHistory),
+      createLogger()
+    ),
+    window.devToolsExtension ? window.devToolsExtension() : f => f
```
chuckharmston, line 51: https://addons.mozilla.org/en-US/firefox/addon/remotedev/ 💓

frontend/src/app/lib/router.js
```diff
@@ -8,24 +8,27 @@ import ExperimentsListPage from '../containers/ExperimentsListPage';
...
-    <Route path="/experiments/:slug" component={ExperimentPage} />
-    <Route path="/legacy" component={LegacyPage} />
-    <Route path="/404" component={NotFoundPage} />
-    <Route path="/share" component={SharePage} />
-    <Route path="/error" component={ErrorPage} />
-    <Route path="/onboarding" component={OnboardingPage} />
-    <Route path="/retire" component={RetirePage} />
-    <Route path="*" component={NotFoundPage} />
+  <Router history={history} component={App}>
+    <Route path="/" component={App}>
```
chuckharmston, line 21: This is how you might use `<App>`; When matching, its `children` prop is set to any nested components that also match. This works infinitely deep, and allows for some really cool DRY patterns.

frontend/src/app/actions/addon.js
```diff
@@ -3,5 +3,6 @@ import { createActions } from 'redux-actions';
 export default createActions(
   {},
   'setHasAddon', 'setInstalled', 'setClientUuid',
-  'enableExperiment', 'disableExperiment'
+  'enableExperiment', 'disableExperiment',
```
chuckharmston, line 6: Filed https://github.com/mozilla/testpilot/issues/1371 to use constants for these.

johngruen:
@chuckharmston dont' sweat the look of the 'reload' page UI too much.

johngruen:
@ckprice please see initial comment regarding instrumentation

chuckharmston:
> @chuckharmston dont' sweat the look of the 'reload' page UI too much.

How's this, @johngruen? (+ different copy when installed from an add-on page)

screen shot 2016-09-18 at 1 23 43 pm

screen shot 2016-09-18 at 1 23 55 pm

I don't think we'd want to do the button-looking non-button.

chuckharmston:
Updated initial comment with videos from d1cced6.

chuckharmston:
For the sake of smaller PRs and quicker testing, I moved two orthogonal pieces of this into their own bugs (#1374, #1375) and am ready for this to be reviewed.

chuckharmston:
Added some testing notes to the PR description; it's a doozy.

lmorchard:
Taking a look

frontend/tasks/server.js
```diff
@@ -2,12 +2,14 @@ const gulp = require('gulp');
 const config = require('../config.js');
 
 const connect = require('gulp-connect');
+const historyApiFallback = require('connect-history-api-fallback');
```
lmorchard, line 5: Nitpick: Didn't want to use this module here, because we're generating static `index.html` landing pages for S3. This can mask missing landing pages that we forgot to add to the generation task. Maybe more suitable for a testing scenario to catch?

frontend/src/app/components/MainInstallButton.js
lmorchard: I watched that video, but I thought the context system was undocumented and not to be relied on? [The docs have a big scary caveat on it](https://facebook.github.io/react/docs/context.html), though I know that's how Redux passes things around with the `<Provider>` / `connect` abstraction. Solves the issue I was having with passing things through the router, though

frontend/src/app/index.js
lmorchard: :+1: 

Nitpick: worth adding a note to dev docs about what this enables, but I wouldn't block this PR for it.

frontend/src/app/actions/addon.js
lmorchard: Commented over there

frontend/src/app/components/MainInstallButton.js
lmorchard: This also couples the component to Redux, right? Seems to make it harder to keep this presentation-only and [use `connect` to pass in the mapped store data & dispatch helpers in props](https://github.com/mozilla/testpilot/blob/b33a19b08295b7d6ef0e789afb6ea1ccfb31e315/frontend/src/app/containers/ExperimentPage.js#L10). That seemed like an important pattern to keep things loosely bound & easily testable. Wondering if we can somehow get this stuff fed in from the `App` component as props?

frontend/src/app/containers/Restart.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
lmorchard, line 1: Nitpick: Should this instead live with the rest of the presentation-only components under `components`?

frontend/src/app/lib/addon.js
```diff
@@ -7,15 +7,19 @@ import experimentActions from '../actions/experiments';
...
   'testpilot.firefox.com',
   'testpilot.stage.mozaws.net',
   'testpilot.dev.mozaws.net'
 ];
 
+let RESTART_NEEDED;
+
 function mozAddonManagerInstall(url) {
   return navigator.mozAddonManager.createInstall({ url }).then(install => {
+    navigator.mozAddonManager.addEventListener('onInstalling', data => {
```
lmorchard, line 20: Does this need to be paired with a `removeEventListener` once we've finished processing here? Guessing the two subscriptions below get garbage collected after the install in question is over, but would this listener catch _all_ install progress events going forward? Probably harmless, but concerned about "dead" code getting called.

frontend/src/app/lib/addon.js
```diff
@@ -86,7 +92,6 @@ function watchForAddonInstallStateChange(store) {
 }
 
 export function sendMessage(type, data) {
-  console.log('TO ADDON', type, data);
```
lmorchard, line 89: (derp.)

frontend/src/app/lib/addon.js
```diff
@@ -106,8 +111,6 @@ export function disableExperiment(dispatch, experiment) {
 function messageReceived(store, evt) {
   const { type, data } = evt.detail;
 
-  console.log('FROM ADDON', type, data);
```
lmorchard, line 109: (herp derp.)

frontend/src/app/components/MainInstallButton.js
lmorchard: Actually, is this what you meant with this bit in `<App>`?

``` javascript
{React.cloneElement(this.props.children, ...this.props)}
```

Could we supply a generic `installAddon` helper function in a prop from `<App>` and leave this component ignorant of the Redux store? Happy not to block this PR and save that as a future improvement if so.

frontend/src/app/components/App.js
lmorchard: I think I grok that, and that sounds like a good follow-on to this!

lmorchard (commented):
Works for me, code looks good. My only things are nitpicks and architectural questions not worth blocking the PR to address.

lmorchard (approved):
Looks like I picked "comment" and not "approve", but I can't change it. Anyway, [this was my real review summary](https://github.com/mozilla/testpilot/pull/1369#pullrequestreview-576724).

frontend/src/app/components/MainInstallButton.js
chuckharmston: I tried to keep the surface area of this PR small because of its urgency, but yep, that's exactly what the promised land looks like.

frontend/src/app/containers/Restart.js
chuckharmston: For the sake of having an easy rule, I'd suggest putting anything that gets handed to the router in `containers`, and anything else in `components`.

frontend/tasks/server.js
chuckharmston: What's the generation process? Are there docs for that? I did this because the route I added for `/restart` wasn't accessible `¯\_(ツ)_/¯`

lmorchard: Ah, yeah docs would be nice, but there are none. So, this is actually something that needs fixing or else the static deploy will result in a 404.

New paths need to be added to [this list over here in the pages task](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/pages.js#L25). (which I missed, and _should_ have been my review comment :) )

frontend/src/app/components/MainInstallButton.js
lmorchard: Okay, cool beans. Was worried I missed something, but nope. Let's save this for future work

lmorchard:
Un-assigning myself, because I think all my concerns are addressed. Good to merge after a rebase IMO

PR #1378, opened 2016-09-18
only prompt users to upgrade firefox on experiment pages if they must

### LINKED ISSUE #1208: NoMore404s page tells me that it requires Firefox 48+, even though I have 51 installed
I was initially confused because I was rocking The Nightly. Not sure if we want to hide the "No More 404s requires Firefox 48 or later" message box if the user is using Firefox >=48.

firefox_test_pilot_-_no_more_404s

johngruen:
Fixes #1208

### DIFF
--- a/content-src/experiments/no-more-404s.yaml
+++ b/content-src/experiments/no-more-404s.yaml
@@ -4,7 +4,7 @@ short_title: 'No More 404s'
 slug: no-more-404s
 thumbnail: /static/images/experiments/no-more-404s/experiments_experiment/a/9/a9e6dc66c6a22851721715ae4d0641fb_thumbnail_1470245154_0349.png
 description: "Tired of dead ends on the web? We'll let you know when there's a saved version of what you're looking for on the Internet Archive's Wayback Machine.\r\n"
-introduction: "<p>This one is simple. If you hit a 404 page while browsing the web, we will let you know if the content you're looking for is archived on the Internet Archive's Wayback Machine.</p>\r\n\r\n<div style=\"padding: 10px; border-radius: 3px; background: rgba(255,147,0,.25); border: 1px solid #FF9300; font-size: 16px; margin-top: 20px;\">No More 404s requires Firefox 48 or later. <br><a href=\"https://support.mozilla.org/lt/kb/find-what-version-firefox-you-are-using\" target=\"_blank\">Make sure you're up to date.</a></div>\r\n"
+introduction: "<p>This one is simple. If you hit a 404 page while browsing the web, we will let you know if the content you're looking for is archived on the Internet Archive's Wayback Machine.</p>"
 image_twitter: /static/images/experiments/no-more-404s/experiments_experiment/a/9/a9e6dc66c6a22851721715ae4d0641fb_image_1472573814_0582.png
 image_facebook: /static/images/experiments/no-more-404s/experiments_experiment/a/9/a9e6dc66c6a22851721715ae4d0641fb_image_1472573814_0278.png
 version: 1.5.6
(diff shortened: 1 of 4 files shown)


johngruen:
Please note, this PR only works with YAML/flat files

content-src/experiments/no-more-404s.yaml
```diff
@@ -4,7 +4,7 @@ short_title: 'No More 404s'
 slug: no-more-404s
 thumbnail: /static/images/experiments/no-more-404s/experiments_experiment/a/9/a9e6dc66c6a22851721715ae4d0641fb_thumbnail_1470245154_0349.png
 description: "Tired of dead ends on the web? We'll let you know when there's a saved version of what you're looking for on the Internet Archive's Wayback Machine.\r\n"
-introduction: "<p>This one is simple. If you hit a 404 page while browsing the web, we will let you know if the content you're looking for is archived on the Internet Archive's Wayback Machine.</p>\r\n\r\n<div style=\"padding: 10px; border-radius: 3px; background: rgba(255,147,0,.25); border: 1px solid #FF9300; font-size: 16px; margin-top: 20px;\">No More 404s requires Firefox 48 or later. <br><a href=\"https://support.mozilla.org/lt/kb/find-what-version-firefox-you-are-using\" target=\"_blank\">Make sure you're up to date.</a></div>\r\n"
+introduction: "<p>This one is simple. If you hit a 404 page while browsing the web, we will let you know if the content you're looking for is archived on the Internet Archive's Wayback Machine.</p>\r\n"
```
chuckharmston: We don't need the newlines here.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -90,6 +90,14 @@ export default class ExperimentPage extends React.Component {
     return currentExperiment[0];
   }
 
+  checkIfValidVersion(min) {
```
chuckharmston: Maybe rename this `isValidVersion`?

frontend/src/app/components/ExperimentPage.js
```diff
@@ -90,6 +90,14 @@ export default class ExperimentPage extends React.Component {
...
+    const version = parseInt(navigator.userAgent.split('/').pop(), 10);
```
chuckharmston, line 94: User agents are the worst.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -90,6 +90,14 @@ export default class ExperimentPage extends React.Component {
...
+    if (typeof min === 'undefined' || version >= min) {
```
chuckharmston: You can simplify this to `return typeof min === 'undefined' || version >= min`

frontend/src/app/components/ExperimentPage.js
```diff
@@ -169,12 +178,18 @@ export default class ExperimentPage extends React.Component {
...
                 {enabled && <button onClick={e => this.renderUninstallSurvey(e)} style={{ width: progressButtonWidth }} data-hook="uninstall-experiment" id="uninstall-button" className={classnames(['button', 'secondary'], { 'state-change': isDisabling })}><span className="state-change-inner"></span><span data-l10n-id="disableExperimentTransition" className="transition-text">Disabling...</span><span data-l10n-id="disableExperiment" data-l10n-args={JSON.stringify({ title })} className="default-text">Disable <span data-hook="title">{title}</span></span></button>}
                 {!enabled && <button onClick={e => this.installExperiment(e)} style={{ width: progressButtonWidth }} data-hook="install-experiment" id="install-button"  className={classnames(['button', 'default'], { 'state-change': isEnabling })}><span className="state-change-inner"></span><span data-l10n-id="enableExperimentTransition" className="transition-text">Enabling...</span><span data-l10n-id="enableExperiment" data-l10n-args={JSON.stringify({ title })} className="default-text">Enable <span data-hook="title">{title}</span></span></button>}
               </div>}
+              {hasAddon && !validVersion &&
```
chuckharmston: Let's separate these logical chunks into separate methods, e.g.

``` javascript
renderMinimumVersionNotice() {
  if (hasAddon && !this.isValidVersion() {
    return (
      <div className="upgrade-notice">...</div>
    );
  }
  return null;
}
```

Then just call them here: `{this.renderMinimumVersionNotice()}`

chuckharmston (changes requested):
Looks good overall!

frontend/src/app/components/ExperimentPage.js
```diff
@@ -90,6 +90,14 @@ export default class ExperimentPage extends React.Component {
     return currentExperiment[0];
   }
 
+  isValidVersion(min) {
+    const version = parseInt(navigator.userAgent.split('/').pop(), 10);
+    if (typeof min === 'undefined' || version >= min) {
```
chuckharmston: We can just `return typeof min === 'undefined' || version >= min;` here, no reason for the conditional.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -325,6 +335,18 @@ export default class ExperimentPage extends React.Component {
     this.didScroll = false;
   }
 
+  renderMinimumVersionNotice(title, hasAddon, min_release) {
```
chuckharmston, line 335: 👌

frontend/src/app/components/ExperimentPage.js
```diff
@@ -338,6 +360,15 @@ export default class ExperimentPage extends React.Component {
     }, 5000);
   }
 
+  clickUpgradeNotice() {
```
chuckharmston, line 360: I think we want this receive the `evt` argument, prevent the default action, and use 

`outboundURL: evt.target.attributes.href`

to prevent a race condition on this request completing and the user navigating away from the page.

chuckharmston (changes requested):

frontend/src/app/components/ExperimentPage.js
johngruen: @chuckharmston i don't think there's a race condition here b/c the link has `target=_blank"

johngruen:
@chuckharmston fixed per your review % the inclusion of an outbound URL. See my comment above.

chuckharmston (approved):

PR #1379, opened 2016-09-19
On add-on first run, open Test Pilot if not already open (closes #1374).

### LINKED ISSUE #1374: On add-on first run, open Test Pilot
Breaking out from #1360:

On add-on first run, if there are no open test Test Pilot tabs, one should be opened. This is primarily intended to cover the restart-required use case, but may be more broadly useful.

chuckharmston:
~~Not ready for review.~~

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -17,6 +17,7 @@ const URL = require('sdk/url').URL;
 
 const { App } = require('./lib/app');
 const ExperimentNotifications = require('./lib/experiment-notifications');
+const FirstRun = require('./lib/first-run');
 const Metrics = require('./lib/metrics');
 const SharePrompt = require('./lib/share-prompt');
 const survey = require('./lib/survey');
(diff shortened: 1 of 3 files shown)


chuckharmston:
This is ready for review now.

addon/index.js
```diff
@@ -358,6 +359,7 @@ exports.main = function(options) {
   ToolbarButton.init(settings);
   ExperimentNotifications.init();
   SharePrompt.init(settings);
+  FirstRun.setup(options.reason, settings);
```
lmorchard, line 362: I can't seem to find the issue now, but I remember you opening one about making this general module pattern cleaner and more like Django. :+1: to this and that in general

addon/test/test-first-run.js
```diff
@@ -0,0 +1,137 @@
+/*
```
lmorchard, line 1: :+1: Yay tests, and these pass for me locally.

lmorchard (approved):
Works for me, looks clean & good

PR #1381, opened 2016-09-19
Instrument the restart-required flows (closes #1375).

### LINKED ISSUE #1375: Insturment the restart-required flows when using mozAddonManager
Breaking out of #1360.

My suggestion from that thread:

> - Event: user sees CTA
>   - Dimension: is user on homepage or experiment detail page?
>   - Dimension: can user use `mozAddonManager`?
> - Event: user clicks install button
>   - Dimension: is restart required?
> - Event: user shown "must restart" modal.
> - Event: add-on opens homepage after restart
>   - We'd want `utm_`s for this.
> 
> This should let us segment different pathways as they're fed into existing funnels to measure the effect on conversion rates. Plus, we would have absolute numbers to pair with the conversion rates to serve as a basis for speculation of how many users we lost in each step.

CC @ckprice

chuckharmston:
r? @ckprice

### DIFF
--- a/frontend/src/app/lib/addon.js
+++ b/frontend/src/app/lib/addon.js
@@ -46,8 +46,9 @@ export function installAddon(store, eventCategory, experimentTitle) {
   };
 
   if (useMozAddonManager) {
-    sendToGA('event', gaEvent);
     mozAddonManagerInstall(downloadUrl).then(() => {
+      gaEvent.eventValue = RESTART_NEEDED ? 1 : 0;
+      sendToGA('event', gaEvent);
       if (RESTART_NEEDED) {
         store.dispatch(addonActions.requireRestart(experimentTitle));
       }
--- a/docs/metrics/ga.md
+++ b/docs/metrics/ga.md
@@ -49,31 +49,34 @@ app.sendToGA('event', {
 
 Here are the current events on the website as of this writing
 
-| Description                                              | eventCategory                      | eventAction       | eventLabel                   |
-|----------------------------------------------------------|------------------------------------|-------------------|------------------------------|
-| Toggle settings menu                                     | Menu Interactions                  | drop-down menu    | Toggle Menu                  |
-| Click Leave from settings                                | Menu Interactions                  | drop-down menu    | Retire                       |
-| Click Discuss from settings                              | Menu Interactions                  | drop-down menu    | Discuss                      |
-| Click Wiki from settings                                 | Menu Interactions                  | drop-down menu    | wiki                         |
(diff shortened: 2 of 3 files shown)


chuckharmston:
Note: the UTMs were added in https://github.com/mozilla/testpilot/pull/1379, where I added the restart-tab-opening.

frontend/src/app/lib/addon.js
```diff
@@ -46,8 +46,9 @@ export function installAddon(store, eventCategory, experimentTitle) {
   };
 
   if (useMozAddonManager) {
-    sendToGA('event', gaEvent);
     mozAddonManagerInstall(downloadUrl).then(() => {
+      gaEvent.eventValue = RESTART_REQUIRED ? 1 : 0;
```
ckprice: `RESTART_NEEDED`?

chuckharmston: ...yeah. I don't know why this worked when I tested it out, I'm going to backtrack to make sure it's doing what's intended.

chuckharmston: At least the tests failed.

lmorchard (approved):
Code looks :100: to me.

pdehaan:
Needs a rebasin', but not sure if we want to fix the "Insturment" typo in commit message while we're at it. :trollface:

frontend/src/app/lib/addon.js
```diff
@@ -46,8 +46,9 @@ export function installAddon(store, eventCategory, experimentTitle) {
   };
 
   if (useMozAddonManager) {
-    sendToGA('event', gaEvent);
     mozAddonManagerInstall(downloadUrl).then(() => {
+      gaEvent.eventValue = RESTART_NEEDED ? 1 : 0;
```
pdehaan, line 50: Is this just casting a Boolean to an int? And if so, couldn't you just do `gaEvent.eventValue = Number(RESTART_NEEDED);`?

``` sh
$ node -e console.log(Number(true));  # 1
$ node -e console.log(Number(false)); # 0
```

chuckharmston: Yeah, but when you're giving numbers magic values, I'd rather be explicit about what the range of values are. Clear > cute.

ckprice:
I pulled the branch down, and there were no errors, but I wasn't able to fully test the small measurement pieces (eventValue and utm values). I ventured down https://github.com/mozilla/testpilot/pull/1330#issuecomment-245821762 to enable `mozAddonManager`, but ended up doing more harm than good here I think. Can we merge this in, and I'll take an action item to test on `dev` (assuming it's any easier there)?

chuckharmston:
Sure. Merging.

PR #1389, opened 2016-09-20
Remove add-on code pinging Django to record experiment enable / disable

### LINKED ISSUE #1039: Replace source for "### people trying experiment" with count derived from metrics
Currently, we expose experiment usage via keeping a record of installations and counting them by distinct user. Once we get rid of the user DB (#1034), that a) removes the uniqueness index and b) means we have to rework that installations table to track client IDs.

It might be better if we also remove that installations table & API altogether, and instead find a way to get a per-experiment usage count from the Telemetry metrics sent from the client. This would also help us along the way to flat files (#988) by removing probably the last thing for which we need our own server & database.

lmorchard:
Fixes #1039

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -11,7 +11,6 @@ const aboutConfig = require('sdk/preferences/service');
 const self = require('sdk/self');
 const store = require('sdk/simple-storage').storage;
 const tabs = require('sdk/tabs');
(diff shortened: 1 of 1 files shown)


chuckharmston (approved):
➖ ➖ ➖ ➖ ➖

Pending passing tests, :shipit:

PR #1390, opened 2016-09-20
update copy on restart page

### LINKED ISSUE #1386: Update style and copy on 'restart required' modal
Had a meeting with Winston today and he had some recommendations on this page. Current implementation: https://vid.me/4Mq9
- Increased font size on the 'Restart' CTA. Maybe bring it to the top.
- Add an 'almost done' implication to the strings.

Current strings ([ref](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl#L146))
- `restartRequiredHeader` = Test Pilot Installed!
- `restartRequiredFromLanding` = Restart Firefox to choose your features.
- `restartRequiredFromExperiment` = Restart Firefox to enable {$experimentTitle}.

@johngruen to confirm strings (and maybe open PR).

johngruen:
Fixes #1386

### DIFF
--- a/frontend/src/app/containers/Restart.js
+++ b/frontend/src/app/containers/Restart.js
@@ -19,16 +19,16 @@ export default class Restart extends React.Component {
     const { experimentTitle } = this.props;
     if (experimentTitle) {
       return (
-        <p data-l10n-id="restartRequiredFromExperiment">
+        <h2 data-l10n-id="restartRequiredFromExperiment">
            data-l10n-args={JSON.stringify({ experimentTitle })}
-          Restart Firefox to enable { experimentTitle }.
-        </p>
+          Please restart Firefox <br/> to enable { experimentTitle }.
+        </h2>
       );
     }
     return (
-      <p data-l10n-id="restartRequiredFromLanding">
-        Restart Firefox to choose your features.
-      </p>
+      <h2 data-l10n-id="restartRequiredFromLanding">
+        Please restart Firefox <br/> to choose your features.
+      </h2>
     );
   }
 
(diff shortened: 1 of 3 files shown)


locales/en-US/app.ftl
```diff
@@ -143,9 +143,9 @@ shareCopy = Copy
 
 eolMessage = <strong>This experiment is ending on {$completedDate}</strong>.<br/><br/>After then you will still be able to use {$title} but we will no longer be providing updates or support.
 
-restartRequiredHeader = Test Pilot Installed!
-restartRequiredFromLanding = Restart Firefox to choose your features.
-restartRequiredFromExperiment = Restart Firefox to enable {$experimentTitle}.
+restartRequiredSubHeader = Almost done . . .
```
chuckharmston, line 146: Can we use `&hellip;` for this?

frontend/src/app/containers/Restart.js
```diff
@@ -37,7 +37,7 @@ export default class Restart extends React.Component {
       <div className="full-page-wrapper space-between">
         <Header forceHideSettings={ false } />
         <div className="centered-banner restart-message">
-          <h2 data-l10n-id="restartRequiredHeader">Test Pilot Installed!</h2>
+          <p data-l10n-id="restartRequiredSubHeader">Almost done . . s.</p>
```
chuckharmston: This doesn't match the string in the localization file. Also, let's use `&hellip;` here, too.

chuckharmston (changes requested):
Looks good other than two small comments.

locales/en-US/app.ftl
johngruen: @chuck Technically, the &hellip; character entity is not correct punctuation in this case: 

http://www.thepunctuationguide.com/ellipses.html

frontend/src/app/containers/Restart.js
johngruen: Fixed this % hellip inclusion.

chuckharmston (approved):

Issue #1387, opened 2016-09-20
If you chose to install Test Pilot before experiments page section is loaded, that area never loads and the redirect is never made after installation


SoftVision-PaulOiegas:
[Affected versions]:
- Nightly 52.0a1

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- The string pref "testpilot.env" with "dev" value was added in "about:config" page on a new Firefox profile.

[Steps to reproduce]:
1. Open Firefox browser.
2. As fast as you can click the "Install the Test Pilot add-on"  button after navigating to testpilot.dev.mozaws.net page.
3. Observe the experiments area behavior on the Test Pilot "Homepage".
4. Proceed with the installation and observe page behavior.

[Expected result]:
- Step 3: The experiments area takes a moment to load but it's correctly displayed after.
- Step 4: After installation, you are correctly redirected to Test Pilot "Experiments" page.

[Actual result]:
- Step 3: The experiments area never loads, the loading animation is displayed indefinitely.
- Step 4: The redirect to "Experiments" page no longer happens after installation due to this behavior.

[Notes]:
- Here is a screen recording proving the issue https://goo.gl/KbpdNv.

chuckharmston:
I'm going to `wontfix` this because it's a fleeting edge case. It will only happen when you get the browser doorhanger to install the add-on, which will only happen if:
1. You are not using SSL.
2. You are testing on dev or stage without `extensions.webapi.testing` set to `true`.

#1 will be remediated shortly when we ship flatfiles, and as this sprint completes, #2 will be a prerequisite for doing any testing. I'll have a document ready for you shortly that outlines all the various test cases that these changes create.

PR #1391, opened 2016-09-20
Wait until experiments are loaded before rendering either landing page

lmorchard:
This seems to also fix issue #1387 and the integration tests as a bonus.


lmorchard:
Basically, all other PRs are blocked from merging until this one gets in or we disable integration tests.

johngruen (approved):
Works

PR #1394, opened 2016-09-20
Try adding HTTPS support to the local dev server

lmorchard:
- new 'example' env for addon that uses example.com
- self-signed certs for example.com
- USE_HTTPS=1 env var for gulp that enables HTTPS

Issue #1329


lmorchard:
@chuckharmston - this one's for you, if it works :)

Some tips:
- `echo '127.0.0.1 example.com' | sudo tee -a /etc/hosts` to create example.com host alias
- Ensure `extensions.webapi.testing` is set to `true`
- Ensure `testpilot.dev` is set to `example`
- Make sure to run `npm run package` in your `addon/` directory
- Run the dev server like this: 
  
  ``` bash
  while [ 1 ]; do USE_HTTPS=1 NODE_ENV=development ENABLE_PONTOON=1 gulp; sleep 1; done
  ```
- Visit https://example.com:8000

lmorchard:
Oh yeah, and I'm sure I'm missing some steps. But, I was able to see the mozAddonManager install flow work in my browser. I think there was another pref to turn on to ensure you could install addons from HTTPS sources with untrusted certs

chuckharmston:
> I think there was another pref to turn on to ensure you could install addons from HTTPS sources with untrusted certs

That was to allow add-on install from non-built-in CAs, which Charles required but this should avoid.

dannycoates:
I'm missing something. When clicking the install button I get this in the console.

screen shot 2016-09-22 at 11 16 09 am

I've installed both the server and ca certs on dev edition, am using the xpi from production, and my about:config setting allow unsigned.

lmorchard:
Hmm, then maybe you do need to also set `extensions.install.requireBuiltInCerts` to `false` [like we had to when using Charles](https://github.com/mozilla/testpilot/pull/1330).

dannycoates (approved):
I think a good followup to this would be an npm script target that sets the envars and maybe a groundcontrol patch that sets the prefs for you when you choose 'example'

PR #1397, opened 2016-09-21
make 45 the minimum version to install test pilot

### LINKED ISSUE #1310: No restriction notification is displayed for users that are trying to instal Test Pilot on incompatible Fx version
"Install the Test Pilot add-on" CTA shows installation in progress even after installation is complete
- Operating system: OS X 10.11.6
- Firefox build 38.0.5

screen shot 2016-08-30 at 12 51 03 pm

johngruen:
Fixes #1310

### DIFF
--- a/frontend/src/app/components/MainInstallButton.js
+++ b/frontend/src/app/components/MainInstallButton.js
@@ -20,41 +20,60 @@ export default class MainInstallButton extends React.Component {
   }
 
   render() {
-    const { isFirefox, hasAddon } = this.props;
+    const { isFirefox, isMinFirefox, hasAddon } = this.props;
     const isInstalling = this.state.isInstalling && !hasAddon;
     const experimentTitle = ('experimentTitle' in this.props &&
                              this.props.experimentTitle);
 
     return (
       <div>
-        {(!isFirefox) ? (
-          <div>
-              <span data-l10n-id="landingDownloadFirefoxDesc" className="parens">(Test Pilot is available for Firefox on Windows, OS X and Linux)</span>
-              <a href="https://www.mozilla.org/firefox" className="button primary download-firefox">
-                <div className="button-icon">
-                  <div className="button-icon-badge"></div>
-                </div>
-                <div className="button-copy">
-                  <div data-l10n-id="landingDownloadFirefoxTitle" className="button-title">Firefox</div>
-                  <div data-l10n-id="landingDownloadFirefoxSubTitle" className="button-description">Free Download</div>
-                </div>
-              </a>
-          </div>
-        ) : (
-          <div>
-            <button onClick={e => this.install(e, experimentTitle)} data-hook="install"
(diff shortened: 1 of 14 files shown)


frontend/src/app/components/MainInstallButton.js
```diff
@@ -20,27 +20,14 @@ export default class MainInstallButton extends React.Component {
...
-                  <div className="button-icon-badge"></div>
-                </div>
-                <div className="button-copy">
-                  <div data-l10n-id="landingDownloadFirefoxTitle" className="button-title">Firefox</div>
-                  <div data-l10n-id="landingDownloadFirefoxSubTitle" className="button-description">Free Download</div>
-                </div>
-              </a>
-          </div>
-        ) : (
+        {(isMinFirefox) ? (
```
chuckharmston: If we're going to use a ternary, can you put both chunks in methods? The end is hard to find.

`(isMinFirefox ? this.renderInstallButton : renderAltButton)()`

frontend/src/app/reducers/browser.js
```diff
@@ -1,9 +1,12 @@
...
+const setIsMinFirefox = (state, { payload: isMinFirefox }) => ({ ...state, isMinFirefox });
 
 export default handleActions({
-  setIsFirefox
+  setIsFirefox,
+  setIsMinFirefox
 }, {
-  isFirefox: false
+  isFirefox: false,
+  isMinFirefox: false
```
chuckharmston: The version of Firefox will never update while the application is running.

Rather than go through the full action => reducer process, just set the value in default props here.

chuckharmston (changes requested):

frontend/src/app/components/MainInstallButton.js
johngruen: Fixed

frontend/src/app/reducers/browser.js
johngruen: @chuckharmston once i figured this out i was able to knock out the browser actions file and remove the dispatches from index.

dannycoates (approved):
LGTM

chuckharmston (approved):

frontend/src/app/reducers/browser.js
```diff
@@ -1,9 +1,12 @@
 import { handleActions } from 'redux-actions';
+import config from '../config';
+import { isMinFirefoxVersion } from '../lib/utils';
 
-const setIsFirefox = (state, { payload: isFirefox }) => ({ ...state, isFirefox });
+const userAgent = navigator.userAgent.toLowerCase();
+const isUserAgentFirefox = userAgent.indexOf('firefox') > -1;
```
lmorchard, line 6: This is already merged, but I think this is something we'll need to fix: This stuff should probably have been in `index.js` where the `setFirefox` action was. Add a `setFirefoxVersion` action and do that from there too

Issue #1328, opened 2016-09-08
Add interstitial message for Tracking Protection experiment when user attempts feedback


johngruen:
With the upcoming tracking protection experiment, we will mostly want people to give feedback via the tracking protection add on. We can cut down on bad interactions by adding an interstitial modal between the feedback button on the experiment details page and the redirect to survey gizmo.

sevaan:
UX: https://mozilla.invisionapp.com/share/Q77S1H6ZC#/187233763_Tracking_Protection_Feedback

Image asset attached.
![asset 2x](https://cloud.githubusercontent.com/assets/473531/18359374/a7d42598-75c7-11e6-8f0c-7dbc87f18d07.png)

sevaan:
So to clarify, user clicked the Give Feedback button on the experiment details page, and then the new tour UI ([Issue #944](https://github.com/mozilla/testpilot/issues/944)) is revised to show a message clarifying where to give feedback.

johngruen:
@sevaan thanks for making this. We're going to need to tweak this asset b/c we've changed the name of the add on to Tracking Protection

johngruen:
I'll get you a new XPI ASAP

johngruen:
@lmorchard I already put together an PR for #1208 that assumes YAML, but I suppose this + that one will require some kind of django migration until we deploy statically. Would it make sense for me to get a YAML only PR together for this one to accompany my other PR then open a follow on issue for you or @dannycoates add the appropriate stuff to django?

lmorchard:
It probably doesn't make sense to duplicate new features like this in Django if we're just going to throw it away in a few weeks (hopefully). 

Here's another option I mentioned in IRC last night: We _can_ actually use the flat files content on Django. There's a waffle switch in admin to swap from using the DB to using content from YAML.

The catch, though, is that we can't update content on the site without doing a full-on tag-and-deployment with ops.

johngruen:
@lmorchard @clouserw straw plan: 

We get everything in place with the YAML for our deployment. We delay tagging until Friday this sprint to acommodate last minute changes to various XPIs and copy. We add the new experiments in the yaml and add a `show_after` flag with our intended launch date (we could even add a `butimspecial`) query sring so soft vision could see the upcoming experiments.

Thoughts?

Issue #1308, opened 2016-08-29
YAML / JSON validation for experiments.json and enforce validation in CI?


lmorchard:
So, we're leaving Django's CMS behind, with all of its required fields and content requirements. That means our YAML-to-JSON content build process is a lot looser and could be prone to errors and neglected fields.

Should we write & maintain a JSON schema that asserts mandatory & optional fields and what kinds of things they should contain, if only as a linting stage in CircleCI so we don't push broken content?

Could be pushed into a future sprint / milestone as long as we're careful with the artisanal YAML for now.

lmorchard:
Going to take this out of TXP-32, because it doesn't directly block migrating the frontend to a static deploy. It just makes things safer

pdehaan:
FWIW, I'm a big fan of Hapi.js' [**joi**](https://github.com/hapijs/joi/) object schema validator.
We'd have to first convert the YAML to JSON, before passing it to Joi, but should be easy peasy.

But yeah, adding it to a `lint` task ensures that it'd get run during `npm test` and on Circe-CI and that you wouldn't theoretically be able to git push if you [poorly] edit the YAML files.

jaredhirsch:
I think the YAML vs JSON decision probably should come down to whether non-technical folks are editing these files regularly. There are plenty of schema validation tools for YAML.

lmorchard:
> I think the YAML vs JSON decision probably should come down to whether non-technical folks are editing these files regularly. There are plenty of schema validation tools for YAML.

The JSON is auto-generated from the YAML. Folks like @johngruen & @ckprice are probably going to be the usual editors of the YAML, and I'd count them as technical folks. But, we all make mistakes, so it would be nice to catch them as early as possible.

Might be better to validate the YAML, since that will be what folks are actually editing.

pdehaan:
I think you'll have much better success finding feature-rich JSON validators instead of YAML validators (but I'm often wrong). In my [oh so brief] research, I did find [**yaml-validator**](https://www.npmjs.com/package/yaml-validator) which seems to do basic schema checking.

For giggles sake, here's a crude attempt at linting via Joi (by converting to JSON on-the-fly using **yamljs**). Apologies to @dannycoates in advance for this :poop::fire: of a schema file. I'm sure there is a much better Joi-ful way, which allows for optional `short_title` empty string values:

<details>
<summary>

Experiments YAML linter via Joi: (click to expand)</summary>



``` js
const Joi = require('joi');
[...]

PR #1398, opened 2016-09-21
add optional pre-feedback notice

johngruen:
#1328

### DIFF
--- a/content-src/experiments/activity-stream.yaml
+++ b/content-src/experiments/activity-stream.yaml
@@ -18,6 +18,9 @@ xpi_url: 'https://moz-activity-streams.s3.amazonaws.com/dist/activity-streams-la
 addon_id: '@activity-streams'
 gradient_start: '#0996F8'
 gradient_stop: '#0675D3'
+min_release:
+pre_feedback_image:
+pre_feedback_copy:
 details:
   -
     headline: ' '
--- a/frontend/src/app/components/ExperimentPage.js
+++ b/frontend/src/app/components/ExperimentPage.js
@@ -13,6 +13,7 @@ import ExperimentDisableDialog from '../components/ExperimentDisableDialog';
 import ExperimentTourDialog from '../components/ExperimentTourDialog';
 import MainInstallButton from '../components/MainInstallButton';
 import ExperimentCardList from '../components/ExperimentCardList';
+import ExperimentPreFeedbackDialog from '../components/ExperimentPreFeedbackDialog';
 
 const CHANGE_HEADER_ON = 105;
 
@@ -24,6 +25,7 @@ export default class ExperimentPage extends React.Component {
     const { isExperimentEnabled } = this.props;
     const experiment = this.findCurrentExperiment(props);
 
+    // TODO: Clean this up per #1367
     this.state = {
       experiment,
       enabled: isExperimentEnabled(experiment),
@@ -34,7 +36,8 @@ export default class ExperimentPage extends React.Component {
       progressButtonWidth: null,
       showDisableDialog: false,
       shouldShowTourDialog: false,
-      showTourDialog: false
+      showTourDialog: false,
+      showPreFeedbackDialog: false
     };
 
     // HACK: Set this as a plain object property, so we don't trigger crazy
@@ -107,7 +110,7 @@ export default class ExperimentPage extends React.Component {
 
     const { experiment, enabled, useStickyHeader, highlightMeasurementPanel,
             showDisableDialog, showTourDialog, isEnabling, isDisabling,
-            progressButtonWidth } = this.state;
+            progressButtonWidth, showPreFeedbackDialog } = this.state;
 
     const { title, version, contribute_url, bug_report_url, discourse_url,
             introduction, measurements, privacy_notice_url, changelog_url,
@@ -140,6 +143,10 @@ export default class ExperimentPage extends React.Component {
           <ExperimentTourDialog experiment={experiment}
             onCancel={() => this.setState({ showTourDialog: false })}
             onComplete={() => this.setState({ showTourDialog: false })} />}
+
+        {showPreFeedbackDialog &&
+          <ExperimentPreFeedbackDialog experiment={experiment} surveyURL={surveyURL}
+            onCancel={() => this.setState({ showPreFeedbackDialog: false })} />}
 
         <Header hasAddon={hasAddon}/>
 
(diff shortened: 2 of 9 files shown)


frontend/src/app/components/ExperimentTourDialog.js
```diff
@@ -89,16 +89,6 @@ export default class ExperimentTourDialog extends React.Component {
...
-  takeTour() {
-    this.setState({ currentStep: 0 });
-
-    sendToGA('event', {
-      eventCategory: 'ExperimentDetailsPage Interactions',
-      eventAction: 'button click',
-      eventLabel: 'take tour'
-    });
-  }
-
```
johngruen, line 101: I should have taken this out on that refactor of the tour UI

johngruen:
If someone wants to check the UI here, just populate the `pre_feedback_image` and `pre_feedback_copy` fields on one of the experiment YAML files

content-src/experiments/activity-stream.yaml
```diff
@@ -18,6 +18,9 @@ xpi_url: 'https://moz-activity-streams.s3.amazonaws.com/dist/activity-streams-la
 addon_id: '@activity-streams'
 gradient_start: '#0996F8'
 gradient_stop: '#0675D3'
+min_release:
```
chuckharmston, line 21: We should probably build the front-end in such a way that we don't need to include empty keys at all.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -34,7 +35,8 @@ export default class ExperimentPage extends React.Component {
       progressButtonWidth: null,
       showDisableDialog: false,
       shouldShowTourDialog: false,
-      showTourDialog: false
+      showTourDialog: false,
+      showPreFeedbackDialog: false
```
chuckharmston, line 40: This should probably be in the redux store, per #1367.

frontend/src/styles/modules/_modals.scss
```diff
@@ -282,14 +282,20 @@
 
   .tour-text {
     background: $white;
+    display: block;
     font-weight: bold;
+    padding: 0 0 $grid-unit * .25;
```
chuckharmston: Throw parens around this? Way easier to read.

locales/en-US/app.ftl
```diff
@@ -71,6 +71,9 @@ isEnabledStatusMessage = {$title} is enabled.
 installErrorMessage = Uh oh. {$title} could not be enabled. Try again later.
 participantCount = <span>{$installation_count}</span> participants
 
+experimentPreFeedbackTitle = {$title} feedback
+experimentPreFeedbackLinkCopy =Give general feedback about {$title}
```
chuckharmston: Needs a space after `=`

frontend/src/app/components/ExperimentPreFeedbackDialog.js
```diff
@@ -0,0 +1,65 @@
...
+                data-l10n-id="experimentPreFeedbackTitle"
+                data-l10n-args={l10nArgs}>{experiment.title} Feedback</h3>
+                <div className="tour-cancel" onClick={e => this.cancel(e)}/>
+          </header>
+            <div className="tour-content">
+              <div className="tour-image">
+                <img src={experiment.pre_feedback_image} />
+                <div className="fade" />
+              </div>
+              <div className="tour-text" dangerouslySetInnerHTML={{ __html: experiment.pre_feedback_copy }} />
```
chuckharmston, line 29: I would really rather not do this. What's the use case?

chuckharmston (changes requested):

frontend/src/app/components/ExperimentPage.js
johngruen: @chuckharmston so, should i do anything about this now, or are you just noting for the future?

chuckharmston: I'd rather you fix it now, but I won't block merge for it.

content-src/experiments/activity-stream.yaml
johngruen: yeah, we don't need to...I just figured we should keep the various files uniform

frontend/src/app/components/ExperimentPage.js
johngruen: Let's do all of this in #1367, since we're going to be adding all of this stuff to the redux store there anyway.

frontend/src/app/components/ExperimentPreFeedbackDialog.js
johngruen: @chuckharmston yeah, not necessary here

johngruen:
Alright, should be resolved % redux store stuff, which should be done collectively in #1367

content-src/experiments/activity-stream.yaml
johngruen: I mean, these fields don;t need to be here...i just figured we could keep the yaml files consistent

frontend/src/app/components/ExperimentPreFeedbackDialog.js
johngruen: @chuckharmston @lmorchard actually, it appears to be, though i'm not exactly sure why...but I'm seeing the same thing elsewhere (ie. in the ExperimentTourDialogue)

frontend/src/app/components/ExperimentPage.js
```diff
@@ -366,15 +372,27 @@ export default class ExperimentPage extends React.Component {
...
+  showPreFeedbackDialog() {
+    this.setState({
+      showPreFeedbackDialog: true
     });
   }
 
+  handleFeedback(evt) {
+    evt.preventDefault();
+    const { pre_feedback_copy } = this.state.experiment;
+    (pre_feedback_copy === null || pre_feedback_copy === 'undefined') ? this.feedback(evt) : this.showPreFeedbackDialog();
```
johngruen: @chuckharmston added the 'undefined' thing to the test here if the pre_feedback fields are not in the model

lmorchard:
Looks like changes make this ready for another review? Will take a peek.

content-src/experiments/activity-stream.yaml
lmorchard: This could be a job for #1308 - can leave things out / decide what's required and what's valid in a linting step. We should, like, document what's in these YAML files someday soon too

frontend/src/app/components/ExperimentPage.js
lmorchard: It's getting kinda hectic, so we can borrow some tech-debt and make #1367 bigger IMO

frontend/src/app/components/ExperimentPage.js
lmorchard: Should this be `typeof(pre_feedback_copy) === 'undefined'`? You could probably get away with just a `!pre_feedback_copy` unless we ever want the copy to literally be "0". Also should be a full if / else statement rather than a ternary condition, since we're not doing anything with the result value.

frontend/src/app/components/ExperimentPreFeedbackDialog.js
lmorchard: Assuming we want markup allowed in `pre_feedback_copy`, [this is the way to do it](https://facebook.github.io/react/tips/dangerously-set-inner-html.html) as far as I understand. It looks ugly as a warning. But, I think it's okay because it's from our own JSON & YAML and we don't need to sanitize our own content.

lmorchard (commented):
Only really one nit, with the ternary/if thing. Adding a little bit for #1367 to do, but that's probably okay for now? Maybe we need a //TODO comment there so we can tally our debt someday?

johngruen:
@chuckharmston @lmorchard all requested changes made

lmorchard (approved):
Works for me

lmorchard:
Not sure how to unblock this for a merge, though. My review approves, but doesn't seem to override the one still requesting changes

chuckharmston (approved):

chuckharmston:
Approving and merging based on Les' r+

PR #1401, opened 2016-09-21
Bugfix for static_experiments_json waffle flag when DEBUG=False

lmorchard:
Should probably have tests for this, too, but should be a throwaway hack until we move off Django entirely.


chuckharmston (approved):
:shipit:

Issue #944, opened 2016-06-03
Create place in experiment details to call out incompatable add-ons


johngruen:
Some experiments collide with other add-ons. We should call these out on experiment details pages.

pdehaan:
HttpFox seems to be incompatible (Ref: #796)

ghost:
Let's do:

1) A standard place in the repositories for a list of incompatible add-ons
2) Something on the details page ( --> needs ux)

johngruen:
@sevaan is on it

sevaan:
Please find the UX attached. This warning box would appear only if:
- the experiment is active
- the user has conflicting Add-ons installed and **enabled**

This box would stack above other boxes (e.g. experiment is ending soon).

Clicking the link would open up the Add-ons manager in a new active tab.

![incompatible add-ons](https://cloud.githubusercontent.com/assets/473531/17982882/68ca49de-6acf-11e6-913b-27002c6cb467.png)
[Assets.zip](https://github.com/mozilla/testpilot/files/437887/Assets.zip)

johngruen:
@chuckharmston wanna pick this up?

ghost:
I was dragging my feet on any frontend bugs, waiting on the flat files patch to land, fwiw.

johngruen:
cool

johngruen:
We may need to make this apply to about config prefs as well, please see: 
https://github.com/mozilla/blok/issues/12

chuckharmston:
I'm not in love with this. The big red box looks like an error message, and it won't apply whatsoever to large proportions of our users. The Test Pilot add-on can:
- Get a list of installed add-ons
- Read prefs
- Change prefs

We should just change prefs for the user (maybe with a callout if we do so), only display the incompatibility message if we know that the user has an incompatible add-on installed, and prevent installation in that event.

johngruen:
@chuckharmston that's the point i think, we don't wanna show this to everybody

johngruen:
@sevaan wants this iff

> the user has conflicting Add-ons installed and enabled

I agree, the red is a little brutal. let's start by getting it in there, and see if we can't soften it.

chuckharmston:
Huh, sorry, I was reading the thread in email and everything but the mock in Sevaan's message got truncated as if it was quoted text.

johngruen:
@chuckharmston, lets user $eol-light-orange as the background and $eol-orange as the border

johngruen:
[...]

PR #1402, opened 2016-09-21
Display list of incompatible add-ons on experiment detail page.

chuckharmston:
Notes:
- Webpages are not allowed to link to `about:*` pages, so we now link to https://support.mozilla.org/kb/disable-or-remove-add-ons
- This assumes we're shipping with YAML.
- #1403 filed as a code quality followup
## Screenshots

Incompatible add-ons installed:

Incompatible add-ons installed.

No incompatible add-ons installed:

No incompatible add-ons installed.

Old version of the Test Pilot add-on installed (falls back on default state, is as if no incompatible add-ons are installed)

Old version of the Test Pilot add-on installed


sevaan:
+r!

chuckharmston:
screen shot 2016-09-21 at 12 28 50 pm

Updated with new colors from Sevaan. Will revert to orange if we're able to do the other half of #944.

chuckharmston:
This now closes all of #944, and is only shown if incompatible add-ons are installed.

r? @lmorchard @johngruen

ericawright:
More incompatible with Tab Center:
id: multipletab@piro.sakura.ne.jp, Multiple Tab Handler
id: tabscope@xuldev.org, Tab Scope
id: bug566510@vovcacik.addons.mozilla.org, Allow multiselect operations on tabs
id: jid1-rOGUyxs2TvBBuA@jetpack, Previous tab highlighting 
[Diigo Toolbar](https://addons.mozilla.org/en-US/firefox/addon/diigo-web-highlighter-and-stic/) (Sorry, I can't find the ID)

meandavejustice:
This looks good and is working for the most part for me.

I've found one issue thought. After seeing that firebug needs to be disabled, I do so but am still prompted with the same message after a page load. It looks like `installedAddons` might not be getting updated.

chuckharmston:
Yeah, I'm not sure what we should do there. I guess we can ping the website from the add-on with every install and uninstall, but sheesh, that's a lot of noise.

Thoughts @johngruen @lmorchard?

meandavejustice:
@chuckharmston Yeah, i dunno. I mean people probably aren't going to be installing/uninstalling a whole lot. It seems reasonable to update the state in those cases. It seems like the type of thing that would future proof against similar bugs as well.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -95,6 +95,42 @@ export default class ExperimentPage extends React.Component {
...
+    const installed = this.getIncompatibleInstalled(incompatible);
+    if (!installed) return null;
+
+    const helpUrl = 'https://support.mozilla.org/kb/disable-or-remove-add-ons';
+
+    return (
+      <section className="incompatible-addons">
+        <header>
+          <h3 data-l10n-id="incompatibleHeader">
+            This experiment may not be compatible with Add-ons you have installed.
```
ghost: Not a big deal, but I don't think Add-ons needs to be capitalized.

chuckharmston: Yeah, I didn't think so either, but I assumed @sevaan knew better than me and went with his mocks.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -95,6 +95,42 @@ export default class ExperimentPage extends React.Component {
...
+    return Object.keys(incompatible).filter(guid => (
+      this.props.installedAddons.indexOf(guid) !== -1
+    ));
+  }
+
+  renderIncompatibleAddons() {
+    const { incompatible } = this.state.experiment;
+    if (!incompatible) return null;
+    const installed = this.getIncompatibleInstalled(incompatible);
+    if (!installed) return null;
```
dannycoates: This needs something like a `|| installed.length === 0` otherwise I see this:

screen shot 2016-09-22 at 11 32 37 pm

dannycoates (changes requested):
LGTM other than the empty array case 👍

frontend/src/app/components/ExperimentPage.js
johngruen: Yeah, this should be lowercase

johngruen:
@chuckharmston updating state live would be preferable, but not a blocker. If you don't do it here, please file a separate issue.

chuckharmston:
Disabled add-ons are currently included in the list; need to filter on that.

chuckharmston:
Updated to:
- Filter out disabled add-ons.
- Include additional add-ons incompatible with Tab Center.
- Monitor disables, enables, uninstalls, installs, and undos, updating live.
- Handle the empty array case Danny pointed out.

Screencap: https://vid.me/gc0B

dannycoates (approved):

PR #1405, opened 2016-09-21
add launch date to yaml (closes #1393)

### LINKED ISSUE #1393: Add start_after field to YAML so we can ship stuff in yaml without it launching

johngruen:
this adds a utc time stamped `launch_date` key to experiment data. 

if you're working on dev or local, this is ignored...on stage and prod, pre-launch experiments are filtered out of `experiments.fetch`

### DIFF
--- a/content-src/experiments/activity-stream.yaml
+++ b/content-src/experiments/activity-stream.yaml
@@ -86,4 +86,5 @@ contributors:
 installation_count: 46732
 created: '2016-03-11T21:05:59.363520Z'
 modified: '2016-09-06T18:30:42.286466Z'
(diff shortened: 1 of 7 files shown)


chuckharmston (changes requested):
Blah, sorry @johngruen, but I don't like this approach. I don't like this approach; the store should be the source of truth, and filters should be happening at the component level, not in actions or reducers.

We should be using [reselect](https://github.com/reactjs/reselect) to do this.

johngruen:
@chuckharmston wanna pick this up and redo it then? this is blocking

Issue #1346, opened 2016-09-14
Missing GA events (and all pageviews)


ckprice:
Copying from https://github.com/mozilla/testpilot/issues/1307#issuecomment-246800267

Can't see a couple GA events/pageviews firing
- Viewing the home page without the add-on installed ([ref](https://github.com/mozilla/testpilot/blob/6bd599eb1c3396883a7a98acaf6671dd49e03132/frontend/src/app/views/landing-page.js#L39)).
- Clicking `Proceed` on retiring modal and going to the retire page ([ref](https://github.com/mozilla/testpilot/blob/6bd599eb1c3396883a7a98acaf6671dd49e03132/frontend/src/app/views/retire-page.js#L45)).
- Viewing the experiment detail pages ([ref](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/views/experiment-page.js#L266)).

Eh, looks like there are no `pageviews` sends in the branch.

ckprice:
The router is a good spot for this!

ghost:
@ckprice please confirm with people when you assign them issues.  thanks!

ckprice:
Eh, this was a dupe of https://github.com/mozilla/testpilot/issues/1362#issuecomment-247641915 which Les took. It was discussed in the meeting, but I can try to be pingier.

ckprice:
Closed in #1406

PR #1406, opened 2016-09-21
Add missing GA metric pings

lmorchard:
- send event for "Proceed" clicked in Retire dialog
- for landing page, experiments list, and experiments page
- add flags in store to indicate whether experiments list and and
  installed status are loaded
- isInstalledLoaded, isExperimentsLoaded, getExperimentBySlug selectors

Issue #1346

### DIFF
--- a/frontend/src/app/actions/addon.js
+++ b/frontend/src/app/actions/addon.js
@@ -1,8 +1,13 @@
 import { createActions } from 'redux-actions';
 
 export default createActions(
-  {},
-  'setHasAddon', 'setInstalled', 'setClientUuid',
+  {
+    setInstalled: installed => ({
+      installed: installed || {},
+      installedLoaded: !!installed
+    })
+  },
+  'setHasAddon', 'setClientUuid',
   'enableExperiment', 'disableExperiment',
   'requireRestart'
 );
--- a/frontend/src/app/actions/experiments.js
+++ b/frontend/src/app/actions/experiments.js
@@ -15,6 +15,7 @@ export default createActions({
       .then(response => response.json())
       .then(counts => ({
         lastFetched: Date.now(),
+        experimentsLoaded: true,
         experiments: experiments.map(experiment => ({
           ...experiment,
           installation_count: counts[experiment.addon_id]
--- a/frontend/src/app/components/ExperimentPage.js
+++ b/frontend/src/app/components/ExperimentPage.js
@@ -21,8 +21,8 @@ export default class ExperimentPage extends React.Component {
   constructor(props) {
     super(props);
 
-    const { isExperimentEnabled } = this.props;
-    const experiment = this.findCurrentExperiment(props);
+    const { isExperimentEnabled, getExperimentBySlug } = this.props;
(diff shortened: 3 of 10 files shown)


lmorchard:
Taking a break from this one for a little bit, but could use some eyes & second opinions. Currently has a problem in that it sends multiple pageview pings as the state of the app updates. So this happens in rough order:
1. Initial page load - addon not yet detected, no known installed experiments, no experiment list loaded
2. Add-on detected.
3. App receives installed experiment list from add-on.
4. Experiments list fetched.

There's a GA ping for each of those stages, as well as possibly a redirect if the user is on the wrong page with respect to add-on installation status.

Ideally, I think we'd just want _one_ pageview event, after all the above has shaken out. Problem is, we don't really have a way to know when that's all done. Maybe the add-on's not really not installed and it's not that we just haven't detected it yet - that happens async, so we don't have a point where we _know_ there are no installed experiments.

Maybe a good hack would be to just set up a debounce timer and accumulate info for a GA ping, but wait to send it until after a second or so?

lmorchard:
Oh, and the experimentsLoaded & installedLoaded flags are there because I thought maybe I could hold off sending a ping until they were both true. But, installedLoaded will _never_ be true if there's no addon. So, I don't use them yet. Thinking of removing that code.

dannycoates:
can we leverage the `navigator.testpilotAddon` flag to determine if we should wait for `installedLoaded`?

_full disclosure_: I haven't read the code 😉

chuckharmston:
> can we leverage the navigator.testpilotAddon flag to determine if we should wait for installedLoaded?

The problem is that we aren't guaranteed that the addon has fully initialized before our page is rendered.

lmorchard:
As far as I can tell from further tinkering, I'm wrong about the `navigator.testpilotAddon` flag being unreliable / delayed at the start of our app code running. That, or it's a race condition I haven't been able to expose.

Changing the default initial value of `addon.hasAddon` in the store from `false` to `!!window.navigator.testpilotAddon` seems to always end up being correct on the first try, with no follow-up state change.

So, got an update here that relies on that being correct at the start and will send just one GA ping after waiting for experiments to load and (if the add-on is present) for installed experiment data to arrive.

One last catch I'm tracking down: When the add-on installation state changes - i.e. after hitting the big green button - state changes go a little nuts again, and we get multiple GA pings

lmorchard:
I think this might be as good as I'm going to get:
- Always wait for `experimentsLoaded`
- On initial page load, rely on the `hasAddon` state to tell us whether or not to wait for `installedLoaded` before sending a GA ping
- Only send one GA ping for a given page path - because when `hasAddon` state changes after page load, all kinds of state changes happen that trigger the measurement handler and cause multiple GA pings

That last point may be where we could drop some page views - that is, if you're on a landing page or an experiment page, there will only be a ping for that first visit and not another one for the refresh / state change after the add-on is installed. There _will_ be another ping, if the user actually refreshes in the browser or navigates to another page, as expected

dannycoates (approved):
I'm not familiar with the previous GA code but I do observe pageviews and events as expected compared with navigating through production. Code looks good 👍

PR #1407, opened 2016-09-21
Disable tracking protection while blok is enabled

### LINKED ISSUE #1333: Toggle Native Tracking Protection Pref off if user installs Test Pilot Tracking Protection
Related to #944, and we may not need a separate issue here, but we can cover this in triage today.

dannycoates:
Fixes #1333

This doesn't cover the edge cases where a user disables Test Pilot from about:addons then disables or removes Blok.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -15,6 +15,7 @@ const { PrefsTarget } = require('sdk/preferences/event-target');
 const URL = require('sdk/url').URL;
 
 const { App } = require('./lib/app');
+const ExperimentHacks = require('./lib/experiment-hacks');
 const ExperimentNotifications = require('./lib/experiment-notifications');
 const FirstRun = require('./lib/first-run');
 const Metrics = require('./lib/metrics');
@@ -157,6 +158,11 @@ function experimentsLoaded(experiments) {
 function setAddonActiveState(addon) {
   const xp = store.availableExperiments[addon.id];
   if (xp) { delete xp.active; }
+  if (addon.isActive) {
+    ExperimentHacks.enabled(addon.id);
+  } else {
+    ExperimentHacks.disabled(addon.id);
+  }
   store.installedAddons[addon.id] = Object.assign({
     active: addon.isActive,
     installDate: addon.installDate
@@ -180,11 +186,6 @@ function installExperiment(experiment) {
 }
 
 function uninstallSelf() {
-  // First, kick out all the experiment add-ons
-  Object.keys(store.installedAddons).forEach(id => {
-    uninstallExperiment({addon_id: id});
-  });
-  // Then, uninstall ourselves
   AddonManager.getAddonByID(self.id, a => a.uninstall());
 }
 
@@ -253,6 +254,7 @@ const addonListener = {
         name: addon.name,
         version: addon.version
       });
+      ExperimentHacks.uninstalling(addon.id);
     }
   },
   onUninstalled: function(addon) {
@@ -361,6 +363,7 @@ exports.onUnload = function(reason) {
     if (store.installedAddons) {
       Object.keys(store.installedAddons).forEach(id => {
         uninstallExperiment({addon_id: id});
+        ExperimentHacks.uninstalled(id);
       });
       delete store.installedAddons;
     }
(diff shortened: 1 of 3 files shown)


addon/index.js
```diff
@@ -180,11 +181,6 @@ function installExperiment(experiment) {
 }
 
 function uninstallSelf() {
-  // First, kick out all the experiment add-ons
-  Object.keys(store.installedAddons).forEach(id => {
-    uninstallExperiment({addon_id: id});
-  });
```
dannycoates, line 186: This is already handled in `onUnload`

addon/index.js
```diff
@@ -361,6 +362,7 @@ exports.onUnload = function(reason) {
     if (store.installedAddons) {
       Object.keys(store.installedAddons).forEach(id => {
         uninstallExperiment({addon_id: id});
+        ExperimentHacks.uninstalled(id)
```
dannycoates: The `addonListener` won't get triggered at this point, so we need to do it manually

addon/lib/experiment-hacks.js
```diff
@@ -0,0 +1,47 @@
...
+
+const PrefsService = require('sdk/preferences/service');
+const store = require('sdk/simple-storage').storage;
+
+const BLOK_ID = 'blok@mozilla.org';
+const TP_PREFS = [
+  'privacy.trackingprotection.enabled',
+  'privacy.trackingprotection.pbmode.enabled',
+  'services.sync.prefs.sync.privacy.trackingprotection.enabled',
+  'services.sync.prefs.sync.privacy.trackingprotection.pbmode.enabled'
```
dannycoates, line 15: Sync prefs weren't in the original bug, but we probably want to turn off syncing of these prefs too. @clouserw @johngruen ?

lmorchard (approved):
:+1: with some nits / questions:
- Could use some tests for the new ExperimentHacks module
- We're adding hacks to the main add-on to disable tracking protection when blok is installed... what happens with users whose Test Pilot add-ons haven't upgraded yet?

lmorchard:
Just thinking out loud if this ends up being a problem: Might make sense to add a bit to the hack that checks after Test Pilot upgrade or browser restart if Blok is installed and turns off tracking protection. Probably a tiny edge case not worth worrying about though

chuckharmston:
Meta-question: why build this as a one-time-use hack, when we could store this information in the yaml for the experiment and handle it more globally?

``` yaml
prefs:
- name: something.whatever
  type: boolean
  enable: true
  disable: false
- name: something.whateverelse
  type: string
  enable: hello
  disable: goodbye
```

This is surely not a unique use case, and this is exactly the sort of thing Test Pilot should be federating for experiments, right?

meandavejustice:
👍 

@chuckharmston That's a good point, personally I would wait until there is another addon that needs to set prefs before including it in the yaml.

The upgrade case that @lmorchard mentions is valid. That seems like a simple enough check in `exports.unLoad` in `index.js`

dannycoates:
> Might make sense to add a bit to the hack that checks after Test Pilot upgrade or browser restart if Blok is installed and turns off tracking protection

Ok, I moved the logic for when hacks get triggered that will catch this case. Expect tests shortly

dannycoates:
tests added

addon/lib/experiment-hacks.js
ghost: Seems right.  If they are set I assume Firefox would keep turning it back on

lmorchard:
Taking another look...

addon/test/test-experiment-hacks.js
```diff
@@ -0,0 +1,123 @@
+/*
```
lmorchard, line 1: :+1: Thanks for circling back to do tests!

lmorchard (approved):
Looks good to me and tests make me feel warm & fuzzy

lmorchard:
> This is surely not a unique use case, and this is exactly the sort of thing Test Pilot should be federating for experiments, right?

Yeah, for this I would say we optimize that when we have a 2nd one that needs the same thing. I also suspect we may need to keep this "hacks" module around in case future experiments need different weird things done for them that can only be done safely/securely/sanely from the main Test Pilot add-on.

PR #1409, opened 2016-09-22
add launch dates

johngruen:
@chuckharmston I think i can add a bunch of stuff to the browser reducer once my other PR gets merged.

### DIFF
--- a/frontend/src/app/components/ExperimentCardList.js
+++ b/frontend/src/app/components/ExperimentCardList.js
@@ -1,22 +1,26 @@
 import React from 'react';
+import moment from 'moment';
 import ExperimentRowCard from './ExperimentRowCard';
 import Loading from './Loading';
 
 export default function ExperimentCardList({
-  navigateTo, isExperimentEnabled, hasAddon, experiments, except, eventCategory
+  navigateTo, isExperimentEnabled, hasAddon, isDev, experiments, except, eventCategory
 }) {
   const experimentsSorted = [].concat(experiments);
+
   experimentsSorted.sort((a, b) => {
     if (a.id > b.id) { return 1; }
     if (a.id < b.id) { return -1; }
     return 0;
   });
 
+  const experimentsReduced = setCurrentExperiments(isDev, experimentsSorted);
+
   return (
     <div className="card-list experiments">
       {(experiments.length === 0) ?
         <Loading /> :
-        experimentsSorted.map((experiment, key) =>
+        experimentsReduced.map((experiment, key) =>
           (!except || except.slug !== experiment.slug) &&
             <ExperimentRowCard key={key}
                                navigateTo={navigateTo}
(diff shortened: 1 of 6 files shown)


frontend/src/app/components/ExperimentPage.js
```diff
@@ -127,6 +129,10 @@ export default class ExperimentPage extends React.Component {
       statusType = 'enabled';
     }
 
+    if (moment(utcNow).isBefore(experiment.launch_date) && typeof experiment.launch_date !== 'undefined') {
+      this.props.history.push('/not-found');
```
lmorchard: Don't use `this.props.history.push` directly, here. Instead use the `navigateTo` store dispatch helper. You probably also want to return empty content, rather then carry on to render the rest of the page. Maybe, something like:

```
navigateTo('/not-found');
return '';
```

frontend/src/app/config.js
```diff
@@ -1,4 +1,6 @@
 export default {
   experimentsURL: '/api/experiments.json',
-  usageCountsURL: 'https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json'
+  usageCountsURL: 'https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json',
+  productionURL: 'https://testpilot.firefox.com',
+  stagingURL: 'https://testpilot.stage.mozaws.net'
```
lmorchard: Maybe this is something to look at later, but something I want to do with this config file is give it a hostname-based overlay. So, like:

``` javascript
const defaultConfig = {
 isDev: true,
 experimentsURL: '/api/experiments.json',
 usageCountsURL: 'https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json'
};

const hostConfig = {
  'testpilot.firefox.com': { isDev: false },
  'testpilot.stage.mozaws.net': { isDev: false }
};

export default Object.assign({}, defaultConfig, hostConfig[location.hostname] || {});
```

Could do this now, or keep it in mind to implement later. Either way, it would be nice not to sprinkle host name checks everywhere if/when we want to do more things that differ between environments

johngruen: @lmorchard yeah, looks good.

chuckharmston: We should stash this somewhere that gets copied to both the addon and site in the build process. Silly to define it twice.

frontend/src/app/components/ExperimentCardList.js
```diff
@@ -27,3 +31,16 @@ export default function ExperimentCardList({
...
+  const currentExperiments = [];
+  const utcNow = moment.utc();
+  if (isDev) return allExperiments;
+  allExperiments.forEach((experiment) => {
+    if (moment(utcNow).isAfter(experiment.launch_date) || typeof experiment.launch_date === 'undefined') {
+      currentExperiments.push(experiment);
+    }
+  });
+  return currentExperiments;
+}
```
dannycoates, line 45: I won't bust your chops over putting this here, but its a good candidate to be a selector on the experiments reducer.

dannycoates (approved):
Looking good @johngruen

johngruen:
@dannycoates @lmorchard just pushed to change commit message.

Issue #1411, opened 2016-09-22
Final configurations for new experiments


ckprice:
In #1392 we pushed out partially complete YAML pages so review could happen outside of a local environment. This is opened as a follow-up and needs to be closed before launch
- [x] Confirm contributors (#1416)
- [x] Confirm tour/detail image copy (#1430)
- [x] Confirm final XPI & ID's (#1436)
- [x] Discourse needs to be set up (Bugs [1304878](https://bugzilla.mozilla.org/show_bug.cgi?id=1304878), [1304879](https://bugzilla.mozilla.org/show_bug.cgi?id=1304879), [1304880](https://bugzilla.mozilla.org/show_bug.cgi?id=1304880))
- [x] Confirm colors and image (#1414)
- [x] Add UTC launch_date to each experiment (#1409)
- [x] Add min_version: 48 to Tracking Protection (#1427)
- [x] ~~Add conflicting addons to Tracking Protection~~ #1402 is no longer a blocker
- [x] Facebook/Twitter images (#1431)
- [x] Maybe hide counts if they don't exist yet ([Ref](https://github.com/mozilla/testpilot/pull/1410/commits/67d2d9e0780d28a171df9928e703ff6488bab662#diff-5eccc665b4f05b0644484ec36a9c9c9cR117))
- [x] If #1398 add pre-feedback info for tracking protection
- [x] add surveygizmo urls (Robert confirmed over email)

johngruen:
contribs added in #1416

johngruen:
user counts dealt with in #1425

johngruen:
Images finalized in: #1426

johngruen:
@clouserw and I just realized we can skip the launch_date field and just throw the waffle flag on prod

johngruen:
Min Version and prefeeback added to tracking protection: #1427

ckprice:
🎉

PR #1414, opened 2016-09-23
WIP finalize sept launch assets

johngruen:
@ckprice PR-ing for visibility, I've reoptimized all of the images and changed the color scheme and order. This should be a part of #1411


ghost (approved):
very optimal

PR #1415, opened 2016-09-23
set isMinFirefox on ExperimentPage

dannycoates:
Without this the "Upgrade Firefox" button is always shown on experiment pages when the addon is not installed.


johngruen:
@dannycoates ah, good catch

johngruen (approved):

johngruen:
R+

PR #1417, opened 2016-09-23
set timeout=60 on integration test

clouserw:


clouserw:
I don't know if this PR will fix our problem.  Other PRs are passing, so it seemed intermittent, but the failure on master has been pretty consistent for the past hour and a half.

chuckharmston (approved):
Not a great solution, but this seems fine to keep it green for now.

Issue #1399, opened 2016-09-21
Backfill frontend tests for new React system


lmorchard:
In #1351, we added a new system for frontend unit tests with React. That PR only came with 2 tests. We should backfill the rest.

lmorchard:
Going to say this is closed by #1488 - though there are still 3 components that need tests, as well as reducers & actions that could use some tests. I'll file separate issues for the remaining components

PR #1425, opened 2016-09-23
resolves low user counts for launch experiments pulling from telemetry

johngruen:


frontend/src/app/components/ExperimentPage.js
```diff
@@ -341,6 +340,23 @@ export default class ExperimentPage extends React.Component {
...
   }
 
+  // this is set to 100, to accomodate Tracking Protection
+  // which has been sending telemetry pings via installs from dev
+  // TODO: figure out a non-hack way to toggle user counts when we have
+  // telemetry data coming in from prod
+  renderInstallationCount(installation_count, title) {
+    if (installation_count <= 100) {
+      return (
+        <span data-l10n-id="userCountContainerAlt" className="bold" data-l10n-args={JSON.stringify({ title })}>
```
johngruen, line 350: @lmorchard just noticed i left the data-l10n-arg in there. i can get rid of this

lmorchard (approved):
We need to start adding unit tests for new stuff like this, but we add this to the debt in #1399 for now

PR #1428, opened 2016-09-23
set a max height for the addon panel

### LINKED ISSUE #1412: Addon panel not fully visible on lower res screens
screen shot 2016-09-22 at 6 14 48 pm

At 1024x640 "View All Experiments" isn't visible and there's no scrolling.

dannycoates:
fixes #1412

Sets a max height to 4¾ experiment tiles to indicate there's more content.

### DIFF
--- a/addon/data/panel.css
+++ b/addon/data/panel.css
@@ -66,11 +66,9 @@ a {
 }
 
 .experiment-list {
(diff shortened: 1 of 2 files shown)


lmorchard (approved):
Works for me!

PR #1437, opened 2016-09-23
Disable integration test on CircleCI for now

lmorchard:
The single integration test we have for add-on installation on Firefox 48 seems to be failing intermittently for no real issue that I can reproduce outside of CircleCI. Maybe we should disable it for now and get it out of our way? With the longer timeout, it's just adding 60 seconds before reporting failure.


dannycoates (approved):
Ok

Issue #1434, opened 2016-09-23
Experiment pages should use the metadata configured in YAML for social shares


ckprice:
Looks like the experiment pages are using the Test Pilot logo on experiment pages for `og:image` and `twitter:image`. It should use [the values configured in the experiment YAML](https://github.com/mozilla/testpilot/blob/master/content-src/experiments/min-vid.yaml#L8).

lmorchard:
Ugh. Looks like our switch over to static YAML isn't fully easy with Django - it's still using the database for those page templates. We don't have any code in Django to read YAML right now, just to switch the JS frontend over to the JSON built from YAML. That doesn't affect server-generated HTML results.

lmorchard:
So, easiest fix (engineering-wise and deployment-wise): Add records for these new experiments in Django admin, use dummy content for all fields except the social thumbnails.

Harder fix, we get Django reading the YAML files to fill in the index.html frontend template when the static switch is turned on.

lmorchard:
Tweaked title, because it's more than images that are the problem

lmorchard:
Looks like we went with the manual solution for this - need to keep title, description, image_facebook, and image_twitter for experiments up to date in Django admin until we finish switching away from Django.

PR #1438, opened 2016-09-23
Fix canonical & social image URL paths for static experiment pages

lmorchard:
- Point canonical header link to prod experiment page path
- Un-escape canonical and social image paths

### DIFF
--- a/frontend/src/templates/index.mustache
+++ b/frontend/src/templates/index.mustache
@@ -11,7 +11,7 @@
     <meta name="viewport" content="width=device-width">
     <link rel="localization" href="/static/locales/{locale}/app.ftl">
 
-    <link rel="canonical" href="https://testpilot.firefox.com/">
+    <link rel="canonical" href="https://testpilot.firefox.com/{{{ canonical_path }}}">
 
     <title>{{ meta_title }}</title>
     <meta property="og:title" content="{{ meta_title }}" />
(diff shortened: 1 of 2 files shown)


lmorchard:
This doesn't help us for #1434, but I think it does fix an issue I think I see on testpilot-static.dev.mozaws.net in trying to validate the sharing headers. In particular:
- Facebook sees the canonical URL pointed to the home page, and uses that for the sharing card instead of the experiment page.
- Twitter sees the HTML-encoded URL path and doesn't recognize it as an image.

frontend/src/templates/index.mustache
```diff
@@ -11,7 +11,7 @@
     <meta name="viewport" content="width=device-width">
     <link rel="localization" href="/static/locales/{locale}/app.ftl">
 
-    <link rel="canonical" href="https://testpilot.firefox.com/">
+    <link rel="canonical" href="https://testpilot.firefox.com/{{{ canonical_path }}}">
```
dannycoates, line 14: Locally, when I open the file from the filesystem I see the correct path here, but if I curl (for example) `http://testpilot.dev:8000/experiments/min-vid` I only see `https://testpilot.firefox.com/`. I can't tell if that's a side effect of the dev server or not.

lmorchard: Yeah, I think that's an artifact of the [connect-history-api-fallback](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/server.js#L13) module showing the root `index.html`. If you add a trailing slash, you should see something different - i.e. `min-vid/`

dannycoates: even with the trailing slash I'm not seeing the full path here, but it seems likely its just a quirk of the local server.

dannycoates (approved):
LGTM

PR #1442, opened 2016-09-23
Use thumbnail paths from YAML when static_experiments_json waffle flag enabled

### LINKED ISSUE #1434: Experiment pages should use the metadata configured in YAML for social shares

lmorchard:
- When static_experiments_json waffle flag is enabled, read the YAML
  source for flat file experiment content, create a fake experiment
  model object sufficient for populating the frontend template
- Add PyYAML dependency
- Quote min-vid adon_id in YAML - breaks in PyYAML parsing

Fixes #1434


lmorchard:
I am... not entirely proud or happy with this code. No tests. It seems to fix the issue for now, though. I won't be sad if we don't launch with it before we go fully static deployment.

lmorchard:
Updated commit to pass linting

testpilot/experiments/models.py
```diff
@@ -33,6 +38,22 @@ def for_request(self, request):
...
+                    data = yaml.load(open(path, 'r').read())
+                    return type('obj', (object,), dict(
+                        title=data['title'],
+                        description=data['description'],
+                        image_facebook=type('obj', (object,),
+                                            dict(url=data['image_facebook'])),
+                        image_twitter=type('obj', (object,),
+                                           dict(url=data['image_twitter'])),
+                    ))
+                except:
```
chuckharmston: Maybe qualify this with any possible exceptions?

chuckharmston, line 55: Maybe qualify this with any possible exceptions?

chuckharmston (approved):
Reaaaalllllly should probably have tests since we're doing file IO, but meh.

lmorchard:
This was an emergency PR and the emergency has long passed. We're doing it manually now until Django is shutdown

PR #1449, opened 2016-09-26
update min-vid xpi url

clouserw:
The 0.2.0 .xpi is at that URL right now, just so this doesn't 404.  I'll update the .xpi as soon as we have another one.


chuckharmston (approved):
Feel free to self-merge when the build finishes.

jaredhirsch (approved):
:+1: lgtm

PR #1454, opened 2016-09-27
Disables first run popup for all clients.

chuckharmston:


lmorchard (approved):
Yup, should turn it off for everyone!

PR #1460, opened 2016-09-27
Ensure frontend state updates when add-on uninstalled

### LINKED ISSUE #1429: Clicking "Home" button from retire page displayed after Test Pilot uninstall, wrongly redirects you to experiments page
[Affected versions]:
- Firefox 49  and up

[Affected Platforms]:
- All Windows

[Prerequisites]:
- The "testpilot.env" string pref was manually added with "dev" value on a new clean profile in "about:config" page.
- Test Pilot add-on was previously installed on the profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to testpilot.dev.mozaws.net/experiments/.
2. Click on "Settings" button from top right, choose "Uninstall" option and click the "Proceed" button.
3. Click the "Home" button and observe the browser behavior.

[Expected result]:
- Test Pilot is successfully uninstalled and you are redirected to Test Pilot homepage.

[Actual result]:
- Test Pilot is successfully uninstalled but you are wrongly redirected to "Experiments" page and the unistall option is still available. 

[Notes]:
- This issue is also reproducible on stage server.
- Added a screen recording with the issue: https://goo.gl/48brhD.

lmorchard:
- Set `window.navigator.testpilotAddon` to `false` on add-on uninstall
- Set `window.navigator.testpilotAddon` to `false` in retire dialog if
  add-on hasn't notified of uninstall within 5 seconds

Fixes #1429

### DIFF
--- a/addon/data/set-installed-flag.js
+++ b/addon/data/set-installed-flag.js
@@ -10,3 +10,8 @@
 
 unsafeWindow.navigator.testpilotAddon = true;
 unsafeWindow.navigator.testpilotAddonVersion = self.options.version;
+
+self.port.on('detach', function() {
+  delete unsafeWindow.navigator.testpilotAddon;
+  delete unsafeWindow.navigator.testpilotAddonVersion;
+});
--- a/addon/package.json
+++ b/addon/package.json
@@ -1,7 +1,7 @@
 {
   "title": "Test Pilot",
   "name": "testpilot-addon",
-  "version": "0.8.5",
+  "version": "0.8.6",
   "private": true,
   "description": "Test Pilot is a privacy-sensitive user research program focused on getting new features into Firefox faster.",
   "repository": "mozilla/testpilot",
--- a/frontend/src/app/components/RetirePage.js
+++ b/frontend/src/app/components/RetirePage.js
@@ -20,6 +20,7 @@ export default class RetirePage extends React.Component {
   componentDidMount() {
     this.fakeUninstallTimer = setTimeout(() => {
       this.setState({ fakeUninstalled: true });
+      window.navigator.testpilotAddon = false;
     }, FAKE_UNINSTALL_DELAY);
   }
 
@@ -32,6 +33,9 @@ export default class RetirePage extends React.Component {
     const { fakeUninstalled } = this.state;
 
     const uninstalled = !hasAddon || fakeUninstalled;
+    if (uninstalled) {
+      clearTimeout(this.fakeUninstallTimer);
+    }
 
     return (
       <div className="full-page-wrapper centered">


lmorchard:
So, this "fixes" the immediate problem - one way or another, the site state changes to reflect add-on uninstall after the retire flow is completed. And in all cases, the add-on does indeed uninstall itself.

But, an underlying problem is exposed: It seems that in many manual attempts, the add-on never manages to run its uninstall code. The `addon-self:uninstalled` message never gets sent to the frontend, and none of the code in `onUnload` gets called.

Sometimes it _does_ all happen as expected, though. I haven't been able to figure out why this happens.

dannycoates:
Using the [detach](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/page-mod#Cleaning_up_on_add-on_removal) event would be more succinct.

lmorchard:
> Using the detach event would be more succinct.

Seems like it's too late at that point. When I try it, I get this error: "Couldn't find the worker to receive this message. The script may not be initialized yet, or may already have been unloaded."

dannycoates:
> Seems like it's too late at that point.

Really? That's not what I get with just this:

``` js
// set-installed-flag.js
unsafeWindow.navigator.testpilotAddon = true;
unsafeWindow.navigator.testpilotAddonVersion = self.options.version;

self.port.on('detach', function () {
  delete unsafeWindow.navigator.testpilotAddon;
  delete unsafeWindow.navigator.testpilotAddonVersion;
});
```

lmorchard:
> Really? That's not what I get with just this

Oh! I was trying to call that in the add-on. Didn't realize the page mod itself got that message too.

lmorchard:
Some tweaks to the commit to make the set-installed-flag page mod clean up on detach. Also tweaked the uninstall process to ensure all the onUnload things happen

addon/data/set-installed-flag.js
```diff
@@ -10,3 +10,8 @@
 
 unsafeWindow.navigator.testpilotAddon = true;
 unsafeWindow.navigator.testpilotAddonVersion = self.options.version;
+
+self.port.on('detach', function() {
+  unsafeWindow.navigator.testpilotAddon = false;
+  unsafeWindow.navigator.testpilotAddonVersion = null;
```
dannycoates: can we delete these instead? It doesn't seem to matter in this case I suppose, except that by deleting them its a clearer signal (to me) that they aren't used by anything else anymore.

addon/index.js
```diff
@@ -186,6 +188,8 @@ function installExperiment(experiment) {
 }
 
 function uninstallSelf() {
+  // HACK: onUnload often doesn't seem to get called, here.
+  exports.onUnload('uninstall');
```
dannycoates: In what case is it not called? When I log the `onUnload` on line 360 I always see it twice, when I remove this line I see it once, albeit not until after this function has completed.
screen shot 2016-09-28 at 10 43 42 am

dannycoates (commented):
I can't reproduce the case that requires tracking `didUnload`. I've tried uninstalling from settings on the webpage and about:addons.

addon/index.js
lmorchard: I don't know what the cases are - I can't reproduce it reliably. About half the time, that code never runs for me on dev edition. That was what caused this issue originally for me, because the `addon-self:uninstalled` didn't always get sent and the `store` data didn't get deleted.

lmorchard: And [line 360](https://github.com/mozilla/testpilot/pull/1460/files#diff-7562eb8eba3c3d86a671045f8c70107eR360) should be preventing the rest of that unload function from running twice, if it already got called on uninstall. That was the only thing that seemed to ensure the onunload happened reliably at least once on my machine. If things are operating correctly, yes that onUnload should get called twice - but a good chunk of the time only the call in uninstallSelf() seemed to be doing it for me

addon/data/set-installed-flag.js
lmorchard: Hmm, weird: [I switched this to delete instead](https://github.com/mozilla/testpilot/pull/1460/commits/349b1b59887dfc2150c9d9f2976ab9e38c1f0489#diff-398f10aebbac83980f6fee6c04815d03R15), which works to change the install state. But, it looks like the key is still present on the `navigator` object. Or, at least `'testpilotAddon' in navigator` is `true` before & after add-on uninstall

addon/index.js
lmorchard: Buuuuut... if I just revert what I did to index.js, the page mod changes on detach seem to resolve this immediate issue fine. So, I removed that stuff to just address the retire page thing for this issue

lmorchard:
Okay, a few more tweaks: Removed all the changes to index.js - will investigate problems with add-on unload code separately. Could even just be an issue on my machine. Either way, the detach event in the set-installed-flag page mod seems to solve the retire page home link issue in #1429

dannycoates (approved):
Awesome!

PR #1466, opened 2016-09-28
Makes ExperimentCardList more idiomatic.

### LINKED ISSUE #1373: Restructure experiments store
The experiments bit of the store currently looks like this:

``` json
{
  "routing": "...",
  "experiments": {
    "experiments": [
      {"slug": "activity-stream"},
      {"slug": "no-more-404s"},
      {"slug": "tab-center"},
      {"slug": "universal-search"},
    ],
    "lastFetched":  1474222448673
  },
  "browser": "...",
  "addon": "..."
}
```

I'd suggest two restructures:
1. Removing the awkward `experiments.experiments` naming.
2. Store experiments in an object, with `slug:  {}`, to make references easy.

It might look like this:

``` json
{
  "routing": "...",
  "experiments": {
    "data": {
      "activity-stream": {},
      "no-more-404s": {},
      "tab-center": {},
      "universal-search": {}
    },
    "lastFetched":  1474222448673
  },
  "browser": "...",
  "addon": "..."
}
```

We lose any sorts, but that is controller logic, not data logic.

chuckharmston:
- Uses reselect for sorting and filtering.
- Restructures experiment store (closes #1373)
- Uses ES6 class structure for ExperimentCardList component.

### DIFF
--- a/frontend/src/app/components/ExperimentCardList.js
+++ b/frontend/src/app/components/ExperimentCardList.js
@@ -1,46 +1,44 @@
 import React from 'react';
-import moment from 'moment';
+
+import { isExperimentEnabled } from '../reducers/addon';
+
 import ExperimentRowCard from './ExperimentRowCard';
 import Loading from './Loading';
 
-export default function ExperimentCardList({
-  navigateTo, isExperimentEnabled, hasAddon, isDev, experiments, except, eventCategory
-}) {
-  const experimentsSorted = [].concat(experiments);
-
-  experimentsSorted.sort((a, b) => {
-    if (a.order > b.order) { return 1; }
-    if (a.order < b.order) { return -1; }
-    return 0;
-  });
-
-  const experimentsReduced = setCurrentExperiments(isDev, experimentsSorted);
-
-  return (
-    <div className="card-list experiments">
-      {(experiments.length === 0) ?
-        <Loading /> :
-        experimentsReduced.map((experiment, key) =>
-          (!except || except.slug !== experiment.slug) &&
-            <ExperimentRowCard key={key}
-                               navigateTo={navigateTo}
-                               experiment={experiment}
-                               enabled={isExperimentEnabled(experiment)}
-                               hasAddon={hasAddon}
-                               eventCategory={eventCategory} />) }
-    </div>
-  );
-}
-
-function setCurrentExperiments(isDev, allExperiments) {
-  const currentExperiments = [];
-  const utcNow = moment.utc();
-  if (isDev) return allExperiments;
-  allExperiments.forEach((experiment) => {
-    if (moment(utcNow).isAfter(experiment.launch_date) || typeof experiment.launch_date === 'undefined') {
-      currentExperiments.push(experiment);
+export default class ExperimentCardList extends React.Component {
+  getExperiments() {
+    if (!this.props.except) {
+      return this.props.experiments;
     }
-  });
-  return currentExperiments;
-}
+    return this.props.experiments.filter(experiment => (
+      experiment.slug !== this.props.except
+    ));
+  }
 
+  renderLoading() {
+    return (
+      <div className="card-list experiments">
+        <Loading />
+      </div>
+    );
+  }
+
+  renderExperiments() {
+    return (
+      <div className="card-list experiments">
+        {this.getExperiments().map((experiment, key) => (
+          <ExperimentRowCard experiment={experiment}
+                             enabled={isExperimentEnabled(experiment)}
+                             key={key} {...this.props} />
+        ))}
+      </div>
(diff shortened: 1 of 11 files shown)


chuckharmston:
I'll write tests for this, but I haven't used enzyme before, so I thought I'd get this out there while I poke around at that.

frontend/src/app/selectors/experiment.js
```diff
@@ -0,0 +1,30 @@
+import { createSelector } from 'reselect';
```
lmorchard, line 1: <strike>Why separate these from the reducer? Thought the [standard advice](https://twitter.com/dan_abramov/status/664581975764766721) was to export selectors from the reducer module.</strike> Oh, looks like the pattern for reselect, which I don't know anything about yet

chuckharmston: Also, Dan's advice you link is still good here: it's about decoupling from actions and reducers, not from keeping them in separate files.

lmorchard: Would be nice to add a little unit test for this new refactored module, since we have the beginnings of a working suite

chuckharmston: [Working on it!](https://github.com/mozilla/testpilot/pull/1466#issuecomment-250239140)

frontend/src/app/selectors/experiment.js
```diff
@@ -0,0 +1,30 @@
...
+import moment from 'moment';
+
+
+// Return all experiments from the store, sorted.
+export const allExperimentSelector = createSelector(
+  store => store.experiments.data,
+  experiments => (
+    experiments.sort((a, b) => {
```
lmorchard: FWIW, `[].sort()` [modifies the array in place](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) and returns the original array. That's why I did [the hack with `concat`](https://github.com/mozilla/testpilot/pull/1466/files#diff-4178a3feb17552468ad69a80a4b66b2cL9) in the old code to copy the array to `experimentsSorted` before sorting.

frontend/src/app/components/ExperimentCardList.js
```diff
@@ -1,46 +1,44 @@
...
-}) {
-  const experimentsSorted = [].concat(experiments);
-
-  experimentsSorted.sort((a, b) => {
-    if (a.order > b.order) { return 1; }
-    if (a.order < b.order) { return -1; }
-    return 0;
-  });
+export default class ExperimentCardList extends React.Component {
+  getExperiments() {
```
lmorchard, line 9: Since you're moving things into separate selectors, would this `except` filtering logic make sense to live there too? Or maybe no, because I think this card list is the only place we use it - so maybe just display logic?

frontend/src/app/selectors/experiment.js
lmorchard: Oh derp. I just jumped straight into the code and didn't see the comment.

lmorchard (changes requested):
Would like to see tests (in progress) and think the array sort thing should be addressed. Other than that, yay reselect!

frontend/src/app/selectors/experiment.js
chuckharmston: ☑️

chuckharmston:
Added tests for the selector in 83cb7bb (will rebase). Working on the new component now.

chuckharmston:
And the component has tests! Squashed and ready to merge.

frontend/test/app/components/ExperimentCardList.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
lmorchard: nit: Been using a `-test.js` suffix on these files so they have something to distinguish them in text editors besides the path that often gets hidden. Seems to be the thing lots of mocha & enzyme examples do

frontend/test/app/components/ExperimentCardList.js
```diff
@@ -0,0 +1,31 @@
...
+import ExperimentRowCard from '../../../src/app/components/ExperimentRowCard';
+import Loading from '../../../src/app/components/Loading';
+
+
+const _exp = [
+  { slug: 'foo' },
+  { slug: 'bat' }
+];
+
+describe('ExperimentCardList', () => {
```
lmorchard: nit: Also been using a pattern of `app/components/ExperimentCardList` in `describe()` to be more explicit about the test subject

lmorchard (approved):
Had a couple of nits about naming things, but don't feel strongly enough about it to request changes. Fix if you want or maybe self merge if no?

Issue #1476, opened 2016-09-28
Mobile experiment view is broken


jaredhirsch:
Single experiment view on mobile safari is looking pretty rough.

Here's the min vid page: 

![image](https://cloud.githubusercontent.com/assets/96396/18931005/33729de8-857f-11e6-8d30-f53d2ebdeb8c.png)

johngruen:
@6a68 oof, i'm not seeing that

what version of mobile safari are you on?

jaredhirsch:
@jgruen This screenshot is from a 6s plus, still on iOS 9. user agent string:

Mozilla/5.0 (iPhone; CPU iPhone OS 9_3_5 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13G36 Safari/601.1

jaredhirsch:
busted in portrait mode, too. I bet we never checked the mobile view with > 3 active experiments.

![image](https://cloud.githubusercontent.com/assets/96396/18931502/5db0ff58-8581-11e6-8f0a-50baef2610e2.png)

johngruen:
@6a68 weirdly not seeing this issue on a similar set up

jaredhirsch:
odd

jaredhirsch:
That's odd. Tried in a private window, just in case there was some weird caching issue, looks busted still:

![image](https://cloud.githubusercontent.com/assets/96396/18931779/74f9cd74-8582-11e6-8f86-c62c39f22045.png)

pdehaan:
Here's what it looks like on my obsolete iPhone 5s (apologies for the watermark, I'll buy Reflector 2 to avoid such embarrassing and unsightly watermarks in the future):

![testpilot iphone5s](https://cloud.githubusercontent.com/assets/557895/18931979/5c762e54-8583-11e6-8a5a-0aa5d985a2d4.gif)
**Figure 1:** 4.8MB GIF: https://cloud.githubusercontent.com/assets/557895/18931979/5c762e54-8583-11e6-8a5a-0aa5d985a2d4.gif

johngruen:
@6a68 seeing it as well on the experiment page on a simulator

jaredhirsch:
:beers:

ckprice:
Fixed in #1478.

PR #1478, opened 2016-09-28
fix mobile details break

johngruen:
@6a68 give 'er a look


johngruen:
Fixes #1476

jaredhirsch:
looking now

jaredhirsch (approved):
:+1: looks great! Before and after on OSX Safari:

PR #1481, opened 2016-09-29
Add S3 deploy script with custom headers

clouserw:
This script syncs our flat file website up to S3.  A couple of open questions:

1) Do we still need to allow framing on sameorigin or can we deny completely?

2) It looks like all the static files have hashes on them (save favicon.ico, but meh), so I'm letting them have default (read: long) expires times, and timing out html and json every 10 minutes.  Concerns or things I missed?

3) I'm under the impression we only need CSP headers on the HTML files.  Can someone verify that?

4) Do we need any special headers on the API endpoints?  (.json files)

5) Passing in these headers via command line is a bit ridiculous.  The CSP in easy to read format is:

```
default-src 'self'; 
connect-src 'self' analysis-output.telemetry.mozilla.org; 
font-src 'self' code.cdn.mozilla.net; 
script-src 'self' www.google-analytics.com; 
style-src 'self' code.cdn.mozilla.net; 
report-uri /__cspreport__;
```

Am I missing anything there?  Do we need pontoon anymore?

6) Any other feedback or headers I'm missing?

/cc for review/feedback @lmorchard @jvehent @muffinresearch @relud

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -1,0 +1,84 @@
+#!/bin/bash
+
+# This file is used to deploy Test Pilot to an S3 bucket.  It expects to be run
+# from the root of the Test Pilot directory and you'll need your S3 bucket name
+# in an environment variable $TESTPILOT_BUCKET
+
+# Questions?  Hit up #testpilot on IRC
+
+if [ ! -d "dist" ]; then
+    echo "Can't find /dist/ directory.  Are you running from the Test Pilot root?"
+    exit 1
+fi
+
+if [ -z "$TESTPILOT_BUCKET" ]; then
+    echo "The S3 bucket is not set. Failing."
+    exit 1
+fi
+
+
+# The basic strategy is to sync all the files that need special attention
+# first, and then sync everything else which will get defaults
+
+
+# For short-lived assets; in seconds
+MAX_AGE="600"
+
+HPKP="\"x-amz-meta-public-key-pins\": \"max-age=300;pin-sha256='WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18=';pin-sha256='r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E='\""
+CSP="\"x-amz-meta-content-security-policy\": \"default-src 'self'; connect-src 'self' analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; object-src 'none'; script-src 'self' www.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
+HSTS="\"x-amz-meta-strict-transport-security\": \"max-age=31536000; includeSubDomains; preload\""
+TYPE="\"x-amz-meta-x-content-type-options\": \"nosniff\""
+FRAME="\"x-amz-meta-x-frame-options\": \"DENY\""
+XSS="\"x-amz-meta-x-xss-protection\": \"1; mode=block\""
+
+# HTML; 10 minute cache
+aws s3 sync \
+  --cache-control "max-age=${MAX_AGE}" \
+  --content-type "text/html" \
+  --exclude "*" \
+  --include "*.html" \
+  --metadata "{${HPKP}, ${CSP}, ${HSTS}, ${TYPE}, ${FRAME}, ${XSS}}" \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  dist/ s3://${TESTPILOT_BUCKET}/
+
+# JSON; 10 minute cache
+aws s3 sync \
+  --cache-control "max-age=${MAX_AGE}" \
+  --content-type "application/json" \
+  --exclude "*" \
+  --include "*.json" \
+  --metadata "{${HPKP}, ${HSTS}}" \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  dist/ s3://${TESTPILOT_BUCKET}/
+
+# XPI; 10 minute cache; amazon won't detect the content-type correctly
+aws s3 sync \
+  --cache-control "max-age=${MAX_AGE}" \
+  --content-type "application/x-xpinstall" \
+  --exclude "*" \
+  --include "*.xpi" \
+  --metadata "{${HPKP}, ${HSTS}}" \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  dist/ s3://${TESTPILOT_BUCKET}/
+
+# RDF; 10 minute cache; amazon won't detect the content-type correctly
+aws s3 sync \
+  --cache-control "max-age=${MAX_AGE}" \
+  --content-type "text/rdf" \
+  --exclude "*" \
+  --include "*.rdf" \
+  --metadata "{${HPKP}, ${HSTS}}" \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  dist/ s3://${TESTPILOT_BUCKET}/
+
+# Everything else; cache forever, because it has hashes in the filenames
+aws s3 sync \
+  --delete \
+  --metadata "{${HPKP}, ${HSTS}}" \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  dist/ s3://${TESTPILOT_BUCKET}/


jvehent:
> 3) I'm under the impression we only need CSP headers on the HTML files. Can someone verify that?
> 
> 4) Do we need any special headers on the API endpoints? (.json files)

We need CSP on non-HTML files if there's a chance of returning HTML, for example on a 404 or 500. If we can enforce that non-HTML endpoints never return `Content-Type: text/html`, we should be fine without CSP.

> 5) Passing in these headers via command line is a bit ridiculous. The CSP in easy to read format is:

I wonder if, rather than using a proxy to serve the CSP header, you could just set it inside HTML files using a meta tag like this one:

``` html
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src 'self' analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; script-src 'self' www.google-analytics.com; style-src 'self' code.cdn.mozilla.net; ">
```

Main downside is you can't use `report-uri` and `frame-ancestors` when the policy is set inside a document. 

> 6) Any other feedback or headers I'm missing?

You might want to add `object-src 'none';` to the policy. I don't think we need plugins here.

bin/deploy.sh
```diff
@@ -0,0 +1,57 @@
...
+  --exclude "*" \
+  --include "*.json" \
+  --metadata '{"x-amz-meta-strict-transport-security": "max-age=31536000; includeSubDomains; preload"}' \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  dist/ s3://${BUCKET}/
+
+# Everything else; cache forever, because it has hashes in the filenames
+aws s3 sync \
+  --delete \
```
lmorchard, line 80: Does this need `--exclude "*.json" --exclude "*.html"` to avoid overwriting what the previous two commands did? Or will `s3 sync` skip the just-updated files?

lmorchard:
Probably should also add Content-Type headers for these extensions / paths:
- `static/addon/*.xpi` = `application/x-xpinstall`
- `static/addon/*.rdf` = `text/rdf`

Also, we should be able to drop this script into [the CircleCI deployment to static dev](https://github.com/mozilla/testpilot/blob/master/circle.yml#L71) to replace the single simple `aws sync` used now.

Might also need to collect the output of these sync commands to build up a file list for a more granular CloudFront invalidation (i.e. rather than `/*`) - but that can be a future improvement, since we're not doing it now.

lmorchard:
More feedback just to cover off on things:

> 1) Do we still need to allow framing on sameorigin or can we deny completely?

No reason for framing that I can think of. Might be cool if someone wanted to do a data mashup with our `experiments.json` but that's so 2006

> 2) It looks like all the static files have hashes on them (save favicon.ico, but meh), so I'm letting them have default (read: long) expires times, and timing out html and json every 10 minutes. Concerns or things I missed?

Yeah, we [have a task that adds hashes to filenames in gulp](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/dist.js#L8), so that should be copacetic.

> 4) Do we need any special headers on the API endpoints? (.json files)

I don't think so - just `application/json` and a low expires time because those URLs won't change or be hashed. 10 minutes seems fine.

> Am I missing anything there? Do we need pontoon anymore?

We'll need pontoon on dev. But, [not on stage or prod](https://github.com/mozilla/testpilot/issues/1356).

bin/deploy.sh
clouserw: nope, they are already up to date, so sync skips them.

ghost:
Thanks for the comments.  New patch:
- Added public-key-pins header
- Added form-action:none to CSP
- Added object-src:none to CSP
- Added content-types to XPI and RDF

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+
+
+# The basic strategy is to sync all the files that need special attention
+# first, and then sync everything else which will get defaults
+
+# This is not pretty, but until S3 supports arbitrary headers, here we are...
+
+# HTML; 10 minute cache
+aws s3 sync \
+  --cache-control "max-age=600" \
```
chuckharmston: This number seems a prime candidate for a variable.

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
+# This file is used to deploy Test Pilot to an S3 bucket.  It expects to be run
+# from the root of the Test Pilot directory and you'll need your S3 bucket name
+# in an environment variable $BUCKET
```
chuckharmston: Can we namespace the environment variables? `$TESTPILOT_BUCKET` should work

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+# first, and then sync everything else which will get defaults
+
+# This is not pretty, but until S3 supports arbitrary headers, here we are...
+
+# HTML; 10 minute cache
+aws s3 sync \
+  --cache-control "max-age=600" \
+  --content-type "text/html" \
+  --exclude "*" \
+  --include "*.html" \
```
chuckharmston, line 39: This is weird. When `--include` is passed, does it not imply `--exclude "*"`?

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+  --metadata '{"x-amz-meta-public-key-pins": "max-age=300;pin-sha256='\''WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18='\'';pin-sha256='\''r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E='\''",
+               "x-amz-meta-content-security-policy": "default-src '\''self'\''; connect-src '\''self'\'' analysis-output.telemetry.mozilla.org; font-src '\''self'\'' code.cdn.mozilla.net; form-action '\''none'\''; object-src '\''none'\''; script-src '\''self'\'' www.google-analytics.com; style-src '\''self'\'' code.cdn.mozilla.net; report-uri /__cspreport__;",
+               "x-amz-meta-strict-transport-security": "max-age=31536000; includeSubDomains; preload",
+               "x-amz-meta-x-content-type-options": "nosniff",
+               "x-amz-meta-x-frame-options": "DENY",
+               "x-amz-meta-x-xss-protection": "1; mode=block"}' \
```
chuckharmston: I don't think that any of these headers change between runs of `aws s3 sync`. Might make sense to also stick these into variables?

``` bash
--metadata "{$HSTS,  $HPKP, $CSP, $NOSNIFF, $FRAME_OPTIONS, $XSS}"
```

At worst, it reads much better.

chuckharmston (changes requested):
Could we also write docs for the deploy process and architecture?

chuckharmston:
Not for this PR, but it strikes me that some sort of declarative configuration format bundled with an nginx docker image would be really useful. We pretty much can't deploy from an S3 bucket without something approximating this functionality.

bin/deploy.sh
relud: no, it does not

relud:
the plan is to have an nginx proxy stripping the x-amz-meta- prefix from headers. the nginx proxy should be as static as possible though, so that normal deployments only require pushing to an s3 bucket.

relud:
`docs for the deploy process and architecture` I will write some on mana when I get a chance, since deploy process and architecture usually fall under ops.

chuckharmston:
> the plan is to have an nginx proxy stripping the x-amz-meta- prefix from headers. the nginx proxy should be as static as possible though, so that normal deployments only require pushing to an s3 bucket.

Right; my suggestion was for some sort of reusable nginx image designed to do just this, plus a companion script that sets the headers similarly to here, only with a declarative configuration format, e.g.

``` json
{
    "rulesets": {
        "hpkp": {
            "public-key-pins": "max-age=300;pin-sha256='WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18=';pin-sha256='r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E='"
        },
        "hsts": {
            "strict-transport-security": "max-age=31536000; includeSubDomains; preload"
        }
    },
    "applications": [
        {
            "exclude": "*",
            "include": "*.json",
            "content-type": "application/json",
            "rulesets": [
                "hsts",
                "hpkp"
            ],
            "rules": {
                "Something": "Custom"
            }
        }
    ]
}
```

relud:
i think coming up with our own declarative configuration format is a bad idea. but i think it would be good to package the static site in a docker container with an nginx config that serves the same headers we are setting in the deploy script. good for testing, and it would make building easier, because i would just copy the files out of the official container, instead of building myself.

ghost:
new patch should address the notes

chuckharmston (approved):

Issue #1364, opened 2016-09-16
Consider separating containers from components


chuckharmston:
It's a common practice for React apps to separate components into two directories: `containers` for anything handed off to a `Route`, and `components` for everything else. We should consider doing this.

lmorchard:
We have containers for a few components in the router - do you mean that _all_ the "pages" should be container components? I skipped a bunch, because they're basically just stateless content. But that is admittedly inconsistent.

I was also trying to figure out if maybe we could do it in the _other_ direction - that is, one App container wrapping the router that passes all the store selector & dispatch props down to page components. Haven't been able to figure that one out though, maybe it's not a good practice?

PR #1487, opened 2016-09-29
Issue #1364: Move redux & browser API calls from components into single App container

lmorchard:
This might be a throwaway PR - more of an architectural proposal / question in code form. @chuckharmston probably has opinions on this, [as a follow up to earlier comments](https://github.com/mozilla/testpilot/pull/1369/files#r79307862) :)

Basically, I'm working through #1399 to backfill tests and ran aground on [the bit in `DiscussDialog` that sets `window.location`](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/DiscussDialog.js#L36) 

[I can mock that out in a test with a global variable.](https://github.com/lmorchard/testpilot/commit/22148d2abb52a1068cabe7845386fc7bef305842#diff-082f19303ba35b040a993008b93a6590R9) But, we have a lot of spots like that. 

It seems interesting to yank all of the Redux & browser knowledge out of the components and leave them as display-only. That could make static page rendering (#1306) easier along with testing (#1399) by way of dependency injection. Then, we can move all that Redux & browser context to the central `App` container for use in the frontend app.

So, here's what I did to thread those changes just from App to Header to DiscussDialog:
- App component tweaks
  - Supplies its own openWindow & sendToGA helpers in props on connect()
  - Inject own props into router-supplied children to pass along helpers
  - Moved from components to containers
- Header component now expects openWindow & sendToGA props
  - no longer loads sendToGA from utils module directly
- DiscussDialog component expects now openWindow & onDismiss props
- Tweaks to all pages using Header to pass openWindow & sendToGA props

Next steps would be to consolidate the rest of what's done in other containers into the central App container and rework the rest of the components to be display-only. 

Anything that touches the Redux store or browser APIs would go through function props injected via App. No component would call or import anything that calls Redux or browser APIs.

Too weird / gratuitous a change?


frontend/src/app/components/ErrorPage.js
```diff
@@ -7,7 +7,7 @@ export default class ErrorPage extends React.Component {
   render() {
     return (
       <div className="full-page-wrapper space-between">
-        <Header />
+        <Header openWindow={this.props.openWindow} sendToGA={this.props.sendToGA} />
```
chuckharmston: You can pass on all props to a child using `{...this.props}`, e.g.

``` jsx
<Header something="nothing" {...this.props} />
```

chuckharmston (approved):
💯 

This is exactly how we should be doing this.

frontend/src/app/components/ErrorPage.js
lmorchard: Hmm, I was a little reticent to do that, since it only needs these 2 props. But, that could be easier as boilerplate

lmorchard: Updated commit to make this change - kind of obscures the particular things passed in, but it's nice consistent boilerplate

lmorchard:
@chuckharmston - Okay, so I just went crazy and applied this process to the rest of the components. I think I managed to pull out and squash nearly everything browser- and Redux-related out of the components, with an exception or two. Can you take another look? I don't want to just merge this stuff on the back of your previous :+1: of my smaller set of changes.

Kind of hate to do all these changes without tests, but all these changes will make things cleaner to test.

lmorchard:
Hmm, this will lead to a ginormous PR - but what if I combine this with a revised #1399 and backfill with tests for all this refactoring?

lmorchard:
Closing in favor of combining with #1488

PR #1488, opened 2016-09-29
[work in progress] Backfill component tests & refactor components for central App container

lmorchard:
This is now a combined PR for Issue #1399 & #1364:
1. refactoring components to work with a central App container all Redux & browser API know-how
2. unit tests to assert that all this changed stuff still works

Tests left to backfill: 
- components
  - <strike>ExperimentRowCard</strike> #1541
  - <strike>ExperimentTourDialog</strike> #1542 
  - [x] LandingPage
  - <strike>MainInstallButton</strike> #1543 
  - [x] Restart
  - [x] RetireConfirmationDialog
  - [x] RetirePage
  - [x] SharePage
- reducers - all, may skip
- actions - all, may skip


frontend/test/app/components/NotFoundPage-test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import NotFoundPage from '../../../src/app/components/NotFoundPage';
+
+describe('app/components/NotFoundPage', () => {
+  it('should render notFoundHeader string', () => {
+    expect(shallow(<NotFoundPage />)
+      // HACK: .find('[data-l10n-id="errorMessage"]') seems not to work
```
lmorchard, line 10: Seems like attribute names with dashes make enzyme unhappy. Maybe better to actually add some descriptive classes to these elements as test handles?

frontend/test/app/components/EmailDialog-test.js
```diff
@@ -0,0 +1,88 @@
...
+
+describe('app/components/EmailDialog', () => {
+
+  const mockClickEvent = {
+    preventDefault() {},
+    stopPropagation() {}
+  };
+
+  beforeEach(function() {
+    global.window = {
```
lmorchard: Hmm this window / ga stuff is coming up a lot, I should refactor it into a shared util if I can

lmorchard:
Looks like #1487 is :+1: so it probably makes sense to merge that and rework some things here to account for that cleaner way of handling things - rather than support all the hacks here for things like GA, cookies, and window location

lmorchard:
This PR is officially crazy, now. But, it amends many systemic downsides from my initial reactpocalypse. Still working through adding tests, but it could use some eyes already.

lmorchard:
Adding in-progress work on tests for ExperimentPage. That one's a biggie.

dannycoates:
I'm following along... and looking good so far. One nit is from proptype warnings when running the tests, for example:

```
  app/components/ExperimentsListPage
    ✓ should render expected content
    ✓ should render a Loading component if no experiments available
Warning: Failed prop type: Required prop `sendToGA` was not specified in `EmailDialog`.
    in EmailDialog
Warning: Failed prop type: Required prop `subscribeToBasket` was not specified in `EmailDialog`.
    in EmailDialog
    ✓ should show an email dialog if the URL contains utm_campaign=restart-required
    ✓ should show an email dialog if the first-run cookie is set
```

dannycoates:
Part of my difficulty in reviewing this is that I'm still trying to grok the react rewrite in general. Everything here looks logical on its own, but I'm still working out the gestalt of the site, so its hard to have solid opinions on changes.

I can easily say that the tests are a wonderful addition though 👍

lmorchard:
Not sure if it helps, but I just rebased / edited the commits in this PR. Main point of interest is I tried [to summarize all the major changes in commit comments](https://github.com/mozilla/testpilot/pull/1488/commits/e94db6b230f985ecf518a951c854b0f7660c220c)

lmorchard:
Also FWIW: Still have a bunch of tests to backfill, but I _think_ the [stuff I've been doing to refactor things](https://github.com/mozilla/testpilot/pull/1488/commits/e94db6b230f985ecf518a951c854b0f7660c220c) is stable now.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -48,14 +48,15 @@ export default class ExperimentPage extends React.Component {
...
     const { shouldShowTourDialog, enabled: prevEnabled } = this.state;
-    const { isExperimentEnabled, getExperimentBySlug } = nextProps;
 
-    const prevExperiment = getExperimentBySlug(this.props.params.slug);
+    const prevExperiment = this.props.getExperimentBySlug(this.props.params.slug);
     const prevInProgress = prevExperiment && prevExperiment.inProgress;
 
-    const nextExperiment = getExperimentBySlug(nextProps.params.slug);
+    const { isExperimentEnabled: nextIsExperimentEnabled,
+            getExperimentBySlug: nextGetExperimentBySlug } = nextProps;
```
dannycoates: 👀 I did a double take when I saw this. To me `nextProps.getExperimentBySlug(...)` is more readable.

dannycoates: actually, there's something unsettling to me about having these functions as props at all. I haven't dug into it but it _feels_ like this component should have an `experiment` prop instead. anyway... maybe a change for another day

dannycoates:
On more of a meta level I wonder what it'd look like if all the `sendToGA` calls were centralized and triggered as side effects to actions instead of passing the function out to the edges. Not for this PR, and not that this here isn't already forward progress, just thinking out loud 😊

frontend/src/app/components/ExperimentPage.js
```diff
@@ -141,13 +149,13 @@ export default class ExperimentPage extends React.Component {
             introduction, measurements, privacy_notice_url, changelog_url,
             thumbnail, survey_url, contributors, details, min_release } = experiment;
 
+    // TODO: #1138 - add optional subtitles the right way
```
dannycoates, line 150: this should be easy now, right? can we Just Do It®?

dannycoates:
> Not sure if it helps, but I just rebased / edited the commits in this PR

It helped 👍

Considering each bullet independently is much easier.

dannycoates:
I'm 👍 on this but I'll let @chuckharmston make it green 😉

frontend/src/app/components/ExperimentPage.js
lmorchard: Yeah, the only reason it doesn't have an `experiment` prop is because it only gets a `params` prop from the router and has to look it up from the list of available experiments. There could be a better way to do that.

This whole function is also kind of a big hack in that it's tracking experiment enable state changes so we can schedule post-install things like showing a tour dialog and ensuring the button states are turned off. Would be great to rework all of this so it's stateless yet just does the right thing.

But, having these functions as props is what I read as the suggested way of injecting external dependencies into components in the React + Redux pattern. The complication I realized while writing tests is that [these functions are closures around a particular Redux state instance in the `connect()` call in `App`](https://github.com/lmorchard/testpilot/blob/e94db6b230f985ecf518a951c854b0f7660c220c/frontend/src/app/containers/App.js#L62). So, when the Redux state updates, `connect()` updates the props with fresh closures for these helper functions. Using the old functions would mean operating on the previous state.

lmorchard: Oh, but I think I will back out this destructuring magic to be more verbose with the `nextProps` reference. Probably trying to be too clever, here

lmorchard:
Going to try wrapping up the backfill of the last component tests there, and see if I can reorg this PR into more commits. And from there I might be able to split it into separate PRs.

Some of the changes mashed together are more related to problems I found while reviewing & writing tests, so maybe those should be untangled and grouped with the backfilled tests

frontend/src/app/components/Header.js
```diff
@@ -152,6 +153,9 @@ export default class Header extends React.Component {
 }
 
 Header.propTypes = {
+  uninstallAddon: React.PropTypes.func.isRequired,
+  sendToGA: React.PropTypes.func.isRequired,
+  openWindow: React.PropTypes.func.isRequired,
   hasAddon: React.PropTypes.bool,
   forceHideSettings: React.PropTypes.bool
```
dannycoates, line 160: `forceHideSettings` looks orphaned now; nothing sets it. I'll add a todo to delete it.

lmorchard: Oh, actually that should be used in the Restart component!

lmorchard: [Added a commit to fix that](https://github.com/mozilla/testpilot/pull/1488/files#diff-014a13e8ad141df734228aa1890d34e8L38).

Missed writing a test for that, because I'm kind of confused [what this bit does](https://github.com/lmorchard/testpilot/blob/4f5bfc1f9da345359b4945d7a656b23df10b7bd9/frontend/src/app/components/Header.js#L22) and then forgot about it.

dannycoates (approved):
👏

PR #1505, opened 2016-10-02
ENV FOR DOCKER-COMPOSE: Use .env file for the keys required in docker…

swami701:
Remove the hard coded keys in docker-compose-base.yml and used .env file for the keys in docker-compose-base.yml.

Please add these values in .env file.

`FXA_CLIENT_ID=<FXA Client ID>`
`FXA_SECRET_KEY=<FXA SECRET KEY>`
`DATADOG_API_KEY=<Datadog Api Key>`
`DATADOG_APP_KEY=<Datadog App Key>`
`INITIAL_ADMIN_USERNAME=<Intial Admin User name>`
`INITIAL_ADMIN_PASSWORD=<Admin Password>`
`INITIAL_ADMIN_EMAIL=<Admin Email>`
`SECRET_KEY=<Your secret key>`
`DATABASE_URL=<Postgres Database Url>`


pdehaan:
@swami701 it may be worth putting all those env vars above in an .env.sample file or something so people can just quickly rename that file to .env (which is already ignored by .gitignore, I think).

chuckharmston:
As part of [this milestone](https://github.com/mozilla/testpilot/milestone/20), we will actually be ridding ourselves of Docker entirely, so I'm going to close this. But we do have a handful of other [good first bugs](https://github.com/mozilla/testpilot/issues?q=is%3Aopen+is%3Aissue+label%3Agood-first-bug), and we would welcome your contributions.

Thanks for the contribution!

PR #1511, opened 2016-10-03
update for new version of tab center

clouserw:


content-src/experiments/tab-center.yaml
```diff
@@ -7,7 +7,7 @@ description: 'What would it be like to move tabs from the top of the browser to
 introduction: '<p>Take your tabs from in the way to tucked away. Tab Center moves your tabs to the side of your browser window so they''re out of sight when you don''t need them, and easy to expand and grab when you do.</p>'
 image_twitter: null
 image_facebook: null
-version: '1.24'
+version: '1.25.0'
```
bwinton: (Maybe just make it `1.25`, for nicer readability?)

clouserw: I was just copying what was in the xpi file.  I think they should match?

bwinton: They haven't so far, and it hasn't been a problem…
(It's 1.25.0 in the package.json because npm complains that it's not semver otherwise.  For user-facing stuff, dropping the .0 is more friendly.)

ghost: a'ight

bwinton (approved):
LGTM.  :shipit:

ghost (approved):
got r+ from bwinton

Issue #1486, opened 2016-09-29
Combine frontend & addon build processes?


lmorchard:
[Building the frontend first requires a separate add-on build run](https://github.com/mozilla/testpilot/blob/master/docs/development/quickstart.md#for-linux--os-x-hosts). That can be (and has been) confusing. Maybe we should find a way to combine all of this into one single set of `npm install && npm {build,test,start}` commands from the project root?

lmorchard:
Alternatively, we could also have the `/static/addon` path in the dev server proxy to the production server. But, I don't think the add-on build is all that onerous if you've already got all the prereqs to build the web frontend

johngruen:
Wontfix: separating concerns seems easier

PR #1516, opened 2016-10-03
Replace eslint-config-airbnb with eslint-config-airbnb-base

### LINKED ISSUE #1514: Replace eslint-config-airbnb with eslint-config-airbnb-base
Currently it looks like we're [extending](https://github.com/mozilla/testpilot/blob/123fa362f66bd92c9db284b7348f08cf8f79cdf8/.eslintrc#L4) the deprecated [**eslint-config-airbnb/base**](https://www.npmjs.com/package/eslint-config-airbnb#eslint-config-airbnbbase) config in our [package.json](https://github.com/mozilla/testpilot/blob/123fa362f66bd92c9db284b7348f08cf8f79cdf8/package.json#L52) file.

We should probably switch to **eslint-config-airbnb-base**:

> #### eslint-config-airbnb/base
> 
> This entry point is deprecated. See [**eslint-config-airbnb-base**](https://npmjs.com/eslint-config-airbnb-base).

&mdash; via https://www.npmjs.com/package/eslint-config-airbnb#eslint-config-airbnbbase

pdehaan:
Fixes #1514


addon/.eslintrc.yml
```diff
@@ -0,0 +1,35 @@
...
+extends:
+  - eslint:recommended
+  - airbnb-base
+
+root: true
+
+rules:
+  import/newline-after-import: off
+  import/no-extraneous-dependencies: off
+  import/no-unresolved: off
```
pdehaan, line 13: ~~These feel a bit hacky since we're not explicitly using this plugin here, it seems to be leaking from the parent .eslintrc.yml file. Feels like an ESLint bug, but I'll have to investigate and boil it down to a smaller test case and file upstream, so we'll just ignore these wrongly inherited rules for right meow.~~

Actually, I think this may just be coming from new rules in airbnb-base config. 🙄

pdehaan:
I have no idea why this is failing. It seems like my deleting of the top level package.json **eslint** dependency shouldn't have broken this. So unless it's something caused by my bumping the addon/ package.json dependency versions, or cached dependencies, or something more sinister/obvious.

Or, we delete this garbage fire PR and just wait for #1486 and focus on combining the linting steps for the site and the add-on so they use a shared config (with some rules overridden in /addon), versus duplicating oodles of dependencies and rules and config file tricks.

lmorchard:
Let's defer this one for now

PR #1531, opened 2016-10-05
[fixes #839] Set GA page explicitly, and update event labels *needs review*

### LINKED ISSUE #839: GA refinements
Met with @garethcull and team today, and we have a couple of requests.
1. Send an identifier on the `Install add-on` page that the user has successfully signed into FxA.
2. Add the experiment title to the Enable/Disable/Feedback events on the experiment details page.

Updating these to (@garethcull ok?)

```
eventCategory: 'ExperimentDetailsPage Interactions',
eventAction: this.model.title,
eventLabel: {'Give Feedback','Enable Experiment','Disable Experiment'}
```

ckprice:
- explicitly set the `page` param in `pageview` calls ([ref](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#tracking_virtual_pageviews)).
- per recommendation from Peter, modify the action/labels for a few event calls

Fixes #839.

### DIFF
--- a/docs/metrics/ga.md
+++ b/docs/metrics/ga.md
@@ -59,9 +59,10 @@ Here are the current events on the website as of this writing
 | Click on experiment from landing page                    | ExperimentsPage Interactions       | Open details page        | `{experiment title}`          |            
 | Click on Install from experiment details page            | ExperimentDetailsPage Interactions | button click             | Install the Add-on            |
 | Click on the "Try out these experiments as well" section | ExperimentsDetailPage Interactions | Open details page        | try out `{experiment title}`  |            
-| Click Enable Experiment                                  | ExperimentDetailsPage Interactions | button click             | Enable Experiment             |            
-| Click Disable Experiment                                 | ExperimentDetailsPage Interactions | button click             | Disable Experiment            |            
-| Click Give Feedback for experiment                       | ExperimentDetailsPage Interactions | button click             | Give Feedback                 |            
+| Click Enable Experiment                                  | ExperimentDetailsPage Interactions | Enable Experiment        | `{experiment title}`          |            
+| Click Disable Experiment                                 | ExperimentDetailsPage Interactions | Disable Experiment       | `{experiment title}`          |            
+| Click Give Feedback for experiment                       | ExperimentDetailsPage Interactions | Give Feedback            | `{experiment title}`          |
+| Click upgrade notice                                     | ExperimentDetailsPage Interactions | Upgrade Notice           | `{experiment title}`          |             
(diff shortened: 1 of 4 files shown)


chuckharmston (changes requested):
This feels like the sort of thing that requires a migration plan. We're basically giving up on all `button click` data before the deploy date. Are we okay doing that?

Also: [better commit message](http://chris.beams.io/posts/git-commit/), please. It's good for someone to at least be able to gain a cursory understanding of what happened by reading the first 50 chars.

ckprice:
> Are we okay doing that?

Yeah, this was suggested by the team responsible for the reporting (Gareth/Peter). Commit message and title updated.

chuckharmston:
Would it make sense to run both in parallel for a while, so we don't just start over on data?

ckprice:
> Would it make sense to run both in parallel for a while, so we don't just start over on data?

The way the events are set up is pretty straight forward, I don't know how much we'd benefit from sending extra pings for a whole sprint. But I've invited that team to comment in this PR.

ckprice:
Talked to Gareth and he said we shouldn't need the two events in parallel with respect to reporting. I also have a checkin scheduled with him tomorrow to verify stage (assuming this merges today).

chuckharmston (approved):

chuckharmston:
Not an r+, but I'm going off for the week and am removing myself as blocker.

ckprice:
Calling out that this **needs review before merge** even though it's green. Ref: https://github.com/mozilla/testpilot/pull/1531#issuecomment-252085391.

Rebased with updated tests (cool stuff).

More context here http://logs.glob.uno/?c=testpilot#c26783.

/edit aside from whether or not we should send another event, it could get pretty ugly to try to send consecutive async GA pings.

dannycoates:
code LGTM 👍

peterGerman:
@ckprice, I'm not seeing updated event values appearing for enable experiments in the development environment. 

Can you ping me when they are updated so that I can make changes to our reporting?

ckprice:
One thing we have in GA is a view for dev, so if we're able to merge this, I can regroup with @peterGerman (who is on the team responsible for running the reports/creating the dashboards) the same day and he can help verify that things look okay before we move to stage (or change things if they have to be changed). Thoughts @chuckharmston?

peterGerman:
@ckprice lets get a placeholder in the calendar for when you think we'll be
meeting about this.

meandavejustice:
👍

chuckharmston:
As I mentioned in IRC, feel free to ship this if you're comfortable losing the data continuity. It will hurt our ability to detect regressions, but that's not a primary concern of our implementation of GA, so feel free to ignore it. Code looks good otherwise.

PR #1532, opened 2016-10-05
added a 'rated' ping

dannycoates:
Looking at the way data is sent to and presented in telemetry and GA using these values for the `object` and `eventName` parameters seems like a good first step to get this stat flowing. Encoding the rating value in the name isn't the most satisfying method from a "purity" standpoint, but it's less disruptive than making deeper changes and will still be easy to work with on the other end.


lmorchard (approved):
Looks good to me, though the Telemetry collection & analysis part of this is still super black-boxy to me

Issue #1469, opened 2016-09-28
Should we require users to be on the latest version of the add-on?


chuckharmston:
We could detect the installed version of the add-on, compare to the latest version, and prompt the user to update.

pdehaan:
Don't the add-ons all auto-update, unless the user specifically turns that off in about:addons?

chuckharmston:
Correct. This would be to handle that latter case.

ghost:
This sounds good.  I don't think we need any prompting to the user - just detect an out of date version and update ourselves.

johngruen:
@clouserw @fzzzy let's discuss this in our monday planning meeting WRT migration to webextensions

johngruen:
WONTFIX

PR #1539, opened 2016-10-06
added 'active' addon ids to 'sync-installed-result'

### LINKED ISSUE #1403: Only send installed add-ons to website once.
#1402 introduces a second call of `AddonManager.getInstalledAddons`, sending the full, unfiltered list of installed add-ons to the website, rather than just a list of the installed experiments.

We should only send one list of installed add-ons, segmented:

``` json
{
  "experiments": [],
  "other": []
}
```

dannycoates:
fixes #1403 

If this gets deployed in this sprint we won't have to worry about addon <-> webapp version compatibility. Old addon -> new webapp: `active` will be `undefined` and yield `[]` in the reducer. New addon -> old webapp: `active` will be ignored.

### DIFF
--- a/frontend/src/app/lib/addon.js
+++ b/frontend/src/app/lib/addon.js
@@ -128,9 +128,7 @@ function messageReceived(store, evt) {
     case 'sync-installed-result':
       store.dispatch(addonActions.setClientUuid(data.clientUUID));
       store.dispatch(addonActions.setInstalled(data.installed));
-      break;
-    case 'sync-installed-addons':
-      store.dispatch(addonActions.setInstalledAddons(data.installedAddons));
+      store.dispatch(addonActions.setInstalledAddons(data.active));
       break;
     case 'addon-self:uninstalled':
       store.dispatch(addonActions.setHasAddon(false));
--- a/addon/index.js
+++ b/addon/index.js
@@ -102,16 +102,7 @@ function changeApp(env) {
     .on('uninstall-self', uninstallSelf)
     .on('install-experiment', installExperiment)
     .on('uninstall-experiment', uninstallExperiment)
-    .on('sync-installed', () => {
-      app.send(
-        'sync-installed-result',
-        {
-          clientUUID: store.clientUUID,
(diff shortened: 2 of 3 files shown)


frontend/src/app/lib/addon.js
```diff
@@ -128,9 +128,7 @@ function messageReceived(store, evt) {
     case 'sync-installed-result':
       store.dispatch(addonActions.setClientUuid(data.clientUUID));
       store.dispatch(addonActions.setInstalled(data.installed));
-      break;
-    case 'sync-installed-addons':
-      store.dispatch(addonActions.setInstalledAddons(data.installedAddons));
+      store.dispatch(addonActions.setInstalledAddons(data.active));
```
lmorchard, line 131: Things get tricky here. We probably need to keep _both_ these messages - or at least fail gracefully - for at least the next add-on upgrade cycle or two. That is, until/unless we require users to have the latest add-on version (#1469). Otherwise, we'll have 24 hours or so of the old add-on version hitting the site and not sending `active` in this message until all clients have done their updates.

lmorchard: But of course I missed the PR description as I started looking at the code :)

lmorchard: So, it sounds like the worst case here is old add-on -> new site and the incompatible add-ons warning is blank until their main add-on is updated. Is that copacetic?

We _shouldn't_ see the new add-on -> old site case since the updated add-on will be coming from the updated site.

lmorchard (approved):
Going to say this is :+1: as long as we're okay with the incompatible add-ons warning being blank until the user has updated their Test Pilot add-on

dannycoates:
> as long as we're okay with the incompatible add-ons warning being blank until the user has updated their Test Pilot add-on

Yeah, #1402 isn't in prod yet, so no regression in this case.

lmorchard:
Oh, one nit / maybe blocker: Need to update the version number for the add-on and then I think we're good to go

PR #1540, opened 2016-10-06
update tab center graphics

johngruen:
@clouserw


lmorchard (approved):
Looks good to me, but don't want to merge if there's checking yet to be done by @clouserw

Issue #1523, opened 2016-10-04
Remove experiment version number from details page


pdehaan:
Currently when we push a new version of an experiment XPI to Test Pilot, we need to update the experiment's YAML file with the new version, and then upstream those to stage and prod.

If we remove the experiment's version number from the Test Pilot site, we'd just need to update the experiment's XPI in S3 and update the update.json.

ghost:
I think this is needs UX.  Right now https://testpilot.firefox.com/experiments/min-vid says:

Version 0.3.0 [changelog]
Last Update 9/22/2016

I think we should drop Version, but we'd also have to drop Last Update, and then where should the changelog link go?

lmorchard:
Issue #547 was closed, but some kind of tool to auto-update experiment details from an .xpi might be handy - if only as a gulp task to generate a pull request

> If we remove the experiment's version number from the Test Pilot site, we'd just need to update the experiment's XPI in S3 and update the update.json.

But we _do_ also have to update YAML content on Test Pilot to bump the update date, which powers our "just updated" messaging on the site. And pushing the site to stage & prod is about to become much simpler, so I don't think this is all that onerous

sevaan:
One option could be to rename the section with the changelog to Additional Info:

<img width="361" alt="screen shot 2016-10-07 at 12 03 28 pm" src="https://cloud.githubusercontent.com/assets/473531/19197102/260d7d2c-8c86-11e6-8cbf-e50ed21acf01.png">

pdehaan:
> One option could be to rename the section with the changelog to Additional Info:

Alternatively, we could just drop the columns and link the text.
So instead of this:

> **Contribute** &nbsp;&nbsp;&nbsp;&nbsp; https://github.com/mozilla/activity-stream
> ...

we just do:

> [**Changelog**](https://github.com/mozilla/activity-stream/blob/master/CHANGELOG.md)
> [**Tour**](#)
> [**Contribute**](https://github.com/mozilla/activity-stream)
> [**Discussion Forum**](https://discourse.mozilla-community.org/c/test-pilot/activity-stream)
> [**Bug Reports**](https://github.com/mozilla/activity-stream/issues)

meandavejustice:
after talking to @clouserw, going to remove release date as well since it often isn't updated on release. If we want to add it back in later, we can grab the `Last-Modified` header from the static xpi file request

PR #1545, opened 2016-10-06
change min-vid URL

clouserw:
Only changing the URL (which is still a 404 right now.  Waiting on @meandavejustice to send me the file).  Not changing the version, that's https://github.com/mozilla/testpilot/issues/1523


clouserw:
I put version 0.3.0 to avoid the 404 confusion. this should be fine to land

ckprice (approved):

PR #1550, opened 2016-10-07
Update yaml for new version of tab center.

bwinton:
(Also make Erica's description more like the rest of the contributors…)


lmorchard (approved):
:+1:

Issue #1551, opened 2016-10-07
Bring back initial email solicitation


johngruen:
It appears to be missing.

PR #1552, opened 2016-10-07
Change copy on email list confirmation message

### LINKED ISSUE #1520: Change copy on email list confirmation message
We don't require a user to verify his or her email address, nor do we send them a confirmation email. Let's change [this language](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/EmailDialog.js#L45):

> When you get a chance, head over to your inbox to verify your email address. Thank you!

Maybe just a "Thank you!"? @johngruen (or maybe @sevaan?)

pdehaan:
Fixes #1520

### DIFF
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -61,7 +61,7 @@ emailOptInInput =
 emailOptInButton = Sign me up
 emailOptInSkip = Skip
 emailOptInConfirmationTitle = Email Sent
-emailOptInSuccessMessage = When you get a chance, head over to your inbox to verify your email address. Thank you!
+emailOptInSuccessMessage = Thank you!
 emailOptInConfirmationClose = On to the experiments...
 
 experimentListPageHeader = Ready for Takeoff!
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -61,7 +61,7 @@ emailOptInInput =
 emailOptInButton = Sign me up
 emailOptInSkip = Skip
 emailOptInConfirmationTitle = Email Sent
-emailOptInSuccessMessage = Thank you!
+emailOptInSuccessMessage2 = Thank you!
 emailOptInConfirmationClose = On to the experiments...
 
 experimentListPageHeader = Ready for Takeoff!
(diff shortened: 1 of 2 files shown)


johngruen:
@pdehaan can you smack a screenshot of your PR in here

locales/en-US/app.ftl
```diff
@@ -61,7 +61,7 @@ emailOptInInput =
 emailOptInButton = Sign me up
 emailOptInSkip = Skip
 emailOptInConfirmationTitle = Email Sent
-emailOptInSuccessMessage = When you get a chance, head over to your inbox to verify your email address. Thank you!
```
lmorchard, line 64: Should we change the l10n ID, since the text has changed so much? We're not actually localizing yet, so it won't do anything practically, but I think this is one of the cases where we want to bump IDs.

chuckharmston: Yeah, we should. I think the general rule of thumb would be to change the ID whenever the meaning of the string changes sufficiently that localizations would change.

chuckharmston (changes requested):

locales/en-US/app.ftl
pdehaan: @lmorchard @chuckharmston What do we want to change the ID to?
I'm guessing it should still have [at least] the `emailOptIn` prefix, but not sure what the best suffix is for this string.

pdehaan:
Ref: #1551

locales/en-US/app.ftl
pdehaan: From IRC:

> @pdehaan: Any ideas on what you want me to change the l10n ID to for https://github.com/mozilla/testpilot/pull/1552#discussion_r82423186?
> 
> @lmorchard: We don't really have a convention, just that it should change to trigger an update (someday) on Pontoon.
> like add a -1 or -2 to the name or the date or something.
> So like emailOptInSuccessMessage2 - not sure what the gaia folks did, but they ran into a need for this.

lmorchard:
Aaaand of course now [the EmailDialog test is looking for the old l10n ID `emailOptInSuccessMessage`](https://github.com/mozilla/testpilot/blob/907b205fff1a0463bad6c20096ddb488cfea567e/frontend/test/app/components/EmailDialog-test.js#L80) - so that needs a tweak to `emailOptInSuccessMessage2`

lmorchard:
Going to close this in favor of PR #1633, where I tweaked the test to pass

PR #1553, opened 2016-10-07
use 'experiment' prop instead of 'getExperimentBySlug' on ExperimentPage

dannycoates:
This feels cleaner to me...

### DIFF
--- a/frontend/src/app/components/ExperimentCardList.js
+++ b/frontend/src/app/components/ExperimentCardList.js
@@ -27,9 +27,10 @@ export default class ExperimentCardList extends React.Component {
     return (
       <div className="card-list experiments">
         {this.getExperiments().map((experiment, key) => (
-          <ExperimentRowCard experiment={experiment}
+          <ExperimentRowCard {...this.props}
+                             experiment={experiment}
                              enabled={isExperimentEnabled(experiment)}
-                             key={key} {...this.props} />
+                             key={key} />
         ))}
       </div>
     );
--- a/frontend/src/app/components/ExperimentPage.js
+++ b/frontend/src/app/components/ExperimentPage.js
@@ -23,12 +23,10 @@ export default class ExperimentPage extends React.Component {
   constructor(props) {
     super(props);
 
-    const { isExperimentEnabled, getExperimentBySlug } = this.props;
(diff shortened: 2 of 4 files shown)


frontend/src/app/containers/App.js
```diff
@@ -59,17 +59,16 @@ function subscribeToBasket(email, callback) {
...
     experiments: experimentSelector(state),
     isFirefox: state.browser.isFirefox,
     isMinFirefox: state.browser.isMinFirefox,
     isDev: state.browser.isDev,
     hasAddon: state.addon.hasAddon,
     installed: getInstalled(state.addon),
     installedAddons: state.addon.installedAddons,
-    getExperimentBySlug: slug =>
-      getExperimentBySlug(state.experiments, slug),
     isExperimentEnabled: experiment =>
```
dannycoates, line 73: you're next 🎯

frontend/src/app/containers/App.js
```diff
@@ -59,17 +59,16 @@ function subscribeToBasket(email, callback) {
 }
 
 export default connect(
-  state => ({
+  (state, ownProps) => ({
     addon: state.addon,
+    experiment: getExperimentBySlug(state.experiments, ownProps.params.slug),
```
lmorchard: I'm kind of leary of this solution. Right now, `/experiment/` is the only page that uses `:slug` in the router. But, that might not always be true and this lookup is specific to the experiment details page. This has us looking up the experiment for _every_ page, whether or not it needs one or has a slug in the URL.

I've been wondering if maybe a cleaner solution would be to put a wrapper around the `ExperimentPage` component in the router that does the experiment lookup & passes it in as a prop. That way, the experiment lookup logic would be limited to the router & the single route.

lmorchard: By wrapper, I mean something like this (which I haven't tried):

``` javascript
      <Route path="/experiments/:slug" 
             component={props => <ExperimentPage {...props} experiment={props.getExperimentBySlug(props.params.slug)}} /> } />
```

dannycoates: Thanks @lmorchard, I didn't know you could do that 😎

lmorchard (approved):

PR #1566, opened 2016-10-10
enable tracking protection in pbmode for blok users

### LINKED ISSUE #1565: Enable Tracking Protection in Private Mode for Tracking Experiment Users
Blok does some stuff to Tracking Protection, we should make sure `privacy.trackingprotection.pbmode.enabled` is set to `true` for experiment users.

dannycoates:
fixes #1565 

This will **enable** native tracking protection for blok users in private browsing mode, meaning blok users will always have protection enabled.

Current blok users will have this setting enabled on browser restart or a disable/enable cycle of the addon. When blok is disabled or uninstalled their previous settings (before enabling blok) will be restored.

### DIFF
--- a/addon/lib/experiment-hacks.js
+++ b/addon/lib/experiment-hacks.js
@@ -8,19 +8,20 @@ const PrefsService = require('sdk/preferences/service');
 const store = require('sdk/simple-storage').storage;
 
 const BLOK_ID = 'blok@mozilla.org';
-const TP_PREFS = [
-  'privacy.trackingprotection.enabled',
-  'privacy.trackingprotection.pbmode.enabled',
-  'services.sync.prefs.sync.privacy.trackingprotection.enabled',
-  'services.sync.prefs.sync.privacy.trackingprotection.pbmode.enabled'
-];
+const TP_PREFS = {
+  'privacy.trackingprotection.enabled': false,
+  'privacy.trackingprotection.pbmode.enabled': true,
+  'services.sync.prefs.sync.privacy.trackingprotection.enabled': false,
+  'services.sync.prefs.sync.privacy.trackingprotection.pbmode.enabled': false
+};
 
 function enableBlok(id) {
   if (id !== BLOK_ID) { return; }
+  const prefnames = Object.keys(TP_PREFS);
   if (!store.tpPrefs) {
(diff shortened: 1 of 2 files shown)


addon/lib/experiment-hacks.js
```diff
@@ -8,19 +8,20 @@ const PrefsService = require('sdk/preferences/service');
...
-  'privacy.trackingprotection.enabled',
-  'privacy.trackingprotection.pbmode.enabled',
-  'services.sync.prefs.sync.privacy.trackingprotection.enabled',
-  'services.sync.prefs.sync.privacy.trackingprotection.pbmode.enabled'
-];
+const TP_PREFS = {
+  'privacy.trackingprotection.enabled': false,
+  'privacy.trackingprotection.pbmode.enabled': true,
+  'services.sync.prefs.sync.privacy.trackingprotection.enabled': false,
+  'services.sync.prefs.sync.privacy.trackingprotection.pbmode.enabled': false
```
groovecoder, line 15: This should be `true` to match the local setting, yes?

groovecoder (changes requested):

addon/lib/experiment-hacks.js
dannycoates: I think we don't want to sync the pref while the experiment is enabled since experiment addons aren't synced. If I have blok on one device but not another I don't want to inadvertently mess with it either way. Make sense (I might be wrong)?

groovecoder: Ah, so this `false` is saying not to sync. I thought it was setting the pref "in the cloud" to be `false`.

groovecoder (approved):

groovecoder:
Approved ... sorry for the confusion about the sync pref.

groovecoder:
FWIW, I'm not comfortable merging this myself in this repo - I haven't spot-checked it locally. Maybe @lmorchard or another core maintainer can review & merge?

lmorchard:
Works for me

groovecoder:
Is this deployed yet? When I open a PBM window with Blok installed, it still opens with TP disabled. 😢

chuckharmston:
@groovecoder our sprint closed on Thursday, this should be deployed on Tuesday.

PR #1568, opened 2016-10-10
update links to point to /latest

clouserw:


chuckharmston:
Make sure we're using 302s here (or are setting low-TTL `Cache-Control` headers), otherwise we may have caching problems.

chuckharmston (approved):
Good start, but we should file a follow-up to backport the other experiments to do this as well.

chuckharmston:
Feel free to merge when CI finishes.

Issue #1302, opened 2016-08-29
Completed experiment page


dannycoates:
Completed experiments have no enable button and special disable button treatment along with different intro content and other elements, as seen: https://mozilla.invisionapp.com/share/Q77S1H6ZC#/screens/181298065

PR #1570, opened 2016-10-11
past experiment list

dannycoates:
I could use some feedback here. See my inline comments.

### DIFF
--- a/frontend/src/app/containers/HomePageWithAddon.js
+++ b/frontend/src/app/containers/HomePageWithAddon.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import classnames from 'classnames';
 import EmailDialog from '../components/EmailDialog';
 import ExperimentCardList from '../components/ExperimentCardList';
 import LoadingPage from './LoadingPage';
@@ -10,11 +11,14 @@ export default class HomePageWithAddon extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { hideEmailDialog: false };
+    this.state = {
+      hideEmailDialog: false,
+      showPastExperiments: false
+    };
   }
 
   render() {
-    const { experiments, getCookie, removeCookie, getWindowLocation } = this.props;
+    const { experiments, getCookie, removeCookie, getWindowLocation, isExperimentCompleted } = this.props;
 
     if (experiments.length === 0) { return <LoadingPage />; }
 
@@ -25,6 +29,9 @@ export default class HomePageWithAddon extends React.Component {
       removeCookie('first-run');
       showEmailDialog = true;
     }
+    const { showPastExperiments } = this.state;
+    const currentExperiments = experiments.filter(x => !isExperimentCompleted(x));
+    const pastExperiments = experiments.filter(isExperimentCompleted);
 
     return (
       <View {...this.props}>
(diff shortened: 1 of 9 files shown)


frontend/src/app/components/ExperimentsListPage.js
```diff
@@ -46,9 +53,14 @@ export default class ExperimentsListPage extends React.Component {
...
         <div className="responsive-content-wrapper">
-          <div data-hook="experiment-list">
-            <ExperimentCardList {...this.props} eventCategory="HomePage Interactions" />
-          </div>
+          <ExperimentCardList {...this.props} experiments={currentExperiments} eventCategory="HomePage Interactions" />
+          {pastExperiments.length > 0 &&
+          <div className={classnames(['button', 'outline'])}
+              style={{margin: '0 auto', display: 'table'}}
+              onClick={() => this.setState({ showPastExperiments: true })}
+              data-l10n-id="viewPastExperiments">View Past Experiments</div>}
```
dannycoates: I need help on how to style this button to get it in the right spot. I had success adding the `style` above but before I put it in scss I'd like to know if that's a _good_ way to do it. @johngruen @chuckharmston ?

frontend/src/app/components/ExperimentsListPage.js
```diff
@@ -46,9 +53,14 @@ export default class ExperimentsListPage extends React.Component {
...
+          {showPastExperiments &&
```
dannycoates: @johngruen do we want the button to go away after clicking? toggle? the spec doesn't say.

johngruen:
Looking at the PR here are things i think we could change: 
- [ ] toggle the button copy to "hide past experiments" when showing the past experiments
- [ ] don't show past experiments if the user does not have the add-on installed
- [ ] past experiments should not show the 'Get started' button if they are not enabled, instead they should have a button that says 'Learn more.' This button should have the `secondary` class assigned to it.

Also, the retired experiment page is sort of an issue, b/c the message implies a past date is in the future. We should fix this, but i'll file  separate bug.

content-src/experiments/universal-search.yaml
```diff
@@ -95,4 +95,5 @@ contributors:
 installation_count: 39943
 created: '2016-04-28T03:57:32.270681Z'
 modified: '2016-07-20T15:42:28.873321Z'
+completed: '2016-10-01T15:42:28.873321Z'
```
dannycoates: I need to remove this before merging, but its useful when reviewing the ui

dannycoates:
This is ready for review, however it builds on top of #1615 so that ought to be reviewed first.

@johngruen I've addressed all your points. The retired experiment page is tracked by #1302 and will have another PR shortly.

dannycoates:
screen shot 2016-10-17 at 11 32 45 am
screen shot 2016-10-17 at 11 33 14 am

frontend/src/app/containers/HomePageWithAddon.js
```diff
@@ -44,9 +51,20 @@ export default class HomePageWithAddon extends React.Component {
...
+              style={{ margin: '0 auto', display: 'table' }}
+              onClick={() => this.setState({ showPastExperiments: true })}
+              data-l10n-id="viewPastExperiments">View Past Experiments</div>}
+          {showPastExperiments &&
+          <div>
+            <div className={classnames(['button', 'outline'])}
+                style={{ margin: '0 auto', display: 'table' }}
+                onClick={() => this.setState({ showPastExperiments: false })}
+                data-l10n-id="hidePastExperiments">Hide Past Experiments</div>
+            <ExperimentCardList {...this.props} experiments={pastExperiments} />
```
lmorchard: Getting a missing required prop warning here for the metrics `eventCategory`: 

```
Warning: Failed prop type: Required prop `eventCategory` was not specified in `ExperimentRowCard`.
```

Do we want to reuse "HomePage Interactions" here, or do we want to measure this differently?

lmorchard (approved):
Works, tests pass, code's good - I'd say the only things are to remove that `completed` timestamp from the YAML and add the `eventCategory` prop to the `ExperimentCardList`. Both seem like small nits, so I'll approve and say go ahead & self-merge after those changes

frontend/src/app/containers/HomePageWithAddon.js
dannycoates: oof. nice catch 🎣

SoftVision-PaulOiegas:
Can't we have one of the experiments or even a sample one as completed on Dev in order to test this before it lands on stage ? Otherwise we could end up with issues on stage that need to be fixed or reverted if major...

PR #1576, opened 2016-10-11
Use double quotes in the PKP header

clouserw:


lmorchard (approved):
Hmm, let's see how this works racing against the other change to deploy.sh

PR #1577, opened 2016-10-11
fix CSP rules for images and ajax post

clouserw:


lmorchard (approved):
Let's go with this and see if it fixes GA

Issue #1372, opened 2016-09-18
Move BaseAppRedirector redirection logic into components


chuckharmston:
The [BaseAppRedirector](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/lib/router.js#L32-L52) logic is sort of gnarly and is probably not the right place for the logic.

A more idiomatic way to do this would be to write wrapping components to encapsulate the logic, acting sort of like Python decorators. In that world, the `ExperimentsListPage` render method might return something like this:

```
<RequireAddon redirect='/'>
    <EverythingElse>
</RequireAddon>
```

The `RequireAddon` component would render `this.props.children`, and its `componentWillMount` method would replace the logic currently encapsulated by `BaseAppRedirector`.

lmorchard:
Yeah, BaseAppRedirector is a sore spot. It works, but it didn't feel like I was doing the right thing. Improvements definitely welcome here.

chuckharmston:
Since this bug was filed, additional logic has been added to `BaseAppRedirector`. Renamed the issue to reflect its original purpose, and filed #1583 to handle the removal of the rest of the logic.

PR #1581, opened 2016-10-11
Trim trailing slash from slug for ExperimentPage pageview GA ping

lmorchard:
Fixes "TypeError: d is undefined" when experiment page is visited with a
trailing slash.

This patches the issue for now, but maybe #1372 can provide a cleaner solution overall. 😞

### DIFF
--- a/frontend/src/app/containers/App.js
+++ b/frontend/src/app/containers/App.js
@@ -49,7 +49,11 @@ class App extends Component {
         dimension1: hasAddon
       });
     } else if (pathname.indexOf(experimentsPath) === 0) {
-      const slug = pathname.substring(experimentsPath.length);
+      let slug = pathname.substring(experimentsPath.length);
+      // Trim trailing slash, if necessary
+      if (slug.charAt(slug.length - 1) === '/') {
+        slug = slug.substring(0, slug.length - 1);
+      }
       const experiment = this.props.getExperimentBySlug(slug);
       this.debounceSendToGA(pathname, 'pageview', {
         dimension1: hasAddon,


ckprice:
Just a note that I'm also seeing this without the trailing slash too

https://testpilot.firefox.com/experiments/activity-stream?utm_source=yes

Parameters are stripped, no ping is sent, and the `TypeError: d is undefined` appears in the console.

lmorchard:
> Parameters are stripped

That's a slightly different issue than what this PR addresses. The server is redirecting all experiment URLs without slashes to add the slash and dropping params. We can't correct for that on the client side. This PR only aims to fix with the slash is already there with params (the error you're seeing)

lmorchard:
Rebased against the refactorings on master

dannycoates (approved):
LGTM 👍

ghost:
This file is completely different than what is in production right now:  https://github.com/mozilla/testpilot/blob/production-static/frontend/src/app/containers/App.js

I can't cherry pick this, but are we even sure it fixes the problem?

lmorchard:
> This file is completely different than what is in production right now

Oh, ugh, yeah I rebased against recent refactorings, which totally made this impossible to cherry-pick. Here's a new branch with my un-rebased previous version cherry-picked on top of production as of right now:

https://github.com/mozilla/testpilot/commits/cherrypick-fix-type-error

> I can't cherry pick this, but are we even sure it fixes the problem?

This fixes _a_ problem that's definitely happening right now: Any visit to an experiment page with a trailing slash in the URL results in a TypeError and no GA ping. 

This PR does _not_ solve the issue of server-side redirects stripping GA params - but even if we have that solved, the client-side issue is still there without this PR.

Issue #1467, opened 2016-09-28
Persist query string through redirect to /experiments or /


ckprice:
STR:
- Have Test Pilot installed.
- Go to https://testpilot.firefox.com/?utm_source=testpilot-addon&utm_medium=firefox-browser&utm_campaign=testpilot-doorhanger&utm_content=not+badged

Expected Result:
- Redirected to https://testpilot.firefox.com/experiments?utm_source=testpilot-addon&utm_medium=firefox-browser&utm_campaign=testpilot-doorhanger&utm_content=not+badged

Actual Result:
- Redirected to https://testpilot.firefox.com/experiments/

This affects our GA attribution.

lmorchard:
As a caveat, that redirect is client-side. Does GA register hits at all for fully client-side window.history changes? If not, then retaining the URL params won't actually solve the issue if recording the page change is what we're after.

ghost:
1) Make the content of /experiments live at /
2) Stop linking to /experiments
3) Redirect /experiments to / for everyone

(stuff will still live at /experiments/______, we're just skipping the middle step)

johngruen:
note @clouserw's comment is now outdated

dannycoates:
no longer an issue as of current master fc53730c390a84c32c6eb7c1287bef2bd102e77c

PR #1584, opened 2016-10-11
Use parent HomePage component.

### LINKED ISSUE #1372: Move BaseAppRedirector redirection logic into components

chuckharmston:
- Redirects `/experiments` to `/`
- Renames `LandingPage` as `HomePageNoAddon`
- Renames `ExperimentsListPage` as `HomePageWithAddon`
- Creates a new `HomePage` component to negotiate which component to render.
- Removes redirection logic from `BaseAppRedirector`

Ultimately I'd prefer that we have all the `HomePage*` components in a single file (e.g.:

``` js
export class HomePageNoAddon {};
export class HomePageWithAddon {};
export default class HomePage {};
```

), but `HomePageNoAddon` is still gnarly enough that it probably belongs in its own file.

Closes #1372, #1467

### DIFF
--- a/frontend/src/app/components/HomePage.js
+++ b/frontend/src/app/components/HomePage.js
@@ -1,0 +1,17 @@
+import React from 'react';
+
+import HomePageNoAddon from './HomePageNoAddon';
+import HomePageWithAddon from './HomePageWithAddon';
+
+export default class HomePage extends React.Component {
+  render() {
+    if (this.props.hasAddon) {
+      return <HomePageWithAddon {...this.props} />;
+    }
+    return <HomePageNoAddon {...this.props} />;
+  }
+}
+
+HomePage.propTypes = {
+  hasAddon: React.PropTypes.bool
+};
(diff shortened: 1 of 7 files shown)


lmorchard (approved):
Much cleaner - works for me!

SoftVision-PaulOiegas:
Let me get this straight. Now homepage and experiments page are displayed the same in address bar (testpilot.firefox.com) ? And only if you navigate to a specific experiment page you'll see /experiments/experiment_name ?

chuckharmston:
Correct! What was formerly at the `/experiments` page was moved to `/`, but is still only displayed if you have the add-on installed.

This let us remove some buggy redirection logic.

PR #1587, opened 2016-10-11
Removes BaseAppRedirector:

### LINKED ISSUE #1583: Move BaseAppRedirector GA logic to App container.
Forked from #1372. See https://github.com/mozilla/testpilot/blob/4bebfa9e1df7313f8c805dc5b9d44791c4ea1999/frontend/src/app/lib/router.js#L42

chuckharmston:
- Disconnects AppRouter.
- Creates wrapping component to lookup + propagate the experiment.
- Moves GA pageview logic to App component.
- Moves AppRouter state processing to App.

Closes #1583.


frontend/src/app/components/ExperimentPage.js
```diff
@@ -18,7 +18,25 @@ import ExperimentPreFeedbackDialog from '../components/ExperimentPreFeedbackDial
 
 const CHANGE_HEADER_ON = 105;
 
+
 export default class ExperimentPage extends React.Component {
+  render() {
+    const { getExperimentBySlug, params } = this.props;
+    const experiment = getExperimentBySlug(params.slug);
```
lmorchard, line 25: This bit just got moved to the router so we can keep router-knowledge (e.g. like params.slug) out of the individual display components. Also let us just pass in an experiment, rather than slug and helper function.

chuckharmston: Hm, not sure I agree with the necessity of that. Information from the router just more state from the store, right? Why is it bad for our components to depend on that? Especially if it's [easily testable](https://github.com/mozilla/testpilot/pull/1587/files#diff-fdc1fc59bc074c06a5e102348cc58cf5R12).

lmorchard: Then again, having that bit in the router means the router has experiment-knowledge. But, I figured that was less of a problem to have the router know how to adapt from slug to experiment to pass into the routed component.

lmorchard: Also considering at some point we'll use this component in static rendering (#1306), easier to just pass in an experiment than a lookup-helper and a param.slug

chuckharmston: That's why I separated `ExperimentPage` and `ExperimentDetail`; the logic for determining the component is in a single, easily-testable component. If we're static-rendering we can detect that and bypass the lookup entirely, e.g.:

``` js
render() {
  if (__ISOMORPHIC__) {
    return <ExperimentDetail experiment={__EXPERIMENT__} {...this.props}/>
  }
  const experiment = getExperimentBySlug(...);
  return <ExperimentDetail experiment={experiment} {...this.props} />
}
```

lmorchard: > That's why I separated ExperimentPage and ExperimentDetail

Ah, yeah, fair enough! I withdraw my raised eyebrow.

lmorchard: (It also looks a lot cleaner than an inline function shoehorned into the router anyway)

chuckharmston: (Short-circuited my above comment, updated with a snippet)

chuckharmston: \+ we get to avoid connecting the router entirely, which gets us down to the single-connected-component ideal.

lmorchard (approved):
Tried to find a gotcha but found none. LGTM

frontend/src/app/lib/router.js
```diff
@@ -1,11 +1,6 @@
 /* global ga */
```
lmorchard: Nit: This could go away since the GA code moved.

Also, now that this is just a clean router component, could it move from `app/lib` to `app/components` and be less of a special snowflake?

chuckharmston: It can and will!

lmorchard:
Found a nit and a question of a nit, but if you don't feel like fixing those go ahead and self-merge

PR #1588, opened 2016-10-12
make experiment details scrolling smooth

johngruen:
@lmorchard this is failing because I'm querying element heights inside the component to modify the scroll snapping. I'm not sure how to mock out element heights in our new test suite. Any advice or suggestions would be most welcome.

### DIFF
--- a/frontend/src/app/components/ExperimentPage.js
+++ b/frontend/src/app/components/ExperimentPage.js
@@ -16,8 +16,6 @@ import MainInstallButton from '../components/MainInstallButton';
 import ExperimentCardList from '../components/ExperimentCardList';
 import ExperimentPreFeedbackDialog from '../components/ExperimentPreFeedbackDialog';
 
-const CHANGE_HEADER_ON = 105;
-
 
 export default class ExperimentPage extends React.Component {
   render() {
@@ -54,7 +52,10 @@ export class ExperimentDetail extends React.Component {
       showDisableDialog: false,
       shouldShowTourDialog: false,
       showTourDialog: false,
-      showPreFeedbackDialog: false
+      showPreFeedbackDialog: false,
+      changeHeaderOn: 125,
+      stickyHeaderSiblingHeight: 0,
+      privacyScrollOffset: 15
     };
 
     // HACK: Set this as a plain object property, so we don't trigger crazy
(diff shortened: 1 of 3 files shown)


lmorchard:
> this is failing because I'm querying element heights inside the component to modify the scroll snapping. I'm not sure how to mock out element heights in our new test suite. Any advice or suggestions would be most welcome.

You can return a value with a function supplied to `sinon.spy()` - so something like `subject.setProps({ getElementOffsetHeight: sinon.spy(sel => 150) })` in a test case will let you mock what height the component gets.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -382,11 +382,56 @@ export class ExperimentDetail extends React.Component {
...
     this.didScroll = false;
   }
 
+  // if there is no content in the .status-bar
+  // adjust the initial snap position and the header isn't yet sticky, adjust
+  // changeHeaderOn to the top of the .details-header
+  // instead of the bottom of the #main-header
+  setChangeHeaderOn(detailsHeaderWrapperHeight) {
+    const { getElementOffsetHeight } = this.props;
+    const statusBar = document.querySelector('.status-bar');
```
lmorchard: You shouldn't need to query the DOM for `.status-bar` here. The logic for displaying the status bar [is here with the `statusType` variable](https://github.com/johngruen/testpilot/blob/3cacd7f58e27eba0a255951dd40042847bd2d857/frontend/src/app/components/ExperimentPage.js#L173) - we might want to split that out into a component method to reuse it. Basically it looks at `this.props.experiment.error` and `this.state.enabled` to decide whether to show that element.

frontend/src/app/components/ExperimentPage.js
```diff
@@ -416,11 +461,18 @@ export class ExperimentDetail extends React.Component {
     return null;
   }
 
+  // added the magic number to the setSrcollY call
+  // to build in a small offset top of the highlight box shadow animation
   highlightPrivacy(evt) {
     const { getElementY, setScrollY } = this.props;
+    const detailsHeaderWrapper = document.querySelector('.details-header-wrapper');
```
lmorchard: Looks like maybe you really want `getElementOffsetHeight('.details-header-wrapper')` here, since the next two calls seem to be looking for a number and not an element

johngruen:
Alright, @lmorchard...I got all the tests working. Thanks for your help here!

lmorchard:
> Thanks for your help here!

I think I was more confus{ed,ing} than helpful, but glad it worked out 😕

lmorchard (approved):
Works, tests pass, looks good :+1:

Issue #1521, opened 2016-10-04
Collect email addresses with persistent form at bottom of page


sevaan:
Our metrics show that email is a powerful tool for converting users for new experiment installations. We should explore options to collect email addresses at a greater number of touch points. One good touch-point would be to have an email collection form at the bottom of each Test Pilot page.

UX: https://mozilla.invisionapp.com/share/SW8TRWTMA
Example of footer on landing page: https://mozilla.invisionapp.com/share/E28TRXP46
Example of footer on details page: https://mozilla.invisionapp.com/share/UN8TRXWVS

This footer exists on a number of different Mozilla pages so it may be possible to use that existing code. You can see an example of it in action here at the bottom of the page: https://www.mozilla.org/en-US/firefox/products/

The asset for the mail icon can be found here: 
![mail](https://cloud.githubusercontent.com/assets/473531/19084157/5fece88e-8a33-11e6-9bcf-380defe37bca.png)

chuckharmston:
Few questions for @sevaan:
- Are these strings final?
- Can we get an SVG of the email icon?
- Is it okay if we block submission of the form if it's not complete, rather than having a distinct error state?
- How does it look if there's a white-background footer, e.g. the [privacy page](https://testpilot.firefox.com/privacy)?

sevaan:
> Are these strings final?

Yes, I am using the same strings that were used to collect email addresses before. The strings in the footer (e.g. the fine print in the form) are on all the other Mozilla web properties so do not need to be altered as they are already vetted with Legal. We removed one string, and Jess Osorio approved it.

> Is it okay if we block submission of the form if it's not complete, rather than having a distinct error state?

All this code exists and is in use on an assortment of Mozilla properties, so the error states, animations, and everything and are already coded. Is it not possible to use that code as it exists?

> How does it look if there's a white-background footer, e.g. the privacy page?

I don't think we should or need to capture emails on the white-background footer pages.

sevaan:
[Asset in SVG!](https://github.com/mozilla/testpilot/files/520076/Email.svg.zip)

chuckharmston:
Thanks!

> All this code exists and is in use on an assortment of Mozilla properties, so the error states, animations, and everything and are already coded. Is it not possible to use that code as it exists?

[...]

PR #1598, opened 2016-10-12
Use <View> component to inject headers and footers.

chuckharmston:
This is a nice code quality PR that should make our views a little more DRY. It:
- Creates a `<View>` component that automatically injects the `<Header>` and `<Footer>`.
  - `View` automatically passes each of its props to all component children.
  - The header and footer can be hidden through a prop.
  - Additional CSS classes can be changed through a prop.
- Implements `<View>` on each appropriate component.
- Moves all page/view-type components + tests to the `containers` directory.
- Cleaned up the `Header` component a bit to better handle configuration via props.
- Renamed `Restart` to `RestartPage` to reflect how it's used.
- Tried out a docstring format used by [react-docgen](https://github.com/reactjs/react-docgen).

Notably, I'll use `View` to globally include the email footer via #1521.


chuckharmston:
This is ready for review. (It's not as scary as it looks; lots of moving files, fixing indentation, etc).

lmorchard:
Oops. I think I just broke this after merging the page scrolling stuff - could use a rebase

frontend/src/app/components/View.js
```diff
@@ -0,0 +1,129 @@
...
+     * proxy object looks like this:
+     *
+     * { '$$typeof': Symbol(react.element), type: 'ComponentName' }
+     *
+     * or like this, if it is a DOM element:
+     *
+     * { '$$typeof': Symbol(react.element), type: 'div' }
+     */
+    return (
+      React.Component.isPrototypeOf.call(React.Component, element) || (
```
lmorchard, line 36: Oof. This looks painful to have worked out. I wonder if there's something we can toss upstream to enzyme? Nasty to have to hack things like this just to satisfy tests, but we do what we must because we can I guess

lmorchard (approved):
Nothing broken I can see; tests pass; code looks good - hooray for simplification and redundancy reduction!

lmorchard:
Just going to leave an approval here - assuming the rebase is easy, go ahead & self-merge after that

Issue #1522, opened 2016-10-04
Switch testpilot.dev.mozaws.net to static deployment from master


lmorchard:
We auto-deploy `master` branch via Docker to an Elastic Beanstalk instance at `testpilot.dev.mozaws.net`

We auto-deploy `development-static` branch via `gulp` &`aws sync` to an s3 bucket distributed through CloudFront at `testpilot-static.dev.mozaws.net`

But, we no longer need these to be separate.

We should: 
1. Switch `master` branch to static deployment to the S3 bucket `testpilot-static.dev.mozaws.net` 
   - (or create a new S3 bucket named `testpilot.dev.mozaws.net`)
2. Switch `testpilot.dev.mozaws.net` to point at the CloudFront host behind `testpilot-static.dev.mozaws.net` 
   - (or create a new CloudFront distro for the new S3 bucket)
3. Switch to using [`bin/deploy.sh`](https://github.com/mozilla/testpilot/blob/master/bin/deploy.sh) rather than [just bare `aws sync` in `circle.yml`](https://github.com/mozilla/testpilot/blob/master/circle.yml#L75)
4. Retire Docker-based build and deployment from our `circle.yml` setup 

Either way, all the above is stuff I / we set up and self-service on Cloud Services AWS. So, this doesn't require bothering ops for help.

lmorchard:
Kind of assumed in many conversations, but not yet mentioned in this issue: Static staging & prod now have an nginx-based proxy between S3 and CloudFront to handle custom headers. dev will need that too

PR #1599, opened 2016-10-13
Only use SSL for Google Analytics (closes #1578).

### LINKED ISSUE #1578: Get our Google Analytics on SSL
In our CSP we allowlist www.google-analytics.com.  I noticed AMO is allowing https://ssl.google-analytics.com.  I like that https restriction.  Let's do that for us too -- why are they using ssl.\* instead of www.* and how do we do that?

chuckharmston:
r? @clouserw

### DIFF
--- a/testpilot/settings.py
+++ b/testpilot/settings.py
@@ -341,6 +341,8 @@ CSP_CONNECT_SRC = (
     'https://*.mozilla.net',
     'http://*.mozilla.org',
     'https://*.mozilla.org',
+    'https://www.google-analytics.com',
+    'https://ssl.google-analytics.com',
 )
 CSP_FONT_SRC = (
     "'self'",
@@ -357,8 +359,8 @@ CSP_IMG_SRC = (
     'https://*.mozilla.org',
     'http://*.gravatar.com',
     'https://*.gravatar.com',
-    'http://*.google-analytics.com',
-    'https://*.google-analytics.com',
+    'https://www.google-analytics.com',
+    'https://ssl.google-analytics.com',
     'https://ssl.gstatic.com/',
     'https://testpilot-dev-media.s3.amazonaws.com',
     'https://testpilot-stage.s3.amazonaws.com',
@@ -370,8 +372,7 @@ CSP_SCRIPT_SRC = (
     'https://*.mozilla.org',
     'http://*.mozilla.net',
     'https://*.mozilla.net',
-    'http://*.google-analytics.com',
-    'https://*.google-analytics.com',
+    'https://ssl.google-analytics.com',
     'https://apis.google.com',
 )
 CSP_STYLE_SRC = (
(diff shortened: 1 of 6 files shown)


testpilot/settings.py
```diff
@@ -357,8 +357,7 @@ def lazy_langs():
     'https://*.mozilla.org',
     'http://*.gravatar.com',
     'https://*.gravatar.com',
-    'http://*.google-analytics.com',
-    'https://*.google-analytics.com',
+    'https://ssl.google-analytics.com',
```
chuckharmston, line 363: Probably not necessary to change this file, but meh.

chuckharmston:
Worth noting that the file from `ssl.google-analytics.com` is hardcoded to use `//www.google-analytics.com` for pings, but since `http://ssl.google-analytics.com` redirects to `https://ssl.google-analytics.com`, it will always use `https`. Accordingly, I've updated `script-src` to only use `https://ssl.` in `script-src`, but still allow `https://www.` in `img-src` and `connect-src`.

testpilot/settings.py
lmorchard: 😐 Django should be excised soon

lmorchard (approved):
LGTM FWIW but don't want to preempt a clouserw review

ghost (approved):
Nice, this looks a lot more secure.

chuckharmston:
🔒 💚

testpilot/settings.py
chuckharmston: https://github.com/mozilla/testpilot/issues/1608

ghost:
Marking needs QA as I think is the convention now.  We want to verify this on stage.  That can be done by basically looking at the console and making sure there aren't any CSP errors.  Extra credit if you can see pings going to google analytics in the network tab.  Thanks!

/cc @SoftVision-PaulOiegas

SoftVision-PaulOiegas:
@wresuolc @chuckharmston Guys, was this deployed on staging or not ? 
I'm asking because since morning we tried to test this on staging and we haven't got any "POST" or "GET" GA events with "https://ssl.google-analytics.com" url. Also no CSP errors displayed while navigating trough Test Pilot pages, only when external pages like GitHub and some footer links were opened.

However, we also tried the same tests on dev server where we found lots of "POST" google analytics collect events with SSL in url but none of the "GET" collect ones had it. If we remember well the "ssl" change was not present yesterday on dev, correct me if I'm wrong. 
So we switched testing on dev server since we found the "ssl" change there and looked also for CSP errors. The problem is that almost every GA event displays a CSP error on dev server. We have seen them for navigating on experiments pages, experiments enable/disable actions and also for tour navigation.

Did we wrongly understood how the events should be displayed or there is an issue here ?

**Stage console**
browser console on stage

**Dev console**
browser console on dev

ghost:
@SoftVision-PaulOiegas sorry for the misunderstanding.  I just meant, when we push to stage at our normal time we should check this.  Stage was not updated.

Additionally, http://testpilot.dev.mozaws.net/ is out of date and testpilot-static.dev.mozaws.net doesn't have the nginx proxy in front of it (which we need to test this).  So, we don't have a working dev server right now apparently.  https://github.com/mozilla/testpilot/issues/1522 should fix that.

PR #1605, opened 2016-10-13
add a comment about 3rd party API access

clouserw:
These URLs changed with our static file conversion and broke our stats.  I'm adding a comment reminding people that they are in use elsewhere.  I think this is the best place for it, but feedback welcome.


lmorchard:
Oh, oof. Yeah we added a `.json`, didn't we?

lmorchard (approved):

Issue #1574, opened 2016-10-11
You are wrongly redirected to Test Pilot homepage while you visit "Privacy" and "Terms" pages and after you click on the Firefox logo


SoftVision-CosminMuntean:
[Affected versions]:
- Firefox 49.0.1 and up

[Affected Platforms]:
- All Windows

[Prerequisites]:
- The "testpilot.env" string pref was manually added with "stage" value on a new clean profile in "about:config" page.
- Test Pilot add-on was previously installed on the profile.

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites and navigate to "https://testpilot.stage.mozaws.net/experiments/" page.
2. Scroll the page and from the bottom click the "Privacy" link.
3. Scroll the page and from the bottom click the "Terms" link.
4. Click the Test Pilot Firefox icon from the top of the page.
5. Observe the browser behavior.

[Expected result]:
- You are redirected to the "https://testpilot.stage.mozaws.net/experiments/" page.

[Actual result]:
- You are wrongly redirected to "https://testpilot.stage.mozaws.net/" page. 

[Notes]:
- Added a screen recording with the issue: https://goo.gl/lCYVuk.

chuckharmston:
I'm going to work on #1372, which will fix this.

chuckharmston:
Apparently #1372 did not fix this; I'll keep investigating.

SoftVision-CosminMuntean:
Verified this after the fix for #1615 was deployed and it's no longer reproducible with the original steps. 
Tested on dev server with the latest Firefox (49.0.1) release and latest Nightly (52.0a1) build.

Base on this, I will close this issue as resolved fixed.

PR #1615, opened 2016-10-13
switch addon detection to polling

### LINKED ISSUE #1606: addon installed state races on home page
I managed to trigger two unfortunate races setting the installed state of the addon that caused the experiment list page to show the no-addon version instead of the (correct) with-addon version. Once the page was in the no-addon state it stayed there.

I added a few `console.warn` lines before each dispatch of `setHasAddon`. Here's the log from a time when the page goes from the loading view to no-addon view.

screen shot 2016-10-13 at 10 42 58 am

And here's a particularly weird one where the page goes from loading flashing to with-addon but then quickly to no-addon.

screen shot 2016-10-13 at 11 06 43 am

dannycoates:
fixes #1606

As noted in the bug, PageMod is a jerk. We can't rely on changes to `navigator.testpilotAddon` to signal when the addon is removed because it can change at other times too.

This PR has the webapp detect the addon by sending `sync-installed` messages until it gets a response.

There's a case this change breaks: when a page is open to the webapp while the addon is removed or disabled from about:addons it won't detect it. Removing the addon via the webapp _does_ work correctly.

### DIFF
--- a/frontend/src/app/containers/App.js
+++ b/frontend/src/app/containers/App.js
@@ -10,7 +10,7 @@ import Clipboard from 'clipboard';
 import { getInstalled, isExperimentEnabled, isInstalledLoaded } from '../reducers/addon';
 import { getExperimentBySlug, isExperimentsLoaded } from '../reducers/experiments';
 import experimentSelector from '../selectors/experiment';
-import { uninstallAddon, installAddon, enableExperiment, disableExperiment } from '../lib/addon';
+import { uninstallAddon, installAddon, enableExperiment, disableExperiment, pollAddon } from '../lib/addon';
 import addonActions from '../actions/addon';
 import RestartPage from '../containers/RestartPage';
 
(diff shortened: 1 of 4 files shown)


dannycoates:
This should also fix #1574

dannycoates:
Side question. We've got two PageMods in the addon. The one that sets the installed flag includes `'https://www.mozilla.org/*,about:home'`. Is the flag used in practice anywhere outside txp or is that list aspirational?

lmorchard:
> Is the flag used in practice anywhere outside txp or is that list aspirational?

Not sure if they're currently in use, but that flag was explicitly requested by Mozilla.org and Snippets teams so they could measure and customize messaging & promos on those sites for Test Pilot users.

lmorchard:
Taking a look at this after many distractions

lmorchard:
Hmm, another effect of this PR is that when multiple tabs with Test Pilot are open, only the foreground tab detects removal of the add-on via the uninstall flow. Install works across tabs, though.

But, this seems to keep the add-on install state from flipping to false during normal site navigation, so I guess that's less edge-case-y than the multiple tab + uninstall case? Think I'll merge this PR and file a followup issue about that case though, in case we can figure out a polling approach to solve that too

lmorchard (approved):

PR #1619, opened 2016-10-14
add nosniff headers to all file types

clouserw:
Add no-sniff headers to all the files.  For most we specify the file type, so it should work fine.  For the "everything else" bucket, amazon attempts to auto-detect the file type and put the right Content-Type header on it.  I spot checked png, ico, js, and css.  They were all correct.  If we find a filetype amazon doesn't detect correctly it's just a matter of setting it ourselves, so let me know if anyone sees one.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -48,7 +48,7 @@ aws s3 sync \
   --content-type "application/json" \
   --exclude "*" \
   --include "*.json" \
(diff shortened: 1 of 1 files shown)


chuckharmston (approved):
Looks good to me! I'll leave it unmerged so you can merge + test it right away.

Issue #1363, opened 2016-09-16
Consider using CSS modules


chuckharmston:
We should consider using [CSS modules](https://github.com/css-modules/css-modules). It feels strange at first, but does a great job of tightly coupling styles to components, avoiding deep nesting, and preventing brittle reliance on cascading.

For React: https://github.com/gajus/react-css-modules

johngruen:
@chuckharmston gonna kick this out of the 37 Milestone. Let's discuss this some time this weel though.

ckprice:
Let's close (our app isn't that complex, so not as helpful).

PR #1620, opened 2016-10-15
Adds newsletter subscription form to footer (closes #1521).

### LINKED ISSUE #1521: Collect email addresses with persistent form at bottom of page

chuckharmston:
Still needs tests, but I thought I'd get it out for @johngruen to look at visuals Monday morning.

Some implementation notes:
- I built the CSS thinking about #1363 (one file per component with a single class' namespace for everything therein).
- I tried to abstract the form from the implementation of the form, so that we can use it in other places (i.e. post-Test Pilot install) with minimal effort.
## TODO
- [x] Rename everything from `email` to `newsletter`.
- [x] Use the `mergeProps` argument to `connect` to namespace `setEmailForm*` to keep everything clean.
- [x] Don't try to plan for future forms, that's premature optimization.
- [x] Make `subscribeToBasket`driven by a dispatched action, rather than a callback.
- [ ] Write tests for all the things I introduced.
## Testing

Using the [redux dev tools extension](https://addons.mozilla.org/en-US/firefox/addon/remotedev/), you can simulate error and success states without submitting the form by dispatching the following action payloads:
### Submitting

``` js
{
  type: 'newsletterFormSetSubmitting'
}
```
### Error

``` js
{
  type: 'newsletterFormSetFailed'
}
```
### Success

``` js
{
  type: 'newsletterFormSetSucceeded'
}
```
### Default

screen shot 2016-10-15 at 1 22 22 am
### After entering a character

screen shot 2016-10-15 at 1 22 51 am
### When trying to submit with an empty email

screen shot 2016-10-15 at 1 23 03 am
### When trying to submit with an invalid email

screen shot 2016-10-15 at 1 23 10 am
### When trying to submit without agreeing to the privacy notice

screen shot 2016-10-15 at 1 23 25 am
### At medium width

screen shot 2016-10-15 at 1 23 37 am
### At small width

screen shot 2016-10-15 at 1 23 49 am
### On success

screen shot 2016-10-15 at 1 27 34 am
### On error

screen shot 2016-10-15 at 1 28 00 am


chuckharmston:
The button and input should maybe be shrunk down a bit at medium widths.

johngruen (dismissed):
@chuckharmston looks like the form is leaking into some extra pages:

screen shot 2016-10-17 at 7 22 56 pm

johngruen:
@chuckharmston also noticed it on the uninstall test pilot interstitial page

johngruen:
Otherwise UX looks solid, i'll leave it to someone else to do a code review

johngruen:
@chuckharmston one more nit... after the email, the string should be "We will only send you Test Pilot related information".

frontend/src/app/components/EmailFooter.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  renderSuccess() {
+    return (
+      <header>
+        <h2 data-l10n-id="emailFooterSuccessHeader">Thanks!</h2>
+        <p data-l10n-id="emailFooterSuccessBody">
+          If you haven't previously confirmed a subscription to a Mozilla-related
+          newsletter you may have to do so. Please check your inbox or your spam
+          filter for an email from us.
```
pdehaan: You say "email" here, but you hyphenated ("e-mail") on line 27 above.

frontend/src/images/email.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="146px" height="127px" viewBox="0 0 146 127" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 40.2 (33826) - http://www.bohemiancoding.com/sketch -->
+    <title>Email</title>
+    <desc>Created with Sketch.</desc>
```
pdehaan: Probably want to run this file through SVGO or something.

locales/en-US/app.ftl
```diff
@@ -163,10 +163,24 @@ restartRequiredFromExperiment = Please restart Firefox <br/> to enable {$experim
...
 incompatibleSubheader = We recommend <a href="{$url}">disabling these add-ons</a> before activating this experiment:
 
+emailFormEmailPlaceholder =
+    [html/placeholder] Your email here
+emailFormDisclaimer = We will only send you Mozilla-related information.
+emailFormPrivacyNotice = I'm okay with Mozilla handling by info as explained in <a href="{$url}">this privacy notice</a>.
+emailFormSubmitButton = Sign Up Now
+emailFormSubmitButtonSubmitting = Submitting...
+
+emailFooterError = There was an error submitting your e-mail address. Try again?
```
pdehaan: e-mail -> email (for consistency w/ line 167 and 177.

frontend/src/styles/components/EmailFooter.scss
```diff
@@ -0,0 +1,104 @@
+$email-icon--width: 146px;
+$email-icon--height: 127px;
+
+$email-icon--responsive-unit: .66;
+
+.email-footer {
+  background: $transparent-white-1;
```
pdehaan:
```
Error in plugin 'sass'
Message:
    frontend/src/styles/components/EmailFooter.scss
Error: Undefined variable: "$transparent-white-1".
        on line 7 of frontend/src/styles/components/EmailFooter.scss
>>   background: $transparent-white-1;
   --------------^
```

chuckharmston: @johngruen had similar issues that were solved by blowing away some old `node_modules`. This variable [exists and is used all over](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=transparent-white-1).

frontend/src/app/components/EmailFooter.js
```diff
@@ -0,0 +1,78 @@
...
+  getData() {
+    return this.props.forms.email ? this.props.forms.email.values : null;
+  }
+
+  submitFailed() {
+    return this.props.forms.email && this.props.forms.email.submitFailed;
+  }
+
+  submitSucceeded() {
+    return this.props.forms.email && this.props.forms.email.submitSucceeded;
```
pdehaan: I think we have some failing Circle-CI tests due to this line:

```
1) app/components/View should pass its props to child components:
     TypeError: Cannot read property 'email' of undefined
      at EmailFooter.submitSucceeded (src/app/components/EmailFooter.js:17:12)
      at EmailFooter.getClassNames (src/app/components/EmailFooter.js:63:21)
      at EmailFooter.render (src/app/components/EmailFooter.js:69:28)
      at /home/ubuntu/testpilot/node_modules/react/lib/ReactCompositeComponent.js:793:21
      at measureLifeCyclePerf (/home/ubuntu/testpilot/node_modules/react/lib/ReactCompositeComponent.js:74:12)
```

chuckharmston: Yeah, I haven't touched tests yet. As i noted in comment 0, I put this up there so John could do a UX review of it while I worked.

frontend/src/styles/components/EmailFooter.scss
pdehaan: OK, I was just copy-pasta-ing that from Circle-CI, so we should see if we can blow away that cache.

pdehaan:
@johngruen Is that the correct email icon image? I vaguely recall a conversation at https://github.com/mozilla-services/pageshot/issues/1650#issuecomment-251809106 where we were looking for a simpler image, due to poor "glance-ability".
Not sure if that applies here as well, or if it was just more of a PageShot icon consistency thing.

johngruen:
@pdehaan i think this one is good here now, it's internally consistent with our other email message

frontend/src/app/components/EmailFooter.js
johngruen: @chuckharmston @pdehaan let's stick to email

frontend/src/app/components/EmailForm.js
```diff
@@ -0,0 +1,92 @@
...
+  renderDisclaimer() {
+    return (
+      <p className={this.makeRevealedClassNames()}
+         data-l10n-id='emailFormDisclaimer'>
+        We will only send you Test Pilot-related information.
+      </p>
+    );
+  }
+
+  subscribeToBasket(values) {
```
lmorchard: FWIW, we have [a `subscribeToBasket` helper in `containers/App.js`](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App.js#L154) that [we pass in props for `components/EmailDialog.js`](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/EmailDialog.js#L75) and [mock out in the test](https://github.com/mozilla/testpilot/blob/master/frontend/test/app/components/EmailDialog-test.js#L48). Could probably use the same thing here?

chuckharmston: Yup, most of the code is lifted from there. As I mentioned in comment 0, I'm trying to appropriately encapsulate all the functionality in the form so it can be reused elsewhere without dependency. I'll likely remove that code when addressing #1551.

chuckharmston:
So, there's a really weird thing here:
- [`redux-form`](http://redux-form.com/) requires that your form components be connected through [a decorator](http://redux-form.com/6.1.1/docs/api/ReduxForm.md/) that wires it to the state.
- Enzyme's shallow rendering does not work when there are nested connected components; it throws an exception complaining that the store isn't present in the context (which of course it isn't; we're trying to make components that can be tested in isolation).
- [Passing context manually](https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md#arguments) doesn't work, for reasons I don't have my head around yet.

This is a problem, because this component (and other forms we might create with the library) will be a child of all of our views. Some related bugs:

https://github.com/airbnb/enzyme/issues/183
https://github.com/producthunt/chai-enzyme/issues/13
https://github.com/airbnb/enzyme/issues/435
https://github.com/airbnb/enzyme/issues/472

I spent most of the day getting to this point wherein the existing tests pass with my patch. I had to do some hackery, effectively not rendering the email footer in all tests. This is less than ideal, but @clouserw mentioned that there's some urgency in this bug, so I'm putting this out there for review now without any additional tests.

Some ideas:
1. I'm going to keep trying to figure out why passing the context manually isn't working for me. The key is likely somewhere in there.
2. I might be able to skip the `reduxForm` entirely if we're able to pass the store down manually. But [it does a lot](https://github.com/erikras/redux-form/blob/428e9f06b8f6e5deace99eb6ab17b61739cb31d9/src/reduxForm.js#L76), and this is likely not feasible.
3. You might be able to wrap the component being tested in `<Provider>`, but many of Enzyme's APIs are restricted to the root-mounted component, so we'd have to drastically rewrite existing tests.
4. We can skip on `redux-form` entirely. It's popular, comprehensive, and well-supported so I'd hate to do that, but it is an option.
5. We can hide these connected components in tests. This is also awful.

@lmorchard @dannycoates: would love to hear your thoughts here.

chuckharmston:
> I'm going to keep trying to figure out why passing the context manually isn't working for me. The key is likely somewhere in there.

Promising: I was able to get this working by setting the static `contextTypes` on `ExperimentDetail`.

lmorchard:
> We can skip on redux-form entirely. It's popular, comprehensive, and well-supported so I'd hate to do that, but it is an option.

I'd lean toward this. It's not a very complex form, is it? Maybe pass on this library until we need it?

chuckharmston:
> I'd lean toward this. It's not a very complex form, is it? Maybe pass on this library until we need it?

`redux-form` makes life a lot easier: it manages fields (which is very complex when you link them to a store), state, and validation. [This](https://github.com/mozilla/testpilot/pull/1620/files#diff-e27553be6868a51686b7624a9a65249fR6) is drastically less (and more grokkable) code than it would have otherwise; definitely a good library to use as far as hygiene goes.

If the problem it creates is related to our testing stack, then I'd say that some investigating is warranted, at the least.

chuckharmston:
I'm going to rewrite it without the library. For posterity, the commit with the passing tests was 413ddfb.

chuckharmston:
Alright, this is updated, rolling it by hand.

I'm not at all thrilled with the approach, especially compared to the relative simplicity of `redux-form`, but it does work. I'll get back to it tomorrow.

frontend/src/app/actions/newsletter-form.js
```diff
@@ -0,0 +1,29 @@
...
+        'Content-Type': 'application/x-www-form-urlencoded'
+      },
+      body: `newsletters=test-pilot&email=${encodeURIComponent(email)}`
+    })
+      .then(() => dispatch(actions.newsletterFormSetSucceeded()))
+      .catch(() => dispatch(actions.newsletterFormSetFailed()));
+  }
+})
+
+export default Object.assign({}, actions, subscribeActions);
```
chuckharmston, line 29: This is sort of hacky, but it works.

The pattern is a little clearer when using vanilla redux for actions: http://redux.js.org/docs/advanced/AsyncActions.html#async-action-creators

frontend/src/app/containers/App.js
```diff
@@ -118,36 +119,52 @@ function subscribeToBasket(email, callback) {
...
-    navigateTo: path => dispatch(routerPush(path)),
-    enableExperiment: experiment => enableExperiment(dispatch, experiment),
-    disableExperiment: experiment => disableExperiment(dispatch, experiment),
-    requireRestart: () => dispatch(addonActions.requireRestart()),
-    setHasAddon: installed => {
-      dispatch(addonActions.setHasAddon(installed));
-      if (!installed) { pollAddon(); }
-    }
-  }),
-  (stateProps, dispatchProps, ownProps) => Object.assign({
```
chuckharmston, line 150: This was getting unwieldy. I broke into parts to make it clearer and to make the actual `connect` call a one-liner.

chuckharmston:
The rewrite is now finished, other than the tests.

@clouserw expressed interest in merging it without tests before the freeze, so I'd like to get this reviewed now. In the interim I'll get started on them.

r? @lmorchard @dannycoates

frontend/src/app/actions/newsletter-form.js
```diff
@@ -0,0 +1,29 @@
...
+    newsletterFormSetEmail: email => email,
+    newsletterFormSetPrivacy: privacy => privacy
+  },
+  'newsletterFormSetFailed',
+  'newsletterFormSetSubmitting',
+  'newsletterFormSetSucceeded'
+);
+
+const subscribeActions = createActions({
+  newsletterFormSubscribe: (dispatch, email) => {
```
chuckharmston, line 14: `dispatch` is added in `mapDispatchToProps` here: https://github.com/mozilla/testpilot/pull/1620/files#diff-d1e3fb4bae03a0b6b298c56e7110d06fR159

dannycoates: I do not like this but I'm not going to block it

frontend/src/app/components/NewsletterForm.js
```diff
@@ -0,0 +1,92 @@
+import classnames from 'classnames';
+import React from 'react';
+
+
+export default class NewsletterForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleEmailChange = this.handleEmailChange.bind(this);
+    this.handlePrivacyClick = this.handlePrivacyClick.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
```
dannycoates, line 10: we aren't consistent in this repo wrt this vs `e => this.foo(e)` and should eventually decide on one or the other

frontend/src/app/actions/newsletter-form.js
```diff
@@ -0,0 +1,29 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded'
+      },
+      body: `newsletters=test-pilot&email=${encodeURIComponent(email)}`
+    })
+      .then(() => dispatch(actions.newsletterFormSetSucceeded()))
+      .catch(() => dispatch(actions.newsletterFormSetFailed()));
+  }
+})
```
dannycoates: ;

frontend/src/app/components/NewsletterForm.js
```diff
@@ -0,0 +1,92 @@
...
+    return (
+      <form className='newsletter-form' onSubmit={this.handleSubmit}>
+        {this.renderEmailField()}
+        {this.renderPrivacyField()}
+        {this.renderSubmitButton()}
+        {this.renderDisclaimer()}
+      </form>
+    );
+  }
+}
```
dannycoates, line 92: Please followup with PropTypes

dannycoates (approved):
LGTM. Needs a followup for tests.

frontend/src/app/components/NewsletterForm.js
chuckharmston: ☑️

frontend/src/app/actions/newsletter-form.js
chuckharmston: ☑️

frontend/src/app/actions/newsletter-form.js
lmorchard: Yeah, I don't really like actions dispatching other actions like this. 

Maybe the form should dispatch the `newsletterFormSetSubmitting` and `newsletterFormSubscribe` actions one after the other itself?

And then you can just yield an appropriate action payload describing success/fail at the end of the fetch without dispatching further actions. [We kind of do this in `fetchExperiments`](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/actions/experiments.js#L7), but sadly without good error handling. 

But, this payload could just be for one single action (`newsletterFormSubscribe`) that the reducer handles appropriately based on whether it has a `statusSuccess==true` or `statusFailed==true` property - or something like this.

chuckharmston: @lmorchard so, I think it does make sense for actions to dispatch other actions; this pattern is even featured in Redux's docs: http://redux.js.org/docs/advanced/AsyncActions.html#async-action-creators

The problem is in the `redux-actions` library, which cuts some boilerplate at the cost of functionality.

lmorchard: Eh, yeah, I think it just rubs me the wrong way for reasons I'm unable to really explain. But, this looks like a pretty textbook example from the docs. So, may just be I need to soak my head in more docs

Issue #1509, opened 2016-10-03
Update /__version__ on static deploys


lmorchard:
Our Docker-based deployments have a `/__version__` JSON resource we can use to check the current deployed version of the site. The new static deployment system does not do this. It probably should.

lmorchard:
Going to have to ping @relud or someone else in ops to fully resolve this. We _were_ doing this in CircleCI with a [build-version-json.sh](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-version-json.sh) script that generated JSON from CircleCI vars. That, in turn, got embedded in the Docker container that we shipped to stage & prod eventually.

[I got this working for a dev S3 bucket](https://github.com/lmorchard/testpilot/commit/c34c407aeb1131bdc121928727d69ed9d5b52f73) - but I realized that's only because I'm using CircleCI to build & deploy to dev. 

For stage & prod, we're using an internal Mozilla service to build & deploy. We'll need that to do the equivalent of [build-version-json.sh](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-version-json.sh) with whatever env vars are available there.

relud:
i can add a script to the build step when i get back next week. Preferrably it would create `dist/__version__` and be labelled with the correct content-type by `bin/deploy.sh`. lmk when there's a script to do that.

relud:
also: you don't get access to the circle env vars, but you do get access to the git repo, which should tell you the branch and commit, and with `git describe --abbrev=0` you can get the most recent tag in the branch

relud:
or if you want tag exactly matching the current commit: `git describe --exact-match --abbrev=0`

relud:
this is fixed

Issue #1575, opened 2016-10-11
Let's turn S3 website mode back off


ghost:
We had to turn website mode on to fix a weird 404 error with the privacy and terms.  I'd like to know why that error was happening, but I'd also like to turn the website mode back off, as it does some different things (http://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html#WebsiteRestEndpointDiff)

lmorchard:
Wondered about this just now in triage: could turn off hosting mode in S3 and add whatever we need to the nginx proxy in the middle? 

I think the main thing we need that mode for is so that we handle trailing slashes - i.e. `/path/` is served up with the object `/path/index.html` from S3. Not sure we need it for anything else

Could bother relud to try this - or after #1522 I could try playing with nginx myself for a bit first so we have a proof-of-concept

lmorchard:
I managed to get [this nginx config](https://github.com/mozilla/testpilot/blob/master/header-proxy/nginx.conf) to work for dev with an S3 bucket that does _not_ have website mode on. 

There's a [`location = /`](https://github.com/mozilla/testpilot/blob/master/header-proxy/nginx.conf#L39) line that takes care of the front page, and [tweaks to deploy.sh](https://github.com/mozilla/testpilot/blob/master/bin/deploy.sh#L96) that upload `path/index.html` as `path` to S3. Between those two, I _think_ maybe we have things covered to consider turning off website mode in S3 for prod & stage? Is there anything missing?

ghost:
We could try it on -dev [early] next sprint to try to flush out problems.  Also /cc @relud

relud:
question: is this really better than using website mode?

lmorchard:
> We could try it on -dev [early] next sprint to try to flush out problems. 

Well, the thing is, testpilot.dev.mozaws.net is working that way _right now_. So, we can beat on it whenever we like or surface issues

ghost:
It's hackier.  My main concern is just that we set custom [security] headers on all the things we care about and everything else just gets served up as XML.  With website mode, injection might be more possible, through 5xx errors, or whatnot.  Anything we don't specify gets the normal, non-secure headers.  It's admittedly a small window and I don't know enough about website mode to point to actual exploit paths.

@jvehent - we talked about this briefly a while ago.  Do you have concrete examples or any concerns with website mode?

lmorchard:
[...]

PR #1629, opened 2016-10-18
Rework dev to use static deployment with nginx proxy

### LINKED ISSUE #1522: Switch testpilot.dev.mozaws.net to static deployment from master

lmorchard:
- Update add-on dev environment to use HTTPS
- Remove {stage,dev}-static envs from add-on
- Master branch uses bin/deploy.sh to upload to S3 bucket
- Upload **version** to S3 (only dev)
- Add Dockerized nginx header filter proxy
- Enable caching for node_modules in circle.yml
- Re-enable setup-test-dependencies.sh to download Firefox binary for
  add-on testing, but disable integration test package install within
  the script (folloup to PR #1624)
- Further bin/circleci cleanup

Fixes #1522, Issue #1509


lmorchard:
I tested out this PR by using `1522-master-static` in place of `master` in `circle.yml` to auto-update https://testpilot.dev.mozaws.net/ on pushes to my own branch.

Also did some things outside of this PR, infrastructure-wise:
- Created a new S3 bucket named `testpilot.dev.mozaws.net` - _without_ web site hosting enabled (see Issue #1575)
- Created a new Docker container for the header transform proxy and [uploaded it to Docker Hub](https://hub.docker.com/r/lmorchard/txp-proxy/) - this proxies requests for the S3 bucket
- Launched a new EC2 instance named `Test Pilot dev HTTP proxy` (ID: `i-0a49cc4d7ea523a5e`) with Docker installed and started an instance of `lmorchard/txp-proxy`
- Created a CloudFront distribution (`E2ERG47PHCWD0Z`) pointing at the proxy in EC2

The EC2 instance should never need to be updated unless we're also making changes to prod & stage, so there's no auto-updating there. The only thing this PR auto-updates is the S3 bucket.

Not entirely sure where / if to document this stuff. I don't think it's exactly worth keeping secret, since you shouldn't be able to do anything to these resourced without a Cloud Services AWS login

addon/index.js
```diff
@@ -41,29 +41,17 @@ const SERVER_ENVIRONMENTS = {
     BADGE_COLOR: '#AA00AA'
   },
   dev: {
-    BASE_URL: 'http://testpilot.dev.mozaws.net',
+    BASE_URL: 'https://testpilot.dev.mozaws.net',
```
ghost, line 44: yesss

ghost (approved):
This looks reasonable to me, and is a pain to test without all the infrastructure set up.  Let's push to -dev!

bin/deploy.sh
```diff
@@ -107,3 +107,12 @@ for fn in $(find dist -name 'index.html' -not -path 'dist/index.html'); do
...
 done
+
+# __version__ JSON; 10 minute cache
+aws s3 cp \
+  --cache-control "max-age=${MAX_AGE}" \
+  --content-type "application/json" \
+  --metadata "{${HPKP}, ${HSTS}, ${TYPE}}" \
+  --metadata-directive "REPLACE" \
+  --acl "public-read" \
+  version.json s3://${TESTPILOT_BUCKET}/__version__
```
relud, line 118: does putting this here mean that on every deploy we delete `/__version__` at line 89 above, and reupload it here?

lmorchard: It shouldn't... `version.json` is generated in the root directory of the project, not `frontend/build`, so it shouldn't get caught up with the sync at line 89. And `__version__` is never actually a local filename

lmorchard: Oh wait, derp, I see what you mean... yeah, it gets deleted & re-uploaded because there's never a local `__version__` file.

relud: the sync at line 89 deletes any files not in `dist/` (because of the --delete flag on line 90)

relud: so `__version__` _should_ be a local filename, to prevent deleting

Issue #1627, opened 2016-10-18
Test Pilot uninstall is not detected across all tabs


lmorchard:
In #1615, we switched to a polling approach to detect installation of the main add-on. The previous approach of relying on a PageMod in the add-on to flip `navigator.testpilotAddon` seemed to have race condition (#1606).

This change introduces a new issue, though: If multiple Test Pilot tabs are open, and the user uninstalls Test Pilot in one of those tabs, the other tabs do not detect the uninstall. Also, uninstalling from about:addons is not detected in any tabs, either.

Maybe a solution would be to add more polling? That is, we're pinging while uninstalled to detect a change to installed. Maybe we add a "heartbeat" ping while installed, and if one or more beats are missed we switch to uninstalled state on the front end?

PR #1630, opened 2016-10-18
send pings from message-bridge to webpage as a keep-alive

dannycoates:
@lmorchard here's a sketch of something for #1627 

I imagine we'd want to deploy this in two parts to give version mismatches time to settle. First to get deployed would be the message-bridge change to begin pinging (the current webapp will receive them but do nothing with them), then the webapp change (next deploy) to begin disconnecting.


dannycoates:
Closing to implement in two parts. Part 1 - #1637

Issue #1249, opened 2016-08-15
Add an About Test Pilot page


johngruen:
- Page should be formatted like the legal pages
- Add a link to the page in the Test Pilot Footer
- Add a link the the page in the Test Pilot Settings Menu
- I owe content here.

Issue #973, opened 2016-06-10
Test Pilot doesn't include SRI or OpenGraph/Twitter metadata on legal pages


pdehaan:
Found in http://testpilot.dev.mozaws.net/__version__ (637abee3e107ecf0aa587c80ef778b72c4578e41):

``` json
{ "commit": "637abee3e107ecf0aa587c80ef778b72c4578e41",
  "version": "",
  "build": "https://circleci.com/gh/mozilla/testpilot/505",
  "source": "https://github.com/mozilla/testpilot" }
```

---
### Steps to reproduce:
1. Go to http://testpilot.dev.mozaws.net/ and view source:
   
   ``` html
   <head>
     <meta charset="utf-8" />
     <link rel="shortcut icon" href="/static/images/favicon.dc6635050bf5.ico">
     <link rel="stylesheet" href="https://code.cdn.mozilla.net/fonts/fira.css"
                            crossorigin="anonymous"
                            integrity="sha384-APhs/OUouhH+ZbBANL3+7a5J1sVSYyfUhGxTWiDMkTuEIO5fXWZonzNEj+CagDB7">
     <link rel="stylesheet" href="/static/styles/main.885355dc102e.css"
                            integrity="sha384-vnDt+fGbk3oCVKyHvGrF3RYzC7h4WdAENt74SgaNvFkdfZxqkwt3lzgJ6gSPe/sP">
     <meta name="defaultLanguage" content="en-US">
     <meta name="availableLanguages" content="en-US">
     <meta name="viewport" content="width=device-width">
     <link rel="localization" href="/static/locales/{locale}/app.l20n">
     <title>Firefox Test Pilot</title>
     <link rel="canonical" href="http://testpilot.dev.mozaws.net/">
     <meta name="description" content="Test new Features. Give us feedback. Help build Firefox." />
     <meta property="og:description" content="Test new Features. Give us feedback. Help build Firefox." />
     <meta property="og:image" content="http://testpilot.dev.mozaws.net/static/images/thumbnail-facebook.1877727a8303.png" />
     <meta property="og:locale" content="en_US" />
     <meta property="og:title" content="Firefox Test Pilot" />
     <meta property="og:type" content="website" />
     <meta property="og:url" content="http://testpilot.dev.mozaws.net/" />
     <meta name="twitter:card" content="summary" />
     <meta name="twitter:description" content="Test new Features. Give us feedback. Help build Firefox." />
[...]

PR #1638, opened 2016-10-20
add an about page

johngruen:
@clouserw the copy here will need to get reviewed before tuesday, so we may need to get a supplemental PR for this, but i'd like to get it reviewed and in today.


johngruen:
Fixes #1249 and (for fun) #973

lmorchard:
I can take a quick look, with regard to the page generation

johngruen:
@lmorchard thanks! To clarify, our copy editor is looking at the text in the about.md page, but it's open season on everything else.

lmorchard (approved):
So, :+1: for the changes to generated pages, and the about page gets built fine.

One question though: All these generated pages have lost their white backgrounds and the "Learn More" links are no longer there to fold up & reveal content. Is that intentional? 

Old: https://www.dropbox.com/s/erfjuwspcj2sher/Screenshot%202016-10-20%2016.35.42.png?dl=0

New: https://www.dropbox.com/s/jd5kmamrgq8cre4/Screenshot%202016-10-20%2016.36.13.png?dl=0

If that's all copacetic, good to self-merge

ckprice:
> and the "Learn More" links are no longer there to fold up & reveal content.

This was an ask from legal, so we may want to get it back in.

ckprice (dismissed):
I think this should be blocked until we can fix the toggle, or confirm with legal that it's not needed anymore.

johngruen:
@ckprice @lmorchard i totally forgot to check the content toggles. The color change is by design...I figured it would be nice if these pages better matched all the others. 

I'll look into re-adding the toggles first thing tomorrow.

PR #1642, opened 2016-10-20
pass browserHistory into createStore

dannycoates:
Clicking experiment cards broke.

I missed this in #1620


chuckharmston (approved):
I missed it too. Good catch.

PR #1646, opened 2016-10-21
add underscores to new sass components

johngruen:
@chuckharmston @dannycoates this should fix the failing sass-build task on master


chuckharmston:
Hm, this is strange.

The build is fine—otherwise we wouldn't see the CSS correctly compiled on staging—the linting is the problem, and apparently the linting and build are calculating builds differently, since the linter thinks a variable isn't available when it actually is.

I'll merge because this is a good change regardless, but it's likely a symptom of a deeper issue for us to keep an eye out for.

chuckharmston (approved):

Issue #1654, opened 2016-10-21
Test Pilot image is no longer displayed in Facebook or Twitter shared posts


SoftVision-PaulOiegas:
[Affected versions]:
- Firefox 49.0.1 and up

[Affected Platforms]:
- All Windows

Prerequisites:
- Facebook or Twitter account already logged.

[Steps to reproduce]:
1. Open Firefox and navigate to "testpilot.firefox.com/share" page.
2. Click on the Facebook or Twitter share button.
3. Follow the process to share / tweet the Test Pilot page.
4. Navigate to Facebook or Twitter account page and observe the shared page.

[Expected result]:
- The shared posts correctly display the Test Pilot image.

[Actual result]:
- The shared posts have no Test Pilot image displayed.

[Notes]:
- Observed this on Twitter first but since last night, Facebook shared posts started also to show this issue.

![facebook share](https://cloud.githubusercontent.com/assets/17258026/19601655/e2c5aa02-97b2-11e6-82e2-5688bc81f09f.png)

lmorchard:
Might have found the issue: https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Ftestpilot.firefox.com

```
Provided og:image URL, https://testpilot.firefox.com/static/images/thumbnail-facebok.png was not valid because it did not meet the minimum size constraint of 200px by 200px.
```

"facebok"

SoftVision-PaulOiegas:
Nope, still not working on dev or stage for either Facebook or Twitter shares. Used a brand new profile so not a caching problem.

![no image in share](https://cloud.githubusercontent.com/assets/17258026/19639774/702c5fcc-99e2-11e6-83c8-2efcdb5df093.png)

SoftVision-PaulOiegas:
In addition some network output.

![warning](https://cloud.githubusercontent.com/assets/17258026/19640168/31573c70-99e4-11e6-93b8-f6e7ef40c0f8.png)

and

![test pilot image maybe](https://cloud.githubusercontent.com/assets/17258026/19640173/36553a1a-99e4-11e6-9c75-c9f9ec836f44.png)

ghost:
Fixed in https://github.com/mozilla/testpilot/pull/1659

SoftVision-PaulOiegas:
Sorry guys, but still not working. Tested on all 3 servers using both Nightly and 49.0.2 release.

chuckharmston:
[...]

PR #1659, opened 2016-10-21
Copy thumbnail-*.png to dist/ for deployment

lmorchard:
Issue #1654

### DIFF
--- a/frontend/tasks/dist.js
+++ b/frontend/tasks/dist.js
@@ -25,8 +25,13 @@ gulp.task('dist-rev-assets', function() {
     .pipe(gulp.dest(config.DIST_PATH));
 });
 
+gulp.task('dist-thumbnail-images', () => {
+  gulp.src(config.DEST_PATH + 'static/images/thumbnail-*.png')
+    .pipe(gulp.dest(config.DIST_PATH + 'static/images'));
+});
+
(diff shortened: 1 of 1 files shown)


lmorchard:
A caveat: This fixes the 404s, but I would like to poke some more at gulp-rev-all to see if we can't get it to update the absolute URLs in the header to use the rev'd URLs. Otherwise, we will have a hard time updating those far-future expiration images.

Spent some time hacking on a [custom replacer](https://github.com/smysnk/gulp-rev-all#replacer). But, I think it's a little more complex than that, so I threw this together as a stopgap in the meantime

ghost:
If I understand this patch, the `**` above works, but it adds a hash to the filename, and the meta tags we have on the site don't know about the hashes.  So, with this patch we end up with two copies of the images, just one that is unused.  Is that right?

lmorchard:
> So, with this patch we end up with two copies of the images, just one that is unused. Is that right?

Yeah, 2 copies for now as a bandaid. In the future - or at least whenever we want to change the long-expiring thumbnail images - we need to figure out how to get the build process to properly replace the URLs in the meta tags with hashed versions. But, for now, this will at least make sure the thumbnail images aren't 404s

ghost (approved):

Issue #1670, opened 2016-10-24
Allow experiments to be hidden from users without specific locales.


chuckharmston:
e.g. Only show Foo Experiment to users who speak Azerbaijani

johngruen:
@chuckharmston clarification on the needs:product label? This seems legit

chuckharmston:
I honestly couldn't say; must have intended to put it on a different issue. Removing.

chuckharmston:
I honestly couldn't say...must have hit the wrong bug on a list. Removing.

PR #1686, opened 2016-10-26
Allow experiments to be blocked from or granted to locales.

### LINKED ISSUE #1602: Allow experiments to be hidden from users with specific locales.
Basically, there's a collision between our upcoming German search experiment and Universal Search. We should decouple the timelines here by adding an optional flag to the YAML that lets us disable certain experiments for certain locales.

chuckharmston:
Closes #1602, #1670.

### DIFF
--- a/docs/experiments/experiment-content/template.yaml
+++ b/docs/experiments/experiment-content/template.yaml
@@ -69,3 +69,7 @@ launch_date: '' #UTM time stamp, this will hide the experiment on production ser
 order: 0 # where your experiment will appear in the list...lower numbers appear higher
 incompatible: # if your experiment has incompatible add ons, repeat the following syntax, otherwise delete this field
   # 'add-on ID': 'Add-on Name'
+locale_blocklist: [] # Prevent the experiment from being shown to users of these locales
+  # - 'de'
+locale_grantlist: [] # Only allow the experiment to be shown to users of these locales
+  # - 'de'
--- a/frontend/src/app/reducers/browser.js
+++ b/frontend/src/app/reducers/browser.js
@@ -5,9 +5,11 @@ import { isMinFirefoxVersion } from '../lib/utils';
 const userAgent = navigator.userAgent.toLowerCase();
 const isUserAgentFirefox = userAgent.indexOf('firefox') > -1;
 const isUserAgentMinFirefox = isMinFirefoxVersion(isUserAgentFirefox, userAgent, config.minFirefoxVersion);
+const locale = (navigator.language || '').split('-')[0];
 
 export default handleActions({}, {
   isFirefox: isUserAgentFirefox,
   isMinFirefox: isUserAgentMinFirefox,
-  isDev: config.isDev
+  isDev: config.isDev,
+  locale
 });
--- a/frontend/src/app/selectors/experiment.js
+++ b/frontend/src/app/selectors/experiment.js
@@ -16,7 +16,7 @@ export const allExperimentSelector = createSelector(
 
 
 // Return launched experiments from the store, sorted.
-export const launchedExperimentSelector = createSelector(
+export const onlyLaunchedExperimentSelector = createSelector(
   allExperimentSelector,
   experiments => experiments.filter(experiment => (
     moment(moment.utc()).isAfter(experiment.launch_date) ||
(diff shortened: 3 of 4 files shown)


docs/experiments/experiment-content/template.yaml
```diff
@@ -69,3 +69,7 @@ launch_date: '' #UTM time stamp, this will hide the experiment on production ser
 order: 0 # where your experiment will appear in the list...lower numbers appear higher
 incompatible: # if your experiment has incompatible add ons, repeat the following syntax, otherwise delete this field
   # 'add-on ID': 'Add-on Name'
+blocklist: [] # Prevent the experiment from being shown to users of these locales
```
pdehaan: `locale_blocklist` and `locale_grantlist`?

chuckharmston: Good idea. Updated.

frontend/src/app/reducers/browser.js
```diff
@@ -5,9 +5,11 @@ import { isMinFirefoxVersion } from '../lib/utils';
 const userAgent = navigator.userAgent.toLowerCase();
 const isUserAgentFirefox = userAgent.indexOf('firefox') > -1;
 const isUserAgentMinFirefox = isMinFirefoxVersion(isUserAgentFirefox, userAgent, config.minFirefoxVersion);
+const locale = (navigator.language || '').split('-')[0];
```
pdehaan, line 8: Should this be `navigator.language` or should we check the `navigator.languages` array?
Are we concerned if the user has the German language, but it isn't the primary? Should they still see the experiment?

chuckharmston: In a meeting, @johngruen said we just wanted to check against the first-choice language.

`navigator.language` [is the first item in `navigator.languages`](https://github.com/mozilla/testpilot/pull/1688).

frontend/src/app/selectors/experiment.js
```diff
@@ -25,6 +25,31 @@ export const launchedExperimentSelector = createSelector(
...
+
+// Return the user's primary language subtag.
+export const localeSelector = store => store.browser.locale;
+
+
+// Passed a locale and set of experiments, filters out the experiments that are
+// blockedlisted in that locale, or grantlisted and not available in that locale.
+export const l10nSelector = (locale, experiments) => experiments.filter(exp => {
+  if ('locale_blocklist' in exp) {
+    return exp.locale_blocklist.indexOf(locale) === -1;
```
pdehaan: Not sure if you want to use the fancy `Array#includes()` here instead. Doesn't really add anything, but you may be able to reduce to:

``` js
return !exp.locale_blocklist.includes(locale);
...
return exp.locale_grantlist.includes(locale);
```

frontend/src/app/reducers/browser.js
lmorchard: What are the `locale` values meant to be, here? For example, it looks like this collapses `en-US`, `en-GB`, & `en-NZ` into just `en` and `es-MX`, `es-ES`, & `es-CO` into just `es`. Is that what we want? They're definitely different locales, despite shared languages.

Actually, just noticed discussion in #1602, will copy this to there

chuckharmston: @johngruen has said (trying to remember where—may have been in a meeting) that we only care about first-choice language, irrespective of region codes. That may change in the future, but would be an easy change to make.

lmorchard: <strike>Could we do this with a glob or pattern or regex? Would be nice if we didn't have to change everything if/when we decide to mind locales in the future</strike>

Edit: Bah, actually, never mind. If we don't need it now, let's worry about it when we do. Not like we have 100s of experiments that will need updating.

chuckharmston: Found it: https://github.com/mozilla/testpilot/issues/1602#issuecomment-253537100

lmorchard (approved):
Looks good, tests pass. Played with block/grant lists and changed languages in my browser, worked as expected

Would be good to have a test somewhere that asserts / documents the locale flattening logic, but I don't see a good place for it except maybe for the browser reducer with a mocked out global. Kind of a nit, though, and we've already made noise about it

PR #1691, opened 2016-10-26
updated circleci build

dannycoates:
The main feature of this change is to run the frontend tests with code coverage and send the results to https://codecov.io

Other changes:
- altered circleci scripts to run on personal forks more easily
- split the dependency install steps from the test steps
  - circleci runs all test steps even if the previous fails which can be useful if there's errors in both addon and frontend code
- moved some build steps that aren't required for testing to the deployment phase
- utilize the circleci "Test Summary" area for frontend tests
- updated node to the LTS version
- fixed logo path on readme

Example circleci run: https://circleci.com/gh/dannycoates/testpilot/15
Example codecov: https://codecov.io/gh/dannycoates/testpilot

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,7 +7,7 @@ machine:
   hosts:
     testpilot.dev: 127.0.0.1
   node:
-    version: 6.2.0
+    version: 6.9.1
   environment:
     TESTPILOT_ENABLE_PONTOON_BRANCHES: master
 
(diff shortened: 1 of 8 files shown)


circle.yml
```diff
@@ -7,40 +7,41 @@ machine:
...
     - "addon/node_modules"
-    - "node_modules"
-
-  pre:
-    # https://discuss.circleci.com/t/nvm-commands-are-not-evaluated/943
-    - nvm use 6.2.0 && nvm alias default 6.2.0
+    # node_modules is included by default
 
   override:
-    - node --version
```
lmorchard, line 25: I think we do still want these versions included in the test output, to be clear about what we're using

circle.yml
```diff
@@ -7,40 +7,41 @@ machine:
...
-    - npm --version
+    - npm install
```
lmorchard: Might want to move these two commands into a `./bin/circleci/setup-node-dependencies.sh' script to be clear

circle.yml
```diff
@@ -7,40 +7,41 @@ machine:
...
+    - cd addon && npm install
     - google-chrome --version
     - ./bin/circleci/setup-test-dependencies.sh
-    - ./bin/circleci/build-version-json.sh
     - ./bin/circleci/build-addon.sh
-    - ./bin/circleci/build-frontend.sh
 
 test:
   override:
+    - ./bin/circleci/test-addon.sh
```
lmorchard, line 28: :+1: on separating test from build. I started to do that at some point with the Django cleanup but got distracted by something shiny and wandered off

lmorchard (approved):
Things look good, though I have two small nits. If you don't feel like making those changes, feel free to merge!

circle.yml
dannycoates: The node and npm version show up in the 'machine' section of circle's output

lmorchard: Ah, cool cool

PR #1695, opened 2016-10-26
Adds warning when localized experiments not available in their language (closes #1600).

### LINKED ISSUE #1600: Provide a method of letting users know if experiments are not localized to their language
When we ship our German experiments, we generally don't want to prevent German speakers from using non-localized experiments. We should provide UI letting German speaking experimenters know if an experiment is EN only.

chuckharmston:
screen shot 2016-10-27 at 8 16 49 am


johngruen:
@chuckharmston made a PR into your branch that should resolve the needs:ux label on #1600...

LMK if you have questions

johngruen:
screen shot 2016-10-27 at 2 15 42 pm

chuckharmston:
Thanks, John! Incorporated your code, this is ready for review now.

locales/en-US/app.ftl
```diff
@@ -182,6 +182,9 @@ newsletterFooterBody = Find out about new experiments and see test results for e
 newsletterFooterSuccessHeader = Thanks!
 newsletterFooterSuccessBody = If you haven't previously confirmed a subscription to a Mozilla-related newsletter you may have to do so. Please check your inbox or your spam filter for an email from us.
 
+localeWarningTitle = This experiment is only available in English.
```
lmorchard, line 185: Hmm, this message is hardcoded to "English" - do we need to include the name of a language here? i.e. German, for a German-localized experiment?

chuckharmston: @johngruen? Copy came from him.

lmorchard: Looks from #1600 like maybe we're only worried about German speakers seeing the English experiments, and not vice versa?

lmorchard (approved):
Tests pass, works, looks good. Just have a question about the hardcoded language message, and this needs a rebase now after merging the other PR. 

I'll say :+1: to self-merge after the rebase, though.

locales/en-US/app.ftl
chuckharmston: For now yeah, we'll be blocklisting some experiments in German and grantlisting others to German. I built this to handle to/from any language though, so the copy should probably support it.

johngruen: @lmorchard @chuckharmston yeah, my guess is that for the foreseeable future experiments (other than the upcoming German experiment) will start in English and be localized later. Should that cease to be the case, we'll need to revisit this string.

PR #1699, opened 2016-10-27
Fixes #1682. Add docs for windows users

### LINKED ISSUE #1682: Add development quickstart guide for Windows hosts.
There's no development quickstart guide for Windows hosts and the steps for setting up dev environment for OSX/Linux are not working on Windows.

dumptyd:
Added development quick start guide for Windows hosts and made some changes to a few files to make testpilot run on all operating systems.
Here are the changes: 

`.eslintrc` : Added `linebreak-style: [0]` option as without that eslint throws error on Windows because Windows uses CRLF as newline and *nix systems use LF. 

`addon/package.json` : Added  `package-win` option for Windows users.

`docs/development/quickstart.md` : Updated quickstart.md

`frontend/tasks/pages.js` : Use `Path.basename` for getting file name from an absolute path instead of `file.path.split('/').pop()` as Windows uses backslashes so the latter only works on *nix systems.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -37,3 +37,4 @@ rules:
   no-unused-expressions: [0]
   no-use-before-define: [2, nofunc]
   yoda: [0]
+  linebreak-style: [0]
--- a/docs/development/quickstart.md
+++ b/docs/development/quickstart.md
@@ -67,7 +67,32 @@ it communicates with the site and grants the ability to enable & disable
 experiments. Setting up your browser with the steps above should make it easier
 for you to get it working.
 
-## Windows hosts
+## For Windows hosts
 
-**Help wanted**: Getting things working on Windows may be similar to OS X,
-but the team has little experience with that environment.
+After installing Node.js for Windows, run these commands to get started: 
+
+```cmd
+:: Set up add-on environment and build an unsigned package
+cd addon
+npm install
+npm run package-win
+
+:: Set up frontend web site environment
+cd ..
+npm install
+```
+
+Now, open a second command prompt window, this time with admin privileges and run this:
+
+```cmd
+:: Add hostname alias to /etc/hosts and start up dev webserver
+echo 127.0.0.1 testpilot.dev >> %WINDIR%\System32\Drivers\Etc\Hosts
+```
+
+Go back to the previous command prompt window and run
+
+```cmd
+npm start
+```
+
+Follow the remaining instructions from **Linux & OS X** section and you're all set.
--- a/frontend/tasks/pages.js
+++ b/frontend/tasks/pages.js
@@ -1,6 +1,6 @@
 const gulp = require('gulp');
 const config = require('../config.js');
-
+const path = require('path');
 const Remarkable = require('remarkable');
 const fs = require('fs');
 const gutil = require('gulp-util');
@@ -95,7 +95,7 @@ function buildExperimentPage() {
 
 function convertToCompiledPage() {
   return through.obj(function compiledConvert(file, encoding, callback) {
-    const filename = file.path.split('/').pop();
+    const filename = path.basename(file.path);
     this.push(new gutil.File({
       path: compiledPagePaths[filename],
       contents: new Buffer(`${compiledTemplates.templateBegin}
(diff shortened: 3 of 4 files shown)


docs/development/quickstart.md
```diff
@@ -2,7 +2,7 @@
 
 # Development Quickstart
 
-Test Pilot uses Node.js [v6.x LTS](https://nodejs.org/dist/latest-v6.x/) for development. You may be able to get by using
+Test Pilot uses Node.js v6.2.0 for development. You may be able to get by using
```
chuckharmston: Danny updated this file yesterday in ece1ba14c1292c4f9763c7f8b08e7006e3e329d7; you may need to rebase against an updated master.

frontend/tasks/pages.js
```diff
@@ -1,6 +1,6 @@
 const gulp = require('gulp');
 const config = require('../config.js');
-
+const Path = require('path');
```
chuckharmston: Since the export of `path` is a module and not a class, we should lowercase it.

frontend/tasks/pages.js
```diff
@@ -95,7 +95,8 @@ function buildExperimentPage() {
 
 function convertToCompiledPage() {
   return through.obj(function compiledConvert(file, encoding, callback) {
-    const filename = file.path.split('/').pop();
+    //const filename = file.path.split('/').pop();
```
chuckharmston: No need to comment this out, if there are problems we can always check the commit log to find the previous version.

docs/development/quickstart.md
```diff
@@ -65,9 +65,34 @@ you should be on your way:
...
+```
+
+Now, open command prompt with admin privileges and run this:
+
+```cmd
+:: Add hostname alias to /etc/hosts and start up dev webserver
+echo 127.0.0.1 testpilot.dev >> %WINDIR%\System32\Drivers\Etc\Hosts
+```
+
+Go back to the previous command prompt window and run
```
chuckharmston, line 92: I'm a bit confused at why there are two command prompts here. Does it require admin privileges to add to hosts? Is there an easier way to do that than a new window??

docs/development/quickstart.md
```diff
@@ -65,9 +65,34 @@ you should be on your way:
...
+cd addon
+npm install
+npm run package-win
+
+:: Set up frontend web site environment
+cd ..
+npm install
+```
+
+Now, open command prompt with admin privileges and run this:
```
chuckharmston: Maybe `open a second command prompt window, this time with` would be clearer?

chuckharmston (changes requested):
This looks great, thanks for the contribution! 👍

lmorchard:
I can also give this a run through, since I have a Windows machine that I hack on at night

docs/development/quickstart.md
dumptyd: I don't think so. I mean there's `runas` command but it asks for credentials. See [this](http://stackoverflow.com/questions/8249705/how-to-run-an-application-as-run-as-administrator-from-the-command-prompt). 

In my opinion, opening a new window is the easiest way.

dumptyd:
@chuckharmston  I made the changes. :)

chuckharmston (approved):

chuckharmston:
I'll hold off on merging until @lmorchard tries it on Windows, but this looks great. Thanks so much for the contribution @dumptyd!

.eslintrc
```diff
@@ -37,3 +37,4 @@ rules:
   no-unused-expressions: [0]
   no-use-before-define: [2, nofunc]
   yoda: [0]
+  linebreak-style: [0]
```
dannycoates, line 40: I don't think we should be mixing line break styles in the source. Windows editors _can_ use LF instead of CRLF.

dumptyd: I don't know much about this, but I didn't change anything or even open a single file in the editor when I tried running `npm start` the first few times (so there's no chance of the editor accidentally changing LF to CRLF). But eslint kept throwing those errors.

dannycoates: Git is probably converting the endings automatically. We can leave this line in for now and fix it later with a `.gitattributes` file.

For reference see https://help.github.com/articles/dealing-with-line-endings/

lmorchard:
Sorry, this fell off my radar. Finally got around to running this on my windows machine, and all looks good!

PR #1707, opened 2016-10-27
[fixes #1678] Use 'location' instead of 'page' for pageviews

### LINKED ISSUE #1678: Update page location with the correct value for GA pageviews
This is an email sent to G support

> We are manually setting the `page` here: https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App.js#L72
> 
> Per the documentation here: https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#tracking_virtual_pageviews
> 
> Example:
> 1. visit https://testpilot.dev.mozaws.net/experiments/page-shot
>    dl: "https://testpilot.dev.mozaws.net/experiments/page-shot"
>    dp: "/experiments/page-shot"
> 2. scroll down on the page and click on the "Tracking Protection" tile.
>    dl: "https://testpilot.dev.mozaws.net/experiments/page-shot"
>    dp: "/experiments/tracking-protection"
> 
> Note: dp changes to the correct page, but dl stays the same.
> 
> When viewing this in the "Real Time" dashboard in GA, the "Active Page" doesn't always represent what is in `dp`. Sometimes it stays with what is in `dl` which is wrong.

This is the response from G

> Thanks for the additional information here. I guess first it's helpful to explain a little bit more about dl, dh, and dp:
> 
> ```
> dl: this parameter is for "Document Location" and takes the full url.
> dp: this parameter is for "Document Path" and takes just the path of the URL.
> dh: this parameter is for "Document Host" and takes just the hostname of the URL. 
> ```
> 
> For any single hit, you need to either have dl or dp AND dh set to determine the URL that you see in report. More information can be found here: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#content
> 
> In your case, I'm guessing the unique architecture of your single page site is causing GA to send conflicting location and path information. The simplest change I can imagine is to directly set "location" rather than "page" in your code, meaning that your virtual pageviews will send a modified dl parameter rather than both dp and the original unmodified dl parameter. More information can be found in the field reference here: https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#content

It may be necessary to also include any `utm` parameters in the `dl` field.

ckprice:
`window.location` has the correct URL host/path/query params in my testing.

Fixes #1678.

### DIFF
--- a/frontend/src/app/containers/App.js
+++ b/frontend/src/app/containers/App.js
@@ -70,7 +70,7 @@ class App extends Component {
     if (window.ga && ga.loaded) {
       const data = dataIn || {};
       data.hitType = type;
-      data.page = (pathname === '/') ? pathname : '/' + pathname;
+      data.location = window.location;
       ga('send', data);
     }
   }


frontend/src/app/containers/App.js
```diff
@@ -70,7 +70,7 @@ class App extends Component {
     if (window.ga && ga.loaded) {
       const data = dataIn || {};
       data.hitType = type;
-      data.page = (pathname === '/') ? pathname : '/' + pathname;
+      data.location = window.location;
```
chuckharmston, line 73: `window.location` would also contain the host. We want `window.location.pathname`, right?

chuckharmston: Separately, the right place to make this change is probably in `measurePageview`, rather than here:

``` js
const { pathname } = window.location;
```

chuckharmston (changes requested):

frontend/src/app/containers/App.js
ckprice: > `window.location` would also contain the host. We want `window.location.pathname`, right?

We need the host so GA can filter pings to the dev/stage/prod profiles. Previously the host was being sent implicitly through the `location` field. We're overriding location here since GA wasn't picking up the path correctly.

ckprice: Thoughts on ^ @chuckharmston? Wasn't sure if your `measurePageView` suggestion would still be valid knowing that we have to use the entire location field.

Issue #1463, opened 2016-09-28
Investigate Webpack


chuckharmston:
Our gulp build process is a bit gnarly and DIY, and some things (i.e. #1376, #1349) are much more challenging to do than it would be when using Webpack. Let's put together a list of requirements for our build system in a post-Django world, and do some informed compare/contrasting.
## Requirements
- Should be entirely contained within the `frontend` directory, as `addon` is.
- Development server
  - Should watch for changes and selectively recompile.
  - With hot reloading for React components.
  - With React and Redux dev tools.
- Assets
  - Should copy locale files, addon XPI and addon RDF to the appropriate locations.
  - Should convert content yaml to JSON and copy to the appropriate location.
  - Should convert legal md to HTML and copy to the appropriate location.
  - Should bundle CSS modules + SCSS.
  - Should bundle Javascript.
    - [Tree-shaking?](https://blog.engineyard.com/2016/tree-shaking)
  - Vendor modules: babel-polyfill, l20n
  - Should build source maps for JS and CSS
  - Should selectively include Pontoon.
  - Should ship with a base configuration and per-environment overrides using `NODE_ENV`
- Linting
  - SCSS and JavaScript
  - Should fail builds on linting errors, notify of warnings.
## Future
- Isomorphic; include experiments JSON blob in `index.html`, prevent need for API call.
## Proposed file structure
- addon
- frontend
  - bin
    - server.js
  - build
    - webpack.config.js
  - config
    - <env>.js
    - index.js
  - dist
    - static
      - (mirrors src)
    - app.css
    - app.js
    - index.html
    - vendor.js
  - server
  - src
    - actions
    - components
    - containers
    - lib
    - reducers
    - selectors
    - static
      - addon
        - addon.xpi
      - images
      - locales
    - app.js
    - index.html
    - routes.js
  - test
    - (mirrors src)

chuckharmston:
Started adding notes above. Would very much appreciate double-checking by people who know better.

chuckharmston:
Experimentalish thing going on over here: https://github.com/chuckharmston/testpilot/compare/master...chuckharmston:webpack-of-lies

chuckharmston:
I've got a (mostly) working port of this over here, with many ideas based on the [react-redux-starter-kit](https://github.com/davezuko/react-redux-starter-kit/): https://github.com/chuckharmston/testpilot-webpack

[...]

PR #1708, opened 2016-10-27
Dynamically generate data-l10n-id for content from YAML (closes #1340).

### LINKED ISSUE #1340: Figure out L10n for YAML flat file content
Moving to flat files & static site generation, we have content in YAML that we want localized. What do?

Here's an idea:
- Take the filename and property names in YAML as L10N string ID prefixes
- Hash the string content, use that as L10N string ID suffix
- Generate an en-us .ftl file from the YAML content files using those L10N string IDs
- Use those L10N string IDs in HTML page content, both in client-side React and static site generation.

So, for example:
- `content-src/experiments/activity-stream.yaml` has a `title` property with content "Activity Stream"
- the L10N string ID for that could be `activity-stream-title-bc6494e4f9589713c63fc6671e19ecd2`
- generate `.ftl` using that string ID
- include that string ID in `data-l10n-id` in markup (not sure how)

The idea here is that we have L10N IDs that naturally change when content changes, so that localizers using Pontoon see when there are new strings that need translation. That is, say the title changes to "Activity Stream is Awesome" - the L10N ID changes to `activity-stream-title-4a490ec33f2887a662f7c31ab29edec7`

Better ideas are totally welcome!

chuckharmston:
As discussed, this has been updated to allow for per-field l10n ID suffixes. So if the title of Activity Stream is updated to be plural again, the YAML can be updated to this:

``` yaml
title: 'Activity Streams'
title_l10nid: 'plural'
```

And its `data-l10n-id` will be changed from `activitystreamTitle` to `activitystreamTitlePlural`, invalidating all the translations. This also works for nested properties, e.g.:

``` yaml
contributors:
  -
    title: 'Staff Engineer'
    title_l10nid: 'staff'
```

Note: when this is merged, we will need to go through Pontoon and translate all the YAML content in English. This will always have to be done for new experiments.

### DIFF
--- a/content-src/experiments/activity-stream.yaml
+++ b/content-src/experiments/activity-stream.yaml
@@ -43,12 +43,10 @@ pre_feedback_image:
 pre_feedback_copy:
 details:
   -
-    headline: ' '
     image: /static/images/experiments/activity-stream/experiments_experimentdetail/1/6/16f2bfb82c63c38573005923f4e08457_image_1462827231_0751.png
     copy: >
       Click on a new tab, and your favorite sites are only one more click away.
   -
-    headline: ' '
     image: /static/images/experiments/activity-stream/experiments_experimentdetail/6/7/67858740793232e7a71f86b4ef272b67_image_1462827231_0657.png
     copy: >
       See where you've been, so you can get on to where you're going.
--- a/frontend/src/app/lib/utils.js
+++ b/frontend/src/app/lib/utils.js
@@ -1,5 +1,8 @@
 import querystring from 'querystring';
 
+import { experimentL10nId, l10nId, l10nIdFormat, lookup } from '../../../tasks/util';
+
+
 export function formatDate(date) {
   let out = '';
   const d = new Date(date);
@@ -42,10 +45,7 @@ export function isMinFirefoxVersion(isFirefox, ua, minVersion) {
 //
 // > l10nIdFormat('Activity Stream');
 // Activitystream
-export function l10nIdFormat(str) {
-  const segment = String(str).replace(/[^A-Za-z0-9]+/g, '');
-  return segment.charAt(0).toUpperCase() + segment.slice(1).toLowerCase();
-}
+export { l10nIdFormat };
 
 
 // Passed an array of pieces, returns a formatted l10n ID comprised of those
@@ -53,11 +53,7 @@ export function l10nIdFormat(str) {
 //
 // > l10nId(['Activity Stream', 'Contributor', 'Title', 'Bryan Bell']);
 // 'activitystreamContributorTitleBryanbell'
-export function l10nId(pieces) {
-  pieces[0] = l10nIdFormat(pieces[0]);
-  const stitched = pieces.reduce((a, b) => `${a}${l10nIdFormat(b)}`);
-  return stitched.charAt(0).toLowerCase() + stitched.slice(1);
-}
+export { l10nId };
 
 
 // Passed three pieces of information, generates and returns an appropriate l10n
@@ -65,27 +61,11 @@ export function l10nId(pieces) {
 //
 // Params:
 // - experiment - the JavaScript object representing the experiment.
-// - pieces - an array of segments comprising the generated l10n ID. If a string
-//            is passed, it's converted to a single-item array.
-// - path - an optional string representing the period-delimited path to the
-//          property of the experiment that the l10n ID references. This is used
-//          to look for a _l10nid version of that field, which is used as a
-//          suffix if present. If omitted, the first item in `pieces` is used.
-export function experimentL10nId(experiment, pieces, path = null) {
-  if (typeof pieces === 'string') {
-    pieces = [pieces];
-  }
-  if (!path) {
-    path = pieces[0];
-  }
-  if (path) {
-    const suffix = lookup(experiment, `${path}_l10nid`);
-    if (suffix) {
-      pieces.push(suffix);
-    }
-  }
-  return l10nId([].concat([experiment.slug], pieces));
-}
+// - pieces - an array of segments used for two purposes:
+//            1) to construct the generated l10n ID
+//            2) to search for an _l10n suffix for the field.
+//            If a string is passed, it's converted to a single-item array.
+export { experimentL10nId };
 
 
 // Passed an object and a period.delimited string indicating a path, returns the
@@ -96,10 +76,4 @@ export function experimentL10nId(experiment, pieces, path = null) {
 //
 // >> lookup({}, 'foo');
 // null
-export function lookup(obj, path) {
-  const pieces = path.split('.');
-  if (pieces.length > 1) {
-    return lookup(obj[pieces[0]], pieces.slice(1).join('.'));
-  }
-  return obj[pieces[0]] || null;
-}
+export { lookup };
--- a/frontend/src/app/lib/utils.js
+++ b/frontend/src/app/lib/utils.js
@@ -34,3 +34,72 @@ export function isMinFirefoxVersion(isFirefox, ua, minVersion) {
   if (!isFirefox) return false;
   return parseInt(ua.split('/').pop(), 10) >= minVersion;
 }
+
+
+// Passed a string, returns a verison of that string sanitized by stripping all
+// non-alphanumeric characters, lowercasing the entire string, then capitalizing
+// the first character.
+//
+// > l10nIdFormat('Activity Stream');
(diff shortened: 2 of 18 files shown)


chuckharmston:
The alternative I was trying to describe is to add a l10n-id field in the YAML for each field. If present, it would be used in the same way that `l10n_version` is used here, but only for that field. The problem is that that's another thing to send over XHR, another piece of state we need to hold, and another thing we need to take into account when looking at changes to the DOM. It would have a noticeable impact on performance.

``` yaml
description: My awesome new description.
description-l10nid: 2016-10-28
```

dannycoates:
original discussion #1340

dannycoates:
> The problem is that that's another thing to send over XHR, another piece of state we need to hold, and another thing we need to take into account when looking at changes to the DOM

Seem like that's the price of admission if we don't want to hash the content and be able to change individual fields without invalidating the whole thing.

> It would have a noticeable impact on performance

How much?

chuckharmston:
Code and PR have been updated to reflect earlier discussions. This is now ready for review.

dannycoates:
I see how this takes care of populating the `data-l10n-id` attributes, but how are we planning on handling the translation files? Is there going to be like `/locales/en-US/tab-center.ftl` or `/locales/en-US/experiments.ftl`? Also how will we populate the default one, generate it from yaml?

chuckharmston:
As I understand it, Pontoon will allow anything with the data-l10n-id attribute to be localized, even if it doesn't have an existing in the .ftl.

dannycoates:
> As I understand it, Pontoon will allow anything with the data-l10n-id attribute to be localized, even if it doesn't have an existing in the .ftl.

Sure, but how do we consume those translations on the site? We have to point to something, right? Like: `<link rel="localization" href="/static/locales/{locale}/app.ftl">` Will all the experiment translations go into `app.ftl`?

lmorchard:
From @chuck:

> Note: when this is merged, we will need to go through Pontoon and translate all the YAML content in English. This will always have to be done for new experiments.

From @dannycoates:

> how are we planning on handling the translation files? Is there going to be like /locales/en-US/tab-center.ftl or /locales/en-US/experiments.ftl? Also how will we populate the default one, generate it from yaml?

FWIW, what I had in mind for #1340 was to auto-generate an en-us `experiments.ftl` from the YAML content files using the L10N string IDs - i.e. as a task in gulp. Manual translation in Pontoon would be redundant, because the original content is already localized to en-US.

chuckharmston:
> FWIW, what I had in mind for #1340 was to auto-generate an en-us experiments.ftl from the YAML content files using the L10N string IDs - i.e. as a task in gulp. Manual translation in Pontoon would be redundant, because the original content is already localized to en-US. 

I don't think there's anything stopping us from doing that.

lmorchard:
> Sure, but how do we consume those translations on the site? We have to point to something, right? Like: <link rel="localization" href="/static/locales/{locale}/app.ftl"> Will all the experiment translations go into app.ftl?

I think L20N is okay with multiple `<link rel="localization">` references, so I figured we could also have `<link rel="localization" href="/static/locales/{locale}/experiments.ftl">` alongside the existing `app.ftl`. How that interacts with Pontoon, though, I'm not sure. That is, would Pontoon commit changes to the appropriate FTL file (`app.ftl` or `experiments.ftl`)

lmorchard:
> I don't think there's anything stopping us from doing that.

Does that mean you plan to do the gulp task as part of this PR? If not, we should keep #1340 open and plan the gulp task as follow-on work.

chuckharmston:
> Does that mean you plan to do the gulp task as part of this PR? If not, we should keep #1340 open and plan the gulp task as follow-on work.

I don't have plans for that, no, but nothing here precludes it. I wouldn't want to invest time into anything related to gulp until we've figured out if we even want to keep it (#1463)

lmorchard:
> I wouldn't want to invest time into anything related to gulp until we've figured out if we even want to keep it 

We're keeping it so far, and we need this l10n stuff before we have time to really move significantly on Webpack - i.e. this sprint or asap. And whatever we do for generating the ftl can probably be ported over to webpack, even if only as a separate script or reusable lib

lmorchard:
Also FWIW - we do need _something_ for the en-US .ftl after merging this PR, because _all_ content for experiments gets replaced with l10n IDs when the strings are missing. I'd rather this were done with an automated task rather than have us copypasta all the YAML strings by hand

screenshot 2016-10-31 15 59 54

chuckharmston:
Yep, covered that in comment 0:

> Note: when this is merged, we will need to go through Pontoon and translate all the YAML content in English. This will always have to be done for new experiments.

lmorchard (changes requested):
TL;DR: We can't ship this as-is, because all the experiment strings are missing from the en-US locale and L20n just replaces content with l10n-ids. 

So I'd want to see an en-US .ftl file ship along with this PR, ideally generated from the YAML via script or build task. Copying & pasting the content into Pontoon is just a roundabout way to get to the same place, IMO (i.e. an en-US ftl file).

lmorchard:
> Yep, covered that in comment 0:

Yes, I know. And I'm saying we need to fix that as part of this PR, or in this sprint / before next deploy to close out #1340.

content-src/experiments/activity-stream.yaml
```diff
@@ -43,12 +43,10 @@ pre_feedback_image:
 pre_feedback_copy:
 details:
   -
-    headline: ' '
```
chuckharmston, line 46: Deleted all these empty headlines, which seem to be an artifact from Django.

frontend/src/app/lib/utils.js
```diff
@@ -1,5 +1,8 @@
 import querystring from 'querystring';
 
+import { experimentL10nId, l10nId, l10nIdFormat, lookup } from '../../../tasks/util';
```
chuckharmston, line 3: In order to reuse this logic in the gulp tasks, I had to move these to their own module that doesn't require babelification. I imported them here for testing and other purposes.

frontend/test/app/containers/ExperimentPage-test.js
```diff
@@ -42,7 +42,7 @@ describe('app/components/ExperimentPage:ExperimentDetail', () => {
     mockExperiment = {
       slug: 'testing',
       title: 'Testing',
-      title_l10nid: 'foo',
+      title_l10nsuffix: 'foo',
```
chuckharmston, line 45: Renamed to `_l10nsuffix` per @clouserw request.

chuckharmston:
Updated to include the FTL generation task. [Example output](https://gist.github.com/chuckharmston/9d78bf95e7082a6b8a7617ad2182dbcf).

This is ready for review again.

dannycoates:
When reordering a list like contributors, it doesn't look obvious that I'd need to add a suffix to `title` when the text isn't changing.

How do we plan on communicating that changing the order of a list requires adding or changing an `_l10nsuffix` to translated strings within those objects? (just reviewing PRs?)

lmorchard (approved):
:+1: LGTM! Tests are thorough & pass, all content looks good, code is good. Looks like it needs a rebase, but good to go otherwise

frontend/src/index.html
```diff
@@ -9,6 +9,7 @@
     <meta name="availableLanguages" content="en-US">
     <meta name="viewport" content="width=device-width">
     <link rel="localization" href="/static/locales/{locale}/app.ftl">
+    <link rel="localization" href="/static/locales/{locale}/experiments.ftl">
 
     <title>Firefox Test Pilot</title>
```
lmorchard: Yeah, I think we realized today that this file doesn't even need to exist :/

chuckharmston:
> When reordering a list like contributors, it doesn't look obvious that I'd need to add a suffix to title when the text isn't changing.

I'm not really sure. My original PR had the ability to replace the index with a slug of sorts (e.g. contributor name), but there wasn't a reliable way to do that and also generate keys in both the Gulp task and component.

lmorchard:
Hmm, the reordering concern is interesting, though maybe more a doco issue? 

I had at one point considered using a hash of the original string as l10nID as default suffix, but that seemed possibly fraught with its own issues

lmorchard:
Also maybe we just try to avoid reordering as much as possible. Say we want to add a contributor and the list was originally in alpha order - just tack the name onto the end and sort on the client side? (Of course then we have to track the original index on the client vs the new sort order, but bleh)

PR #1709, opened 2016-10-27
[fixes #1461] Add ID to Survey Gizmo URLs

### LINKED ISSUE #1461: Send the `clientUUID` in the URL to SurveyGizmo
Request from rrayborn

> While I wasn't actively involved with the discussion, it's my understanding that there was talk of passing the user's UUID into SG earlier on in the TxP project for later correlations.
> 
> My team is still interested in this and it would be a good test for potential similar experiments in Firefox desktop.  Would this be something that we could try?  Is there an existing bug that I'm not aware of?

ckprice:
Would like verification from someone else that the add-on heartbeat survey is populating correctly.

Fixes #1461.

### DIFF
--- a/addon/lib/survey.js
+++ b/addon/lib/survey.js
@@ -121,6 +121,7 @@ function launchSurvey(options) {
             if (clicked) {
               const urlParams = {
                 id: experiment.addon_id,
+                cid: store.clientUUID,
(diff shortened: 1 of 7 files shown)


chuckharmston (approved):
Looks and works perfectly! Great job, @ckprice.

Issue #1698, opened 2016-10-27
Wrong cards in list after clicking a card


dannycoates:
With no addon, from an experiment page clicking one of the cards navigates to the new experiment but it remains in the 'other experiments' section.

SoftVision-PaulOiegas:
Hey Danny, this probably a dupe of #1518. If you refresh the page after each selected card, the right content is displayed.

dannycoates:
> this probably a dupe of #1518.

I think this is different. 

It looks like this is another side effect of the `TypeError: t is null` reported in #1696

The error appears to be caused by a bad interaction between react and l20n. When I remove the `data-l10n-id` from the installation counter [span](https://github.com/mozilla/testpilot/blob/9fb88ef777019d6a9b43cda2735e8833d7d94feb/frontend/src/app/containers/ExperimentPage.js#L457) the error goes away, GA sends the pageview, and the card list renders correctly.

@lmorchard @chuckharmston any ideas how to work around these two libs fighting over the dom?

PR #1712, opened 2016-10-28
fixed l20n/react conflict on user count text

### LINKED ISSUE #1696: Console error in `measurePageview` when experiment cards are clicked from an experiment detail page
STR:
1. Uninstall Test Pilot
2. Visit https://testpilot.firefox.com/experiments/tab-center
3. Scroll down to the bottom and click on the "Page Shot" card.

Expected Result:

User is taken to the Page Shot experiment page, and a `pageview` is sent.

Actual Result:

User is taken to the Page Shot experiment page, no `pageview` is sent, and the following console error is served

> TypeError: t is null

Observations:
- This only seems to happen when you click on the "footer" cards on an experiment details page. If you click on a card from the homepage, it works as expected.
- After a bit of digging, [`this.props` appears to be undefined](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App.js#L27) in the function when clicking a card from an experiment detail page. /edit on a closer look, this appears to be erroring all the way back to `ExperimentPage.js`.

dannycoates:
This is a semi-hacky way to fix #1696 and #1698 

The problem is that react adds comment nodes that it relies on later for diffing that l20n clobbers when it translates. 

screen shot 2016-10-27 at 4 33 03 pm

By not rendering the text in react (just the elements) the replacement by l20n matches react's virtual dom close enough to not cause problems.

Having l20n and react both modify the dom independently isn't ideal. In the future we ought to look into a more integrated approach. This thread might have some leads: https://groups.google.com/forum/#!topic/mozilla.tools.l10n/XtxHgBEokCA

### DIFF
--- a/frontend/src/app/containers/ExperimentPage.js
+++ b/frontend/src/app/containers/ExperimentPage.js
@@ -471,8 +471,10 @@ export class ExperimentDetail extends React.Component {
     }
     const installCount = installation_count.toLocaleString();
     return (
-      <span data-l10n-id="userCountContainer" data-l10n-args={JSON.stringify({ installation_count: installCount, title })}>There are <span data-l10n-id="userCount" className="bold">{installCount}</span>
-      people trying {title} right now!</span>
+      // Note: this doesn't include the text content because of a conflict
+      // in how l20n and react modify the dom.
+      // https://github.com/mozilla/testpilot/pull/1712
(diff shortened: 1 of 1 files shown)


chuckharmston (approved):
Eeesh, this really isn't ideal. But it works.

R+, but I'll let you self-merge after you file a new issue to find that more integrated approach.

PR #1713, opened 2016-10-28
Add Self-Destructing Cookies to incompatible add-on list

ianb:
While we have some workarounds for Self-Destructing Cookies, by wiping cookies it breaks My Shots and generally makes things difficult.  SInce incompatible add-ons only give a warning, it seems appropriate to add the add-on there.


chuckharmston (approved):
Feel free to self-merge when CI finishes.

PR #1637, opened 2016-10-20
Part 1 of addon sending pings to webapp

dannycoates:
Part 1 of 2 fixing #1627 as described in #1630

PR #1715, opened 2016-10-28
part 2 of detecting addon uninstallation

### LINKED ISSUE #1627: Test Pilot uninstall is not detected across all tabs

dannycoates:
Follow up to #1637 fixes #1627

### DIFF
--- a/addon/data/set-installed-flag.js
+++ b/addon/data/set-installed-flag.js
@@ -10,8 +10,3 @@
 
 unsafeWindow.navigator.testpilotAddon = true;
 unsafeWindow.navigator.testpilotAddonVersion = self.options.version;
-
-self.port.on('detach', function() {
-  delete unsafeWindow.navigator.testpilotAddon;
-  delete unsafeWindow.navigator.testpilotAddonVersion;
-});
--- a/addon/package.json
+++ b/addon/package.json
@@ -1,7 +1,7 @@
 {
   "title": "Test Pilot",
   "name": "testpilot-addon",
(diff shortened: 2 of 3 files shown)


lmorchard (approved):
Hmm, the one thing I found about this is that it behaves rather badly with a very old version of the add-on. It ping-pongs between installed and uninstalled every 3 seconds. That was expected, though, and actually why we broke this up into two parts.

But, I wonder if maybe we can peek at `window.navigator.testpilotAddonVersion` and not enable the disconnect timer unless the add-on is definitely past a certain version?

Maybe I'm being extra paranoid? Going to approve this and say it might be too edgy to care about? If you think so to, go ahead & merge

dannycoates:
> I wonder if maybe we can peek at window.navigator.testpilotAddonVersion and not enable the disconnect timer unless the add-on is definitely past a certain version?

👍 I went ahead and bumped minor version for this.

PR #1718, opened 2016-10-31
add dev flag; fixes 1690

clouserw:
/cc @relud -- can you confirm our dev environment has this flag on it now so we can deploy this without hiccuping?  Thanks.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -3,7 +3,12 @@
 # This file is used to deploy Test Pilot to an S3 bucket.  It expects to be run
 # from the root of the Test Pilot directory and you'll need your S3 bucket name
 # in an environment variable $TESTPILOT_BUCKET
-
+#
+# It takes a single argument:  If you're deploying to a dev instance, pass in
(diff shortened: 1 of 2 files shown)


relud:
I don't own the testpilot dev environment. I think @lmorchard does.

lmorchard:
Yeah, dev is owned by devs on Cloud Services AWS - hopefully not just me, bus factor and all that. It's deployed via circle.yml and [`deploy.sh` gets run here](https://github.com/mozilla/testpilot/blob/master/circle.yml#L44). Should able to add the `dev` param there

lmorchard (changes requested):
Realized my last comment wasn't a change request - you should be able to wrap this up with adding the `dev` param to [the deploy.sh call in circle.yml](https://github.com/mozilla/testpilot/blob/master/circle.yml#L44)

ghost:
Thanks @lmorchard . new patch is pushed

chuckharmston (approved):

PR #1720, opened 2016-10-31
Fix typo in the command that watches frontend/src/index.html.

fzzzy:


padilla360:
it was very smooth opening up and transitioning over form chrome.  I was hesitant at first, once I hit the Enable button it was and is a whole new experience for me.  Thank You Team!

PR #1725, opened 2016-11-01
[fixes #904] Add Optimizely

### LINKED ISSUE #904: Add Optimizely to Test Pilot.
This issue is created to just add Optimizely to the website.

Optimizely has some [developer notes on single page applications](https://help.optimizely.com/Build_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications). It looks like there's a bit of special things we have to do on both the app and in the Optimizely console.

@johngruen as a first step, could you provide the Optimizely snippet?

ckprice:
This PR does ~~two~~ three things - one optional, two required - for Optimizely integration (in addition to adding the code snippet)

(1) Set JavaScript variables to explicitly set the current page state ([optional:ref](https://help.optimizely.com/Build_Campaigns_and_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications_in_Optimizely_Classic#1._Identify_states_in_your_Single_Page_App)).

As noted in the docs, we can scrape the page's DOM elements to determine page state in lieu of setting these variables. For instance, in the Optimizely dashboard, if our targeting has `document.getElementById('feedback-button')` we would know that the page state is most likely an experiment page.

The environment variables give us a pretty certain way to determine page state now and going forward. But again, this is optional, so I would welcome input here on removing. There may be another way we can determine this that I missed.

(2) Send a custom event to Optimizely when things happen on the page ([required:ref](https://help.optimizely.com/Build_Campaigns_and_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications_in_Optimizely_Classic#3._Use_Custom_Event_Goals)).

Piggy-backing on the `sendToGA` function used for event tracking seemed like the best way to go.

(3) Update CSP to give Optimizely framing rights ([required:ref](https://help.optimizely.com/Troubleshoot_Problems/Update_your_site's_Content_Security_Policies_(CSP))).

(this one got a bit sticky)

Updated the prod/stage CSP rule to include the `frame-src` directive and add Optimizely.

Updated the dev CSP rule to include Optimizely in the `frame-src` directive.

I removed the `X-FRAME-OPTION` header because 1. [it looks to be deprecated](https://www.owasp.org/index.php/Content_Security_Policy_Cheat_Sheet#Preventing_Clickjacking), and 2. multiple URLs aren't fully supported.

It seems as though the `frame-ancestors` directive is the preferred way to add this, but there's a note that you can't use it in a `<meta>` tag, which - from what I gathered [in `deploy.sh`](https://github.com/mozilla/testpilot/blob/master/bin/deploy.sh#L72) - is what we are doing. If it's not, I can update this to use `frame-ancestors` instead.

@clouserw noted you can't test this stuff locally so I couldn't verify that this change made the Optimizely dashboard work, but we can in dev. Very much open to opinions here as I haven't done much CSP stuff before.

Fixes #904.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -34,18 +34,16 @@ fi
 MAX_AGE="600"
 
 HPKP="\"public-key-pins\": \"max-age=300;pin-sha256=\\\"WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18=\\\";pin-sha256=\\\"r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E=\\\"\""
-CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
+CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org https://*.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self' https://*.optimizely.com; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
 HSTS="\"strict-transport-security\": \"max-age=31536000; includeSubDomains; preload\""
 TYPE="\"x-content-type-options\": \"nosniff\""
-FRAME="\"x-frame-options\": \"DENY\""
 XSS="\"x-xss-protection\": \"1; mode=block\""
 
 # Our dev server has a couple different rules to allow easier debugging and
 # enable localization.  Also expires more often.
 if [ "$DEST" = "dev" ]; then
     MAX_AGE="15"
-    CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-src 'self' https://pontoon.mozilla.org; img-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com; style-src 'self' https://pontoon.mozilla.org code.cdn.mozilla.net; report-uri /__cspreport__;\""
-    FRAME="\"x-frame-options\": \"ALLOW-FROM https://pontoon.mozilla.org/\""
+    CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org https://*.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self' https://pontoon.mozilla.org https://*.optimizely.com; img-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com; style-src 'self' https://pontoon.mozilla.org code.cdn.mozilla.net; report-uri /__cspreport__;\""
 fi
 
 # build version.json if it isn't provided
@@ -69,7 +67,7 @@ aws s3 sync \
   --content-type "text/html" \
   --exclude "*" \
   --include "*.html" \
-  --metadata "{${HPKP}, ${CSP}, ${HSTS}, ${TYPE}, ${FRAME}, ${XSS}}" \
+  --metadata "{${HPKP}, ${CSP}, ${HSTS}, ${TYPE}, ${XSS}}" \
   --metadata-directive "REPLACE" \
   --acl "public-read" \
   dist/ s3://${TESTPILOT_BUCKET}/
@@ -135,7 +133,7 @@ for fn in $(find dist -name 'index.html' -not -path 'dist/index.html'); do
     --content-type "text/html" \
     --exclude "*" \
     --include "*.html" \
-    --metadata "{${HPKP}, ${CSP}, ${HSTS}, ${TYPE}, ${FRAME}, ${XSS}}" \
+    --metadata "{${HPKP}, ${CSP}, ${HSTS}, ${TYPE}, ${XSS}}" \
     --metadata-directive "REPLACE" \
     --acl "public-read" \
     $fn s3://${TESTPILOT_BUCKET}/${s3path}
--- a/frontend/src/app/containers/App.js
+++ b/frontend/src/app/containers/App.js
@@ -36,6 +36,10 @@ class App extends Component {
 
     const experimentsPath = 'experiments/';
 
+    // These are used to expose page state for Optimizely, testing, etc.
+    window.testPilotPathname = pathname;
+    window.testPilotHasAddon = hasAddon;
+
     if (pathname === '/') {
       const installedCount = Object.keys(installed).length;
       const anyInstalled = installedCount > 0;
@@ -95,12 +99,19 @@ function sendToGA(type, dataIn) {
       document.location = data.outboundURL;
     }
   };
+  sendToOptimizely(data.eventCategory + '/' + data.eventAction + '/' + data.eventLabel);
   if (window.ga && ga.loaded) {
     data.hitType = type;
     data.hitCallback = hitCallback;
     ga('send', data);
   } else {
     hitCallback();
+  }
+}
+
+function sendToOptimizely(evt) {
+  if (window.optimizely) {
+    window.optimizely.push(['trackEvent', evt]);
   }
 }
 
--- a/frontend/src/app/lib/optimizely-snippet.js
+++ b/frontend/src/app/lib/optimizely-snippet.js
@@ -1,0 +1,18 @@
+/*eslint-disable*/
+// HACK: Optimizely lives here, because CSP won't let it live inline
+
+import config from '../config';
+
+if (config.optimizelyEnabled) {
+	(function() { 
+	    var projectId = '5941720679';
+	    var scriptTag = document.createElement('script');
+	    scriptTag.type = 'text/javascript';
+	    scriptTag.async = true;
+	    scriptTag.src = 'https://cdn.optimizely.com/js/' + 
+	    projectId + '.js';
+	    var s = document.getElementsByTagName('script')[0];
+	    s.parentNode.insertBefore(scriptTag, s);
+	})();
+}
+/*eslint-enable*/
--- a/frontend/src/app/config.js
+++ b/frontend/src/app/config.js
@@ -3,7 +3,8 @@ const defaultConfig = {
   minFirefoxVersion: 45,
   experimentsURL: '/api/experiments.json',
   usageCountsURL: 'https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json',
-  ravenPublicDSN: 'https://5ceef9a20a6e4cdd93cc9a935be78c73@sentry.prod.mozaws.net/169'
+  ravenPublicDSN: 'https://5ceef9a20a6e4cdd93cc9a935be78c73@sentry.prod.mozaws.net/169',
+  optimizelyEnabled: true
 };
 
 const hostname = (typeof window === 'undefined') ? '' : window.location.hostname;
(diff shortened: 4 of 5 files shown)


ckprice:
Looks like I may have to propose more [CSP updates](https://help.optimizely.com/Troubleshoot_Problems/Update_your_site's_Content_Security_Policies_%28CSP%29) to get this to play right with the Optimizely dashboard (comments also welcomed here). Blocking this from merging, but the non-CSP bits are ready for review.

bin/deploy.sh
```diff
@@ -25,7 +25,7 @@ fi
...
+CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org https://cdn.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-src 'self' https://pontoon.mozilla.org; img-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com; style-src 'self' https://pontoon.mozilla.org code.cdn.mozilla.net; report-uri /__cspreport__;\""
```
ghost: There are now two CSP lines in this file (one for -dev).  You'll need to update them both

ckprice: @wresuolc thanks! Do CSP updates in `deploy.sh` affect your local environment too?

ghost: They only would if you were deploying to S3 and using that.  If you're on our local box they will not.

jaredhirsch:
Adding stable page IDs in the DOM would make integration testing a lot easier. :+1: to that idea

ckprice:
Rebased with the master `deploy.sh`, and added more CSP context to comment 0 that's ready for review.

ghost:
> I removed the X-FRAME-OPTION header because 1. it looks to be deprecated, 

The new way only works in browsers that support CSP 2.0.  The security team is still recommending we use x-frame-options (https://wiki.mozilla.org/Security/Guidelines/Web_Security#X-Frame-Options ).  r? @jvehent for a recommendation here

ckprice:
> The new way only works in browsers that support CSP 2.0. The security team is still recommending we use x-frame-options

Okay. Maybe @jvehent could also comment if multiple uris are supported with `X-FRAME-OPTIONS`. I couldn't find anything confirming this.

ckprice:
Hmm, found this on [that Wiki](https://wiki.mozilla.org/Security/Guidelines/Web_Security#X-Frame-Options)

> ALLOW-FROM uri: deprecated; instead use CSP's frame-ancestors directive

chuckharmston:
Can we add a configuration flag to not include Optimizely until we know we're going to use it?

``` js
if (config.isDev || config.optimizelyEnabled) {
  import './lib/optimizely-snippet';
}
```

As is, we'd be including a third-party library for no reason.

frontend/src/app/lib/optimizely-snippet.js
```diff
@@ -0,0 +1,15 @@
+/*eslint-disable*/
+// HACK: Optimizely lives here, because CSP won't let it live inline
+(function() { 
+    var projectId = '5941720679';
+    var protocol = ('https:' == document.location.protocol ? 
+    'https://' : 'http://');
+    var scriptTag = document.createElement('script');
+    scriptTag.type = 'text/javascript';
+    scriptTag.async = true;
+    scriptTag.src = protocol + 'cdn.optimizely.com/js/' +
```
chuckharmston: We should hardcode `https` here.

bin/deploy.sh
```diff
@@ -34,18 +34,16 @@ fi
...
-CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
+CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org *.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-src 'self' *.optimizely.com; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
```
chuckharmston: We should also hardcode `https` into the CSP.

frontend/src/app/containers/App.js
```diff
@@ -95,6 +103,11 @@ function sendToGA(type, dataIn) {
       document.location = data.outboundURL;
     }
   };
+  // Duplciate the event to Optimizely
```
chuckharmston: Typo

frontend/src/app/containers/App.js
```diff
@@ -65,8 +65,16 @@ class App extends Component {
   }
 
   debounceSendToGA(pathname, type, dataIn) {
+    const { hasAddon } = this.props;
+
     if (this.lastPingPathname === pathname) { return; }
     this.lastPingPathname = pathname;
+    // Keep an environment variables to identify page state
+    // https://help.optimizely.com/Build_Campaigns_and_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications_in_Optimizely_Classic#1._Identify_states_in_your_Single_Page_App
+    if (window.optimizely) {
```
chuckharmston: Can we move this to its own function, rather than stick it into GA?

chuckharmston:
> could also comment if multiple uris are supported with `X-FRAME-OPTIONS`

I believe you can send multiple `X-Frame-Options` headers with a single request.

frontend/src/app/containers/App.js
```diff
@@ -65,8 +65,16 @@ class App extends Component {
   }
 
   debounceSendToGA(pathname, type, dataIn) {
+    const { hasAddon } = this.props;
+
     if (this.lastPingPathname === pathname) { return; }
     this.lastPingPathname = pathname;
+    // Keep an environment variables to identify page state
+    // https://help.optimizely.com/Build_Campaigns_and_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications_in_Optimizely_Classic#1._Identify_states_in_your_Single_Page_App
```
chuckharmston: From this page:

> This article is about Optimizely Classic. Optimizely X contains improved support for Single Page Applications.

Does this still apply? Should we be using Optimizely X if we aren't already?

g-k:
Hey, @jvehent asked me to comment here since I've been doing some CSP stuff lately.

We still recommend using X-Frame-Options for older browsers:

> As frame-ancestors is not yet supported in IE11 and older, Edge, Safari 9.1 (desktop), and Safari 9.2 (iOS), it is recommended that sites employ X-Frame-Options in addition to using CSP.

http://caniuse.com/#search=csp
http://caniuse.com/#search=x-frame

You can't have multiple URIs in an ALLOW-FROM.  https://tools.ietf.org/html/rfc7034#section-2.3.2.3 recommends returning a different origin in the header based on a query param. That's kind of gross.  I'll look into returning multiple X-Frame-Options headers with different origins is an option.  Even if we can return multiple headers the wildcard for `*.optimizely.com` won't be a valid  serialized origin https://tools.ietf.org/html/rfc6454#section-7.1.

ckprice:
Let's not land this this sprint (Sprint 13), and spend some time making sure the CSP stuff is good on dev.

ckprice:
Yeah, multiple URIs is a requirement. On dev, we want to allow framing from Pontoon and Optimizely. And I think Optimizely serves from a couple subdomains, hence the wildcard.

> I'll look into returning multiple X-Frame-Options headers with different origins is an option. 

Thanks @g-k, looking forward to hearing options.

frontend/src/app/containers/App.js
ckprice: We currently are not using Optimizely X. Upgrading costs $, and is controlled by marketing. Don't think we'd be a driver for them to upgrade.

ckprice:
@chuckharmston's suggestions are all in.

> Adding stable page IDs in the DOM would make integration testing a lot easier. 👍 to that idea

These are global variables `window.testPilotPathname` and `window.testPilotHasAddon`. So if someone finds them useful they will be available outside of just Optimizely usage.

For now, `X-FRAME-OPTIONS` is still removed, and this bug is blocked based on https://github.com/mozilla/testpilot/pull/1725#issuecomment-258457298.

g-k:
OK I didn't find anything suggesting multiple headers with different URIs would be honored and OWASP seems pretty clear:

>  Multiple options not supported There is no way to allow the current site and a 3rd party site to frame the same response -- browsers only honour one X-Frame-Options header and only one value on that header. 

https://www.owasp.org/index.php/Clickjacking_Defense_Cheat_Sheet#Limitations_2

I also I don't know if Pontoon and Optimizely would let us pass a query param.

chuckharmston:
> These are global variables `window.testPilotPathname` and `window.testPilotHasAddon`. So if someone finds them useful they will be available outside of just Optimizely usage.

I'm not thrilled about bleeding scope from the app, but need to spend some time understanding how we'd use it before offering suggestions.

g-k:
I was thinking the best we could do for older browsers is remove X-Frame-Options and fall back to a frame busting snippet with a referrer check (like https://www.nczonline.net/blog/2013/04/16/getting-the-url-of-an-iframes-parent/).

But dveditz said we could dynamically add XFO based on referrer (which would work in case JS is blocked).  We can change nginx to include the right XFO for whitelisted origins and `Vary: Referer` from the CDN.

Lets leave XFO deny in S3 and change nginx to overwrite it when we see a referrer whitelisted in CSP frame-ancestors.

cc @jvehent if that sounds reasonable

Edit: After looking at the traffic numbers, it's not worth the effort since almost all of the traffic comes from browsers that support CSP frame-ancestors.

g-k:
How much traffic to testpilot is from non-Firefox browsers and older browsers?

ckprice:
From @g-k 

> I think we can safely drop the X-Frame-Options header.
> 
> Given that:
> - testpilot is primarily for Firefox users
> - >98% of traffic comes from browsers that support CSP frame-ancestors
> - XFO is deprecated
> - testpilot doesn't expose functionality to non-FF users beyond
>   subscribing to the testpilot email list and downloading Firefox
>   (stop me if I'm wrong about that), so successful clickjacking
>   attacks on IE and Edge users could at most trick them into learning
>   about or downloading a better browser

The only question I have regarding `frame-ancestors` is a note from [the spec on MDN](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#frame-ancestors):

> This directive is not supported in the `<meta>` element or by the Content-Security-Policy-Report-Only header field.

@lmorchard @chuckharmston can you confirm whether or not we are using a `<meta>` element? Judging from [the variable names in `deploy.sh`](https://github.com/mozilla/testpilot/blob/master/bin/deploy.sh#L72) it looks like we may be.

ghost:
We are not using a `<meta>` tag with CSP.  (We're using headers)

ckprice:
These updates are in and ready for review. I'm going to keep the `blocked` label on until after we push Sprint 13 to production. I'd like to let this stay on dev for a bit while I confirm it's working with the Optimizely dashboard.

ckprice:
Ready for review/merge.

lmorchard:
> Can we add a configuration flag to not include Optimizely until we know we're going to use it?
> 
> ``` js
> if (config.isDev || config.optimizelyEnabled) {
>   import './lib/optimizely-snippet';
> }
> ```
> 
> As is, we'd be including a third-party library for no reason.

FWIW, this doesn't work in ES6 - `import` is a static / compile-time thing and I think a syntax error when found in a conditional. But, [there is a flag checked within that module](https://github.com/mozilla/testpilot/pull/1725/files#diff-aabca5d9bea52b94fee07360ebacca98R6), so I think we're good?

ckprice:
> But, there is a flag checked within that module, so I think we're good?

I thought the idea was good, but yea, it wouldn't let me add the check there. I moved it to the snippet as you noted.

frontend/src/app/containers/App.js
```diff
@@ -36,6 +36,10 @@ class App extends Component {
 
     const experimentsPath = 'experiments/';
 
+    // These are used to expose page state for Optimizely, testing, etc.
```
lmorchard, line 39: How are these used by Optimizely, exactly? 

Wondering in part because these variables won't get set immediately on page load, because of the conditions that come before. Namely, experiment data has to be loaded, and then if the add-on is installed it will wait for the installed add-ons list to be received.

I see there's a `sendToOptimizely` call later... is that where these get picked up?

lmorchard:
I think this looks okay. I don't know a whole lot about Optimizely myself to see what all it does. But seems like this is limited to testpilot.dev.mozaws.net for now, so we can play with it there and back it out if we have to

frontend/src/app/containers/App.js
ckprice: See [here](https://developers.optimizely.com/classic/javascript/code-samples/index.html#conditional-activation-js-variable) (linked to from [the SPA documentation](https://help.optimizely.com/Build_Campaigns_and_Experiments/AngularJS,_Backbone.js,_React,_and_Single_Page_Applications_in_Optimizely_Classic#1._Identify_states_in_your_Single_Page_App)).

In the Optimizely dashboard, you have to configure a set of conditions that will activate the experiment.

Like, if we want to run an A/B test on the Page Shot experiment page for users who do not have Test Pilot installed, the condition would be

``` js
window.testPilotPathname === ('/experiments/page-shot') && (!window.testPilotHasAddon)
```

I don't know the technical bits about when Optimizely goes in and checks for the condition. If there's a more appropriate spot/way to accomplish this I'd be happy to implement.

`sendToOptimizely` will tell us when a user converts on the experimentation page. For instance, we can say that whenever `ExperimentDetailsPage Interactions / button click / Install the Add-on` is fired, consider this a successful conversion.

Note: I haven't been able to do testing against Optimizely because of CSP and local dev.

(Aaaand it's merged as I'm typing my long-winded response. Adding it anyway)

PR #1726, opened 2016-11-02
added 365 day cache max-age to hashed assets

dannycoates:
I just noticed that our hashed assets aren't being cached


ghost (approved):
I think cloudfront has some default caching, but this wfm

dannycoates:
> I think cloudfront has some default caching

idk... i do know that chrome dev tools complained about it on testpilot.firefox.com

PR #1729, opened 2016-11-02
Run ZAP baseline when deploying to dev environment

psiinon:
This change (hopefully;) allows us to run the ZAP baseline scan whenever testpilot is deployed to the dev environment. This perform a 1 min spider of testpilot and then checks for whatever we can check for on the security checklist https://wiki.mozilla.org/Security/CloudSec#Security_Checklist 
For more info about the ZAP baseline scan see https://github.com/zaproxy/zaproxy/wiki/ZAP-Baseline-Scan
The current checks are actually defaulted and need to be tuned via a config file - any suggestions as to where this could live in this repo?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,6 +6,8 @@
 machine:
   hosts:
     testpilot.dev: 127.0.0.1
+  services:
+    - docker
   node:
     version: 6.9.1
   environment:
@@ -44,6 +46,8 @@ deployment:
       - TESTPILOT_BUCKET=testpilot.dev.mozaws.net ./bin/deploy.sh dev
       - aws configure set preview.cloudfront true
       - aws cloudfront create-invalidation --distribution-id E2ERG47PHCWD0Z --paths '/*'
+      - docker pull owasp/zap2docker-weekly
+      - timeout 300 docker run -t owasp/zap2docker-weekly zap-baseline.py -t https://testpilot.dev.mozaws.net/
 
 # Only notify of builds on master branch.
 experimental:


lmorchard:
Tests are passing, but I think this will fail on merge to dev because we recently took the Docker service out of circle.yml. 

I think you'll need to re-add the Docker service in the `machine` section, [like we used to have it here](https://github.com/psiinon/testpilot/blob/af610e373e017a492d7ffb2b2e31d8d75fc0ab2a/circle.yml#L13) before we dropped the container build. For example:

``` yaml
machine:
  hosts:
    testpilot.dev: 127.0.0.1
  services:
    - docker
```

psiinon:
Thanks, I'll do that...

psiinon:
Added the docker service.

pdehaan:
Neat.
Question: Will this do a ZAP scan of the existing TestPilot dev environment? Or of the files in the PR being submitted?

psiinon:
It will scan whatevers on https://testpilot.dev.mozaws.net/ so if that hasnt been updated at that point in the deployment then I've put it in the wrong place :) 
Suggestions for a better place to put the changes appreciated.

chuckharmston:
@psiinon Would you mind outlining exactly what this will do and how it will surface problems? Best I can tell, it proxies incoming and outgoing requests on the build to simulate attacks, which seems unnecessary for a static site with a good CSP.

psiinon:
@chuckharmston sure :) The baseline scan doesnt actually perform any attacks. It just spiders the site for (by default) 1 minute and then reports any issues it finds passively. The tests include a range of things like missing or incorrect headers, cookies set incorrectly etc. These can be issues with static sites. I agree that static sites are much safer to run than dynamic ones, but we want to run the baseline scan against all of our sites :)
If everything passes then it returns 0, if there are any failure it returns 1 and 2 for warnings. It also outputs a summary. We have a dashboard on a private Mozilla repo (I can send you the url on irc if you're interested) - a public example is https://github.com/zaproxy/community-scripts/wiki/Baseline-Summary - click on the Status buttons to see an example of the output.

circle.yml
```diff
@@ -44,6 +46,8 @@ deployment:
       - TESTPILOT_BUCKET=testpilot.dev.mozaws.net ./bin/deploy.sh dev
       - aws configure set preview.cloudfront true
       - aws cloudfront create-invalidation --distribution-id E2ERG47PHCWD0Z --paths '/*'
+      - docker pull owasp/zap2docker-weekly
+      - timeout 300 docker run -t owasp/zap2docker-weekly zap-baseline.py -t https://testpilot.dev.mozaws.net/
```
chuckharmston, line 50: I'm not sure this will spider the most recent build as you might expect; `aws cloudfront create-invalidation` isn't synchronous, it just [creates the invalidation batch request](http://docs.aws.amazon.com/cli/latest/reference/cloudfront/create-invalidation.html#output), which may take several minutes to actually take effect.

psiinon: Is there a better place this could go then? Or do we not get to know when its actually been deployed?

lmorchard: There _is_ an [`invalidation-completed`](http://docs.aws.amazon.com/cli/latest/reference/cloudfront/wait/invalidation-completed.html) command we can use to wait until the thing is done, but sometimes that can take up to 10 minutes

chuckharmston: It's technically been deployed, but the CDN doesn't know about it until later, and the CDN + nginx proxy that sits between the CDN and our S3 bucket are important for adding security headers.

Our build does not currently know when the CDN invalidation has been finished. ~~You'd need to poll the URL returned by the command on `aws cloudfront create-invalidation`, monitoring for `status` to change to `Completed`.~~ Ah, as @lmorchard noted above, there's a command that will do this for you.

Is it possible to run this asynchronously after a build has completed, kicking off by HTTP request or similar, and reporting results in IRC? I'd probably push back on any attempt to wait for the invalidation to finish, as it would likely double our build times.

chuckharmston: I've actually been meaning to file a bug against [http-observatory](https://www.github.com/mozilla/http-observatory) to do similar regression monitoring against its rating via webhook. Perhaps it's worth setting up a joint service with that project?

CC @marumari

pdehaan: Would there be a way to just spin up a server locally [on Circle-CI] and just test localhost:8888 or something directly without having to wait for the code to be deployed to a remote AWS instance?

chuckharmston: @pdehaan not easily, since Cloudfront is part of the mechanism that adds the security headers. We'd need to create a new distribution just for purposes of testing, which is even slower than an invalidation would be.

ghost:
Is this slowing builds down by however long it takes to run? (60s?)

lmorchard:
> Is this slowing builds down by however long it takes to run? (60s?)

Well, that, plus however long it takes to wait for a CloudFront invalidation to complete before kicking off that run - sometimes upwards of 10 minutes. 

Of course, that bit isn't in this PR. As-is, this will test the _previous_ master merge, because it will run against the stale CloudFront distribution before the invalidation is done.

chuckharmston:
> As-is, this will test the previous master merge, because it will run against the stale CloudFront distribution before the invalidation is done.

Or worse, the possibility that the invalidation finishes during the scan; then it would test against previous master merge for part of the test, and current master merge for another part.

jaredhirsch:
Here's an idea: what if we instead ran the ZAP baseline when we pushed to the staging branch? We'd get the benefits of finding security errors before they hit production, and also avoid adding latency to the dev CI build queue.

psiinon:
So we already run the baseline against testpilot, you can see the results here: https://github.com/mozilla-services/cloudsec/wiki/Baseline-services
Well, you can if you have the right perms - if not let me know :)
We have also added it to some of the jenkins deploy jobs, and the plan is to add it to all of them.
Do you see / keep an eye on the jenkins build results? If so then that could be all we need.
If not then whats the best way to surface the results? And can we do that prior to it being pushed to staging?

ghost:
We don't watch any jenkins builds right now.  The best way to surface results would be to have your script file an issue, I think.

psiinon:
@wresuolc so how would that work in practice? We wouldnt want it to raise new issues every time it ran :/ I guess it could raise an issue, record it locally and then only raise another issue if the existing one is closed and the scan still fails?

ghost:
That sounds right.  It could also comment on the closed issue with a "verified" when things work correctly.

jaredhirsch:
@psiinon @wresuolc OK, based on [the linked security bug](https://github.com/mozilla-services/cloudsec/issues/53), it seems like there's interest in finding a good first test case for ZAP baseline testing. Since test pilot is static, doesn't use Jenkins, and Circle-CI integration looks to be problematic, I'm inclined to say Test Pilot is not a good fit for a first attempt at ZAP integration.

I'm also very skeptical of the idea of automatically filing bugs. It sounds like the bug-filing code would be awfully complex:
- manage a long-lived database of which bugs have been filed for which ZAP scan issues
- automatically file bugs as needed
- mark fixed bugs as verified

This sounds like IMO a lot of complexity compared to the value it may deliver.

I'm going to close this PR for now. If there's more discussion to be had, let's schedule a meeting or get together in Hawaii.

PR #1730, opened 2016-11-02
Install raven-js on the client side.

fzzzy:


frontend/src/app/config.js
```diff
@@ -1,15 +1,23 @@
+
```
chuckharmston: Looks like an extra newline snuck in here.

chuckharmston (approved):
Looks great! Feel free to self-merge after the small tweak + passing CI.

chuckharmston:
Also looks like you may need to squash everything; there are a few merge commits in here.

PR #1731, opened 2016-11-02
added firstRun flag to the addon store

### LINKED ISSUE #1457: Track first-time add-on runs.
Despite #1456 being likely WONTFIX, we should still continue to store something in `simple-storage` the first time our add-on is initialized. Then, in the future, we can do a better job of detecting actual first runs vs upgrades.

dannycoates:
fixes #1457

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -330,6 +330,9 @@ exports.main = function(options) {
     // Generate a UUID for this client, so we can manage experiment
     // installations for multiple browsers per user. DO NOT USE IN METRICS.
     store.clientUUID = require('sdk/util/uuid').uuid().toString().slice(1, -1);
+    store.firstRun = true;
+  } else {
+    store.firstRun = false;
   }
 
   Metrics.init();


addon/index.js
```diff
@@ -330,6 +330,9 @@ exports.main = function(options) {
     // Generate a UUID for this client, so we can manage experiment
     // installations for multiple browsers per user. DO NOT USE IN METRICS.
     store.clientUUID = require('sdk/util/uuid').uuid().toString().slice(1, -1);
+    store.firstRun = true;
```
chuckharmston, line 333: Hm, if `!store.clientUUID` is a sufficient test, can't we just say this if we need to know it? I'm not sure how permanent `simple-storage` is.

`const isFirstRun = !store.clientUUID;`

dannycoates: > can't we just say this if we need to know it?

If we only care about first run in this function, sure. TBH I'm not really sure what the requirement is so I created a flag that any part of the app can check.

> I'm not sure how permanent simple-storage is

I did a bit of testing and simple-storage is persistent through a disable/enable but gets immediately deleted on uninstall (clicking undo is like a fresh install).

jaredhirsch: The [simple-storage docs](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/simple-storage) do say that the data is persisted across Firefox sessions.

IIUC the intent of #1457 was to help us differentiate between fresh installs and upgrades. Looks like this branch would do that. Merging.

PR #1734, opened 2016-11-02
Restore integration tests

jaredhirsch:
Get our lone integration test working again on circle-CI.

TL;DR: mostly just needed to bump deps.

Some breaking changes occurred between Firefox and the Marionette python client. It looks like https://bugzil.la/1302707 (landed around the time #1437 was filed) removed `marionette.timeouts`, which broke the self-test that runs inside marionette 3.1.0. I was seeing this error locally today, not sure if others did or not. Upgrading marionette to 3.2.0 (le sigh, a minor version number bump for a breaking API change...) and updating firefox-puppeteer to latest seems to have fixed everything.

Also tweaked the integration test to stop looking for the now-defunct `/experiments` redirect, and cleaned up the docs and .gitignore a bit. :beers:

### DIFF
--- a/bin/circleci/run-integration-tests.sh
+++ b/bin/circleci/run-integration-tests.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-npm run server &
+npm start &
 STATIC_SERVER_PID=$!
 
 # Wait until the server is available...
@@ -12,7 +12,7 @@ done
 python integration/runtests.py \
     --binary=$HOME/firefox/firefox-bin \
     --verbose \
-    integration
+    integration/test_installation.py
 TEST_STATUS=$?
 
 kill $STATIC_SERVER_PID
--- a/.gitignore
+++ b/.gitignore
@@ -1,21 +1,28 @@
-frontend/build/
-dist/
-build/
-tmp/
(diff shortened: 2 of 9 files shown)


dannycoates:
screen shot 2016-11-02 at 8 50 55 pm

👍 

squash maybe?

bin/circleci/run-integration-tests.sh
```diff
@@ -12,7 +12,7 @@ done
 python integration/runtests.py \
     --binary=$HOME/firefox/firefox-bin \
     --verbose \
-    integration
+    integration/test_installation.py
```
lmorchard, line 15: If we add more test cases, do we have to list them all here?

jaredhirsch: I'm hoping to figure that out while I'm writing the second integration test :-)

jaredhirsch: Or rather, I'm hoping to figure out a working glob pattern while writing the second test

jaredhirsch:
@dannycoates cool, thanks for looking. squashed

lmorchard (approved):
LGTM, will merge after this test run finishes

PR #1736, opened 2016-11-03
Make URLs non-localizable

stasm:
Remove the { $url } external arguments from the localizable content.  The URL
can be defined as the href attribute in the HTML source.  L20n will copy it to
the corresponding <a> element in the translation.

### DIFF
--- a/frontend/src/app/components/NewsletterForm.js
+++ b/frontend/src/app/components/NewsletterForm.js
@@ -41,13 +41,12 @@ export default class NewsletterForm extends React.Component {
   renderPrivacyField() {
     const fieldName = 'privacy';
     const url = '/privacy';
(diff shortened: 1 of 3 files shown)


stasm:
It looks like you're already doing this in other strings ([example](https://github.com/mozilla/testpilot/blob/5cdb516b79a18dad744cff6b4957f3a36e675706/locales/en-US/app.ftl#L30)), so this is just making the rest of the translations consistent with this good practice.

dannycoates:
LGTM. thanks @stasm

stasm:
I just scrolled up and read the initial description of this PR.  I didn't mean for it to sound so dry and rude :/  Github just used my commit message and I didn't stop to review it before hitting Create. Thanks for your understanding!

PR #1746, opened 2016-11-03
Fix broken build by explicitly listing python in CI config

jaredhirsch:
Circle-CI automatically creates a virtualenv if a python version is
included in the circle.yml file:
https://circleci.com/docs/language-python/#version

Try adding a listing for the default version.


lmorchard (approved):
:+1: Let's see what this does

Issue #557, opened 2016-03-30
Localize the Test Pilot add-on


lmorchard:
THere's HTML content in the add-on, now. We should start using l20n to localize it, if possible.

meandavejustice:
https://developer.mozilla.org/en-US/Add-ons/SDK/Tutorials/l10n

lmorchard:
That tutorial doesn't use l20n, though. It uses `sdk/l10n`. If we're l20n on the front end, we should figure out how to use it in the add-on too. That way, we're not splitting localizations between `*.l20n` and `*.properties` formats

meandavejustice:
Here is an example from activity-streams https://github.com/mozilla/activity-streams/pull/429/files

ckprice:
Strings in the add-on will have to be localized ("View all experiments", "New experiment", etc) for our German launch in Q4. We recently landed a big localization PR (#1708), not sure if any of that touched the add-on, but pinging this issue in case it didn't. /cc @chuckharmston

johngruen:
@chuckharmston @ckprice I will file a follow up when i hear final confirmation from Winston. We will probably hand localize the add-on via the SDK API for the near term and then figure out a better solution going forward.

chuckharmston:
@ckprice it wasn't, and this bug won't land in this sprint.

Without a lot of engineering work, we won't be able to use Pontoon for this; strings will have to get added manually. So the sooner we can get the German translations necessary, the sooner we can start testing this. I'm not sure who to talk to about that.

johngruen:
@chuck I can coordinate any german strings we need with the German team

ghost:
We chatted about this briefly on IRC.  We can just support .properties files in the add-on and pontoon will work fine.  This should be straight forward.  You can see Min Vid doing this at https://github.com/meandavejustice/min-vid/commit/4341c32e7e2212c1ca5f008bd0859c7defc6f489 (although note that the .properties files need to be in /locale/%LOCALE%/whatever.properties instead of what min vid is doing.

lmorchard:
Just tweaked the title to get us less snagged on the literal l20n issue (or maybe just me, because I was scratching my head since March)

chuck:
Hi folks, just a note that I'm @chuck, not the Chuck that works on this project. :)

chuckharmston:
Sorry about that, Chuck! Great name though. Solid. Sounds dependable.

johngruen:
[...]

PR #1755, opened 2016-11-03
First pass at l10n on add-on

lmorchard:
Issue #557

I read through all the JS modules in the add-on, and I think I found all the strings. Might have missed some, though.

My goal was just to see if I could get these changes working with the standard `sdk/l10n` approach in [the Localization tutorial](https://developer.mozilla.org/en-US/Add-ons/SDK/Tutorials/l10n). But, this seems to be a requirement:

> although note that the .properties files need to be in /locale/%LOCALE%/whatever.properties 
> instead of what min vid is doing.

So, this might need a follow up / more work to get things working in Pontoon?

### DIFF
--- a/addon/locale/en-US.properties
+++ b/addon/locale/en-US.properties
@@ -1,0 +1,20 @@
+experiment_list_enabled = Enabled
+experiment_list_new_experiment = New Experiment
+experiment_list_view_all = View all experiments
+
+experiment_eol_tomorrow_message = Ending Tomorrow
+experiment_eol_soon_message = Ending Soon
+experiment_eol_complete_message = Experiment Complete
+
+installed_message = <strong>FYI:</strong> We put a button in your toolbar <br> so you can always find Test Pilot.
+
+# LOCALIZER NOTE: Placeholder is experiment title
+survey_show_rating_label = Please rate %s
+# LOCALIZER NOTE: Placeholder is experiment title
+survey_rating_thank_you = Thank you for rating %s.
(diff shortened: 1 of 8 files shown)


lmorchard:
Hmm, failed an add-on test in circleci that passed locally. This could still use some general review, though. I'll poke and see why that failed.

addon/locale/en-US.properties
```diff
@@ -0,0 +1,19 @@
...
+
+# LOCALIZER NOTE: Placeholder is experiment title
+survey_show_rating_label = Please rate %s
+# LOCALIZER NOTE: Placeholder is experiment title
+survey_rating_thank_you = Thank you for rating %s.
+survey_rating_survey_button = Take a Quick Survey
+# LOCALIZER NOTE: Placeholder is experiment title
+survey_launch_survey_label = The %s experiment has ended. What did you think?
+
+notification_via = via testpilot.firefox.com
```
pdehaan: Should this URL be a placeholder to prevent localizers from potentially changing it?

lmorchard:
Think I fixed the broken test and placeholder-ized the host name in the notification message.

lmorchard:
Grr, lint.

chuckharmston:
I don't have any substantive feedback; code looks good, seems to work well.

chuckharmston (approved):

lmorchard:
Going to merge this and follow up on [making the path Pontoon-friendly](https://github.com/mozilla/testpilot/issues/557) in another PR today.

PR #1772, opened 2016-11-07
Copy Pontoon-friendly L10n resources to SDK add-on friendly paths

### LINKED ISSUE #557: Localize the Test Pilot add-on

lmorchard:
- Move add-on locales to Pontoon-friendly paths

- Gulp task to copy Pontoon-friendly resources to add-on friendly paths

- Update add-on build in CircleCI to run the gulp task

- Update .gitignore

- Update docs

Fixes #557


lmorchard:
This is mildly awkward. Could be made less awkward with a solution for #1486 to simplify building across the whole project. But, this seems to work for now.

lmorchard:
Just noticed the lint failure, updated.

addon/README.md
```diff
@@ -37,23 +37,29 @@ A relatively easy path for working on this addon involves the following steps:
...
    and `xpinstall.whitelist.required`.
 
-3. Install the [Extension Auto-Installer][autoinstaller] Add-on in Firefox
+1. Install the [Extension Auto-Installer][autoinstaller] Add-on in Firefox
    Developer Edition.
 
-4. Run `npm start` to fire up a watcher that will build the Test Pilot add-on
-   whenever files change and auto-update the installed version in Firefox.
+1. In the top-level directory, run `npm run addon:locales` to copy over
+   translated string resources into the addon project. [If you have the frontend
```
pdehaan: `addon` => `add-on` for consistency.

jaredhirsch:
LGTM and seems to be working :+1:

flodolo:
Why are we mixing .properties and .ftl in the website?

It would also be nice to stop trickling new strings every few days to localizers, and try to do some sort of scheduled pushes if there are strings to expose.

BTW, this just broke the only dashboard we have outside of Pontoon, because it doesn't support mixed formats in a project.

(cc @clouserw @peiying2)

lmorchard:
> Why are we mixing .properties and .ftl in the website

For what it's worth, the .ftl is for the website and the .properties is for the add-on.

flodolo:
I managed to fix the dashboard, luckily the l20n parser works for .properties too.

The concern about adding strings too frequently remain though, especially without any kind of warning.

ghost:
Test Pilot does pushes to production every two weeks, but commits land on -dev as they are reviewed and committed.  What kind of warning would you want?

flodolo:
Re: warning. I'd like to avoid unreviewed landings like experiments.ftl: we could have caught those issues earlier and avoided all the pain we're having now to fix it.

Localizers don't work only on Test Pilot, they have dozens of other projects to follow, included Firefox itself. Checking day by day for new strings is not something I want to establish as a workflow.
That's also not the workflow of other projects: if there's a 2 weeks cycle, new strings are exposed at the beginning of the cycle to localizers, not a bit every few days.

lmorchard:
Should we open a new issue to address changing the cadence of string releases? Maybe we _do_ need a separate repo / branch for them? Not sure otherwise how we could ever separate new strings from new features as we review & merge new code into master branch.

Could be we need to set up a "deployment" of strings to a different repo to match a better cadence for localizers, then merge those back into master as they're completed? How are other projects handling this satisfactorily that don't run on trains / versions / etc?

flodolo:
I've talked with @wresuolc on IRC, and I think we can keep the existing workflow.

I need to draft an email to send to localizers to explain that they will see new strings frequently, but the project will keep shipping every 2 weeks or so. Hopefully I'll manage to do that in the next 24 hours (I want it reviewed by my team).

As for other projects, they usually merge new strings manually every X weeks, but they mostly work on .po and can live without the strings being available for en-US. I'm afraid that's not the case for Test Pilot.

ghost:
I chatted with @flodolo off github.  For the time being we're going to continue committing as we are now, and we'll try some messaging changes with the localizers regarding cadence and priorities.  Thanks.

edit: jinx!

Issue #1778, opened 2016-11-08
Experiment content not being picked up by Pontoon


chuckharmston:
We [generate](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/content.js#L58-L63) the en-US [`experiments.ftl`](https://testpilot.dev.mozaws.net/static/locales/en-US/experiments.ftl) from our [content files](https://github.com/mozilla/testpilot/tree/master/content-src/experiments) in our build process, and then appropriate [include it](https://github.com/mozilla/testpilot/blob/master/frontend/src/templates/index.mustache#L13) on the page. Unfortunately, it appears like it's not being picked up by Pontoon.

@mathjazz any chance you know why this is happening? Would the default one need to be included in source control?

mathjazz:
Yup, the file needs to be available here:
https://github.com/mozilla/testpilot/tree/master/locales/en-US

chuckharmston:
#1779 moved a copy of the experiment localization file into the repository as a short-term fix, and it's now being picked up by Pontoon:

https://pontoon.mozilla.org/de/test-pilot-website/experiments.ftl/

Closing this, and I'll follow up with #1781. Thanks for the assist @mathjazz!

mathjazz:
As pointed out by @flodolo, there's a lot of HTML tags in those strings. Can we get rid of them?

chuckharmston:
Yeah, it's not awesome. Normalizing those to remove as much HTML as possible is one of the tasks a task in https://github.com/mozilla/testpilot/issues/1724, which we've slated to do this sprint (ends 11/18). I'll prioritize it.

flodolo:
The problem is that we have already exposed those strings to localization, and people are already working on them. If we clean up en-US string, we should have something to do the same safely in l10n files.

Issue #1768, opened 2016-11-07
Localization Capability for Legal + About pages


johngruen:
We need to localize the ToS, PN and About Pages. We could move these to the legal docs repo, but it's not required (and the About page isn't a legal page).

johngruen:
Related to  #738

lmorchard:
Hmm, this is different than just localizing strings in the common UI. Maybe we need to create markdown & generate whole new separate HTML pages with the locale in the path? That is, you hit `/terms/` and get bounced to `/de/terms`? And that's generated from `compiled-templates/de/terms-of-use.md`?

lmorchard:
Also: are these legal docs meant to be localized via Pontoon? Or will we need a different way to localize them?

ghost:
We'll figure out a timeline in our monday meeting w/ legal

ghost:
We talked with legal and this is blocking.  We weren't expecting that, so this is on us for having it appear in the middle of a sprint. :-/

The docs are at https://github.com/mozilla/legal-docs/tree/master/firefox_testpilot_PrivacyNotice and https://github.com/mozilla/legal-docs/tree/master/firefox_testpilot_Terms .  We'll never have to worry about partial localizations or anything there -- if the file exists, it's good to show.

Thanks.

flodolo:
I don't know much about the background discussion, but legal docs are in that repo for a specific reason: they're translated by an external agency (usually via @peiying2), localizers don't have direct access to them, they can file bugs for minor fixes.

Personally I think it would be pretty confusing to have Test Pilot legal docs in this repo instead of the legal-docs one. And if we do, they shouldn't be exposed for localization through unreviewed channels.

On the other hand, the about page should be translated, and that can't really happen with a .md file.

peiying2:
These documents are being localized as we speak.  We will have the localized files on Thursday.  There should be a "de" equivalent file in both folders.

peiying2:
https://bugzilla.mozilla.org/show_bug.cgi?id=1317493 tracks legal l10n progress.

Regarding the About doc, this comes from the product team, not legal team.

PR #1782, opened 2016-11-08
Remove HTML from tour steps.

chuckharmston:


lmorchard:
Is there an associated issue with some background? At one point, we explicitly wanted HTML in these tour steps - e.g. in case we wanted to link out to more info on some feature or another. Don't want to take out the capability if we still want to use it some day.

Might also want to let the initial wave of localization work to die down unless this is an urgent change.

chuckharmston:
It was [requested](https://github.com/mozilla/testpilot/issues/1778#issuecomment-259136231) by @mathjazz and @flodolo of the localization team. HTML makes localization really challenging, as changes to structure are tightly coupled to the localizations. We could still have the ability to link out if that becomes a need—we'd just need to re-add `dangerouslySetInnerHtml` to the appropriate spot.

jaredhirsch:
I see a _lot_ of other strings that still have html in them. Is there any reason not to fix those strings at the same time?

chuckharmston:
@6a68 I'm doing just that (see #1768), just trying to keep PRs small to make them more reviewable.

lmorchard:
Yeah, would rather get these in a big single batch and wait until after Pontoon activity dies down. Would also like to be sure we don't want HTML in those fields, because it seems harder to add it back in once we have localizations.

chuckharmston:
I guess that's fine, but:

1) We need to make sure we don't invalidate localizers' hard work, and the longer we wait, the harder this will be.
2) Dibs on not massaging all the files by hand.

jaredhirsch:
HTML in l10n strings is a bummer for multiple reasons. Biggest one is, if localizers typo the tags, that stuff gets checked in without a review--so QA has to check that, eg., links aren't broken, for every locale. That's a lot of needless work.

Another thing is, changing the HTML structure suddenly depends on an army of volunteers.

If we instead assemble the HTML ourselves, like "grab this key and that key and shove into this template string", life is easier.

chuckharmston:
> it seems harder to add it back in once we have localizations.

It's really not; we aren't giving up any capability here, just adding more structure and making localization easier.

If we want to add links to tour steps, You just need to change this line:

``` jsx
<p data-l10n-id={this.l10nId(['tour_steps', idx, 'copy'])}>{step.copy}</p>
```

to this:

``` jsx
<p data-l10n-id={this.l10nId(['tour_steps', idx, 'copy'])} dangerouslySetInnerHTML={createMarkup(step.copy)}></p>
```

I'm doing that here: https://github.com/mozilla/testpilot/pull/1786

jaredhirsch:
Closing, blocked on #1811 for now. See [my comment in other PR](https://github.com/mozilla/testpilot/pull/1786#issuecomment-263340519) for discussion.

Issue #1724, opened 2016-11-01
Audit experiment YAML files, template, and documentation


chuckharmston:
Now that we're localizing YAML content (#1340), we should audit and reconsider the experiment YAML files, template, and documentation with localization in mind. Some tasks:

- Ensure that all fields are currently being used (e.g. #1723)
- Do a better job delineating optional and required fields in the template and docs.
- Minimize the use of HTML in strings; this makes localization much more challenging.
- Try to standardize the structure of fields that are currently semi-freeform (e.g. [`measurements`](https://github.com/mozilla/testpilot/blob/master/content-src/experiments/tab-center.yaml#L21-L33))
- Remove blank keys and null values where appropriate.
- Make l10n arguments available to strings where appropriate, e.g. [the experiment name in the measurements paragraph](https://github.com/mozilla/testpilot/blob/master/docs/experiments/experiment-content/template.yaml#L24)
- Make sure templates correctly handle missing optional fields.
- Move all the inline comments in [`template.yaml`](https://github.com/mozilla/testpilot/blob/master/docs/experiments/experiment-content/template.yaml) to [`getting_started.md`](https://github.com/mozilla/testpilot/blob/master/docs/experiments/experiment-content/getting_started.md)
- Move the docs in `docs/experiments/experiment-content` (intended to be docs for experiment authors) to `docs/content`

ghost:
Related: https://github.com/mozilla/testpilot/issues/1308

chuckharmston:
I'm moving this from sprint 14 to 15, since it's sort of a meta-bug that I've been addressing with small PRs as I have time.

ghost:
Are enough of these done to turn this into separate bugs?  and/or can we turn this into checkboxes so we can see how much is left?

chuckharmston:
This is good to be closed, actually.

PR #1786, opened 2016-11-08
Remove HTML from measurements section.

chuckharmston:


content-src/experiments/activity-stream.yaml
```diff
@@ -18,22 +18,20 @@ contribute_url: 'https://github.com/mozilla/activity-stream'
...
+    We collect basic usage data regarding how you interact with experimental
+    new tab and timeline pages, such as the types of elements that you click
+    and their position on the page.
+  - >
+    We also collect basic information about your browser profile that affect
+    Activity Stream, such as the number of bookmarks and history items.
+  - >
+    To provide Activity Stream, we fetch metadata about websites you have
+    visited, such as images and icons, using a service called <a
+    href="http://embed.ly/" target="_blank">Embedly</a>. We send these requests
```
jaredhirsch, line 32: Hmm, we still have html here.

I think it might be good to have a parent bug that documents a single approach we'll use to remove html globally, then tie the PRs back to it

chuckharmston: We don't have a good way to strip HTML from experiment content generated like this, unfortunately. The best we can do with current capabilities is to add as much structure to the content as possible, which is what I'm doing here.

#1724 is the parent bug for this.

chuckharmston:
Note: since none of the original l10n keys for this section are in tact, there's no need to touch any of the existing non-`en-US` files

content-src/experiments/activity-stream.yaml
jaredhirsch: Sure, but we're the ones who built this yaml config file thing, so we should be able to tweak it a bit to support l10n best practices.

For example, we could instead have the yaml use markdown link style (the string literal `[link text](link URL)`), then use our yaml parser to replace the link with a placeholder before sending to localizers.

So experiment authors would write something like `...a service called [Embedly](https://embedly.com)`, which is easy for them, while localizers would see two separate strings, and a [standard localization note](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Add_localization_notes):

```
# LOCALIZATION NOTE(activity-stream-measurements): {{embedly}} will 
# be replaced at run-time by an active link. The string 'embedly-link' will be
# used as text for this link.
activity-stream-measurements= a service called {{embedly}}
embedly-link=Embedly
```

Finally, when templating the site, we'd just need to parse the localized strings for the double-mustachio'd placeholders, and convert into an anchor tag with the correct text infixed.

Since most of our HTML is links, I think it'd make sense to support this use case.

This is just a sketch of an idea; anything that works with l10n best practices, and keeps things manageable for experiment yaml maintainers, would do.

chuckharmston: That'd be a great enhancement and I would invite you to file a bug for it, but it's out of scope for #1724, and I'd like to get an incremental improvement in front of localizers sooner rather than later.

lmorchard:
I'm really leery of pushing these changes without talking them through in status or something. They're touching on product-y things and aren't just quick tweaks. And we've got localizations coming in fast right now, so I'm really averse to making changes there

lmorchard:
Or I mean, at least tag in folks like @johngruen & @ckprice who've also been working on this content

ckprice:
I'm kind of onboard with not tweaking things mid l10n rush. What is the risk of us keeping things the way they are? Have we received complaints about too much HTML?

So this generalizes our measurement intro for all experiments to:

`In addition to the <a>data</a> collected by all Test Pilot experiments, here are the key things you should know about what is happening when you use { $experimentTitle }`

A couple questions:
1. Would we ever want to customize this intro per experiment? /cc @johngruen 
2. I'm no language expert, but does the `experimentTitle` always makes sense in the same position for every language/experiment name? /cc ?
3. What happens to the [existing `*Measurements` IDs](https://github.com/mozilla/testpilot/blob/master/locales/de/experiments.ftl#L3) that have been localized?

FWIW it's not a huge deal on my side either way with editing YAML content.

chuckharmston:
> Have we received complaints about too much HTML?

Yes, the localization team [has asked us to pare it down](https://github.com/mozilla/testpilot/issues/1778#issuecomment-259136231).

> 1. Would we ever want to customize this intro per experiment?

We don't currently; it's even [referred to as "boilerplate"](https://github.com/mozilla/testpilot/blob/493c2c630e1c70f736c105e54d8c069498a57cf1/docs/experiments/experiment-content/template.yaml#L20) in the content template.

> 1. I'm no language expert, but does the experiment title always makes sense in the same position for every language/experiment name? 

No, but because it's in a variable, localizers have the freedom to move it where it makes sense.

> 1. What happens to the existing *Measurements IDs that have been localized?

I was hoping to get this in fast enough that we could massage it to the new format. At this point it's probably too much work.

ckprice:
> No, but because it's in a variable, localizers have the freedom to move it where it makes sense.

Yeah, I was thinking more about the structure of the string changing depending on the nature of the experiment title for some languages. Like does "Page Shot" make sense at the end of the string, but "No More 404s" make sense in the middle of the string. It seems like it wouldn't change, but someone with more experience may be able to comment (also willing to move past this if the group is good with these changes).

johngruen:
@chuckharmston I'm actually pretty okay with all of this. I can check out the branch to review, but this all seems legit to me.

johngruen:
and to @ckprice question. Possibly, but we haven't so far, and I don't anticipate the need to do so at this time. Keeping these short privacy notices consistent is a good thing. My view is that doing this simply enforces some consistency.

johngruen:
UI looks fine to me.

jaredhirsch:
@chuckharmston Could you summarize the current status of this PR and #1782? Looks like they both need a rebase. Do we still need to have more discussion?

chuckharmston:
We still want to land these changes, but a script (or a bunch of manual work) is necessary to do so without invalidating all these strings.

jaredhirsch:
@chuckharmston Aha, I get it - this PR and #1782 are blocked on #1811. In that case, since #1811 isn't assigned for this sprint, these will be lingering for at least another week. I'd like to keep the PR queue clear, so I'm just going to close this and #1782 for now. Please reopen as soon as they are cleared to land.

PR #1793, opened 2016-11-09
[ref #904] Add Optimizely to 'script-src'

ckprice:
Ref #904.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -34,7 +34,7 @@ fi
 MAX_AGE="600"
 
 HPKP="\"public-key-pins\": \"max-age=300;pin-sha256=\\\"WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18=\\\";pin-sha256=\\\"r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E=\\\"\""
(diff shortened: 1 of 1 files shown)


bin/deploy.sh
```diff
@@ -34,7 +34,7 @@ fi
...
-CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org https://*.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self' https://*.optimizely.com; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
+CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org https://*.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self' https://*.optimizely.com; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com https://*.optimizely.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
```
lmorchard: I think we're strongly advised to never use wildcards here... I see cdn.optimizely.com, are there other domains?

lmorchard: Probably should also only put it on the `$DEST = dev` part, until / unless we want this to work in stage & prod

ckprice:
@lmorchard removed from stage/prod, removed wildcards.

PR #1794, opened 2016-11-09
[ref #904] Add app.optimizely.com to script-src and update style-src

ckprice:
- add app.optimizely.com to `script-src`
- ~~add `unsafe-inline` and `unsafe-eval` to `script-src`~~ 
- add `unsafe-inline` to `style-src`

Ref #904.

Closer! I'm reading console errors from [the Optimizely dashboard now](https://cloud.githubusercontent.com/assets/8632184/20158107/58c3875e-a68d-11e6-88d1-5583b601c8c7.png). Additions are based on [these recommendations](https://help.optimizely.com/Troubleshoot_Problems/Update_your_site's_Content_Security_Policies_(CSP)#2._Script).

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -43,7 +43,7 @@ XSS="\"x-xss-protection\": \"1; mode=block\""
 # enable localization.  Also expires more often.
 if [ "$DEST" = "dev" ]; then
     MAX_AGE="15"
-    CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org https://cdn.optimizely.com; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self' https://pontoon.mozilla.org https://app.optimizely.com; img-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com https://cdn.optimizely.com; style-src 'self' https://pontoon.mozilla.org code.cdn.mozilla.net; report-uri /__cspreport__;\""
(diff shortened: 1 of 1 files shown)


ckprice:
"unsafe" sounds...not very safe. Pinging @g-k to see if he wants to comment on these.

g-k:
Thanks for bringing this up!  Sorry for the delay (and wall of text); I'm in MTV for new hire onboarding this week.

The unsafe sources are accurately named and very not safe.  Adding `unsafe-inline` and `unsafe-eval` script sources defeat most of the XSS protection CSP provides.

I'm assuming this is Optimizely JS since this was in their recommendation and might have a better idea what's going on if you can send me the snippet that triggered the violation or a har file of a session with the Optimizely editor.

So many questions:
- What does Optimizely need unsafe eval and inline scripts for?
  - Can they add hash or nonce sources to the inline scripts?  [This article by Mark Goodwin](https://blog.mozilla.org/security/2014/10/04/csp-for-the-web-we-have/) gives a good overview of which to use when.
  - Can they serve the inline scripts and snippets to eval as separate scripts so we can use `strict-dynamic` which ["allows the execution of scripts dynamically added to the page, as long as they were loaded by a safe, already-trusted script"](https://csp.withgoogle.com/docs/strict-csp.html)?
- Can we avoid adding either unsafe src without breaking Optimizely's core functionality?  e.g. do thye only need them in the experiment editor or is it a performance thing like [angular using unsafe-eval to speed up `$parse`](https://docs.angularjs.org/api/ng/directive/ngCsp)? Their site says the unsafe srcs are for "Some of Optimizely's functionality," but it'd be good to know what.
- Could we set up a dev instance behind the VPN with a relaxed CSP to work with the Optimizely editor and maintain a stricter CSP in prod?

I'd be happy to talk to Optimizely about this. 

Also, depending on the urgency of getting Optimizely working I can see what the rest of cloudsec thinks about the risk of weakening CSP for awhile to not hold you up.

I'm less concerned about `unsafe-inline` `style-src`.  It'd be good to move them out of the inline html, but less critical since style based attacks usually relate to defacement and vandalism (tricky data exfiltration stuff is possible too).

http://stackoverflow.com/questions/30653698/csp-style-src-unsafe-inline-is-it-worth-it#31759553
https://lists.w3.org/Archives/Public/public-webappsec/2012Mar/0046.html
https://mikewest.org/2013/09/xss-no-the-other-s-cssconfeu-2013
https://html5sec.org/#css

ckprice:
Thanks. I've updated this PR to take out the unsafe stuff from `script-src`, and will open a follow-up issue (and keep this context) if required.

> I'm less concerned about unsafe-inline style-src. 

I kept the update to `style-src`.

The reason I added these in the first place is because CSP is reporting a violation that mentioned `eval()` at https://testpilot.dev.mozaws.net/

```
"script-sample":"call to eval() or related function blocked by CSP",
"violated-directive":"script-src https://testpilot.dev.mozaws.net https://pontoon.mozilla.org https://ssl.google-analytics.com https://cdn.optimizely.com"
```

I may be misinterpreting this report and maybe we don't need the unsafe stuff. But again, I can only test the Optimizely dashboard on dev so won't know until we're merged :/ r? @lmorchard @g-k

PR #1796, opened 2016-11-10
Support a fake dev-only experiment for feature QA

### LINKED ISSUE #1762: make a fake experiment on -dev
Let's add a fake experiment on -dev which we can graduate early

lmorchard:
This turned out a little bit more involved than I'd anticipated, after thinking through localization issues and such. Hopefully this catches things, should let us create fake experiments that a) only get built for local & dev environments and b) never show up in our .ftl files as noise for localizers.

I was also lazy on creating the first fake experiment: I copied min-vid and only changed a few fields. We can tweak the rest as needed to try out features.

- `dev: true` YAML property to mark an experiment as dev-only

- ENABLE_DEV_CONTENT=1 env var to include dev content in build

- Update `npm start` to use dev environment vars

- New dev-example.yaml for a fake experiment

- Localization disabled for dev-only experiments
  - excluded from .ftl string extraction
  - data-l10n-id generation returns null

- onlyLaunchedExperimentSelector selector also excludes dev-only
  experiments on stage & prod - reduntant measure alongside build
  process ENABLE_DEV_CONTENT default

- Test updates

Fixes #1762

### DIFF
--- a/bin/circleci/build-frontend.sh
+++ b/bin/circleci/build-frontend.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
-if [[ " $TESTPILOT_ENABLE_PONTOON_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
-    NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_LOCALES=1 npm run static
+if [[ $CIRCLE_BRANCH == 'master' ]]; then
+    NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 ENABLE_DEV_LOCALES=1 npm run static
 else
-    NODE_ENV=production ENABLE_PONTOON=0 ENABLE_DEV_LOCALES=0 npm run static
+    NODE_ENV=production ENABLE_PONTOON=0 ENABLE_DEV_CONTENT=0 ENABLE_DEV_LOCALES=0 npm run static
 fi
--- a/content-src/experiments/dev-example.yaml
+++ b/content-src/experiments/dev-example.yaml
@@ -1,0 +1,89 @@
+dev: true
+id: 999
+title: 'Dev Example'
+slug: dev-example
+thumbnail: /static/images/check.png
+description: >
+  This is an example experiment that you should never see in production.
+introduction: >
+  <p>Do you like Test Pilot experiments? Well, this one should exercise all the
+  features of the site. You should never see it on production!</p>
+  <ul>
+    <li><strong>Markup in introduction:</strong> Look at this HTML list!</li>
+    <li><strong>Hidden from production:</strong> If you see this on production
+      or staging, something has gone horribly wrong!</li>
+    <li><strong>Lorem ipsum dolor sit amet:</strong> Etiam lacinia, turpis sit
+      amet sagittis viverra, est erat tincidunt mi, a cursus leo orci et
+      orci.</li>
+  </ul>
+image_twitter: /static/images/experiments/min-vid/social/mv-twitter.jpg
+image_facebook: /static/images/experiments/min-vid/social/mv-facebook.jpg
+version: '0.3.1'
+changelog_url: 'https://github.com/meandavejustice/min-vid/blob/master/CHANGELOG.md'
+contribute_url: 'https://github.com/meandavejustice/min-vid'
+bug_report_url: 'https://github.com/meandavejustice/min-vid/issues'
+discourse_url: 'https://discourse.mozilla-community.org/c/test-pilot/min-vid'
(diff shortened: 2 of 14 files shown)


lmorchard:
Oh and you should be able to see the site with and without the dev experiment in your local environment like so:

``` bash
# With content, by default
npm start 

# Without content, by customizing env vars to gulp
NODE_ENV=development ENABLE_DEV_CONTENT=0 gulp
```

content-src/experiments/dev-example.yaml
```diff
@@ -0,0 +1,89 @@
+dev: true
```
pdehaan, line 1: Do we need to add this new key to the experiment template?
https://github.com/mozilla/testpilot/blob/master/docs/experiments/experiment-content/template.yaml

bin/circleci/build-frontend.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
-if [[ " $TESTPILOT_ENABLE_PONTOON_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
-    NODE_ENV=production ENABLE_PONTOON=1 npm run static
+if [[ $CIRCLE_BRANCH == 'master' ]]; then
+    NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 npm run static
```
chuckharmston: Yeah, after making a similar comment on #1797, I'm convinced that these should be governed by a single `DEV` flag.

content-src/experiments/dev-example.yaml
```diff
@@ -0,0 +1,89 @@
...
+    title: 'Staff Engineer'
+    avatar: /static/images/experiments/min-vid/avatars/jared-hirsch.jpg
+  -
+    display_name: 'Jen Kagan'
+    title: 'Engineering Intern'
+    avatar: /static/images/experiments/min-vid/avatars/jen-kagan.jpg
+installation_count: 1
+created: '2016-09-22T00:07:28.847430Z'
+modified: '2016-10-06T22:19:14.129Z'
+order: 999
```
chuckharmston, line 89: A few things I think we'd want to always be testable:
- `locales`
- `locale_blocklist`
- `locale_grantlist`
- `incompatible`

frontend/tasks/util.js
```diff
@@ -39,6 +39,9 @@ function lookup(obj, path) {
 }
 
 function experimentL10nId(experiment, pieces) {
+  if (experiment.dev) { // For dev-only experiments, data-l10n-id=null is omitted from the DOM
+    return null;
+  }
```
chuckharmston, line 44: :+1:

frontend/test/app/containers/ExperimentPage-test.js
```diff
@@ -143,6 +143,15 @@ describe('app/components/ExperimentPage:ExperimentDetail', () => {
     expect(findByL10nID('testingDescription')).to.have.property('length', 1);
   });
 
+  it('should have l10n IDs for content omitted if the experiment is dev-only', () => {
```
chuckharmston: This is tough to parse. Perhaps `should omit l10n IDs for dev-only content`?

chuckharmston (changes requested):
Looks good overall!

bin/circleci/build-frontend.sh
lmorchard: I replied over there - tl;dr: local & remote dev environments have enough differences to separate them, and IMO I've found turning these things on/off individually to be handy.

frontend/test/app/containers/ExperimentPage-test.js
lmorchard: Made this change

content-src/experiments/dev-example.yaml
lmorchard: Updated commit to add this change

chuckharmston (approved):
I'm not too far from finishing for the week, but this is close, so I'll approve it. 

tumblr_li1ocze0gv1qbiz5t

lmorchard:
Rebased against current master, will merge after tests are green

SoftVision-PaulOiegas:
Since the fake experiment was deployed on dev and since we have only 2 days left till the stage deploy. Could you guys set a graduation date for it as follows in order to test both the pre-graduation and graduation flows ? Currently it appears as "Just launched".

I was thinking at setting graduation today (Tuesday) around 20:00-21:00 hours your time, to be in 24 hours. This will allow us to test the pre-graduation flow on our Wednesday morning. And the actual graduation flow on Thursday our time until you guys reach the office for Stage deploy.

How does it sound ?

ghost:
I landed https://github.com/mozilla/testpilot/compare/b4dc5b802960...4836bf7947f8 which should let you see it on -dev

PR #1797, opened 2016-11-10
Enable new locales in the frontend

### LINKED ISSUE #1784: Make German show up on testpilot.dev.mozaws.net
I said German, but I really mean all the languages who have committed translations.  Chuck says we need to update the availableLanguages meta tag.

Let's do that part right away and land it, but bonus points for thinking about how to notify us of new languages so we can enable them.

lmorchard:
- Default enabled locales in a manual list to be updated as translation
  work reaches 100% in each locale

- New `ENABLE_DEV_LOCALES=1` env var for gulp which enables all known
  locales in Pontoon

- Add locale set to availableLanguages meta header for L20n

- Update package.json to enable dev locales on local env

- Update bin/circleci/build-frontend.sh to enable dev locales on dev env

- Prod & stage use the default of known completed locales

Fixes #1784

### DIFF
--- a/bin/circleci/build-frontend.sh
+++ b/bin/circleci/build-frontend.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 if [[ " $TESTPILOT_ENABLE_PONTOON_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
-    NODE_ENV=production ENABLE_PONTOON=1 npm run static
+    NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_LOCALES=1 npm run static
 else
-    NODE_ENV=production ENABLE_PONTOON=0 npm run static
+    NODE_ENV=production ENABLE_PONTOON=0 ENABLE_DEV_LOCALES=0 npm run static
 fi
--- a/frontend/config.js
+++ b/frontend/config.js
@@ -3,6 +3,11 @@ module.exports = {
   IS_DEBUG: (process.env.NODE_ENV === 'development'),
   USE_HTTPS: (process.env.USE_HTTPS == 1),
   ENABLE_PONTOON: (process.env.ENABLE_PONTOON === '1'),
+  AVAILABLE_LOCALES: (process.env.ENABLE_DEV_LOCALES === '1') ?
+    // All locales on Pontoon for local & dev
+    'en-US,zh-CN,de,ja,fy-NL,fa,zh-TW,ast,uk,fr,sv-SE,bn-BD,pt-BR,tr,kab,es-ES,sl,cs,ru,bs,es-CL' :
+    // Enabled locales for stage & production - update as they reach 100%
+    'en-US,zh-CN,de',
 
   // TODO: Move addon build to a better path
   ADDON_SRC_PATH: './addon/',
--- a/compiled-templates/compiled-templates.js
+++ b/compiled-templates/compiled-templates.js
@@ -1,3 +1,5 @@
+const config = require('../frontend/config.js');
+
 module.exports.templateBegin = `
 <!doctype html>
 <html>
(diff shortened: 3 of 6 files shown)


pdehaan:
I don't think the availableLocales are getting set on all the templates:

### GOOD:
- http://testpilot.dev:8000/
- http://testpilot.dev:8000/share
- http://testpilot.dev:8000/onboarding
- http://testpilot.dev:8000/experiments/page-shot

firefox_test_pilot

---

### BAD:
- http://testpilot.dev:8000/about
- http://testpilot.dev:8000/privacy
- http://testpilot.dev:8000/terms

firefox_test_pilot

---

pdehaan:
Is the add-on localized?

test_pilot_and_menubar_and_firefox_test_pilot_and_browser_console

add-ons_manager

ghost:
I'm excited about this, if we can land this today, let's do, so QA has as much lead time as they can get on it.

Re: the legal documents, @johngruen said we don't need to localize them yet at https://github.com/mozilla/testpilot/issues/738#issuecomment-258257919 -- I'm not sure what the timeline is though.  Also I see https://github.com/mozilla/testpilot/issues/1768 exists, so...I'll follow up on those.

Re: the add-on, I think that's another kettle of fish (that is, it's not ready yet, but I can't find the issue), Les will know for sure

bin/circleci/build-frontend.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 if [[ " $TESTPILOT_ENABLE_PONTOON_BRANCHES " =~ " $CIRCLE_BRANCH " ]]; then
-    NODE_ENV=production ENABLE_PONTOON=1 npm run static
+    NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_LOCALES=1 npm run static
```
chuckharmston, line 4: Would it be better to just have a dev flag that governs each of these? I can't imagine they'll ever be out of sync.

chuckharmston (approved):
One comment that may or may not be worth considering. Looks good to me, otherwise!

bin/circleci/build-frontend.sh
lmorchard: There is one big place where these are out of sync:
- Local env needs `NODE_ENVIRONMENT=development` to skip minification & such.
- Dev env needs `NODE_ENVIRONMENT=production` to be prod-like in most ways.

So, most of the other vars apply to both local & remote dev environments, but I've been keeping them separate just in case other differences pop up. Also, it's handy to turn them on & off individually during local development IMO.

lmorchard: Oh, and also [ENABLE_PONTOON is omitted (default 0) in the `npm start` command](https://github.com/mozilla/testpilot/pull/1797/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R105)

lmorchard: So, TL;DR, the vars between local, dev, stage, and prod are effectively:
- Local - `NODE_ENVIRONMENT=development ENABLE_PONTOON=0 ENABLE_DEV_LOCALES=1`
- Dev - `NODE_ENVIRONMENT=production ENABLE_PONTOON=1 ENABLE_DEV_LOCALES=1`
- Stage & Prod - `NODE_ENVIRONMENT=production ENABLE_PONTOON=0 ENABLE_DEV_LOCALES=0`

lmorchard:
> I don't think the availableLocales are getting set on all the templates

Huh, that's weird... [the locales should be included in those pages' template](https://github.com/mozilla/testpilot/pull/1797/files#diff-a6e0e557c0f2919134e33cd0c13a5c2bR12).

Edit: Oh, but I don't think those pages actually load L20N.js!

> Is the add-on localized?

[Yes](https://github.com/mozilla/testpilot/blob/master/locales/en-US/addon.properties), and [with a build-time hack in gulp](https://github.com/mozilla/testpilot/commit/b9cfcc8dcda9866277e983c1053043e166f10bcd), but not as a part of this PR.

frontend/config.js
```diff
@@ -3,6 +3,11 @@ module.exports = {
   IS_DEBUG: (process.env.NODE_ENV === 'development'),
   USE_HTTPS: (process.env.USE_HTTPS == 1),
   ENABLE_PONTOON: (process.env.ENABLE_PONTOON === '1'),
+  AVAILABLE_LOCALES: (process.env.ENABLE_DEV_LOCALES === '1') ?
+    // All locales on Pontoon for local & dev
+    'en-US,zh-CN,de,ja,fy-NL,fa,zh-TW,ast,uk,fr,sv-SE,bn-BD,pt-BR,tr,kab,es-ES,sl,cs,ru,bs,es-CL' :
```
dannycoates, line 8: it'd be cool if we could automate this all locales list. maybe with a script that gets the directory names under `/locales`

PR #1818, opened 2016-11-16
navigate to localized terms and privacy

### LINKED ISSUE #1768: Localization Capability for Legal + About pages

dannycoates:
What a mess...

This gives us l10n capability for terms, privacy, and about pages. You'll notice the directory layout has changed a bit. Source markdown files live under `compiled-templates` with a dir per page and file per locale which matches the [legal repo structure](https://github.com/mozilla/legal-docs/tree/master/firefox_testpilot_Terms) so files can be copied over easily.

On build they wind up in the same place as before with the addition of a dir per locale, like `/terms/de` for German. `/terms` goes to an English page where the `locale.js` redirects to the best match language in a way that matches how l20n.js chooses elsewhere.

<strike>Currently it sets `availableLanguages` using `config.AVAILABLE_LOCALES` which can cause a problem if there isn't localized markdown, leading to a 404. We should either not consider a language available until they have legal pages, or maintain a separate list. For now I opted for the former.</strike>

`availableLanguages` is set at build time. If the directory for a page has a locale file it will be included.

This is the most straightforward way I can think to get this done, but it's not great. <strike>One obvious problem is that the `data-l10n-id` strings don't get translated because l20n isn't included (a relatively easy thing to fix... maybe)</strike>. Another is how the method clashes with l20n from the rest of the site. L20n isn't flexible enough to handle this case elegantly, so we end up having to do something custom.

fixes #1768


chuckharmston (approved):
Code is good, this works great. Good work!

I have some overall concerns about adding Yet Another Localization Approach, but we discussed these in IRC and they were deemed not worth blocking.

flodolo:
We should not ask people to work on a markdown file, because that needs to happen outside of tools and directly in the repository.

If you want the About page localized, it should be converted to an HTML template and strings exposed through tools.

I agree it's not blocking, but it **is** blocking if you want the page localized.

PR #1820, opened 2016-11-16
componentize eol warning

### LINKED ISSUE #1741: Let's use the Warning component for the ExperimentPage renderEolBlock
we should add an optional field to the yaml for an eol message passed in as a child of the warning. It seems like we will need to tailor these messages to the experiments themselves.

johngruen:
Fixes #1741


dannycoates:
needs a rebase but looks good

johngruen:
@dannycoates this should be good now

PR #1831, opened 2016-11-17
[ref #904] Update CSP script-src policy

ckprice:
Per [Optimizely documentation](https://help.optimizely.com/Troubleshoot_Problems/Update_your_site's_Content_Security_Policies_(CSP)#2._Script).

Ref: #904.


chuckharmston:
Please don't merge yet; chatting about this with @jvehent backchannel.

april:
Please note that allowing Optimizely and `'unsafe-eval'`/`'unsafe-inline'` has the following effects, respectively:
- Anybody at Optimizely (or anybody who breaks into them) will have the ability to install plugins, given Test Pilot's escalated privileges
- Any cross-site scripting (XSS) bug on Test Pilot -- by far our most common class of websec bug -- will have the ability to install plugins

Note that AMO specifically doesn't use Optimizely (and those CSP flags) for the very high risk and impact that a problem like one of the above could cause.  You can certainly accept this risk, but it should absolutely be noted in your Rapid Risk Assessment form, and that might end up triggering extra requirements from an information security perspective.

jvehent:
In my initial assessment with Cory, I had forgotten about testpilot's addons installation powers. Thank you @chuckharmston for raising this point. I think that, as @marumari points out, the impact of a compromise should prevent us from allowing 3rd party scripts to run on `testpilot.firefox.com`.

r-

ckprice:
Okay. I'll open a follow-up PR to remove Optimizely.

Issue #1809, opened 2016-11-14
Automate list of dev vs prod locales?


lmorchard:
From PR #1803:
> The list of locales is available here
> https://l10n.mozilla-community.org/webstatus/api/?product=testpilot&txt
>
> Technically there's an API also to get complete locales
> https://l10n.mozilla-community.org/webstatus/api/?product=testpilot&txt&type=complete

With those resources, we *could* automate updating [the lists of locales for local/dev & prod/stage](https://github.com/mozilla/testpilot/blob/master/frontend/config.js#L6) at build time. The only caveat I'd have is whether we still want to have a manual QA & testing speed bump on enabling locales marked completed.

flodolo:
I'm not sure we want the production list to be automated. A complete locale would get out of that list when we add new strings, and we don't want that to happen.
Also worth doing a quick test of a brand new locale on stage or locally before enabling it.

As for the stage list, we could use that link or directly get a list of folders in the repo (in /locales), they should be identical. The latter doesn't introduce an external dependency, so it might be preferable.

ghost:
We could automate dev and not prod

PR #1851, opened 2016-11-21
Update locales for DEV and PROD

flodolo:
I'm adding dsb and hsb, even if they aren't available on stage yet without this PR (we need #1809 for that).

I'll send him an email to check the stage as soon as this is merged.


flodolo:
@chuckharmston r?

chuckharmston (approved):
Looks good! Will merge when CI finishes.

PR #1863, opened 2016-11-22
Ensure .properties files are copied to add-on before prod build & sign

### LINKED ISSUE #1858: Localization missing for specific elements on the Test Pilot website
[Affected versions]:
- Firefox 50.0 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "testpilot.env" string pref was manually added with "stage" value on a new clean profile in "about:config" page.
- The "experiments.webapi.testing" boolean pref was manually set to true.

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites and navigate to "https://testpilot.stage.mozaws.net" page.
2. Click the "Install Test Pilot add-on" green button in order to install the add-on.
3. Choose an available experiment and enable it.
4. Click the Test Pilot icon from the browser toolbar.
5. Observe the strings displayed in doorhanger.

[Expected result]:
- All the strings are correctly displayed with no format issue in doorhanger.

[Actual result]:
- The "experiment_list_enabled" string is displayed instead of "Enabled" for the enabled experiment.
- The "experiment_list_view_all" string appears instead of "View All Experiments" one at the end of doorhanger. 

[Notes]:
- This issue is also reproducible for Test Pilot massage displayed on the no-experiment notification prompt after 24h from Test Pilot installation.
- This reproduces on all locales including en-US default one.
- This issue could affect other strings too but some of them are unreachable to test at the moment.
- Here are the screenshots with the issue:
no experiment notification
doorhanger strings

lmorchard:
- Add `npm run addon:locales` call to `npm run sign` for add-on

- Bump add-on version to 0.9.1

Fixes #1858

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -1,7 +1,7 @@
 {
   "title": "Test Pilot",
   "name": "testpilot-addon",
-  "version": "0.9.0",
+  "version": "0.9.1",
   "private": true,
   "description": "Test Pilot is a privacy-sensitive user research program focused on getting new features into Firefox faster.",
   "repository": "mozilla/testpilot",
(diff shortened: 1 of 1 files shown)


lmorchard:
This feels a bit janky, but I think it should fix the issue. Might be a good idea in the near future to revisit & normalize how we're doing builds between dev & stage/prod because they're on different systems. This could also include #1486 to combine frontend & add-on build processes into the same gulp setup - or at least the same set of package.json scripts

chuckharmston (approved):
This seems like it probably makes sense. Let's try it!

Circle does the building on dev, so we'd need to make a new tag and push this to staging, right?

PR #1865, opened 2016-11-22
Remove installations_url for experiments

### LINKED ISSUE #1857: installations_url for experiments 404s?
Casually surfing through https://testpilot.firefox.com/api/experiments.json and I see an `installations_url` field for each experiment. If I try and ping that endpoint, I get a 404:

https://testpilot.firefox.com/api/experiments/2.json -- works
https://testpilot.firefox.com/api/experiments/2/installations/ -- no works

Not sure if we're using that `installations_url` for anything, or if we're getting the usage counts via telemetry (see https://github.com/mozilla/testpilot/blob/4ab8fdacc9f1a796115ba3dc54c95d6a9066db6a/frontend/src/app/config.js#L5)?

pdehaan:
Fixes #1857

### DIFF
--- a/addon/experiment-data.json
+++ b/addon/experiment-data.json
@@ -71,7 +71,6 @@
         "avatar": "\/\/www.gravatar.com\/avatar\/abe193c513e181a8a3ee40980928af39?s=64"
       }
     ],
-    "installations_url": "http:\/\/testpilot.dev:8000\/api\/experiments\/1\/installations\/",
     "installation_count": 1,
     "created": "2015-12-09T15:45:42.395000Z",
     "modified": "2015-12-10T22:10:11.640000Z",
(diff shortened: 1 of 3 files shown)


addon/experiment-data.json
```diff
@@ -71,7 +71,6 @@
         "avatar": "\/\/www.gravatar.com\/avatar\/abe193c513e181a8a3ee40980928af39?s=64"
       }
     ],
-    "installations_url": "http:\/\/testpilot.dev:8000\/api\/experiments\/1\/installations\/",
```
pdehaan, line 74: Not sure why all these URLs have have escaped slashes in the JSON, or if they need em, but that can be a separate bug/PR.

lmorchard: [Technically](http://www.json.org/), forward slashes *should* be escaped in JSON, but many serializers don't do it and most parsers don't care.

Issue #1771, opened 2016-11-07
Graduate Universal Search


ghost:

johngruen:
Here's a timeline: we will graduate Universal Search at 12:00 EST on November 30th, 2016. Working backward from this, we will need the following:
- [x] Nov 30th: @chuckharmston ships a version of Universal Search That's self-uninstalling. Goodnight.
- [x] Nov 23rd: Test Pilot begins to show per-retirement UI.
- [x] Nov 22nd: @clouserw pushes Test Pilot to Prod with all Universal Search graduation data in place. 
- [x] Nov 19th-22nd: @SoftVision-PaulOiegas leads QA on the graduation UI + Universal Search Add-on.
- [x] Nov 18th: Final copy for the Test Pilot graduation summary due, all final graduation assets in YAML for Sprint-14 tag @javaun
- [x] Nov 18th @chuckharmston and @6a68 provide @SoftVision-PaulOiegas with the final version of the Universal Search add-on
- [x] Nov 14th: @javaun's draft of Test Pilot graduation summary due, YAML updated with draft and completion date ahead of Sprint-14 tag. @jgruen and @chuckharmston will review and our copywriter will give it a few passes.

How does this sound to people, specifically, @javaun, @chuckharmston, @clouserw

javaun:
SGTM. I better start writing.

clouserw:
The current version of https://addons.mozilla.org/en-US/firefox/addon/test-pilot/?src=ss opens a tab and uninstalls itself.  Should just be a matter of replacing the URL and add-on ID.

chuckharmston:
We also need to ensure that prefs we changed are reverted to defaults, including one-off searches. We should probably also make sure the separate search box is restored.

ghost:
Good point.  That add-on also deletes its old prefs so that should be straight-forward too.

jaredhirsch:
> We also need to ensure that prefs we changed are reverted to defaults, including one-off searches. We should probably also make sure the separate search box is restored.

[...]

PR #1868, opened 2016-11-23
[ref #1771] Graduation page validation fixes

ckprice:
@pdehaan r? It's passing your validator now.

Here's what the page rendered looks like https://pageshot.net/2WdRDMu6PhrmcngH/testpilot.dev


pdehaan:
LGTM. 👍

Issue #1860, opened 2016-11-22
"Your Privacy" section looses localization when resizing window (mobile portrait size)


SoftVision-PaulOiegas:
[Affected versions]:
- Firefox 50.0 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Test Pilot add-on was previously installed on a new clean Firefox profile.
- A localization language other than English was added on top of the list (eg German), from "Firefox > Options > Content > Language" section.

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites and navigate to "https://testpilot.stage.mozaws.net" page.
2. Click on one of the available experiments cards.
3. After redirected to details page, scroll down to the "Your Privacy" section and observe the strings language.
4. Resize the browser window until mobile portrait view size is reproduced and observe the section strings again.

[Expected result]:
- The browser window size does not affect the strings localization.

[Actual result]:
- Step 3: The strings are correctly displayed in selected language.
- Step 4: After resize the strings loose the localization and they get displayed in English.

[Notes]:
- This issue is easily reproducible if you have Tab Center enabled with default functionality and you use your browser on half of the screen.

![your privacy looses localization](https://cloud.githubusercontent.com/assets/17258026/20528849/7108bca0-b0d6-11e6-8157-aeac6ec7791e.gif)

flodolo:
I would add that the section is in English if you start directly with the small window size, so it's not related to the resize, but to the viewport size.

flodolo:
Looks like the section doesn't have the data-l10n-id set when the window is small or resized.

SoftVision-CosminMuntean:
This issue is no longer reproducible on dev server. Marking it as verified.

PR #1870, opened 2016-11-24
Fix missing data-l10n-id in "Your Privacy" section at small viewport

flodolo:
Fixes issue #1860


flodolo:
I confess that I'm absolutely unfamiliar with the code, so I don't know why this fragment of HTML is invoked at small viewport, but I just noticed the missing data-l10n-id. Tested locally, and it works as expected.

jaredhirsch:
@flodolo Thanks for the pull request! Looks like you need to update the tests.

PR #1873, opened 2016-11-26
[fix #1869] Remove inline style from graph legend

### LINKED ISSUE #1869: Graphs legends are wrongly rendered on the Universal Search graduation page
[Affected versions]:
- Firefox 50.0 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "testpilot.env" string pref was manually added with "dev" value on a new clean profile in "about:config" page.
- The "experiments.webapi.testing" boolean pref was manually set to true.
- Test Pilot add-on was previously installed on a new profile from "https://testpilot.dev.mozaws.net".

[Steps to reproduce]:
1. Change the system date to 1 December 2016.
2. Open Firefox with the profile from prerequisites and navigate to "https://testpilot.dev.mozaws.net" page.
3. Click on "View Past Experiments" button and click the Universal Search experiment card.
4. Observe the graphs legends from graduation the page.

[Expected result]:
- The graphs legends are correctly rendered.

[Actual result]:
- The graphs legends are wrongly rendered.

[Notes]:
- Here is how the graphs legends should be rendered: https://pageshot.net/vNkLUsxtaekGC76I/testpilot.dev.
- Here is the screenshots with the issue:
universal search graduation page

ckprice:
I couldn't reproduce this issue locally, but did see it on dev on Chrome with a CSP message about inline styles (which was odd). Probably good form to try to take out inline stuff anyway.

Fyi this rule is needed to override [the `img` width setting](https://github.com/mozilla/testpilot/blob/master/frontend/src/styles/modules/_experiment-details.scss#L189) in the details section.

Fixes #1869.

### DIFF
--- a/frontend/src/styles/modules/_experiment-details.scss
+++ b/frontend/src/styles/modules/_experiment-details.scss
@@ -250,6 +250,10 @@
     figcaption {
       color: $transparent-black-7;
       font-size: $font-unit;
+
+      img {
+        width: auto;
+      }
     }
   }
 }
--- a/content-src/experiments/universal-search.yaml
+++ b/content-src/experiments/universal-search.yaml
@@ -146,8 +146,8 @@ graduation_report: >
   <figure>
   <img src="/static/images/experiments/universal-search/graduation/graph1.png" alt="Search Results Displayed and Clicked" />
   <figcaption>
-  <img style="width:36px;" src="/static/images/experiments/universal-search/graduation/graph1_key1.png" alt="Displayed" /> Displayed<br />
-  <img style="width:36px;" src="/static/images/experiments/universal-search/graduation/graph1_key2.png" alt="Clicked" /> Clicked<br /><br />
+  <img src="/static/images/experiments/universal-search/graduation/graph1_key1.png" alt="Displayed" /> Displayed<br />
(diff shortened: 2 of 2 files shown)


frontend/src/styles/modules/_experiment-details.scss
```diff
@@ -251,6 +251,10 @@
       color: $transparent-black-7;
       font-size: $font-unit;
     }
+
+    figcaption img {
```
jaredhirsch: @ckprice Two questions:

First, a minor sass syntactical question: why not just put this img block inside the above figcaption block?

```
     figcaption {
        color: $transparent-black-7;
        font-size: $font-unit;

        img {
          width: auto;
        }
      }
```

Second, why are you setting the width to auto? It looks like it was previously set to `36px`.

ckprice: > First, a minor sass syntactical question: why not just put this img block inside the above figcaption block?

Updated!

> Second, why are you setting the width to auto? It looks like it was previously set to 36px.

I was thinking it would make this rule a bit more useful if any future images that wanted to break out of [the width set in `.details-description img`](https://github.com/mozilla/testpilot/blob/master/frontend/src/styles/modules/_experiment-details.scss#L189) could use it. Like if someone wanted to add an image in the future to a caption, it could be any width.

PR #1877, opened 2016-11-28
Localize page titles

### LINKED ISSUE #1824: Page titles not localized
A quick test shows the `<title>` element not being localized:

screen shot 2016-11-16 at 5 38 21 pm

In fact, it appears that none of our page title strings are even used in the code: [pageTitleDefault](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=pageTitleDefault), [pageTitleLandingPage](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=pageTitleLandingPage&type=Code), [pageTitleExperimentListPage](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=pageTitleExperimentListPage&type=Code), [pageTitleExperiment](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=pageTitleExperiment&type=Code)

Let's either get page titles localized, or decide that it's not necessary and remove the strings.

lmorchard:
- Use data-l10n-id=pageTitleDefault for all static page titles

- setPageTitleL10N helper to change page title string

- Change to pageTitleExperiment with placeholder for experiment title on
  experiment pages

Fixes #1824

### DIFF
--- a/frontend/src/app/containers/App.js
+++ b/frontend/src/app/containers/App.js
@@ -174,6 +174,11 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
     subscribeToBasket,
     clipboard,
     userAgent: navigator.userAgent,
+    setPageTitleL10N: (id, args) => {
+      const title = document.querySelector('head title');
+      title.setAttribute('data-l10n-id', id);
+      title.setAttribute('data-l10n-args', JSON.stringify(args));
+    },
     openWindow: (href, name) => window.open(href, name),
     getWindowLocation: () => window.location,
     addScrollListener: listener =>
--- a/frontend/src/app/containers/ExperimentPage.js
+++ b/frontend/src/app/containers/ExperimentPage.js
@@ -187,13 +187,17 @@ export class ExperimentDetail extends React.Component {
   }
 
   render() {
-    const { experiment, experiments, installed, isAfterCompletedDate, isDev, hasAddon, setExperimentLastSeen, clientUUID } = this.props;
+    const { experiment, experiments, installed, isAfterCompletedDate, isDev,
+            hasAddon, setExperimentLastSeen, clientUUID,
+            setPageTitleL10N } = this.props;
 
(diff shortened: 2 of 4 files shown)


lmorchard:
This is a rather minimal fix: We already had strings for [generic page title](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl#L5) and [page title with experiment title placeholder](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl#L8). So, only experiment pages have the title actually change - all other pages should end up with the default title, though.

In the future, we might want to further customize titles on other pages - legacy, about, terms, error, retire, 404, share, etc. But, I skipped those here because we didn't have strings for those & I didn't want to guess yet

chuckharmston (changes requested):
I'm not seeing this locally; might be a race condition between l20n and our `<App>` container. Might need to [manually](https://github.com/l20n/l20n.js/tree/master/docs) do the localization?

screen shot 2016-11-29 at 8 42 57 am

pdehaan:
Unrelated, but I like how the "availableLanguages" meta tag in @chuckharmston's screenshot above starts with a ".DS_Store" locale. We'll probably want to review that `readdirSync()` code to filter directories only.

**UPDATE:** Filed as #1893, but probably not worth pursuing since it's a local/dev-only papercut.

lmorchard:
> I'm not seeing this locally

Not sure what you're not seeing... the tab title in the screenshot looks correct. The data-l10n-id is pageTitleExperiment and the data-l10n-args are the experiment data. If that's the German locale, then [the string seems correct too](https://github.com/mozilla/testpilot/blob/master/locales/de/app.ftl#L10)

chuckharmston:
> If that's the German locale, then the string seems correct too

Oh, hah. I thought it was the fallback English, never checked the string.

chuckharmston (approved):

Issue #1811, opened 2016-11-14
Script to convert existing strings to new split up strings


ghost:
The changes in https://github.com/mozilla/testpilot/pull/1786 and https://github.com/mozilla/testpilot/pull/1782 are helpful and appreciated (thanks, Chuck!).  Paragraphs are not a deal breaker for L10n, but multiple paragraphs should be broken up (as in the PRs -- the bulleted lists are a great example).  Any time we can keep HTML out of the localized strings, we should.

So, I'd like to land those PRs, but I don't want to ask the localizers who just spent all that time localizing to essentially do it all again in smaller chunks.  This bug is to look into converting at least the main parts of the .ftl files to use the new strings.  

/cc @mathjazz for any ideas since he already parses these files
/cc @flodolo for L10n interest

mathjazz:
Looking at the `.ftl` files, #1786 only _adds_ one string. If that's correct, there's nothing we need to change in locale files.

#1782 removes `<p>` tags from most of the strings containing them, so those strings need to be modified in locale files, too.

Perhaps @chuckharmston already has something in place, given he already created a zh-CN `.ftl` file. If not, one possible approach is to identify keys with stripped `<p>` tags and then [parse](https://github.com/l20n/python-l20n/blob/master/l20n/format/parser.py), strip and [serialize](https://github.com/l20n/python-l20n/blob/master/l20n/format/serializer.py) locale `.ftl` files.

ghost:
HTML was removed from tour steps in https://github.com/mozilla/testpilot/commit/1c97ab06d4b0fb2496ead7a50d50818d159cee3f

PR #1879, opened 2016-11-28
add contrib extras

### LINKED ISSUE #1875: Update Tracking Protection copy to differentiate it from the Firefox feature
@johngruen to provide copy and placement.

johngruen:
Fixes #1875


content-src/experiments/tracking-protection.yaml
```diff
@@ -86,6 +86,7 @@ contributors:
     display_name: 'Rebecca Billings'
     title: 'Senior QA Engineer'
     avatar: /static/images/experiments/tracking-protection/avatars/rebecca-billings.jpg
+contributors_extra: 'This experiment is based on Firefox Tracking Protection technology built by Mozilla employees and contributors over the past several years. Learn more about <a href="https://mzl.la/2fYneDq" target="blank">Firefox Tracking Protection</a>.'
```
chuckharmston: We shouldn't be putting URLs in strings like this, it allows localizers to break links in our site in a way we wouldn't find easily.

Could we prospectively move this section to the introduction section that we plan to have localized professionally?

chuckharmston (changes requested):

content-src/experiments/tracking-protection.yaml
johngruen: No can do, it needs to be there, unfortunately.  I thought community L10Ns could navigate URLs, no

johngruen: @chuckharmston I have a fix for this incoming.

chuckharmston: It's a bad practice, CC @flodolo to discuss further.

The right way to do it is to put the href on the element with the `data-l10n-id`, e.g.

```jsx
{contributors_extra && <p className="disclaimer"
    dangerouslySetInnerHTML={createMarkup(contributors_extra)}
    data-l10n-id={this.l10nId('contributors_extra')}
    href="https://whatever" />} 
```

Then in the string omit the argument entirely:

```
contributors_extra: 'This experiment is based on Firefox Tracking Protection technology built by Mozilla employees and contributors over the past several years. Learn more about <a>Firefox Tracking Protection</a>.' 
```

l20n will handle moving the `href` attribute to the inner link for you when localizing the page. Unfortunately, we don't have a good system for handling that in experiment content right now.

If it's a 100% hard requirement that it be here, done in this way, feel free to dismiss my review and merge. I'm just reticent to approve it when it's such a hard hole to dig out of (see #1782, #1786, #1811).

pdehaan: nit: I'm a huge jerk, so I'll also point out that it's `"_blank"`, not `"blank"`.

Somebody smarter than me (pretty much everybody reading this), should probably also skim https://dev.to/ben/the-targetblank-vulnerability-by-example and/or https://mathiasbynens.github.io/rel-noopener/ and see if we should be doing something extra w/ `rel="noopener"` or whatever.

johngruen:
@chuckharmston I think this is fixed now. thanks, we need to land this asap so if you could confirm that would be dope.

chuckharmston (approved):

chuckharmston:
This will only work with one link per section. If you're cool with that, ship it when CI passes!

johngruen:
@chuckharmston yeah, in all likelihood this is a one-off or near one-off

content-src/experiments/tracking-protection.yaml
flodolo: Why it's a bad idea:
* You put something that should not be localized in a localizable string. It makes the string more confusing (because of markup), easier to break, potentially dangerous.
* If you need to update the URL, you would need to change the entire string, while the actual string doesn't change at all.

PR #1896, opened 2016-11-30
Update HPKP headers. Fixes #1885

### LINKED ISSUE #1885: Update HPKP header
https://bugzilla.mozilla.org/show_bug.cgi?id=1315033#c4 has a new HPKP header to use

clouserw:
Just updating the HPKP headers.  Some notes:

* I know this changes the case of the header, I think that's fine
* This keeps the same, low, max-age.  I'm planning on upping that once things work as expected
* This adds a semi-colon to the end.  I think that's fine too.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -33,7 +33,7 @@ fi
 # For short-lived assets; in seconds
 MAX_AGE="600"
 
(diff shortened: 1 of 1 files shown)


lmorchard:
LGTM r+

As a follow up someday, it would be good to have some doco on why & how to generate these headers. I keep meaning to read up on it and never get around to it :/

PR #1900, opened 2016-12-01
Note new version, and my title

ianb:


pdehaan:
I'm never 100% certain on this l20n stuff, but do we need to change the YAML file _AND_ the job title in the [locales/en-US/experiments.ftl](https://github.com/mozilla/testpilot/blob/814ba1671cd1033e359ebc467d2e3215390aa256/locales/en-US/experiments.ftl#L59) file?

chuckharmston:
Ugh, you're right, that's on me for not catching that. Fix incoming.

flodolo:
Also note, that you want people to update that, you'll need a different data-l10n-id to reforce translation.

chuckharmston:
@flodolo #1902 gets that.

PR #1914, opened 2016-12-09
update regex in hosting docs

clouserw:


clouserw:
just getting this off my todo list.  we changed the regex last week

PR #1921, opened 2016-12-19
WIP DO NOT MERGE One click install

fzzzy:


addon/index.js
```diff
@@ -347,9 +347,13 @@ exports.main = function(options) {
   SharePrompt.init(settings);
   NoExperiments.init(settings);
 
+/*
```
fzzzy, line 352: This needs UX. The test pilot first run doesn't really work well with the experiment first run. If this block is uncommented, the test pilot first run page is opened in a background tab, which just seems really weird.

frontend/src/app/components/MainInstallButton.js
```diff
@@ -31,13 +35,16 @@ export default class MainInstallButton extends React.Component {
...
+    let { experimentTitle } = this.props;
     return (
       <div>
         <button onClick={e => this.install(e)}
                 className={classnames('button extra-large primary install', { 'state-change': isInstalling })}>
           {hasAddon && <span className="progress-btn-msg" data-l10n-id="landingInstalledButton">Installed</span>}
           {!hasAddon && !isInstalling &&
-            <span className="default-btn-msg" data-l10n-id="landingInstallButton">Install the Test Pilot Add-on</span>}
+            <span>
+              <span className="default-btn-msg"><span data-l10n-id="landingInstallButton">Install the Test Pilot Add-on</span> { experimentTitle && "+" } { experimentTitle }</span>
```
fzzzy, line 46: This is an amazing hack that I came up with. I changed the button text to "Install the Test Pilot Add-on + Experiment Name" which doesn't require any localization changes. However, we probably want to localize this differently, instead.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -649,19 +650,34 @@ export class ExperimentDetail extends React.Component {
...
-    enableExperiment(experiment);
-
-    sendToGA('event', {
-      eventCategory: 'ExperimentDetailsPage Interactions',
-      eventAction: 'Enable Experiment',
-      eventLabel: experiment.title
+    installAddonPromise.then(() => {
+      // XXX installAddonPromise shouldn't resolve until the add-on is ready
+      // to accept messages, instead of using a racy setTimeout here.
+      setTimeout(() => {
```
fzzzy: We shouldn't do this setTimeout here, instead waiting on a promise that resolves on line 143 of frontend/src/app/lib/addon.js, where the console.log statement is below.

frontend/src/app/lib/addon.js
```diff
@@ -138,6 +139,9 @@ function messageReceived(store, evt) {
       store.dispatch(addonActions.setInstalled(data.installed));
       store.dispatch(addonActions.setInstalledAddons(data.active));
       break;
+    case 'addon-self:installed':
+      console.log("ADDON_SELF_INSTALLED");
```
fzzzy: I think this is the right place to resolve a promise indicating all the Test Pilot add-on event handlers are installed and it is safe to install the experiment, but I haven't been able to hook everything up properly yet.

frontend/tasks/scripts.js
```diff
@@ -32,9 +32,9 @@ function shouldLint(opt, task) {
 
 gulp.task('scripts-lint', () => {
   return gulp.src([config.SRC_PATH + '*.js', config.SRC_PATH + '{app,tests}/**/*.js'])
-    .pipe(eslint())
+    /*.pipe(eslint())
```
fzzzy, line 35: This just turns off eslint in this branch so I can use console.log debug printing. It will not be in the final patch.

frontend/src/app/components/MainInstallButton.js
flodolo: I wish I had some background on this PR (don't see an issue linked), but this is really bad. 

You should not hard-code pieces of strings (+), nor assume that all languages will have the same structure as English, and produce meaningful results from this kind of concatenation.

frontend/src/app/containers/ExperimentPage.js
chuckharmston: The right way to do this would be to poll for the existence of the [`navigator.testpilotAddon` flag](https://github.com/mozilla/testpilot/blob/e209f400492e6ddc87aae15814d36447ba50569c/addon/data/set-installed-flag.js#L11), which the add-on adds on init [in a PageMod](https://github.com/mozilla/testpilot/blob/276d2021bbd474309e37971d62a7d4b07d6f2a5b/addon/lib/app.js#L43-L50).

The promise might look like this (rejecting after 10 seconds under the assumption that something went wrong; see #660,  #1227):

```js
  let i = 0;
  const interval = setInterval(() => {
    i++;
    if (window.testpilotAddon) {
      clearInterval(interval);
      resolve();
    } else if (i > 100) {
      clearInterval(interval);
      reject();
    };
  }, 100);
});
```

frontend/src/app/components/MainInstallButton.js
fzzzy: It's just a wip. We will get the localization stuff ironed out.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -649,19 +650,34 @@ export class ExperimentDetail extends React.Component {
     // Ignore subsequent clicks if already in progress
     if (isEnabling) { return; }
 
+    let installAddonPromise = Promise.resolve();
+    if (! this.props.hasAddon) {
```
meandavejustice: Nit: I don't think the space following `!` follows our eslint rules

frontend/src/app/containers/ExperimentPage.js
fzzzy: Thanks! Your code snippet had window.testpilotAddon instead of navigator.testpilotAddon, but I finally figured it out... :)

johngruen:
@fzzzy just writing this down to make sure i remember when i get back after the holiday:

This is looking really good. We should figure out a better button style, make sure we're sending the right GA pings and squash the commits before we merge but, otherwise this is awesome!

johngruen:
@fzzzy, it occurs to me we're going to have to bring the short privacy notice on the page somehow...i'll scratch my head about that. In the meantime, here's a button for the double install.

Use the `.primary` and `.button` classes. Padding is `10px 40px`. Fonts are 14px and 20px.

double install button

jaredhirsch:
@fzzzy Hi! Can this WIP PR be closed? Is there more you need to discuss?

fzzzy:
I have fixed all the feedback I got from opening this pr. I will close it and reopen a new pr once the ux has settled.

Issue #1916, opened 2016-12-14
Gzip and cache headers missing for some resources


digitarald:
Time to Meaningful Paint is about 10sec on DSL, which probably causes higher bounce rates than necessary. I talked to @ckprice about performance before about tracking this; so I'll just start with small nits:

Missing gzip: https://www.webpagetest.org/result/161214_EP_4XC6/1/performance_optimization/#compress_text
Missing caching: https://www.webpagetest.org/result/161214_EP_4XC6/1/performance_optimization/#cache_static_content

digitarald:
The 10sec are painfully highlighted in this [slow-mo video for DSL](https://www.webpagetest.org/video/view.php?id=161214_747e61f24c41c57df5a14118ee2c3e57f6d1d951&embed=1&width=520&height=448)

ckprice:
Chatted a bit with Harold. Taking this for first steps.

lmorchard:
Probably some checkboxes need checking in our stage & prod CloudFront distributions managed by ops? Would be interesting to see if this is an issue on testpilot.dev.mozaws.net, since that's a CloudFront distro we manage on the dev team

ckprice:
@relud is this something you can configure?

relud:
gzip is probably a flag on my end somewhere, but caching should be set by bin/deploy.sh

lmorchard:
TL;DR: We can knock off some of those problems WebPagetest reports, but we won't get to 100%

* We are indeed missing a `--cache-control "max-age=${MAX_AGE}"` on [deploying `*.svg` assets](https://github.com/mozilla/testpilot/blob/master/bin/deploy.sh#L108), so we can add that and resolve most of the issues.

* The [`experiments.json` resource](https://testpilot.firefox.com/api/experiments.json) *should* have a short cache time - the data changes whenever we deploy, but the URL can't change because 3rd parties and add-ons depend on it.

* We don't really want to use long-expiration on [the Telemetry counts](https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json) as that's a frequently-updated data source - and it's not under the control of TxP deployment anyway

relud:
found where to put the gzip flag. if mozilla-services/cloudops-deployment#308 is acceptable then i will deploy that to stage, and then to prod on tuesday morning or something?

lmorchard:
FWIW, I think that's a private repo I can't see

relud:
[...]

PR #1923, opened 2016-12-19
[ref #1916] Add cache-control for SVG assets

ckprice:
Ref #1916.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -31,18 +31,22 @@ fi
 
 
 # For short-lived assets; in seconds
-MAX_AGE="600"
+TEN_MINUTES="600"
+
+# For long-lived assets; in seconds
+ONE_YEAR="31536000"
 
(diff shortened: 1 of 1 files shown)


bin/deploy.sh
```diff
@@ -107,6 +107,7 @@ aws s3 sync \
 
 # SVG; cache forever, assign correct content-type
 aws s3 sync \
+  --cache-control "max-age=${MAX_AGE}" \
```
jaredhirsch: @ckprice It looks like `MAX_AGE` is a deceptively-named constant which only equals 600 seconds. To cache SVG forever, you want the bigger value found in the catch-all below (line 122), 31536000 seconds.

Here's what I'd suggest, since we're already in this file: 
- rename `MAX_AGE` to something more descriptive, like `TEN_MINUTES`, and update all existing instances
- add a new variable, maybe call it `ONE_YEAR`, with the value 31536000
- you'll also want to reassign `ONE_YEAR` to something dramatically shorter, like 60, in the "if dev" clause around line 45
- use `ONE_YEAR` in the line you added, and replace the number with the variable in the catch-all on line 122

This is a bit of a refactoring chore, but I probably would have made the same mistake, so it'll be worth the trouble. Ping me when it's ready for another look :+1:

jaredhirsch (dismissed):

ckprice:
@6a68 updates are in, thanks!

jaredhirsch:
@ckprice Looks great! Thanks for taking the time to clean things up :+1:

PR #1969, opened 2017-01-04
[fixes #1959] Update experiment page warning for 'locales' mechanism

### LINKED ISSUE #1959: Update experiment page warning for `locales` mechanism
STR:

- Add a `locales` field to experiment page YAML for `de` ([ref](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#locales)).

```
locales:
  - 'de'
```

- Set language to `en`
- Visit experiment page

Expected Result:

- Message states that the experiment is *not* available in English.

OR

- Message states that the experiment is *only* available in German.

Actual Result:

- Message states that the experiment is *only* available in English.

Looks like this string has been hard-coded to English regardless of what's in the `locales` field.

ckprice:
Thought about using `titleL10nArgs` with the `locale` property, but I couldn't find any nice map between 2-digit locale, and pretty language name.

/cc @flodolo (for routing string change), @chuckharmston (in case there's a better way)

Fixes #1959.
~~Ref 1968 (if this lands in time, we should add `locales` to Cliqz)~~


flodolo:
I don't think there's a safe way to convert a locale code to a pretty language name.

Examples: in Italian it would be "German" at the beginning of a sentence, "german" in the middle. Locales with grammar cases would need a different declination depending on the role of the noun, e.g. "German is great" (nominative) vs "I like German" (accusative).

The easiest way out is "This page is not supported for your language (LOCALE_CODE)". Not pretty, but safe.

flodolo:
```
localeUnavailableWarningTitle = This experiment is not available in {{lang}}.
```
Jared suggestion is great in theory, but based on my experience it will fail badly, because some localizers don't pay enough attention to comments. They will keep the placeholder, and we put "English" they'll translate English.

ckprice:
@flod thanks! Agreed it's not pretty, but it will be the most consistent. Updated.

jaredhirsch (approved):
@ckprice Thanks for opening this PR. This does fix the case where I'm viewing the Cliqz experiment page with any non-German locale.

However, if I view the Activity Stream page with the `de` locale, no warning is shown. Should #1959 reflect the blocklist as well?

Also, is there any reason we need both `locales` and `locale_grantlist`? Would those two ever differ? If not, I think it'd be good to just stick with one.

Since Cliqz is the main thing to worry about short-term, I'm ok to land this now and file followup bugs to add blocklist/grantlist support, and remove the redundant whitelist yaml keys.

ckprice:
> However, if I view the Activity Stream page with the de locale, no warning is shown. Should #1959 reflect the blocklist as well?

Thanks @6a68, this does make sense. Updated.

> Also, is there any reason we need both locales and locale_grantlist? Would those two ever differ? If not, I think it'd be good to just stick with one.

Probably not, but I didn't address in this PR.

ckprice:
I think we should hold on this since it's a bit late in the game to try and change a string everywhere. Going to block until post launch.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -157,12 +157,13 @@ export class ExperimentDetail extends React.Component {
 
   renderLocaleWarning() {
     const { experiment, locale } = this.props;
-    if (experiment.locales && locale && !experiment.locales.includes(locale)) {
+    if (locale && ((experiment.locales && !experiment.locales.includes(locale)) || (experiment.locale_blocklist && experiment.locale_blocklist.includes(locale)))) {
```
jaredhirsch, line 160: @ckprice Is there a better place we could put this filter code? Maybe reuse the filters in `src/app/selectors/experiment.js`?

jaredhirsch (changes requested):
If there's now no time pressure to land this before Cliqz launches, I think it would be better to reuse the existing selectors than use an inline check.

Maybe, in the style of [l10nSelector](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/selectors/experiment.js#L40-L42), we could just have a general `validExperimentSelector` that, given a locale and a single experiment, checks if the experiment should be visible to the user (checks launch_date and locale)? I'm not sure this is the most idiomatic React-ish implementation advice, but it would keep the experiment validity logic in one place.

frontend/src/app/containers/ExperimentPage.js
ckprice: @6a68 maybe. But this PR is only rendering a warning message to users, so it's not excluding it from that user's list. Do you think it still makes sense for a selector here? Like `renderLocalWarningSelector`, then test to see if the experiment exists in that collection? Seems like a bit of overhead.

jaredhirsch: We can just land this, it's been sitting long enough. It's minor duplication at most.

jaredhirsch (approved):

PR #1978, opened 2017-01-06
revert package upgrades

clouserw:
Upgrading our packages messes up our deployment (which isn't tested on PRs, but is after the code is merged).  It's 5pm, so here is a patch to revert back to a stable master.


clouserw:
For posterity, here are the failures with the new packages:  https://circleci.com/gh/mozilla/testpilot/2410

PR #1984, opened 2017-01-06
Trigger the 'new' badge based on the locale/launch filtered experiment list

### LINKED ISSUE #1973: Toolbar button 'new' label should be locale-aware
Some new experiments might not be available in the user's locale, so a 'new' label may be incorrectly shown over the toolbar button.

Given that we're going to remove the doorhanger, the fix is probably to wait till that is done, then just relocate the filtering code into the part of lib/toolbar-button that checks experiment freshness.

So, I guess this is related to #1922.

jaredhirsch:
Unfortunately, the 'new' badge isn't recalculated dynamically when the user's language preference changes. But it will only trigger in the case where the language isn't blocked and the launch_date is in the future. We _could_ dynamically recalculate the 'new' badge, but that would expand the scope of this patch.

Fixes #1973.

### DIFF
--- a/addon/lib/toolbar-button.js
+++ b/addon/lib/toolbar-button.js
@@ -206,17 +206,18 @@ const ToolbarButton = module.exports = {
     }
 
     // Look through available experiments for anything newer than the last
-    // toolbar button click.
-    let hasNew = false;
-    if (store.availableExperiments) {
-      Object.keys(store.availableExperiments).forEach(id => {
-        const experiment = store.availableExperiments[id];
-        const created = new Date(experiment.created);
-        if (created.getTime() > store.toolbarButtonLastClicked) {
-          hasNew = true;
-        }
-      });
-    }
+    // toolbar button click. The list of experiments is filtered based on
+    // the criteria used by `getExperimentList`.
+    const availableExperiments = getExperimentList(
+      store.availableExperiments || {},
+      store.installedAddons || {});
+    // If some experiment is new, show the new badge.
+    const hasNew = availableExperiments.some(experiment => {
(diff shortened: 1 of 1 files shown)


addon/lib/toolbar-button.js
```diff
@@ -206,17 +206,18 @@ const ToolbarButton = module.exports = {
...
+    // the criteria used by `getExperimentList`.
+    const availableExperiments = getExperimentList(
+      store.availableExperiments || {},
+      store.installedAddons || {});
+    // If some experiment is new, show the new badge.
+    const hasNew = availableExperiments.some(experiment => {
+      const created = new Date(experiment.created);
+      const launched = experiment.launch_date && new Date(experiment.launch_date);
+      return (created.getTime() > store.toolbarButtonLastClicked) ||
+             (launched.getTime() > store.toolbarButtonLastClicked);
```
dannycoates: `launched` might not be a Date here

pdehaan: I'm a bit confused on this line.
I _think_ line 217 above is saying `experiment.launch_date &&` which looks like it could almost return a false/undefined if `launch_date` wasn't somehow defined for an experiment.
Which if that boolean logic fails, and `launched` doesn't get set to a Date object, this line could fail as `.getTime()` would be trying to access a non-Date object here.

Maybe?

jaredhirsch: :+1: thanks

jaredhirsch: yup, dcoates found the same potential bug :+1:

dannycoates (approved):

Issue #1977, opened 2017-01-05
Update screenshot description on Cliqz page


ckprice:
We received some feedback on the Cliqz page, looks like this was missed during the general copy review.

> Thank you, looks good. Just a minor edit to the text for screenshot 2: It should read
> 
> Sehen Sie in Echtzeit zusammengefasste Informationen, zum Beispiel über das Wetter, direkt unter der Adressleiste.
> 
> ((Removed "und Sendezeiten“ = "and showtimes". We do not offer this feature.))

I think we'll need to modify the YAML as referenced in https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization

/cc @flodolo 

Ref #1968.

flodolo:
Argh. Since this string is basically targeting only German, I suggest to fix the string without a new ID, I'll fix the localization in Pontoon in the meantime.

flodolo:
@koehlermichael and @Archaeopteryx
https://pontoon.mozilla.org/de/test-pilot-website/all-resources/?search=zusammengefasste&string=160975

See comment above, the original string is wrong (no showtimes). I was going to change it directly, but it seems quite different.

Current: "Sehen Sie zusammengefasste Informationen – wie Wetter und Sendezeiten – in Echtzeit direkt unter der Adressleiste."

Proposed: "Sehen Sie in Echtzeit zusammengefasste Informationen, zum Beispiel über das Wetter, direkt unter der Adressleiste."

Can you update the translation in Pontoon?

koehlermichael:
@flodolo @Archaeopteryx

I just made the change in Pontoon though the incorrect original string is still visible there. Flod, your translation works fine, nice work as usual.

flodolo:
Thanks Michael, I see it just landed in the repo
https://github.com/mozilla/testpilot/commit/1b03375d8272c09693cb4697190e14ddd813dd90

johngruen:
@koehlermichael thanks! I'll tweak the original string now!

ckprice:
Fixed in #1986 and 1b03375

PR #1986, opened 2017-01-06
update image desc and create date

johngruen:
#1977


content-src/experiments/cliqz.yaml
```diff
@@ -96,5 +96,5 @@ contributors:
     display_name: 'Cliqz GmbH'
     avatar: /static/images/experiments/cliqz/avatars/cliqz-avatar.jpg
 installation_count: 1
-created: '2017-01-11T00:06:00.00Z'
+created: '2016-12-01T00:06:00.00Z'
```
johngruen, line 99: @ckprice i set this to an inoffensive date for people who have old versions of the add-on

ckprice (approved):

content-src/experiments/cliqz.yaml
ckprice: FYI to @6a68 I think we are going to do this so we don't have to rely on any late translations of "New" for our launch on Tuesday.

jaredhirsch: :+1:

PR #1988, opened 2017-01-06
negotiated a ceasefire between l20n and react on the experiment page

### LINKED ISSUE #1889: "TypeError: t is null" when disabling an experiment
Found in https://testpilot.stage.mozaws.net/__version__ (e42d9cb5ec4e75f45fe6123ade507bce14037fea):

```json
{ "commit": "e42d9cb5ec4e75f45fe6123ade507bce14037fea",
  "version": "2016-11-29-1",
  "source": "https://github.com/mozilla/testpilot"}

```

---

I was testing uninstalling the Universal Search add-on and as soon as the "Vielen Dank!" dialog popped up, my DevTools console spewed "TypeError: t is null" error, citing some error on line 8 of the minimized vendor.ec3d1ad6.js file. I tried tracking that back to some unminimized source. Not sure if we are generating/linking source maps to make debugging minified source easier.

firefox_test_pilot_-_universal_search

dannycoates:
fixes #1889 ... for now

Anytime we have `data-l10n-id` and react templated text in that tag we're going to have problems.

We need a better way

### DIFF
--- a/frontend/src/app/containers/ExperimentPage.js
+++ b/frontend/src/app/containers/ExperimentPage.js
@@ -539,7 +539,7 @@ export class ExperimentDetail extends React.Component {
     if (hasAddon && !this.isValidVersion(min_release)) {
       return (
         <div className="upgrade-notice">
-          <div data-l10n-id="upgradeNoticeTitle" data-l10n-args={JSON.stringify({ title, min_release })}>{title} requires Firefox {min_release} or later.</div>
+          <div data-l10n-id="upgradeNoticeTitle" data-l10n-args={JSON.stringify({ title, min_release })}></div>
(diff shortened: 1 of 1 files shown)


lmorchard:
> We need a better way

IIRC, the L20N advice is to never have text in DOM elements and just let L20N handle it. Then, React only ever sets `data-l10n-args` properties. This gets tricky, though, if we ever want to generate static pages via React components unless we also get L20N working in the static generation build too.

dannycoates:
> the L20N advice is to never have text in DOM elements and just let L20N handle it

That's my understanding as well. I've got mixed feelings about doing that...

If I had my druthers I'd only have react managing the dom and have it "sub out" translations as needed. Whether as a component, function call, or something else I don't know.

lmorchard:
LGTM

PR #2002, opened 2017-01-09
Remove the doorhanger with a quick hack.

fzzzy:
Switch to an ActionButton instead of a ToggleButton, and open settings.BASE_URL on click.


dannycoates (approved):
Just the lint, then LGTM!

Issue #1989, opened 2017-01-06
Send locale when submitting footer newsletter form


ckprice:
We fixed the install welcome email dialog in #1987. This is opened to add it to the newsletter form email.

Ref: https://github.com/mozilla/testpilot/blob/master/frontend/src/app/actions/newsletter-form.js#L17

ckprice:
Closed in #2003

Issue #1987, opened 2017-01-06
Send correct locale when submitting welcome email 


ckprice:
This could be a thing that blocks our launch.

STR:

- Uninstall Test Pilot
- Set language to de
- Browse to dev
- Install Test Pilot
- Submit welcome email modal

Expected Result:

- You are registered in email system as a German speaker, and receive German version of welcome email

Actual Result:

- Receive English version of welcome email

You can test multiple times by adding "+" after your username, e.g. cprice+20160106@mozilla.com.

I think we have to send a locale here: https://github.com/mozilla/testpilot/blob/b3ce96502f5f93113a66b662ac57b2197d10d213/frontend/src/app/actions/newsletter-form.js#L24

ckprice:
We should either fix this, or suppress the dialog for launch.

ckprice:
Including `lang=de` to the basket call fixes this. Ref: http://basket.readthedocs.io/newsletter_api.html#news-subscribe

jaredhirsch:
@ckprice Do we have a German-localized newsletter? That is, is this bug about disabling the newsletter dialog, or adding the `lang` parameter to the basket call?

ckprice:
@6a68 working with wbowden to confirm one is ready. I've confirmed with him that if we send `lang=de`, the record comes into the email platform as DE. So to answer your question, we're looking more like this is going to do the fix (adding `lang` param)

jaredhirsch:
Also, are we just going to have one German newsletter, or will there be one per region (`de-de` vs `de-ch`)?

ckprice:
> Also, are we just going to have one German newsletter, or will there be one per region (de-de vs de-ch)?

Just one (de).

Issue #1758, opened 2016-11-03
Use NewsletterForm component in EmailDialog


pdehaan:
In the spirit of easier localization, would it make sense to use the same strings for the two email signup forms?

![window_and_firefox_test_pilot](https://cloud.githubusercontent.com/assets/557895/19989237/60c66020-a1e2-11e6-9ba1-443f6a087389.png)

### Popup:

- `email goes here :)`
- `Sign me up`

### Footer:

- `YOUR EMAIL HERE`
- `Sign Up Now`

chuckharmston:
Then plan is actually to move the modal to use the [`<NewsletterForm>` component](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/NewsletterForm.js); just haven't gotten around to it yet. Actually, I'm not sure there's a bug for it yet. I'll hijack this one, thanks.

chuckharmston:
Make sure to delete all the unnecessary strings when doing this; there are quite a few only used by this component.

johngruen:
Not going to get to this today...I'll save it for when i want to hide in a corner for awhile in hawaii.

PR #2003, opened 2017-01-09
add lang param to footer email

johngruen:
#1989

### DIFF
--- a/frontend/src/app/actions/newsletter-form.js
+++ b/frontend/src/app/actions/newsletter-form.js
@@ -14,14 +14,14 @@ const actions = createActions(
 );
 
 const subscribeActions = createActions({
-  newsletterFormSubscribe: (dispatch, email) => {
+  newsletterFormSubscribe: (dispatch, email, locale) => {
     dispatch(actions.newsletterFormSetSubmitting());
     fetch(basketUrl, {
       method: 'POST',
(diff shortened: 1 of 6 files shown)


chuckharmston:
Note that this doesn't include the modal email form; see #1758.

chuckharmston:
Needs tests.  😑

johngruen:
@chuckharmston that got fixed here #1987 (re your first comment)

chuckharmston:
You'll need to add `locale` to `View.propTypes`, `NewsletterFooter.propTypes`, and `NewsletterForm.propTypes`.

johngruen:
@chuckharmston should i set a default?

chuckharmston:
> @chuckharmston should i set a default?

Nah, just make sure it's a required prop; it should be [set automatically](https://github.com/mozilla/testpilot/blob/a4b08eaafefe15db6720ce3b8b0eba2952ca240b/frontend/src/app/reducers/browser.js#L9).

Speaking of that bit, it's mutating `navigator.language`. What does Basket want out of this? It may be that instead of using a prop, we should use `(navigator.language || '')`.

chuckharmston:
Tests look fine ~; pending the screenshot I requested in IRC~, r+wc

PR #2016, opened 2017-01-12
add en locale

### LINKED ISSUE #2006: The Testpilot add-on, and the testpilot website appear in the wrong locale.
My browser language is set to English (en-GB).
My operating system, (Windows 10) is set to English.
"Choose your preferred language for displaying pages" is set to en-GB.

yet, both the add-on and https://testpilot.firefox.com/ appear in _Dutch_, without any option to change the language.

johngruen:
@lmorchard @clouserw this is a bit of a hack to fix #2006, but it seems expedient and I'd rather not wait for our next deployment to get a fix in. If you like we can merge this and keep the issue open.

### DIFF
--- a/frontend/tasks/content.js
+++ b/frontend/tasks/content.js
@@ -22,6 +22,17 @@ gulp.task('content-experiments-data', function generateStaticAPITask() {
     .pipe(gulp.dest(config.DEST_PATH));
 });
 
+// The next two tasks are hacks to copy the en-US locale 
+// to an en locale. See issue #2006 for background.
+gulp.task('content-watch-en', () => {
+   gulp.watch('./locales/en-US/**/*', ['content-build-en']);
+});
+
+gulp.task('content-build-en', () => {
+  return gulp.src('./locales/en-US/**/*')
+    .pipe(gulp.dest('./locales/en'));
+});
+
 gulp.task('import-api-content', (done) => {
   fetch(config.PRODUCTION_EXPERIMENTS_URL)
     .then(response => response.json())
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ media/
 node_modules/
 static/
 tmp/
+locales/en/
 
 # Files to ignore within specific directories
 addon/*.rdf
--- a/frontend/config.js
+++ b/frontend/config.js
@@ -8,9 +8,10 @@ module.exports = {
   ENABLE_DEV_CONTENT: (process.env.ENABLE_DEV_CONTENT === '1'),
   AVAILABLE_LOCALES: (process.env.ENABLE_DEV_LOCALES === '1') ?
     // All locales on Pontoon for local & dev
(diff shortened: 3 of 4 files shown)


chuckharmston:
@johngruen you should be able to symlink `en-US` to `en`, right?

flodolo:
> @johngruen you should be able to symlink en-US to en, right?

Exactly. I haven't tested it, but we should just add the symlink if that works.

frontend/tasks/content.js
```diff
@@ -22,6 +22,17 @@ gulp.task('content-experiments-data', function generateStaticAPITask() {
...
 
+// The next two tasks are hacks to copy the en-US locale 
+// to an en locale. See issue #2006 for background.
+gulp.task('content-watch-en', () => {
+   gulp.watch('./locales/en-US/**/*', ['content-build-en']);
+});
+
+gulp.task('content-build-en', () => {
+  return gulp.src('./locales/en-US/**/*')
+    .pipe(gulp.dest('./locales/en'));
```
pdehaan, line 33: So we're copying ./locales/en-US/\* into ./locales/en/\*, but I see that we have a bunch of committed locales/en/*.{properties,ftl} files below. Since we're copying these assets in the `gulp build` step, should these files be committed to the repo?

ghost:
s3 doesn't support symlinks, right?

johngruen:
@clouserw nope: http://stackoverflow.com/questions/29459885/does-amazon-s3-support-symlink#29462540

@pdehaan kind of a wash...all the other locales are committed. I set up the gulp tasks to copy if changes are saved into the yaml or en-US.ftl file rather than when the app starts.

lmorchard:
Starting to take a look... $0.02 on "just use symlinks" though: probably a bad idea - not a feature of S3; supported by Git on *nix machines but problematic on Windows. I'd lean toward an explicit copy during build, even if it looks like a redundant hack for now

lmorchard (changes requested):
Yeah, agreeing with pdehaan: Should add `locales/en/*` to `.gitignore` and not check those files in, since they're being generated via copy from a build task.

johngruen:
kk

frontend/tasks/content.js
johngruen: Should be fixed in my latest tweaks

johngruen:
@lmorchard should be good now

lmorchard (approved):

addon/tasks/locales.js
```diff
@@ -1,4 +1,4 @@
-const gulp = require('gulp');
+qconst gulp = require('gulp');
```
lmorchard: Oh, whoops! Typo here.

lmorchard (changes requested):
Looked good, but noticed the test failure - one 'q' typo

lmorchard (approved):
Actually, since disapproval blocks this, I'll just approve - r+ and merge after typo fix.

Issue #1770, opened 2016-11-07
Add flow to entice zero-experiment-users to enable an experiment


johngruen:
@sevaan will provide UX here ASAP.

sevaan:
UX: https://mozilla.invisionapp.com/share/Q77S1H6ZC#/203225298_No_Experiments_Enabled_After_N

chuckharmston:
Few questions about the UX:
1. The link in the heartbeat banner appears to be a version of our experiment listing page with slightly different copy. Is that the only way you can see that copy? Or is it what we show all users with no experiments?
2. Minor nit, but there isn't enough contrast in the "Let us know why." link color; it'd fail every accessibility test.
3. Where does that link go? Discourse?
4. How would you like this instrumented @ckprice?

sevaan:
1) Yes, the only way you would see the copy is by going through the heartbeat info bar.

@johngruen What do you think of making this what all users without experiments see when they go to the Test Pilot landing page after, say, a week?

2) Agreed. I’ve updated it.

3) Not sure, I think a SurveyGizmo survey was what @johngruen mentioned. Could you clarify, John?

sevaan:
Orange for link: #FF9D16

johngruen:
@chuckharmston
- I think we should do it for all new users who haven't installed an experiment after one day.
- Obviously, when we ship there will be a sizable cohort who are already in Test  Pilot, but have no experiment installed. We should trigger this for those folks also.
- We should only trigger this thing once to see how it works, so there should be a flag set to ensure that it doesn't happen again until we do something to change that.
- looks like @ckprice filed the metrics issue

johngruen:
@rayborn In this issue we'd like to add a survey to test pilot to ask users not involved in any Test Pilot experiments why they aren't interested. In order to complete this work @chuckharmston will need to point users to a Survey Gizmo URL. Can you provide this?

ghost:
Great idea, sorry for being slow getting you the link.

[Test link](https://qsurvey.mozilla.com/collab/TxP-User) - for full survey click "I didn't like the experiments".  We don't dig into the individual experiments much, but we ask about: 
- Reason for not being in an experiment
- Satisfaction
- If they are interested in future experiments
- What experiments they tried
- What they didn't like in general about experiments
- General feedback
- Tell them how to uninstall if they're in a group highly likely to not want Test Pilot based on not liking it, not being interested, and not being aware of it.

[...]

PR #2023, opened 2017-01-17
fix miscolored button on win-10 addon

### LINKED ISSUE #2007: Button label in feedback bar has wrong color on Windows
The button label for taking a survey is really hard to read on Windows (10). I assume that either the label should have been black or that the button color should have been blue.
snip_20170110151452

johngruen:
Fixes #2007 

Uses an inset box-shadow to git around windows weird dislike of colored XUL buttons

### DIFF
--- a/addon/lib/survey.js
+++ b/addon/lib/survey.js
@@ -217,7 +217,7 @@ function showSurveyButton(options) {
     }
     const button = box.getElementsByClassName('notification-button')[0];
     if (button) {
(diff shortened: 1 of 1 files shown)


johngruen:
screen shot 2017-01-17 at 10 08 42 am

chuckharmston:
I don't think it's a winning battle to try to fight native platform styles for these things; for #1770 @sevaan and I just abandoned the button CSS entirely. Probably worth considering here.

johngruen:
meh, technique works, button stands out more...I'm not sure i understand the downside when the Default botton style is far less eye-catching

jaredhirsch:
Looks like this PR has gone stale. I'm inclined to just merge this and see what happens, so, here we go

Issue #1807, opened 2016-11-14
The redirect to experiments page is not performed after Test Pilot add-on is installed (dev server)


SoftVision-CosminMuntean:
[Affected versions]:
- Firefox 49.0.2 and up

[Affected Platforms]:
- All Windows
- All Mac

[Prerequisites]:
- The "testpilot.env" string pref was manually added with "dev" value on a new clean profile in "about:config" page.

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites and navigate to "https://testpilot.dev.mozaws.net" page.
2. Click the "Install the Test Pilot Add-on" green button and install the add-on.
3. Observe the browser behavior.

[Expected result]:
- Test Pilot add-on is successfully installed and you are redirected to the experiments page.

[Actual result]:
1. On latest Firefox release:
- After the browser is restarted, Test Pilot add-on is successfully installed, but the installation page is still available. 
2. On latest Nightly build:
- Test Pilot add-on is successfully installed, but the button remain in the "Installing" state.

[Notes]:
- The continuous "Installing..." state issue also occurs on experiment details page where you are not redirected to the page from which you can enable the experiment.
- If the page is refreshed, you are correctly redirected to the experiments page. 
- Here are the browser console errors:
1. On the latest Firefox release: https://goo.gl/Ui1JkU.
2. On the latest Nightly build: https://goo.gl/31CVuI
- Here is a screen recording with the issue: https://goo.gl/AQ3Lmz.

chuckharmston:
@SoftVision-CosminMuntean Can you verify that this is still reproducible if `extensions.webapi.testing` is set to `true`?

Any time you're testing the add-on installation in dev or stage, that needs to be set.

chuckharmston:
Now that we're in Firefox 50, any time you see the doorhanger on install (as you would if `extensions.webapi.testing` is either unset of `false`), that's probably the issue.

SoftVision-PaulOiegas:
@chuckharmston Cosmin is out of office today but I can help here. I've added manually the pref since it's not by default or installed with Test Pilot but this doesn't change the behavior. 
However I can give you some extra bits of information on this. Since after you refresh the page you are correctly redirected after all, uninstalling and reinstalling the add-on after, doesn't display this behavior again.

ckprice:
Cosmin to check the pref and confirm repro still.

SoftVision-PaulOiegas:
[...]

PR #2034, opened 2017-01-18
Allow installation of individual experiments without requiring the Test Pilot add-on to already be installed.

fzzzy:
This has now been properly localized and the tests and linter pass, so it should be ready to land.

John says there are some things we are going to have to change after this lands to satisfy legal, but he is ok with doing them in separate commits since it is still early in the sprint.

### DIFF
--- a/frontend/src/app/containers/ExperimentPage.js
+++ b/frontend/src/app/containers/ExperimentPage.js
@@ -276,10 +276,11 @@ export class ExperimentDetail extends React.Component {
                 <h2 className="banner">
                   <span data-l10n-id="experimentPromoHeader" className="block">Ready for Takeoff?</span>
                 </h2>
-                <p data-l10n-id="experimentPromoSubheader">We're building next-generation features for Firefox. Install Test Pilot to try them!</p>
+                <p data-l10n-id="experimentPromoSubheader"></p>
                 <MainInstallButton {...this.props}
                                    eventCategory="HomePage Interactions"
-                                   experimentTitle={title} />
+                                   experimentTitle={title}
+                                   installCallback={ this.installExperiment.bind(this) } />
               </div>
             </div>
           </div>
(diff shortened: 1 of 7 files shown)


frontend/src/app/containers/ExperimentPage.js
```diff
@@ -276,10 +276,11 @@ export class ExperimentDetail extends React.Component {
                 <h2 className="banner">
                   <span data-l10n-id="experimentPromoHeader" className="block">Ready for Takeoff?</span>
                 </h2>
-                <p data-l10n-id="experimentPromoSubheader">We're building next-generation features for Firefox. Install Test Pilot to try them!</p>
+                <p data-l10n-id="experimentPromoSubheader">We&apos;re building next-generation features for Firefox. Install Test Pilot to try them!</p>
```
flodolo: This looks strange: there's a change in the template but not in the FTL.

Is the text still localized if you switch to a different locale, like Italian or German? If it's not, I would suggest to drop this change.

flodolo: Never mind, I forgot that what counts is the data-l10n-id. Given that the best practice would be to leave the text empty, it doesn't look like a relevant change.

dannycoates:
`addon/package.json` could use a patch version bump.

*note to self - remove onboarding from x16*

frontend/src/app/containers/ExperimentPage.js
fzzzy: The dangling apostrophe was just screwing up diffs on github. I think a better change is just to remove the text instead.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -648,20 +649,57 @@ export class ExperimentDetail extends React.Component {
...
+      finishEnabling();
+      return;
+    }
+
+    installAddonPromise.then(() => {
+      return new Promise((resolve, reject) => {
+        let i = 0;
+        const interval = setInterval(() => {
+          i++;
+          if (window.navigator.testpilotAddon) {
```
dannycoates, line 693: This looks good. In the future it'd be interesting to see if we can get away with removing this flag and use `mozAddonManager.getAddonByID('@testpilot-addon')`

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -648,20 +649,57 @@ export class ExperimentDetail extends React.Component {
...
-      eventLabel: experiment.title
-    });
+      sendToGA('event', {
+        eventCategory: 'ExperimentDetailsPage Interactions',
+        eventAction: 'Enable Experiment',
+        eventLabel: experiment.title
+      });
+    }
+
+    if (installAddonPromise === null) {
```
dannycoates, line 683: I wonder if setting `installAddonPromise` to `Promise.resolve()` by default and removing this would be cleaner? It would force the test loop below to run though.

meh. its fine either way

dannycoates (approved):
Awesome stuff 👍🏻

I noticed a few UI oddities around uninstalling from `about:addons` while testing this, but they're unrelated to this.

frontend/src/app/containers/ExperimentPage.js
fzzzy: This is actually exactly how I originally implemented it, but the existing test requires that `enableExperiment` get called synchronously, so I switched to this which causes the existing test to pass, and also allows for enableExperiment to be called asynchronously when the test pilot add-on wasn't already installed.

SoftVision-CosminMuntean:
Hi guys,

We started testing this functionality on dev server, but from the start we hit an old issue that makes our tests unsure. Issue #1807 is still reproducible on dev server and affects testing the new one click install & enable functionality. 

When you are on an experiment page and choose to install both Test Pilot and the experiment, Test Pilot is successfully installed but the installation page remains hanging. 
If the page is refreshed you are correctly redirected to the experiment page but the experiment is not enabled. The issue still persists even if the "experiments.webapi.testing" boolean pref is created and set to "true".

In order to install both the add-ons, you need to uninstall test Pilot and redo the installation steps. But this also is not working every time.Sometimes the experiment remains in the enabling state even if the Test Pilot installation issue is bypassed. Here are the browser console errors: https://goo.gl/hovQMS.

Here is a screen recording with the issue: https://goo.gl/kIK8Ts.

fzzzy:
Note that the pref is "extensions.webapi.testing", not "experiments.webapi.testing"

fzzzy:
I had wil push to stage, and it works a little more reliably there, but still seems a bit flaky. We may be able to improve the way we check to see if test pilot has been installed, or we may just have to figure out the actual root cause of #1807 and fix that. It seems like sometimes the web page does not receive messages sent from the add-on, but it is very intermittent.

Issue #2051, opened 2017-01-24
Experiment Install sometimes hangs with one-click install method


johngruen:
I'm seeing a high rate of add-on install failure with the new one-click method. Noticed this more on Page Shot and Min Vid, though that might be a coincidence. No errors in the web console, but the browser console is showing this: 

<img width="1012" alt="screen shot 2017-01-24 at 11 30 18 am" src="https://cloud.githubusercontent.com/assets/3323249/22256571/2b3e6ce0-e229-11e6-8e3e-51c406c30da4.png">

fzzzy:
Are you testing locally, on dev, or on stage?

johngruen:
locally

dannycoates:
Here's my theory...

The addon takes a certain amount of time to initialize. On startup it makes a request to get the experiments.json from the server. Before that request is finished it knows nothing, so if an `install-experiment` comes in during that time, [this code](https://github.com/mozilla/testpilot/blob/master/addon/index.js#L159) will return false and just ignore it.

johngruen:
@dannycoates did a test where i disabled my internet before clicking the add-on install. I see an actual error in this case, not just an infinite barberpole
<img width="1305" alt="screen shot 2017-01-24 at 2 58 32 pm" src="https://cloud.githubusercontent.com/assets/3323249/22264392/c342347e-e245-11e6-82a7-0dc6abfc704d.png">

PR #2055, opened 2017-01-24
wait on hasAddon for single-click-install

dannycoates:
helps #2051

In my testing `navigator.testpilotAddon` becomes true several ticks before `props.hasAddon`. Now, reading the code, it still looks *possible* for there to be a window where `install-experiment` might not work. Guaranteeing the timing works 100% will require a larger patch, but this has turned the issue from easily reproducible to  not reproducible for me.


fzzzy:
This is great! r+

dannycoates:
@johngruen give this a shot

johngruen:
@dannycoates cool will pull down in a sec

Issue #1945, opened 2016-12-22
Cannot navigate directly between experiments


jaredhirsch:
STR
1. visit TxP website (without addon installed)
2. visit universal-search page
3. scroll to bottom, click on link to any other experiment, say, min-vid

Expected: the min-vid page should be rendered correctly.

Actual: the min-vid title and image are shown, but most of the page is not updated: the universal-search team and description (graduation report) are shown, and the list of experiments at the bottom of the page excludes universal search, not min-vid.

jaredhirsch:
This only happens when the addon is not installed.

SoftVision-PaulOiegas:
This is already fixed on dev by #1920. Graduated experiments should not be displayed without TxP add-on installed. I think we can close this one if you don't see any remaining cases when this could occur.

jaredhirsch:
> Graduated experiments should not be displayed without TxP add-on installed

I'm confused. Why would we bother writing up a graduation report if we want to hide past experiments from new users? Reopening.

jaredhirsch:
I'm going to move the product question to a separate bug and leave this open as a bug about the app not updating the page when navigating away from a graduation report.

johngruen:
So this problem is actually much bigger than what jared saw... it seems like we cannot navigate directly between individual experiments. This is happening on dev...

## STR

- go to any experiments details page
- navigate to another experiment page by click on the experiment cards at the bottom of the page


## Results 

- URL matches new experiment's slug but previous page content is in experiment details

## Expected

- You go to the page of the experiment you clicked on. 

## Notes

So i was fiddling around trying to fix this tonight, but didn't have much luck. With some console logging, it looks like that when you navigate to the experiment page from the home page or the experiments list page, the render method for `ExperimentPage` is called twice. The first time the props aren't available to be passed in and the second time they are (this happens instantly so it isn't noticeable unless you log it) so the page renders correctly. If you go from experiment to experiment to experiment, render only gets fired once per navigation event, and only the prop-less render gets called. That's why @6a68 saw the initial bug he reported here where the previous page's content stayed visible.

## Fix

I'm a little green WRT Redux component life cycle, routing, etc, so I'm not sure the best way to fix the core issue. We could just take out the footer cards for the next sprint if this is tricky to figure out. I bet its like a 2 line fix though if somebody knows what they're doing. 

[...]

PR #2063, opened 2017-01-25
 past experiments on home page

### LINKED ISSUE #1951: Past experiments should be visible without addon installed
Looks like the 'view past experiments' button is gone from dev, so there's no way for new users to see the universal search graduation report.

We also need to make past experiments visible from the single experiment details page, so users can browse to all the experiments regardless of which page they are on.

Split off from #1945.

johngruen:
@dannycoates could you take a look since this is mostly refactoring your code?

Fixes #1951

### DIFF
--- a/frontend/test/app/components/PastExperiments-test.js
+++ b/frontend/test/app/components/PastExperiments-test.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import PastExperiments from '../../../src/app/components/PastExperiments';
+
+describe('app/components/PastExperiments', () => {
+
+  const pastExperiments = [{ title: 'foo' }, { title: 'bar' }];
+  let subject;
+
+  beforeEach(function() {
+    subject = shallow(<PastExperiments pastExperiments={ pastExperiments } />);
+  });
+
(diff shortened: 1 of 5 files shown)


frontend/test/app/components/PastExperiments-test.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { expect } from 'chai';xe
```
pdehaan: Looks like a bonus "xe" at the end there.

johngruen: BONUS!

johngruen:
Given the ongoing difficulties in #1945, we might not want to use this PR yet

jaredhirsch:
@dannycoates sorry about the notification noise, I was trying to figure out how to get the reviewer avatar visible from the main PR list. I guess assignees, not reviewers, are the scannable thing

dannycoates (approved):
sure, why not 👍

johngruen:
Oh shit

johngruen:
I don;t think we should have merged this right now

dannycoates:
> I don;t think we should have merged this right now

no regrets. keep pushing forward

johngruen:
YOLO

PR #2071, opened 2017-01-26
Fix race condition between React & L20N in install button

### LINKED ISSUE #1945: Cannot navigate directly between experiments

lmorchard:
Fixes #1945


fzzzy:
r+. If the react node is empty, then react doesn't have any reason to try to go in and touch a node that l20n touched the first time the page was rendered, avoiding a "node is null" error.

PR #2077, opened 2017-01-26
remove conflicting react/l20n content

### LINKED ISSUE #2076: Resolve Race Conditions Between L20N and React
The original bug noticing this issue comes from #1945, where @6a68 noted that navigating from a graduated experiment to a current experiment does not fully update a page's state.

It turns out that there is larger issue, specifically our current l20n and pontoon integration hits a race condition on tags that accept `data-l10n-args` and pass in React props.

```
<span data-l10n-id="foo" data-l10n-args={json.stringify({foo})}>something {foo}</span>
```

Thus far, this issue affects navigation when users navigate to different pages that share the same React container  (as in navigating from one experiment directly to another, where both experiments are rendered into the `ExperimentPage` container), but there may be other issues we haven't seen yet. 

There are a few options to fix this issue:
- disable L20N's triggers somehow, then just run it on the DOM after componentDidUpdate in <App>
- go through react components and audit all the places where react and l20n are fighting by removing the react-y pieces (this doesn't fix the underlying problem)

This would cause some problems that should prevent merging #2063

dannycoates:
fixes #2076

### DIFF
--- a/frontend/src/app/components/ExperimentEolDialog.js
+++ b/frontend/src/app/components/ExperimentEolDialog.js
@@ -13,10 +13,10 @@ export default class ExperimentEolDialog extends React.Component {
           <form>
 
             <div className="modal-content modal-form">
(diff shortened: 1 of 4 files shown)


dannycoates:
i just grepped for `data-l10n-args` and removed the content

johngruen:
works!

Issue #1335, opened 2016-09-09
Do a better job monitoring for error states in extension installation


chuckharmston:
We don't do a great job of monitoring for error states in add-on installations. I see at least three places for improvement:
- Listen for exceptions in the Test Pilot add-on when `mozAddonManager` is used (#1223)
- [Listen for exceptions](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/InstallListener#onInstallFailed%28%29) in experiment installations via the Test Pilot add-on
- Set a timeout on the traditional link-to-XPI install.

sevaan:
@chuckharmston I'm not sure I fully understand the UX ask here. Could you clarify further (screenshots could be helpful if applicable)?

chuckharmston:
@sevaan 

Right now, while installing either the Test Pilot add-on or experiments, we don't do the best job of surfacing that there are errors. Sometimes that's because we haven't implemented it yet, but sometimes we also can't know that there's an error (for security/other reasons). That means that in some error states, our install button spins infinitely. I'm proposing that:

1) We do a better job of listening for errors and telling the user when they happen.
2) We assume that if the Test Pilot add-on or experiment hasn't been successfully installed after `n` seconds (not sure what that should be...20?) we assume something went wrong and that we tell the user as much.

johngruen:
Whelp, @sevaan went and had a baby. I'll do UX for this issue by EOW

chuckharmston:
@johngruen putting this back on your radar

johngruen:
@chuckharmston, noted. Let's see if this doesn't sneak into 14 tomorrow...

Caspy7:
I've just helped a user who was receiving "Installing..." forever.
After _much_ troubleshooting, it was determined that the user already had Test Pilot installed but disabled.

We should do better about failing gracefully and/or informing the user of the situation.

sevaan:
I have a proposal here: https://mozilla.invisionapp.com/share/Q77S1H6ZC#/203288154_Landing_Page_-_Error_Installing

If the install process hangs for say, 30 seconds, we could refresh the page and display this error message at the top. The "Still not working?" link could take the user to the Github issues, or maybe even just a readme somewhere. Thoughts?

Caspy7:
> maybe even just a readme somewhere

Perhaps a sumo page.

johngruen:
I'll need to add an FAQ, to link to, but this is ready to go

vpomerleau:
[...]

PR #2101, opened 2017-01-28
2090 wait longer

fzzzy:


dannycoates (approved):
I think it's fine to extend this timeout, but I don't think it will help in the "slow download" case that was mentioned earlier. This code doesn't run until after the mozAddonManager has emitted `onInstallEnded` in [addon.js](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/lib/addon.js#L24-L27)

What this might help with though is slow addon initialization, so 👍 

Incidentally, I don't see us handling the `onInstallFailed` case either. We should probably handle that and show the "something went wrong" [banner](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/ExperimentPage.js#L298) or something.

chuckharmston:
> Incidentally, I don't see us handling the `onInstallFailed` case either.

See #1335

PR #2120, opened 2017-02-01
Add eslint config to /addon/

pdehaan:
Still a couple warnings (`no-console`) and a couple of errors, but I don't think that our Travis-CI is running /addon/tests/ yet, so shouldn't break the CI build.

```
$ [addon/] npm run lint

> testpilot-addon@1.0.0 lint /Users/pdehaan/dev/github/mozilla/testpilot/addon
> eslint .


/Users/pdehaan/dev/github/mozilla/testpilot/addon/src/lib/actionCreators/env.js
  19:32  warning  Unexpected console statement  no-console

/Users/pdehaan/dev/github/mozilla/testpilot/addon/src/lib/actionCreators/InstallManager.js
  64:15  error  'baseUrl' is assigned a value but never used  no-unused-vars

/Users/pdehaan/dev/github/mozilla/testpilot/addon/src/lib/metrics/webextension-channels.js
  145:9  warning  Unexpected console statement  no-console

/Users/pdehaan/dev/github/mozilla/testpilot/addon/src/lib/middleware/Hub.js
  18:7  warning  Unexpected console statement  no-console

/Users/pdehaan/dev/github/mozilla/testpilot/addon/src/lib/reducers/clientUUID.js
  13:50  error  'action' is defined but never used  no-unused-vars

/Users/pdehaan/dev/github/mozilla/testpilot/addon/src/lib/Telemetry.js
  51:5  warning  Unexpected console statement  no-console

/Users/pdehaan/dev/github/mozilla/testpilot/addon/tools/control-panel/main.js
  15:3  warning  Unexpected console statement  no-console

✖ 7 problems (2 errors, 5 warnings)
```

r? @dannycoates 

Fixes #2115

PR #2121, opened 2017-02-01
fixed addon lint issues

dannycoates:
follow up to #2120

### DIFF
--- a/addon/src/lib/middleware/Hub.js
+++ b/addon/src/lib/middleware/Hub.js
@@ -14,8 +14,8 @@ export default class Hub {
   dispatch: Dispatch;
   ports: Set<EventEmitter>;
   constructor() {
-    this.dispatch = () =>
-      console.error('Hub cannot use dispatch() before middleware()');
+    this.dispatch = () => // eslint-disable-next-line no-console
+    console.error('Hub cannot use dispatch() before middleware()');
     this.ports = new Set();
   }
 
--- a/addon/test/test-migrations.js
+++ b/addon/test/test-migrations.js
@@ -1,4 +1,4 @@
-/* global describe beforeEach it */
+/* global describe it */
 import assert from 'assert';
 import { migrate } from '../src/lib/migrations';
 
--- a/addon/package.json
+++ b/addon/package.json
@@ -35,7 +35,7 @@
     "eslint": "^3.14.1",
     "eslint-plugin-flowtype": "^2.30.0",
     "flow-bin": "^0.38.0",
-    "flow-coverage-report": "^0.2.0",
+    "flow-coverage-report": "^0.3.0",
     "flow-typed": "^2.0.0",
     "jpm": "^1.2.2",
     "mocha": "^3.1.2",
@@ -50,7 +50,7 @@
   "dependencies": {
     "lodash": "^4.15.0",
     "redux": "^3.6.0",
-    "redux-logger": "^2.6.1",
+    "redux-logger": "^2.8.0",
     "seedrandom": "^2.4.2"
   },
   "scripts": {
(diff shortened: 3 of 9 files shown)


addon/src/lib/middleware/Hub.js
```diff
@@ -14,8 +14,8 @@ export default class Hub {
   dispatch: Dispatch;
   ports: Set<EventEmitter>;
   constructor() {
-    this.dispatch = () =>
-      console.error('Hub cannot use dispatch() before middleware()');
+    this.dispatch = () => // eslint-disable-next-line no-console
```
pdehaan, line 17: Had to read this a couple times... looks like an arrow function to nowhere, but I assume it in fact references the `console.error` on the following line (which is now unindented below. Not sure if we want to throw some free `{`..`}` around the arrow function to make it clearer (or not, whatever).

But it passes `npm run lint` and still seems to lint after running `npm run format`.

addon/test/test-migrations.js
```diff
@@ -1,4 +1,4 @@
-/* global describe beforeEach it */
+/* global describe it */
```
pdehaan, line 1: If you don't want to manage these `global` variables per-file, we could also set an additional `env` of `mocha: true` in the addon/.eslintrc.yml file (or even just add extra env for tests in addon/test/.eslintrc.yml).

addon/src/lib/middleware/Hub.js
```diff
@@ -14,8 +14,8 @@ export default class Hub {
...
+    console.error('Hub cannot use dispatch() before middleware()');
```
pdehaan, line 18: Offtopic: Should we do something more than `console.error()` here, like emit some error to Sentry or whatever?
Or do we care if errors get sent into a Browser Console abyss?

pdehaan (approved):
R+ with a few nit comments which can all be safely mocked and ignored. Merge at your convenience.

pdehaan:
Also, a few undoubtedly unrelated warnings buried in the Circle-CI logs:

```
./bin/circleci/test-frontend.sh
+ npm run lint

> testpilot@1.0.0 lint /home/ubuntu/testpilot
> gulp scripts-lint styles-lint

[22:18:56] Using gulpfile ~/testpilot/gulpfile.js
[22:18:56] Starting 'scripts-lint'...
[22:18:56] Starting 'styles-lint'...
[22:19:00] Finished 'styles-lint' after 4.14 s
[22:19:01] Finished 'scripts-lint' after 4.98 s
+ MOCHA_FILE=/tmp/circle-junit.x889vdj/junit/test-results.xml
+ npm run test:ci

> testpilot@1.0.0 test:ci /home/ubuntu/testpilot
> NODE_ENV=test nyc mocha --recursive --reporter mocha-junit-reporter --require frontend/test-setup.js frontend/test

Warning: Failed prop type: The prop `locale` is marked as required in `NewsletterForm`, but its value is `undefined`.
    in NewsletterForm
Warning: Failed prop type: The prop `locale` is marked as required in `NewsletterFooter`, but its value is `undefined`.
    in NewsletterFooter
Warning: Failed form propType: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `NewsletterForm`.
Warning: Failed prop type: The prop `locale` is marked as required in `View`, but its value is `undefined`.
    in View
Warning: Each child in an array or iterator should have a unique "key" prop. Check the top-level render call using <ul>. See https://fb.me/react-warning-keys for more information.
    in li
Warning: Failed prop type: The prop `subscribeToBasket` is marked as required in `EmailDialog`, but its value is `undefined`.
    in EmailDialog
Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `ExperimentDetail`. See https://fb.me/react-warning-keys for more information.
    in li (created by ExperimentDetail)
    in ExperimentDetail (created by Unknown)
    in Unknown

=============================== Coverage summary ===============================
Statements   : 96.46% ( 1797/1863 )
Branches     : 86.73% ( 340/392 )
Functions    : 94.41% ( 591/626 )
Lines        : 96.65% ( 1733/1793 )
================================================================================
```

Not sure if any of those are important, but I'll shut up now since this should be a separate bug for easier tracking.

---

**UPDATED:** Filed as #2122

addon/src/lib/middleware/Hub.js
dannycoates: yeah, that got formatted weirdly by `prettier`

addon/src/lib/middleware/Hub.js
dannycoates: This falls into the "should never happen" bin. I'll even tempt fate and say it'd never make it past QA. So its just a hint to me if I grossly bork it up.

PR #2132, opened 2017-02-05
clean up styles

### LINKED ISSUE #2104: CSS + Component Cleanup
I've started working on a branch that removes a lot of dead Sass and abstracts a few components into simple functional elements. Not sure if this will go anywhere but my goals are as follows:

- clear out no-longer-used Sass (which is surprisingly plentiful)
- abstract commonly repeated HTML blocks (copter, responsive-content-wrapper) into stateless functional components.
- simplify the relation between React and Sass using BEM syntax for the new component styles.
- provide documentation about how to use the above setup

The big idea is reduce some of the complexity and redundancy currently in our style and JSX code, and to make it easier to create new blocks of content with a greatly simplified set of tools.

johngruen:
Moved layout stuff into React components to try and simplify things across the site. Also modified header styles a bit. Fixes #2104 

There's a lot here so we might want to merge now and make sure it's all good.

screen shot 2017-02-05 at 7 45 12 pm
screen shot 2017-02-05 at 7 45 15 pm
screen shot 2017-02-05 at 7 45 18 pm
screen shot 2017-02-05 at 7 45 26 pm
screen shot 2017-02-05 at 7 45 35 pm


jaredhirsch:
> There's a lot here so we might want to merge now and make sure it's all good.

I'll help make sure it's all good before we land it :+1:

frontend/src/app/components/Banner.js
```diff
@@ -0,0 +1,37 @@
...
+import LayoutWrapper from './LayoutWrapper';
+
+const renderCondensedBanner = ({ children, dataL10nId = null }) => {
+  return (
+    <div className="banner banner--condensed">
+      <LayoutWrapper flexModifier="row-between-reverse">
+        <h2 className="banner__title" data-l10n-id={dataL10nId}>{ children }</h2>
+        <Copter small="true" />
+      </LayoutWrapper>
+    </div>
```
jaredhirsch: I think we could get rid of this special case. The other uses of `Banner` do not include the `LayoutWrapper`, `Copter`, or `h2` elements. If you remove those, the only difference is that the condensed banner uses the `banner--condensed` class instead of the `banner--expanded` class.

I think you could get rid of the special case using something like this (note that you don't need the dataL10nId any more, either):

```js
const Banner = ({ children, condensed = false, background = false }) => {
  return (
    <div className={classnames('banner', {
      'banner--condensed': condensed,
      'banner--expanded': !condensed,
      'banner--background': background
    }>
      { children }
    </div>
  );
};
```

Then you could replace that [one condensed instance](https://github.com/mozilla/testpilot/pull/2132/files#diff-d3b4b0683f2471eaec03fd641e6de552R69):

```js
// before
<Banner condensed={true} dataL10nId="experimentsListCondensedHeader">
         Pick your experiments!
</Banner>

// after
<Banner condensed={true}>
  <LayoutWrapper flexModifier="row-between-reverse">
    <h2 className="banner__title" data-l10n-id="experimentsListCondensedHeader">
        Pick your experiments!
    </h2>
    <Copter small="true" />
```

frontend/src/app/components/Header.js
```diff
@@ -74,12 +74,12 @@ export default class Header extends React.Component {
     return (
       <div>
         {this.renderRetireDialog()}
-        <header id="main-header" className="responsive-content-wrapper">
+        <LayoutWrapper id="main-header" flexModifier="row-between-top">
```
jaredhirsch: What about keeping the `header` el and just putting the `LayoutWrapper` inside it? Would be consistent with how `Footer` behaves, and it's nicer to keep the semantically-meaningful `header` vs replacing with a generic `div id="main-header"`.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -452,18 +454,18 @@ export class ExperimentDetail extends React.Component {
...
                   </div>}
-              </div>
+              </LayoutWrapper>
             </div>
           </div>
-          {!hasAddon && <div>
-            <h2 className="card-list-header" data-l10n-id="otherExperiments">Try out these experiments as well</h2>
+          {!hasAddon && <Banner>
+            <h2 className="banner__subtitle centered" data-l10n-id="otherExperiments">Try out these experiments as well</h2>
             <div className="responsive-content-wrapper delayed-fade-in">
```
jaredhirsch: How about replacing this div with a `<LayoutWrapper helperClass="delayed-fade-in">`?

.sass-lint.yml
```diff
@@ -1,4 +1,5 @@
 rules:
+  class-name-format: 0
```
jaredhirsch: Instead of getting rid of the class name formatting rules, what about using the [hyphenatedbem](https://github.com/benthemonkey/sass-lint/blob/feature/class-name-format/docs/rules/class-name-format.md#example-6) formatting convention instead? I'm fine to change the standard, but I do think it makes sense to enforce some kind of consistency.

frontend/src/app/components/LayoutWrapper.js
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+const LayoutWrapper = ({ children, id = null, flexModifier, helperClass = null }) => {
```
jaredhirsch: (very optional suggestion) The only cases where the `id` is needed are the `Header` and `Footer` components. Maybe just add wrapper divs with `id`s to those components, and remove that option here?

jaredhirsch (changes requested):
Hi @johngruen - Did a first pass, mostly focused on the React components. I added some notes suggesting simplifications for the `Header` and `Banner` components. I also think it'd be good to pick a new sass formatting convention, and not just stop enforcing it altogether.

Ping me when you've updated the PR, and I'll do a final nitpicky pass at the JS and the styles.

frontend/src/app/containers/ExperimentPage.js
johngruen: @6a68 doing the layout wrapper thing...killing off most of the animations for now just to have a re-think about them

johngruen:
@_6a68 i think i fixed all of your comments! Thanks for the review!

jaredhirsch:
looks great @johngruen, merging :tada:

PR #2134, opened 2017-02-06
expose clientUUID in the PageMod

dannycoates:
This will allow us to use the `clientUUID` on the frontend w/o messaging the addon.

### DIFF
--- a/addon/data/message-bridge.js
+++ b/addon/data/message-bridge.js
@@ -9,6 +9,7 @@
 // Page script acts as messaging bridge between addon and web content.
 unsafeWindow.navigator.testpilotAddon = true;
 unsafeWindow.navigator.testpilotAddonVersion = self.options.version;
+unsafeWindow.navigator.testpilotClientUUID = self.options.clientUUID;
 
 // New channel
 self.port.on('action', function(data) {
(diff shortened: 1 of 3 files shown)


lmorchard:
This is just on the current Test Pilot environment website, right? (Looks like it to me, but just kind of weakly waving a red flag.)

dannycoates:
> This is just on the current Test Pilot environment website, right?

correct

fzzzy (approved):
r+ with a request for comment: Can you explain why you think doing it this way is better than waiting for a message from the addon?

Also, is there any point in committing this if nothing uses it yet? Should we wait until something wants to use this before committing it?

dannycoates:
> Can you explain why you think doing it this way is better than waiting for a message from the addon?

Yeah, I should have led with that... When we can rely on folks having Firefox 49 we'll be able to do most of the heavy lifting in the webapp. In doing that we'll be able to cut down the data the webapp needs from the addon to just the clientUUID. If we set it here with this PR we should be able to sever the comm channel and remove a bunch more code from the addon in the future.

> Also, is there any point in committing this if nothing uses it yet? Should we wait until something wants to use this before committing it?

This needs to land at least a sprint before we intend to use it in the webapp so people have time to update. I've got some code in a branch that uses it, but it'll need to wait another sprint to land.

fzzzy:
Sounds good! Thanks!

PR #2140, opened 2017-02-07
added option in experiment yaml for disabling rating prompts

### LINKED ISSUE #2107: Permit experiments to opt out of Heartbeat-style Test Pilot prompts.
Pulse will be prompting users about their experiences with websites using a Heartbeat-style prompt:

screen shot 2017-01-30 at 3 05 32 pm

There's potential for confusion with the Test Pilot prompts asking for users' experience with experiments (which don't really apply to Pulse anyway, since it doesn't introduce a feature). It'd be great if those can be opted out of on a per-experiment basis.

dannycoates:
closes #2107 

Disable rating prompts in experiment yaml with:

```yml
testpilot_options:
    ratings: disabled
```

cc @chuckharmston

### DIFF
--- a/addon/src/lib/Experiment.js
+++ b/addon/src/lib/Experiment.js
@@ -15,6 +15,14 @@ function toAbsoluteUrl(baseUrl, path) {
   return path[0] === '/' ? baseUrl + path : path;
 }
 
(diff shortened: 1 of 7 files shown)


fzzzy:
Chuck is probably busy. This looks straightforward enough to me.

PR #2141, opened 2017-02-07
Implement variant testing for Test Pilot itself, and perform a very small test to make sure it is working

fzzzy:


fzzzy:
I guess I better fix the failures first.

Issue #2017, opened 2017-01-13
If browser language is set to German/Germany [de-de], Cliqz experiment is available but the Test Pilot pages are displayed in English


SoftVision-CosminMuntean:
[Notes]:
- The issue is also reproducible for all the others locales of German language.

[Affected versions]:
- Firefox 50.1.0

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Test Pilot add-on was previously installed on the profile.
- Have the German/Germany [de-de] and English [en] languages set in "about:preferences#content".

[Steps to reproduce]:
1. Open Firefox with the profile from prerequisites and navigate to "https://testpilot.firefox.com" page.
2. Observe the page language and the available experiments.

[Expected result]:
Depending on the desired behavior
- The page is displayed in German and the Cliqz experiment is available since [de-de] is a variation of default German [de] language.
or
- The page is displayed in English [en] (jumps to second in list) and Cliqz is not available since we don't have German [de-de] language localization.

[Actual result]:
- The page is displayed in English and the Cliqz experiment is available.

[Additional Notes]:
- The same behaviour can be observed for most languages that have language variations from default. Probably for all of them.
- Here is a screen recording with the issue: https://goo.gl/ItZTq4.

flodolo:
If you set the browser to "German (de)" does it work as expected?

Default accept-language for German users would be: 'de, en-US, en'.

SoftVision-CosminMuntean:
> If you set the browser to "German (de)" does it work as expected?

Yes, if the browser language is set to German [de] everything works as expected.

ckprice:
Top 3 locales over the last week (% of traffic)

1. de 66.92%
2. en-us 19.85%
3. **de-de 5.27%**

So some traffic, but not the majority of our German users.

pdehaan:
@ckprice @johngruen Should we do a similar en-US=>en copy-pasta for de=>de-de?

johngruen:
@pdehaan maybe, might want to think about being more formal since it's the second time this is happening

flodolo:
I think we need a better solution, there are plenty of other locales in this situation.

Also CC @stasm in case he has any ideas to solve this.

flodolo:
Also filed https://bugzilla.mozilla.org/show_bug.cgi?id=1331721

ghost:
I asked for a priority/timeline in the bug.  If it's a long ways out, let's do a quick fix on our end.

[...]

PR #2149, opened 2017-02-08
copy de l10n to de-DE

dannycoates:
"fixes" #2017

### DIFF
--- a/frontend/config.js
+++ b/frontend/config.js
@@ -11,7 +11,7 @@ module.exports = {
     // en is copied from en-US at build time, so we have to add it special
     'en,' + fs.readdirSync('./locales').join(',')  :
     // Enabled locales for stage & production - update as they reach 100%
-    'en-US,en,cs,de,dsb,es-ES,es-MX,fr,fy-NL,hsb,hu,it,ja,kab,nl,pt-BR,ru,sq,sv-SE,uk,zh-CN,zh-TW',
+    'en-US,en,cs,de,de-DE,dsb,es-ES,es-MX,fr,fy-NL,hsb,hu,it,ja,kab,nl,pt-BR,ru,sq,sv-SE,uk,zh-CN,zh-TW',
 
   // TODO: Move addon build to a better path
   ADDON_SRC_PATH: './addon/',
--- a/frontend/tasks/content.js
+++ b/frontend/tasks/content.js
@@ -22,7 +22,7 @@ gulp.task('content-experiments-data', function generateStaticAPITask() {
     .pipe(gulp.dest(config.DEST_PATH));
 });
 
-// The next two tasks are hacks to copy the en-US locale 
+// The next two tasks are hacks to copy the en-US locale
 // to an en locale. See issue #2006 for background.
 gulp.task('content-watch-en', () => {
    gulp.watch('./locales/en-US/**/*', ['content-build-en']);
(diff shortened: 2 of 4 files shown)


frontend/config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     // en is copied from en-US at build time, so we have to add it special
     'en,' + fs.readdirSync('./locales').join(',')  :
     // Enabled locales for stage & production - update as they reach 100%
-    'en-US,en,cs,de,dsb,es-ES,es-MX,fr,fy-NL,hsb,hu,it,ja,kab,nl,pt-BR,ru,sq,sv-SE,uk,zh-CN,zh-TW',
+    'en-US,en,cs,de,de-DE,dsb,es-ES,es-MX,fr,fy-NL,hsb,hu,it,ja,kab,nl,pt-BR,ru,sq,sv-SE,uk,zh-CN,zh-TW',
```
pdehaan, line 14: Do we also need to add it to line 12 above? (although "meh" since I guess it's really only for dev server development)
```js
// en is copied from en-US at build time, so we have to add it special
```

frontend/tasks/content.js
```diff
@@ -33,6 +33,11 @@ gulp.task('content-build-en', () => {
     .pipe(gulp.dest('./locales/en'));
 });
 
+gulp.task('content-build-de', () => {
+  return gulp.src('./locales/de/**/*')
+    .pipe(gulp.dest('./locales/de-DE'));
+});
+
```
pdehaan, line 40: Do we need a matching `content-watch-de` task? Or is "en" a special special case?

dannycoates: no, we don't edit translations live

frontend/config.js
dannycoates: not gonna do it

jaredhirsch:
lgtm :+1:

SoftVision-CosminMuntean:
Considering  the fact that this commit was merged, I have retested the issue on dev server, but is still reproducible. 
If you have the German/Germany [de-de] and English/United State [en-US] languages set in "about:preferences#content", the Test Pilot pages are displayed in English.

fzzzy:
I wonder if the problem is that in the browser the language string is "de-de" but in this pr the language string is "de-DE"

SoftVision-CosminMuntean:
I am not sure if this is the problem, because I have set the Spanish/Mexico [es-mx] language in "about:preferences#content" and the Test Pilot pages are correctly displayed in Spanish.  Even if in the browser the language string is "es-mx" and in pr the language string is "es-MX" .

pdehaan:
I _think_ the problem is that testpilot.dev must be running w/ `ENABLE_DEV_LOCALES === '1'`, since I'm not seeing de-DE in the list of `availableLanguages`:

```
<meta name="defaultLanguage" content="en-US">
<meta name="availableLanguages" content="en,ar,ast,az,bn-BD,bs,cs,de,dsb,en-US,es-AR,es-CL,es-ES,es-MX,fa,fr,fy-NL,hsb,hu,it,ja,kab,nl,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,te,tl,tr,uk,zh-CN,zh-TW">
```

Ref: https://github.com/mozilla/testpilot/pull/2149#discussion_r100182064

PR #2155, opened 2017-02-09
Just use production Ping Centre endpoints

jaredhirsch:
per convo with @tspurway, seems better to just ship stuff to prod


jaredhirsch:
just going to land this :+1:

PR #2157, opened 2017-02-09
rework addon initialization

dannycoates:
this is a bikeshed PR.

Initialization is currently spread out, some in `main.js` some in `sideEffects.js` triggered by actions. This makes it more complicated than it ought to be.

Here I've moved all that into `main.js` and removed those actions that only happen on load/unload. Its more straightforward this way in my opinion.


addon/src/lib/reducers/sideEffects.js
```diff
@@ -78,11 +76,14 @@ export function reducer(
...
         loader.schedule();
-        Object.keys(payload.experiments)
+        Object
+          .keys(payload.experiments)
           .map(id => payload.experiments[id])
           .filter(x => x.uninstalled && new Date(x.uninstalled) < new Date())
-          .forEach(experiment => dispatch(actions.UNINSTALL_EXPERIMENT({ experiment })))
-      }
+          .forEach(
+            experiment => dispatch(actions.UNINSTALL_EXPERIMENT({ experiment }))
```
jaredhirsch, line 84: Is this like an 80 chars max thing? If we're going to start enforcing line lengths, I think it'd be good to work out a standard as a team and add something to the eslintrc.

addon/src/main.js
```diff
@@ -43,37 +45,61 @@ const webapp = new WebApp({
...
+    t: 'event',
+    ec: 'add-on Interactions',
+    ea: 'browser startup',
+    el: Object.keys(activeExperiments(store.getState())).length
+  });
+  Services.obs.addObserver(
+    () => {
+      telemetry.ping(
+        'daily',
+        Object.keys(activeExperiments(store.getState())).length
```
jaredhirsch, line 84: Was this `addObserver` parameter auto-indented? Seems like kinda unusual formatting

addon/src/main.js
```diff
@@ -43,37 +45,61 @@ const webapp = new WebApp({
...
+    'idle-daily',
+    false
+  );
 }
 
 export function onUnload(reason: string) {
-  installManager.selfUnloaded(reason);
+  switch (reason) {
+    case 'uninstall':
+      telemetry.ping(self.id, 'disabled');
```
jaredhirsch, line 95: I think 'disable' should be fired before 'uninstall', so that this will double-count disable pings. I'll take a closer look to be sure.

addon/src/lib/reducers/sideEffects.js
dannycoates: I'm using [prettier](https://www.npmjs.com/package/prettier) so I don't have to think about it. Use `npm run format`. I haven't automated it beyond that. Maybe its should run on `npm start`? or test?

addon/src/main.js
dannycoates: prettier did it

addon/src/main.js
dannycoates: cool. I'm matching the behavior of the old addon [here](https://github.com/mozilla/testpilot/blob/master/old-addon/index.js#L356-L358)

PR #2168, opened 2017-02-10
[fixes #2136] Add completed dates to graduating experiments

### LINKED ISSUE #2136: Uninstall Tracking Protection experiment
We are graduating the Tracking Protection experiment out of Test Pilot with plans to do a "Phase 2" tracking protection experiment based on ideas we've learned here.

Since the Tracking Protection add-on is a WebExtension, it cannot use the AddonManager to uninstall itself. (I.e., [`getAddonByID()`](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/AddonManager#getAddonByID()) and [`uninstall`](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/Addon#uninstall()) like [universal-search](https://github.com/mozilla/universal-search/commit/f1655d6674db79750542c653388a9d79641a1ed6) did.)

So, we need the Test Pilot add-on to uninstall Tracking Protection for us.

ckprice:
This assumes the following plan:

1. Both TP and NM404s is being removed from Test Pilot on 2/15 (6am PST)
2. We will also be shipping an empty add-on to unship Tracking Protection
3. NM404s is moving to AMO

One question: if we're shipping an add-on to remove TP, why are we configuring an `uninstall` date? @clouserw @dannycoates

Fixes #2136

### DIFF
--- a/content-src/experiments/no-more-404s.yaml
+++ b/content-src/experiments/no-more-404s.yaml
@@ -84,6 +84,8 @@ contributors:
     avatar: /static/images/experiments/no-more-404s/avatars/adam-miller.jpg
 installation_count: 20377
 created: '2016-08-03T17:25:54.612365Z'
+completed: '2017-02-15T14:00:00.000000Z'
+eol_warning: The No More 404s experiment will not be removed from your browser and the results will be reported here when it ends. Future updates will be available on addons.mozilla.org.
 locale_blocklist:
   - 'de'
 order: 6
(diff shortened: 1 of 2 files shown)


ghost:
Just as a double precaution.  It was originally an easy alternative, but @dannycoates rocked this patch so fast that it's probably not needed, but doesn't hurt.  Empty add-on is easy to do.

ghost (approved):

PR #2081, opened 2017-01-26
added 'immutable' to long lived cache-control headers

dannycoates:
https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/

Pretty Cool!

PR #2173, opened 2017-02-11
l10n files should not be cached forever with immutable

dannycoates:
Since the last deploy I added `immutable` to our long lived, hashed assets in #2081. Unfortunately, FTL files are in that group despite not being hashed or immutable. This would break future changes to strings. FTL files must not get the immutable cache-control flag.

@ckprice this needs to be merged before we deploy to production.


ckprice:
👍

Issue #1480, opened 2016-09-29
Make optional external assets non-breaking


karlcow:
Hello.
(this was initially filed at https://github.com/meandavejustice/min-vid/issues/352 )

If you go to the current page 
https://testpilot.firefox.com/experiments/min-vid

with this settings in uMatrix (aka blocking the telemetry), the page content never ever initializes.

![capture d ecran 2016-09-29 a 12 08 34](https://cloud.githubusercontent.com/assets/505230/18939755/95ea0df8-863d-11e6-9264-88977b78517b.png)

It stays in this state:

![capture d ecran 2016-09-29 a 12 10 07](https://cloud.githubusercontent.com/assets/505230/18939764/b1e8aec4-863d-11e6-8a87-dd24dc39ef6a.png)

ghost:
The page fails to render when the analysis-output.telemetry.mozilla.org ping fails

johngruen:
Fira loaded in the `index.moustache` file is also offending

ghost:
PR in #2184

fzzzy:
https://github.com/mozilla/testpilot/pull/2184 has landed, is this fixed?

dannycoates:
I don't know about the font issue that @johngruen mentioned. I've been too <strike>lazy</strike> busy to actually test this

fzzzy:
I just tested this on master, and the original issue is fixed. I don't see any issues with the font thing. If there is an issue with loading the font, let's file a new issue for it and provide STR.

PR #2184, opened 2017-02-14
made the user counts request not block rendering

dannycoates:
This mitigates #1480 wrt the user counts. I'm not sure the best way to test it. I just changed the `usageCountsURL` to something unresolveable to check it.

### DIFF
--- a/bin/circleci/test-frontend.sh
+++ b/bin/circleci/test-frontend.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 set -ex
 
+# build experiments.json
+npm run content
+
 npm run lint
 MOCHA_FILE="$CIRCLE_TEST_REPORTS/junit/test-results.xml" npm run test:ci
--- a/frontend/src/app/actions/experiments.js
+++ b/frontend/src/app/actions/experiments.js
@@ -4,22 +4,23 @@ export default createActions({
 
   updateExperiment: (addonID, data) => ({ addonID, data }),
 
-  fetchExperiments: (experimentsUrl, countsUrl) => {
-    let experiments = [];
+  fetchExperiments: (experimentsUrl) => {
     return fetch(experimentsUrl)
       .then(response => response.json())
-      .then(data => {
-        experiments = data.results;
-        return fetch(countsUrl);
-      })
-      .then(response => response.json())
-      .then(counts => ({
+      .then(data => ({
         lastFetched: Date.now(),
         experimentsLoaded: true,
-        data: experiments.map(experiment => ({
-          ...experiment,
-          installation_count: counts[experiment.addon_id]
-        }))
+        data: data.results
+      }));
+  },
+
+  fetchUserCounts: (countsUrl) => {
+    return fetch(countsUrl)
+      .then(
+        response => response.json(),
+        () => ({}))
+      .then(counts => ({
+        data: counts
       }));
   }
 
--- a/frontend/src/app/actions/experiments.js
+++ b/frontend/src/app/actions/experiments.js
@@ -4,20 +4,16 @@ export default createActions({
 
   updateExperiment: (addonID, data) => ({ addonID, data }),
 
-  fetchExperiments: (experimentsUrl) => {
-    return fetch(experimentsUrl)
-      .then(response => response.json())
-      .then(data => ({
-        lastFetched: Date.now(),
-        experimentsLoaded: true,
-        data: data.results
-      }));
-  },
-
   fetchUserCounts: (countsUrl) => {
     return fetch(countsUrl)
       .then(
-        response => response.json(),
+        response => {
+          if (response.ok) {
+            return response.json();
+          }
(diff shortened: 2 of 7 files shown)


dannycoates:
i forgot about the counts returning before experiments case... i can make this better

dannycoates:
I took a different tack on the async loading race... I made experiments.json get browserified so there is no race.

frontend/src/app/store.js
```diff
@@ -10,6 +10,8 @@ import experimentsReducer from './reducers/experiments';
 import newsletterFormReducer from './reducers/newsletter-form';
 import varianttestsReducer from './reducers/varianttests';
 
+import experiments from '../../build/api/experiments.json';
```
dannycoates, line 13: this is where the 🐰🎩 happens

bin/circleci/test-frontend.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -ex
 
+npm run content
```
dannycoates: since the app now uses `experiments.json`directly we need to build it in order to run the tests

frontend/src/app/actions/experiments.js
```diff
@@ -4,22 +4,13 @@ export default createActions({
...
-      .then(data => {
-        experiments = data.results;
-        return fetch(countsUrl);
-      })
-      .then(response => response.json())
+  fetchUserCounts: (countsUrl) => {
+    return fetch(countsUrl)
+      .then(
+        response => response.json(),
+        () => ({}))
```
dannycoates: meh on errors

chuckharmston: :| I'm not. This is a really good way to suppress exceptions. Can we add some error handling here?

dannycoates: I dunno, error handling here feels excessive. If an error happens should we just not have the counts (like this) or fall back to something else (which can get complicated)? If we reported an error somewhere, who's going to look at it and be able to act on it? Also there's zero error handling in the current code either (not an excuse, but we're not having problems with it either)

That's why I went with 'meh'. If you have ideas for handling it I'm all ears.

jaredhirsch:
Ugh, sorry I missed this! Looking now

frontend/src/app/actions/experiments.js
```diff
@@ -4,22 +4,13 @@ export default createActions({
...
-      .then(response => response.json())
-      .then(data => {
-        experiments = data.results;
-        return fetch(countsUrl);
-      })
-      .then(response => response.json())
+  fetchUserCounts: (countsUrl) => {
+    return fetch(countsUrl)
+      .then(
+        response => response.json(),
```
jaredhirsch: I think you'll want to either check `response.ok`, or, equivalently, check that `response.status` is >= 200 and <300, before proceeding: https://mdn.io/Using_Fetch#Checking_that_the_fetch_was_successful

frontend/src/app/actions/experiments.js
jaredhirsch: Good discussion here. My thoughts:

> If an error happens should we just not have the counts (like this) or fall back to something else (which can get complicated)?

I agree that we should fail gracefully, and continue rendering the page without the counts.

> If we reported an error somewhere, who's going to look at it and be able to act on it? 

If the counts can't be loaded, it might just be a local network error, or it might indicate a busted release. How about [forwarding the error to sentry](https://docs.sentry.io/clients/javascript/integrations/react/#expanded-usage), so Ops can determine if it's a global or local thing?

While you're at it, would you mind also adding a sentry ping in the [`subscribeToBasket` error handler](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App.js#L123)? (Since this is a really nice way to detect problems with other services, I'll probably change #2231 to straight-up replace the `sendBeacon` ping centre call with a `fetch` that forwards network errors to Sentry.)

(Optional) If you're up for getting a little fancy, since I'm suggesting adding a `Raven.captureException` call in two spots, you could centralize that into some kind of app-global network error logging function, or DRY things up by wrapping the `fetch` API in a function that forwards exceptions to Sentry automatically...whatever, your call.

bin/circleci/test-frontend.sh
jaredhirsch: :+1: Worth documenting in the shell script? idk, your call, could be useful

jaredhirsch (changes requested):
LGTM, couple minor things to address, then good to land yourself (or ping me to do a final pass + merge)

frontend/src/app/actions/experiments.js
dannycoates: Logging is fine but should be worked on as a separate issue (#2253). I'm not going to do it in this PR.

fzzzy:
Does this mean experiments.json will be synchronously available on page load? That will be fantastic.

fzzzy (approved):
Stoked about this change. r+

Issue #2182, opened 2017-02-14
Add graduation reports for Tracking Protection and No More 404's


ckprice:
This will need to be merged in and pushed to production tonight (a bit behind on timing here =/). Hopefully it won't need much QA since we're just updating an existing YAML field.

pdehaan:
@johngruen @clouserw This was completed [and deployed to prod], yes?

PR #2185, opened 2017-02-14
[ref #2182] Push back graduation time

ckprice:
We're behind schedule, need to move this time back to 12pm pst.

Ref #2182


johngruen:
R+ merge when circled

PR #2191, opened 2017-02-15
Fix #2154 Record chosen variant test in google analytics.

### LINKED ISSUE #2154: Actually record variant test cohorts in google analytics
We need to nail down the strategy for storing a user's chosen cohort for a test in ga. I think we can look at Page Shot since it just did the same thing recently.

fzzzy:
Also, only allow a user to be in one test at a time.

### DIFF
--- a/frontend/src/app/reducers/varianttests.js
+++ b/frontend/src/app/reducers/varianttests.js
@@ -25,7 +25,7 @@ const tests = [
     name: 'installButtonBorder',
     getValue: function getValue() {
       if (!window.navigator.language.startsWith('en')) {
-        return false;  // User gets whatever the DefaultCase is.
+        return 'default';  // User gets whatever the DefaultCase is.
       }
       return random({
         bigBorder: 1,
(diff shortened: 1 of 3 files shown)


frontend/src/app/reducers/varianttests.js
```diff
@@ -38,9 +38,34 @@ const tests = [
 const chosenVariants = {};
 const identityReducers = {};
 
+let chosenTest = null;
+
 tests.forEach(test => {
-  chosenVariants[test.name] = test.getValue();
+  // Only put each user in one test. If we previously found a non-default test,
+  // put this user in default for the rest of the tests.
```
dannycoates, line 45: I'm not saying this is wrong, but it makes me wonder how the order of `tests` affects the populations of users they have access to. For example, 2 tests with a 50% chance on a variant. The first test will get 50% of all users as expected, but the second only 25% (50% of the remaining 50%). Am I right? Does that matter? I'm certainly no expert.

fzzzy: That's correct, and I think the weights have to be chosen to make all the populations equal. That's why I chose to replace the white-border-test with the toggle-switch-on-homepage-test in my toggle switch pr.

Who can we ask about how to choose the weights properly when there are more tests?

fzzzy: So, for example, if there are two tests, the first test should have a 1/3 chance of getting a user, and a 2/3 chance of default, and then the second test should have 1/2 and 1/2.

dannycoates (approved):
Code looks good 👍

PR #2207, opened 2017-02-17
fixed daily ping format

### LINKED ISSUE #2206: Daily telemetry ping is misformatted
The daily telemetry ping is an integer in the event field (which breaks our pipeline schema expectations of `event` being a string):
`"events":[{"timestamp":851080,"event":0,"object":"daily"}]`

Looks like the [parameters might be swapped](https://github.com/mozilla/testpilot/blob/e16941fdf687239d41a1a4e82822a5ffaa2d3214/addon/src/main.js#L82-L85), but also, the [telemetry doc for testpilot](https://github.com/mozilla/testpilot/blob/df90f3b05c4a90ee73ee4a92d1625fa99e77a2bb/docs/metrics/telemetry.md#testpilot-summary-ping) specifies that the datatypes for both `event` and `object` should be strings.

dannycoates:
fixes #2206

### DIFF
--- a/addon/src/lib/Telemetry.js
+++ b/addon/src/lib/Telemetry.js
@@ -49,7 +49,7 @@ export default class Telemetry {
     }
   }
 
-  ping(object: any, event: any, time?: number) {
+  ping(object: string, event: string, time?: number) {
     const payload = {
       timestamp: makeTimestamp(),
       test: self.id,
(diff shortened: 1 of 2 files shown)


addon/src/lib/Telemetry.js
```diff
@@ -49,7 +49,7 @@ export default class Telemetry {
     }
   }
 
-  ping(object: any, event: any, time?: number) {
+  ping(object: string, event: string, time?: number) {
```
fzzzy, line 52: flow === 👍

PR #2208, opened 2017-02-17
Simplify Telemetry ping message for WebExtension experiments

### LINKED ISSUE #2197: New level of nesting on webextensions pings is inconsistent with SDK pings
While writing the ETL for SnoozeTabs I noticed we have a new level of nesting in the payload just for webextensions pings:
https://github.com/mozilla/testpilot/blob/master/addon/src/lib/metrics/webextension-channels.js#L141

Unless there's a different reason we have this field, senderAddonId is redundant with the addon ID in `payload/test`, and `testpilotPingData` is an unnecessary level of nesting which is also inconsistent with pings sent from SDK addons. Preferably we'd return to the previous format (data from the test inside `payload/payload` without nesting) but, barring that, we should at least make sure webextensions and SDK addons send their pings in similar formats.

lmorchard:
- Retain legacy format for Tracking Protection experiment

- Simplify ping payload for all future WebExtension experiments

- Test updates

Fixes #2197

### DIFF
--- a/addon/src/lib/metrics/experiment.js
+++ b/addon/src/lib/metrics/experiment.js
@@ -20,7 +20,7 @@ import type Variants from './variants';
 
 export type ExperimentPingData = {
   subject: string,
-  data: string /* JSON { senderAddonId: string, testpilotPingData: any }*/
+  data: string /* JSON { any } */
 };
 
 const EVENT_SEND_METRIC = 'testpilot::send-metric';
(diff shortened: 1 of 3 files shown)


lmorchard:
I *think* this should fix #2197 while keeping the old format for Tracking Protection

dannycoates (approved):
cool

PR #2221, opened 2017-02-22
add min releases

johngruen:
@clouserw


ckprice:
I think this was covered for Pulse and Containers in 9fcc2609e569fee1552a2759c568c4b661d86d22 and 1ca824a19bebd35ccf6232fa195e7a66b00daf4e. Not sure if Snooze Tabs needs to be updated too.

PR #2247, opened 2017-02-27
make social urls absolute

### LINKED ISSUE #2244: Twitter Images aren't working for new experiments
## expected
tweeting an experiment link shows an image

## actual
screen shot 2017-02-27 at 2 38 26 pm

johngruen:
Fixes #2244 
I think...hard to review this until it's on dev

@pdehaan @lmorchard

### DIFF
--- a/frontend/config.js
+++ b/frontend/config.js
@@ -19,9 +19,10 @@ module.exports = {
   SRC_PATH: './frontend/src/',
   DEST_PATH: './frontend/build/',
   DIST_PATH: './dist/',
-  DJANGO_OLD_STATIC:  './testpilot/frontend/static/',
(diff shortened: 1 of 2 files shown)


lmorchard (approved):
:+1: to merge, if only to see if this fixes things.

Main issue with making these absolute is [rev-all](https://www.npmjs.com/package/gulp-rev-all) ignores them in markup [processed at the end of the build](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/dist.js#L8). So, we lose the cache-busting filename hashes. Failing finding a way to make a special case in rev-all to handle them automatically, we could manually change the names of these images if/when we need to update them.

PR #2248, opened 2017-02-27
copy contribute.json back into build

johngruen:
@lmorchard this may be a naive approach to adding this back into public view, but it seems to work.


lmorchard (approved):
Eh, sometimes the naive way is the best way

PR #2256, opened 2017-02-28
fixes #1590: Moved Settings menu to it's own component

### LINKED ISSUE #1590: Move settings menu into its own component
We currently have the settings menu fully contained inside of the [`Header`](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/Header.js) component. We should break it out into its own component.

pmespresso:


package.json
```diff
@@ -48,6 +48,7 @@
     "babelify": "^7.3.0",
     "browserify": "^11.0.1",
     "chai": "^3.5.0",
+    "chai-enzyme": "^0.6.1",
```
fzzzy: This isn't actually used, better remove it.

frontend/src/app/components/Header.js
```diff
@@ -1,19 +1,24 @@
...
 import RetireConfirmationDialog from './RetireConfirmationDialog';
+import Settings from './Settings';
 
 export default class Header extends React.Component {
 
   constructor(props) {
     super(props);
+
     this.closeTimer = null;
+    this.showSettingsMenu = this.showSettingsMenu.bind(this);
```
fzzzy: I prefer to put the call to bind at the place where this.showSettingsMenu is being passed, but I know I am probably in the minority about that.

It just seems gross to me to have some functions on this that are bound, and some that are not.

frontend/test/app/components/Header-test.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { expect } from 'chai';
 import sinon from 'sinon';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
```
pmespresso: shallow never used, removing import.

package.json
pmespresso: Got it

frontend/src/app/components/Header.js
pmespresso: I can make that fix if it's necessary.

pmespresso: But perhaps I should leave it for the sake of consistency with other components?

fzzzy:
Thanks for the pull request!

pmespresso:
Awesome! Happy to contribute!

PR #2263, opened 2017-03-01
Add AMO link for No More 404s add-on

### LINKED ISSUE #2196: Replace AMO link in No More 404s graduation report with actual add-on URL
Ref: [./content-src/experiments/no-more-404s.yaml:130](https://github.com/mozilla/testpilot/blob/7d2a6b6cc167e8708af8d7c0eadf5597b073d624/content-src/experiments/no-more-404s.yaml#L130):

> `<p>While we’re graduating No More 404s from Test Pilot, and will be moving the add-on to the <a href="https://addons.mozilla.org/firefox/" target="_blank">Mozilla Add-ons</a> store so it’s accessible to all Firefox users. We won’t remove it from users who have it installed. If you currently have the add-on installed, you don’t need to do anything. The add-on will continue to update as usual. The Internet Archive team will also be adding some new features so users can add content to the Wayback Machine directly through the add-on! On the Firefox side, we will promote No More 404s on the homepage of the browser.`

A link like `https://addons.mozilla.org/firefox/` isn't too useful to me, I'd still need to search AMO to find the add-on again. It'd be nice if we can put the full URL to the extension... once it's in AMO.

pdehaan:
Fixes #2196

### DIFF
--- a/content-src/experiments/no-more-404s.yaml
+++ b/content-src/experiments/no-more-404s.yaml
@@ -127,8 +127,7 @@ graduation_report: >
     <figcaption> Breakdown of the number of 404s encountered by users each week.</figcaption>
   </figure>
   <p><strong>Here's what happens next</strong></p>
-  <p>While we’re graduating No More 404s from Test Pilot, and will be moving the add-on to the <a href="https://addons.mozilla.org/firefox/" target="_blank">Mozilla Add-ons</a> store so it’s accessible to all Firefox users. We won’t remove it from users who have it installed. If you currently have the add-on installed, you don’t need to do anything. The add-on will continue to update as usual. The Internet Archive team will also be adding some new features so users can add content to the Wayback Machine directly through the add-on! On the Firefox side, we will promote No More 404s on the homepage of the browser.
-  </p>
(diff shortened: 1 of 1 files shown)


content-src/experiments/no-more-404s.yaml
```diff
@@ -127,8 +127,7 @@ graduation_report: >
...
-  </p>
+  <p>While we’re graduating No More 404s from Test Pilot, and will be moving the add-on to the <a href="https://addons.mozilla.org/firefox/addon/wayback-machine_new/" target="_blank">Mozilla Add-ons</a> store so it’s accessible to all Firefox users. We won’t remove it from users who have it installed. If you currently have the add-on installed, you don’t need to do anything. The add-on will continue to update as usual. The Internet Archive team will also be adding some new features so users can add content to the Wayback Machine directly through the add-on! On the Firefox side, we will promote No More 404s on the homepage of the browser.</p>
```
pdehaan, line 130: The only change in this mess was changing the general AMO link with a locale-agnostic add-on URL (plus moving the `</p>` to the end of the line).

chuckharmston (changes requested):
Make sure to [update the `experiments.ftl`](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization). You'll also want to invalidate all the existing translations with an `_l10nsuffix`.

pdehaan:
> Make sure to update the experiments.ftl. You'll also want to invalidate all the existing translations with an _l10nsuffix.

@chuckharmston But the `graduation_report`s aren't translated, are they?

chuckharmston (approved):

chuckharmston:
Oh, right 🤦‍♀️

PR #2264, opened 2017-03-01
Fix #2259 Remove redux-thunk.

### LINKED ISSUE #2259: Remove redux-thunk dependency.
It is unused. Any use cases that require redux-thunk functionality should use redux-promise instead.

fzzzy:
If you need to dispatch an async action, use redux-promise instead.


dannycoates (approved):
😎

PR #2276, opened 2017-03-03
Replace sendBeacon with Fetch

### LINKED ISSUE #2231: Replace `sendBeacon` with `fetch`
We use `navigator.sendBeacon` to send GA pings from the Test Pilot addon, because it's more polite than xhr or `fetch`: `sendBeacon` waits till the browser is idle before making a network request.

Some tracking protection addons, like uBlock Origin, set the `beacon.enabled` pref to `false`. This globally removes the `sendBeacon` function from the `navigator` object.

In #2151, I [added some code](https://github.com/mozilla/testpilot/blob/336cb2d/addon/src/lib/Telemetry.js#L30) that flips the `beacon.enabled` pref to `true`. Unfortunately, it turns out that the `beacon` pref change isn't reflected in the `navigator` object until the browser is restarted. So, we can't just flip the pref and assume `sendBeacon` is available; we'll need to fall back to `fetch` if it's missing. Probably still a good idea to flip the pref, since it'll mean a better user experience once Firefox is restarted.

_Refs #2230._

jaredhirsch:
Fixes #2231.

I'm swallowing errors and responses here, because `sendBeacon` did, but we could log them or introduce sentry to the addon.

### DIFF
--- a/addon/src/lib/Telemetry.js
+++ b/addon/src/lib/Telemetry.js
@@ -111,9 +111,11 @@ export default class Telemetry {
       topic: 'testpilot'
     };
 
-    Services.appShell.hiddenDOMWindow.navigator.sendBeacon(
-      'https://tiles.services.mozilla.com/v3/links/ping-centre',
-      JSON.stringify(pcPayload)
-    );
+    const req = new Request({
+      url: 'https://tiles.services.mozilla.com/v3/links/ping-centre',
+      contentType: 'application/json',
+      content: JSON.stringify(pcPayload)
+    });
+    req.post();
   }
 }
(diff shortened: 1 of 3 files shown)


addon/src/lib/Telemetry.js
```diff
@@ -111,9 +111,18 @@ export default class Telemetry {
...
-    Services.appShell.hiddenDOMWindow.navigator.sendBeacon(
+    const fetchConfig = {
+      method: 'POST',
+      mode: 'cors',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(pcPayload)
+    };
+    Services.appShell.hiddenDOMWindow.fetch(
```
dannycoates: why not use `sdk/request`?

fzzzy: Yeah, I think sdk/request might be a good idea.

jaredhirsch: :+1: updating now

jaredhirsch:
Updated. @dannycoates, mind taking a look?

dannycoates (approved):
lovely 🌻

jaredhirsch:
:beers:

PR #2279, opened 2017-03-03
[fixes #2278] Remove locale_grantlist from new experiments

### LINKED ISSUE #2278: Can't see latest experiments
Hi I love FF and would like to contribute to the new experiments. Alas I can't see them. I see only page shot, min vid, activity stream, and tab center. This morning I also tried to remove and install again test pilot but no change. With Test pilot add-on remove if I brows page https://testpilot.firefox.com I see only those four.
I am on Ubuntu 16.04.2 64 bit with Firefox 51.0.1
Thanks,
Dario Sestero

ckprice:
@flodolo  @johngruen  r?

Fixes #2278


ckprice:
Rebased with master.

flodolo:
Thanks, we could definitely do this since we also have fallback. It will make sense to use it again if we launch new experiments last minute.

Note that `flod` is a random inactive guy on GitHub, not me ;-)

Issue #2266, opened 2017-03-01
Test Pilot add-on not served with X-Target-Digest


chuckharmston:
Ref: #1474, #1537, [this Discourse thread](https://discourse.mozilla-community.org/t/firefox-add-on-test-pilot-wont-install/13822).

We should be serving the Test Pilot add-on with the same redirect/headers as we do the experiments.

lmorchard:
We can set headers for the XPI here, I think: https://github.com/mozilla/testpilot/blob/master/bin/deploy.sh#L90

Edit: Oh, wait, no - [we can't set the header on the XPI, we have to send the header in a 302 redirect to the .xpi](https://github.com/mozilla/testpilot/issues/1537#issuecomment-252095226). That complicates things 😞

ghost:
I think we should just use the system we have for hosting all our other add-ons to host this one too.

ghost:
One thing to consider using the other script is that the add-on could get out of sync from the site for a few minutes while one completes before the other.

Also, the script is expecting the branch to be called 'production' not 'production-static'.  I don't know how easy that is to change in the script per-repository.

ghost:
@relud The /latest/ URL exists now, but it looks like the x-amz-meta- prefixes aren't being stripped.  Are those rules running on this S3 bucket like they are for /files/?

```
$ curl -v 'https://testpilot.dev.mozaws.net/static/addon/latest'
[...]
< x-amz-meta-x-target-digest: sha256:1e0b66604d0ef92108d2632d73815f09383b831466d8df65c653de47a10e63ad
< x-amz-meta-location: /static/addon/addon.xpi
```

relud:
I don't run dev, only stage and prod

relud:
that said, the problem is that you prefixed the metadata tag with x-amz-meta-, but amazon also prefixes the tag with x-amz-meta-, so the double-prefix is properly reduced to a single prefix.

ghost:
Oh, that would be cool.  Fixed in ^, will check after it lands.

PR #2282, opened 2017-03-03
Add a /latest/ redirect for the xpi. Connected to #2266

clouserw:
I talked to @relud about this and we agreed the simplest way was just to add a /latest/ redirect to the site instead of hooking this up into our second AWS bucket.  The attached should do that, I think.  If this works, we can land and verify and then we can write a second patch for pointing the site to use /latest/


dannycoates (approved):
😎

Issue #2249, opened 2017-02-28
[l10n] Find a way to fall back to English strings when translation is missing


flodolo:
Right now we display the label, which is far from acceptable in a website. 

SnoozeTabs and Pulse are disabled on the home page because of this, while several locales are completely translated.

Possible solutions:
* Build a full FTL file as part of the build process, i.e. add English if a string is missing
* Fall back to English at run-time

CC @stasm too

flodolo:
I would strongly suggest to deal with this as a high priority issue to fix. 

We currently have about 20 languages enabled in production, not all of them can always be at 100% at the end of the cycle (especially if strings are added a few hours before tagging), and we can't keep adding and removing languages from the list.

Even for experiments, the `locale_grantlist` [approach](https://github.com/mozilla/testpilot/pull/2269/files) doesn't really scale.

CC also @zbraniecki in case he has ideas.

stasm:
I have a few ideas about how to fix this. I'll test them early next week.
In the meantime, perhaps we could introduce a simple build step in which
every locale file is concatenated with en-US (en-US first, other local
second). This will increase the download size of localization files a bit
but will fix the problem. I'll create a proper solution next week.

03.03.2017 06:30 "Francesco Lodolo" <notifications@github.com> napisał(a):

stasm:
Actually, it occurred to me that an even simple hotfix could be deployed.  In https://github.com/mozilla/testpilot/blob/master/frontend/src/templates/index.mustache#L12 we could link the en-US translation files, like so:

```diff
+    <link rel="localization" href="/static/locales/en-US/app.ftl">
+    <link rel="localization" href="/static/locales/en-US/experiments.ftl">
     <link rel="localization" href="/static/locales/{locale}/app.ftl">
     <link rel="localization" href="/static/locales/{locale}/experiments.ftl">
```

flodolo:
Oh wow, that would even help getting rid of the awful en hack while we fix fallbacks
https://github.com/dannycoates/testpilot/blob/master/frontend/tasks/content.js#L25

flodolo:
> Oh wow, that would even help getting rid of the awful en hack while we fix fallbacks

Scratch that, it was introduced to serve English to users with an accept languages like 'en,nl'.

PR #2287, opened 2017-03-06
Load en-US FTL files to provide fallback for incomplete locales

flodolo:
This is a good temporary fix for issue #2249, while waiting for l20n to provide fallback.

Also whitelisting Russian for new experiments, since it's complete.

### DIFF
--- a/frontend/src/templates/index.mustache
+++ b/frontend/src/templates/index.mustache
@@ -9,6 +9,9 @@
     <meta name="defaultLanguage" content="en-US">
     <meta name="availableLanguages" content="{{{ available_locales }}}">
     <meta name="viewport" content="width=device-width">
+    <!-- Load en-US first as fallback for incomplete locales -->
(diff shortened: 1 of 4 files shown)


flodolo:
Screenshot of es-ES before

schermata 2017-03-06 alle 20 12 03

and with this patch applied

schermata 2017-03-06 alle 20 21 18

@stasm plans to write on Discourse about a long term fix, but this should be good enough to ship a decent experience.

frontend/src/templates/index.mustache
```diff
@@ -9,6 +9,9 @@
     <meta name="defaultLanguage" content="en-US">
     <meta name="availableLanguages" content="{{{ available_locales }}}">
     <meta name="viewport" content="width=device-width">
+    <!-- Load en-US first as fallback for incomplete locales -->
```
flodolo, line 12: Not sure if you prefer to lose the comment here.

PR #2257, opened 2017-02-28
Fix #680 Fix #1300 Add copy for when an experiment has been manually disabled in the add-on manager

fzzzy:

PR #2296, opened 2017-03-09
WIP: more mozAddonManager

dannycoates:
This is a **rough** (but functional) sketch of using `mozAddonManager` for most frontend tasks.

@fzzzy and I have had a frustrating time with #2257 so I wanted to explore something different.

First thing, I kept the `lib/addon.js` api intact so the rest of the app doesn't need to change but I moved it to `lib/InstallManager.js` which just delegates to either a mozAddonManager (mam) wrapper `lib/moz.js` or the txp addon `lib/addon.js`.

When mam is available it'll use that for everything except tracking the txp addon state and the `installedAddons` list (which is a bit more of a pain with mam than I wanted to take on with this). The additions to `lib/addon.js` are to handle those specific tasks and ignore the rest, while keeping the existing logic for when mam isn't available.

One thing worth noting is that enabling an experiment that is disabled in about:addons works nicely if mam is available.


fzzzy:
We're going to revisit this through the lens of requiring mozAddonManager soon.

Issue #1309, opened 2016-08-29
Notification without a title.


mhoye:
I get this notification: /edit by ckprice http://i.imgur.com/N6p3C3W.png

in Nightly on OSX (Both current latest) and that seems like it's not the right thing.

Clicking that directs me to the No More 404s test pilot page - No More 404s is the only Test Pilot add-on I have installed.

digitarald:
I have the same issue (empty notification) but don't have No More 404 installed

ckprice:
I've deleted the current notification until this is investigated.

lmorchard:
@ckprice - Did you happen to get a screenshot of the notification in admin before it was deleted? I never saw it, would help to diagnose things

lmorchard:
Seems like [our code](https://github.com/mozilla/testpilot/blob/master/addon/lib/experiment-notifications.js#L109) called the SDK notification library correctly: A notification was displayed with our hardcoded "via testpilot.firefox.com" and clicking it went to the right experiment URL. But, it seems like [the notification we randomly picked](https://github.com/mozilla/testpilot/blob/master/addon/lib/experiment-notifications.js#L102) had an empty or undefined title & text. Not sure how that would happen. Could the random array index come up with something empty if there's just one notification?

[I've got a PR open to add unit tests for notifications](https://github.com/mozilla/testpilot/pull/1297), but nothing like this came up in there. Would be very handy to reproduce this issue in a test (after someone reviews & merges that PR pretty please!) and see about fixing it

lmorchard:
For good measure, [I just added a unit test to my PR that works with just one notification](https://github.com/mozilla/testpilot/pull/1297/files#diff-2b3639e8f7a812b3bddbfc4edcb2e0fdR87). That's as close as I can come to what happened in this issue off the top of my head. Bad news is, the test didn't reproduce this issue. So, I still don't know what happened.

ckprice:
> Did you happen to get a screenshot of the notification in admin before it was deleted? I never saw it, would help to diagnose things

No, sorry 😞 It was a dup of the last one - copied the two fields from the existing notification.

dannycoates:
@lmorchard I saw the json from https://testpilot.firefox.com/api/experiments before it was deleted. It looked fine to me:

``` json
{
          "id": 2,
          "title": "Tired of 404s?",
          "text": "Check out our latest experiment!",
          "notify_after": "2016-08-29T20:15:45Z"
}
```

[...]

PR #2303, opened 2017-03-10
[ref #1309] Push test notifications to Dev

ckprice:
This is to try to see if our notifications are working on dev/stage since we were unable to repro #1309 locally.

I'll open a `CRITICAL` issue once this lands on stage to remove this commit before we push.


lmorchard:
Is there a reason why this shouldn't be merged?

ckprice:
@lmorchard yes let's merge! I'm ready to test and will follow-up with removing the test notification after the testing.

lmorchard:
Okay, might as well merge now and then remove before the end of the sprint

PR #2306, opened 2017-03-10
Remove mustache and use ReactDOMServer instead

fzzzy:


frontend/src/templates/index.mustache
```diff
@@ -1,71 +0,0 @@
...
-            </div>
-          </div>
-          <div class="copter">
-            <div class="copter__inner"></div>
-          </div>
-        </div>
-      </div>
-    </noscript>
-
-    <div class="full-page-wrapper centered overflow-hidden">
```
lmorchard, line 53: FWIW, looks like these loader bars got left out of the new template - intentionally?

fzzzy: Yes, because I thought we didn't need loading bars any more, but as we discussed on the call I will put them back in. Thanks for looking!

fzzzy: I added it back in.

dannycoates (approved):
🤘

Issue #2313, opened 2017-03-13
Improve Ping Centre integration


jaredhirsch:
Test Pilot's integration with Ping Centre is changing: rather than use a single JS schema to send all experiment pings to Ping Centre, we will have one client-side JS schema per experiment, which will map to the experiment's Redshift table in Ping Centre.

Steps here:

- add snoozetabs schema to ping-centre repo (https://github.com/mozilla/ping-centre/issues/59)
- integrate ping-centre library into test pilot addon (#2310)
- update test pilot metrics code to send experiment-specific pings to Ping Centre, starting with Snooze Tabs (#2312)

Refs https://github.com/mozilla/ping-centre/issues/57.

PR #2314, opened 2017-03-13
Remove 'raw' field from Ping Centre pings

jaredhirsch:
Only potential issue here is that the [corresponding ping-centre PR](https://github.com/mozilla/ping-centre/pull/60) needs to be released before this change. I'll work with the Ping Centre team to orchestrate the change.

Related to #2313 and mozilla/ping-centre#58.


dannycoates (approved):
@6a68 r+. fire when ready

PR #2320, opened 2017-03-14
Stop hashing en-US ftl files in template

flodolo:
Following @lmorchard suggestion on IRC


lmorchard (approved):
Works for me! But, someone else might want to verify

lmorchard:
Eh, I'll just merge.

PR #2322, opened 2017-03-14
Fix #2261 Remove redux-actions

### LINKED ISSUE #2261: Remove redux-actions dependency.
While this will result in much more boilerplate, we don't have that many actions and I think it will improve the approachability of the codebase and make it easier to follow the control flow.

fzzzy:


frontend/src/app/reducers/addon.js
```diff
@@ -23,42 +30,46 @@ function setInstalledAddons(state, { payload: installedAddons }) {
   return { ...state, installed: newInstalled, installedAddons: actuallyInstalledAddons };
 }
 
-const setClientUuid = (state, { payload: clientUUID }) => ({ ...state, clientUUID });
+function setClientUuid(state, { payload: clientUUID }) {
+  return { ...state, clientUUID };
+}
 
-const manuallyEnableExperiment = (state, { payload: experiment }) => {
+function manuallyEnableExperiment(state, { payload: experiment }) {
```
dannycoates, line 37: nice to see `function` making a comeback here 👍

dannycoates (approved):
Great step! In the future it might be useful to make our action types flow [unions](https://flowtype.org/docs/union-intersection-types.html)

PR #2324, opened 2017-03-15
add sourcemaps; fixes #1892

### LINKED ISSUE #1892: Add JavaScript sourcemaps in Stage/Production?
Ref [./frontend/tasks/scripts.js:49-55](https://github.com/mozilla/testpilot/blob/979875a113bee445170f04f49e3d2ab2c575eddf/frontend/tasks/scripts.js#L49-L55):

```js
gulp.task('scripts-misc', () => {
  return gulp.src(config.SRC_PATH + 'scripts/**/*')
    .pipe(gulpif(config.IS_DEBUG, sourcemaps.init({loadMaps: true})))
    .pipe(gulpif(!config.IS_DEBUG, uglify()))
    .pipe(gulpif(config.IS_DEBUG, sourcemaps.write('./')))
    .pipe(gulp.dest(config.DEST_PATH + 'static/scripts'));
});
```

It looks like we may not enable source maps while not in debug mode (where [`IS_DEBUG`](https://github.com/mozilla/testpilot/blob/03061ad53f1f07b1cc8b7a0bfbbedf02ff8075a8/frontend/config.js#L5) is defined as `IS_DEBUG: (process.env.NODE_ENV === 'development')`.

Oddly, I wasn't seeing sourcemaps working on https://testpilot.dev.mozaws.net/ when I tried browsing the app.7e8065e0.js or vendor.ec3d1ad6.js files and formatting the minified code.

https://blog.sentry.io/2015/10/29/debuggable-javascript-with-source-maps.html

clouserw:
It looks like we specifically avoided generating the sourcemaps in production and I'm not sure why.  I ran this locally and got the sourcemaps, but I'm not sure if we'd have to do anything else to get them out there.

r? @lmorchard at a minimum since I think he wrote the original code

it looks like it generates maps with the hashes in the name (off the JS files with those hashes) but it also generates the files without the hashes.  I'm not sure which one the browser loads and currently the CDN will cache .map files forever.  Happy to add the patch that expires .map files in a reasonable amount of time if needed.

### DIFF
--- a/frontend/tasks/scripts.js
+++ b/frontend/tasks/scripts.js
@@ -48,9 +48,9 @@ gulp.task('scripts-watch', () => {
 
 gulp.task('scripts-misc', () => {
   return gulp.src(config.SRC_PATH + 'scripts/**/*')
-    .pipe(gulpif(config.IS_DEBUG, sourcemaps.init({loadMaps: true})))
+    .pipe(sourcemaps.init({loadMaps: true}))
     .pipe(gulpif(!config.IS_DEBUG, uglify()))
-    .pipe(gulpif(config.IS_DEBUG, sourcemaps.write('./')))
+    .pipe(sourcemaps.write('./'))
     .pipe(gulp.dest(config.DEST_PATH + 'static/scripts'));
 });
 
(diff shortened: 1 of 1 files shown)


lmorchard:
LGTM, can't remember why we wouldn't want sourcemaps on prod

PR #2336, opened 2017-03-20
Fix bustage

fzzzy:


dannycoates (approved):
👍 and let's follow up with finding where `state.installed` can be undefined

PR #2346, opened 2017-03-22
2337 fix user counts

fzzzy:
The isomorphic-fetch upgrade changed the response.json() api from synchronously returning to returning a promise :(


chuckharmston (changes requested):
Can you please add a regression test for this?

chuckharmston (approved):

PR #2351, opened 2017-03-24
Remove react router

fzzzy:


frontend/src/app/containers/App.js
```diff
@@ -1,23 +1,28 @@
-/* global ga */
+/* global ga, location */
```
chuckharmston: Generally a good practice to refer to window-scoped variables by that prefix, so `window` here and `window.location`, maybe?

fzzzy: Good call.

frontend/src/generate-static-html.js
```diff
@@ -2,23 +2,38 @@
 import React from 'react';
 import ReactDOMServer from 'react-dom/server';
 
+import Loading from './app/components/Loading';
 
-function makeStaticString(headComponent, bodyComponent) {
+function makeStaticString(
```
fzzzy: This module is a mess, it really should be cleaned up.

fzzzy:
The number of users of each experiment is not properly being fetched in this branch. I think it just got left out.

fzzzy:
Fixed fetching the number of users of each experiment.

frontend/src/app/actions/newsletter-form.js
```diff
@@ -1,5 +1,4 @@
 
-import 'isomorphic-fetch';
```
fzzzy, line 2: Actually I think we still want this in here, to polyfill old browsers.

fzzzy: I guess it should go in bootstrap.js instead.

fzzzy: I added it to bootstrap.js.

frontend/src/app/actions/newsletter-form.js
```diff
@@ -32,6 +31,9 @@ function newsletterFormSubscribe(dispatch, email, locale) {
       }
     })
     .catch(() => dispatch(actions.newsletterFormSetFailed()));
+  return {
+    type: 'NEWSLETTER_FORM_SUBSCRIBE'
+  };
```
fzzzy, line 37: This was a bugfix, I'm not sure what exposed it or introduced it. redux doesn't like `undefined` as an action.

compiled-templates/compiled-templates.js
```diff
@@ -1,65 +0,0 @@
-module.exports.render = function (options) {
```
fzzzy, line 1: I was able to completely replace this other way we had of building static pages from a .md file with generate-static-html.

frontend/src/app/actions/browser.js
```diff
@@ -6,13 +6,9 @@ import type {
   SetStateAction
 } from '../reducers/browser';
 
-function setState(payload: BrowserState): SetStateAction {
+export function setState(payload: BrowserState): SetStateAction {
```
fzzzy, line 9: Using `export default { setState }` as down below prevents using `import { setState } from 'module'`. It just comes back as undefined. It could be used as `import foo from 'module'; foo.setState`, but the {} form is nice to be able to use.

frontend/src/app/actions/experiments.js
```diff
@@ -3,17 +3,18 @@
 import type {
   Experiment,
   UpdateExperimentAction,
-  FetchUserCountsAction
+  FetchUserCountsAction,
+  SetSlugAction
```
fzzzy, line 7: Since we are rendering the experiments.js at build time in node, we don't synchronously know the slug (until we are running in the browser where we know the full url). So, we need a SetSlugAction to update the state.

frontend/src/app/actions/varianttests.js
```diff
@@ -0,0 +1,8 @@
+
+import type { ChooseTestsAction } from '../reducers/varianttests';
+
+export function chooseTests(): ChooseTestsAction {
```
fzzzy, line 5: As we don't want to synchronously run the chooseTests code, but want to wait until we are running in the browser to choose them, we now need a chooseTests action which we can trigger in the browser to update the test cohort state.

frontend/src/app/components/ExperimentRowCard.js
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'react-router';
```
fzzzy, line 2: We don't need Link any more, we just link to the other pages normally.

frontend/src/app/components/ExperimentRowCard.js
```diff
@@ -156,7 +155,7 @@ export default class ExperimentRowCard extends React.Component {
 
 ExperimentRowCard.propTypes = {
   experiment: React.PropTypes.object.isRequired,
-  hasAddon: React.PropTypes.bool.isRequired,
+  hasAddon: React.PropTypes.any,
```
fzzzy, line 158: hasAddon is now trinary, with null indicating that we are being rendered at build time, and true and false being reserved for when we are running in the browser.

frontend/src/app/components/Loading.js
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
 
 const Loading = () => {
-  return (
+  return <div className="full-page-wrapper centered overflow-hidden">
```
fzzzy, line 4: This just lets us use the same <Loading /> component in a few places, instead of having the same structure copied into a few files as it was before.

frontend/src/app/components/Routes.js
```diff
@@ -1,36 +0,0 @@
-import React from 'react';
```
fzzzy, line 1: We don't need the Routes file any more, since now the routes are just specified by .js and .md files in the frontend/src/pages directory.

frontend/src/app/containers/App.js
```diff
@@ -1,23 +1,29 @@
...
+import { chooseTests } from '../actions/varianttests';
 import addonActions from '../actions/addon';
 import newsletterFormActions from '../actions/newsletter-form';
 import RestartPage from '../containers/RestartPage';
+import { isFirefox, isMinFirefoxVersion, isMobile } from '../lib/utils';
+import config from '../config';
 
-const clipboard = new Clipboard('button');
+let clipboard = null;
+if (typeof document !== 'undefined') {
```
fzzzy, line 24: Since this code runs at build time in node now, sometimes we don't want to do some things if we are at build time, but just want to wait until we are running in the browser.

frontend/src/app/containers/App.js
```diff
@@ -26,12 +32,12 @@ class App extends Component {
...
 
   measurePageview() {
-    const { routing, hasAddon, installed, installedLoaded } = this.props;
+    const { hasAddon, installed, installedLoaded } = this.props;
 
     // If we have an addon, wait until the installed experiments are loaded
     if (hasAddon && !installedLoaded) { return; }
 
-    const { pathname } = routing.locationBeforeTransitions;
+    const pathname = window.location.pathname;
```
fzzzy, line 40: We just get our path from the browser instead of react-router now.

frontend/src/app/containers/App.js
```diff
@@ -74,7 +80,19 @@ class App extends Component {
     }
   }
 
-  componentDidUpdate() {
+  componentDidMount() {
```
fzzzy, line 83: We have to do a lot more stuff in App.componentDidMount, since we have to wait until here to be able to access anything on the window object.

frontend/src/app/containers/App.js
```diff
@@ -125,6 +143,7 @@ function subscribeToBasket(email, locale, callback) {
 const mapStateToProps = state => ({
   addon: state.addon,
   experiments: experimentSelector(state),
+  slug: state.experiments.slug,
```
fzzzy, line 146: We get passed the slug as a prop instead of in the react-router params.

frontend/src/app/containers/App.js
```diff
@@ -147,7 +167,12 @@ const mapStateToProps = state => ({
 
 
 const mapDispatchToProps = dispatch => ({
-  navigateTo: path => dispatch(routerPush(path)),
+  setBrowserState: state => dispatch(setBrowserState(state)),
+  chooseTests: () => dispatch(chooseTests()),
+  fetchUserCounts: (url) => dispatch(fetchUserCounts(url)),
+  navigateTo: path => {
+    window.location = path;
```
fzzzy, line 174: This works well as a quick hack!

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -280,7 +280,7 @@ export class ExperimentDetail extends React.Component {
 
         <View {...this.props}>
 
-        {(!hasAddon && !graduated) && <section id="testpilot-promo">
+        {hasAddon !== null && (!hasAddon && !graduated) && <section id="testpilot-promo">
```
fzzzy, line 283: if hasAddon is null, we're being rendered at build time

frontend/src/app/index.js
```diff
@@ -1,48 +0,0 @@
-import 'babel-polyfill/browser';
```
fzzzy, line 1: Everything that used to happen in app.js now happens in bootstrap.js, which is common across all pages, and the actual app bundle for the page, which gets rendered by running pages/<page>.js:create() and using inject.js:inject() on the result.

frontend/src/app/lib/inject.js
```diff
@@ -0,0 +1,37 @@
+
```
fzzzy, line 1: inject.js takes care of building the store and the app, and renders the component into the page once we are running in the browser. When running on the server, the return value is rendered to a static string.

frontend/src/app/reducers/varianttests.js
```diff
@@ -2,6 +2,10 @@
```
fzzzy, line 2: All that changed in this file is instead of the test-choosing happening synchronously at the top level of the module, it happens in a chooseTests function that is triggered by an action once we are running in the browser.

frontend/src/generate-static-html.js
```diff
@@ -1,24 +1,43 @@
 
 import React from 'react';
 import ReactDOMServer from 'react-dom/server';
+import ReactMarkdown from 'react-markdown';
 
+import Loading from './app/components/Loading';
 
-function makeStaticString(headComponent, bodyComponent) {
+function makeStaticString(
```
fzzzy: Maybe it is not too bad. It is big and unweildy, and could be refactored with better function names and arguments, but it works for now.

frontend/src/pages/error.js
```diff
@@ -0,0 +1,9 @@
+
```
fzzzy, line 1: A generic route is just a .js file in the frontend/src/pages directory or any subdirectory that exports a function which does `return inject(<MyComponent />)`

frontend/src/pages/privacy
```diff
@@ -0,0 +1 @@
+../../../node_modules/legal-docs/firefox_testpilot_PrivacyNotice
```
fzzzy, line 1: We can also have .md files in the `frontend/src/pages` directory (or subdirectories thereof) and they will be generated like compiled-templates used to do it.

ghost:
I have a copy of this running at http://d1s1pwobne7vcc.cloudfront.net/ .  Differences and assumptions:

* Deployed with DEV flag (15s cache limits)
* No CSP
* add-on not included, since I didn't want to sign a copy of it

chuckharmston:
Videos of [the above](https://irccloud.mozilla.com/file/4ayVlVy7/without-router.mov) vs [existing](https://irccloud.mozilla.com/file/UvLop3vD/with-router.mov).

Update: existing on dev instead of prod. https://irccloud.mozilla.com/file/4awW1DYT/without-router-dev.mov

fzzzy:
Here is a summary of the pros and cons of this branch, and a plan to fix the problems that Chuck pointed out. TLDR: We're going to get some help from the UX team and I am going to polish the rough edges, so this branch won't land for another two sprints still.

- Pros
	- react-router often caused problems with CSP involving the use of pushState, and we didn't have steps to reproduce. This branch just uses normal page loads, and thus fixes this bug.
	- react-router also interacted badly with l20n, making it very easy to introduce bugs causing the entire site to break when navigating to another link on the site. Again this branch just uses normal page loads, fixing these bugs.
	- In the remove-react-router branch, the page's DOM nodes are rendered at build time and present in the html page. This makes them much quicker to show up and removes the loading bar entirely.
	- Each page has it's own bundle; since each page bundle only contains the Component for that page instead of all the components. 
	- Dynamic pages and precompiled pages are unified in the new branch. Previously, dynamic stuff was built using react and static stuff was built using test-pilot/compiled-templates. Compiled templates are now .md files in the frontend/src/pages directory and dynamic pages are now .js files in the same directory, rendered using the same code path in generate-static-html.js
- Cons
	- Clicking on a link does a full page refresh. react-router will always win here because it does a partial page refresh and everything it needs is already loaded.
	- The UX of hits with a cold cache is bad; many things jump around during loading.
- Need to fix before landing
	- Currently, the new site is larger than the old site, even though 3 dependencies have been removed. The new site bundles should be smaller, so I will work on fixing this before landing.
	- The UX on the home page must be fixed to not pop between two very different states between html load and js load. The UX team is going to redesign the 'has-addon' and 'no-addon' experiences so they are visually more similar to prevent this being so jarring.
	- The noscript experience on the new site needs polishing.
	- The import dependency graph needs to be cleaned up so that each bundle actually only contains components used in that bundle; currently still including everything.
- Future
	- Partial page loads and the use of pushState is possible in the new code; it just needs to be done. To prevent this PR from taking even longer to land, I'll do this some time later.

frontend/src/app/components/ExperimentRowCard.js
lmorchard: Hmm, this seems like trouble to me. Maybe a second prop or some other way of detecting static generation would be clearer? Though I guess the trinary thing means yes / no / dunno, which seems fair enough. :shrug:

frontend/src/app/actions/browser.js
lmorchard: Huh. I can't remember why I did `export default { setState }` now - seems dumb in retrospect.

frontend/src/app/containers/App.js
```diff
@@ -1,23 +1,29 @@
-/* global ga */
+/* global ga, window */
```
lmorchard: I think `window` is already covered in the [global `.eslintrc`](https://github.com/mozilla/testpilot/blob/master/.eslintrc#L14).

frontend/src/app/containers/App.js
```diff
@@ -175,11 +200,12 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
     sendToGA,
     subscribeToBasket,
     clipboard,
-    userAgent: navigator.userAgent,
     setPageTitleL10N: (id, args) => {
-      const title = document.querySelector('head title');
-      title.setAttribute('data-l10n-id', id);
-      title.setAttribute('data-l10n-args', JSON.stringify(args));
+      if (typeof document !== 'undefined') {
```
lmorchard: Could do this and save a level of indent:
```javascript
if (typeof document === 'undefined') { return; }
```

jaredhirsch:
I'm generally in favor of landing this branch, though we'd need to figure out a timeline for addressing the 'need to fix' bullet points in @fzzzy's comment above.

I think it might be worthwhile to invest in improving the no-JS experience (put the locale slug in the URL and pre-render localized pages); maybe that would be better as a follow-on, since it'll involve a lot of hacking on the build steps, and, I guess, configuring nginx to redirect based on the Accept-Languages header.

## load time / performance

**TL;DR: Faster time-to-content in remove-router branch**, especially on the experiment pages. The page is larger and takes longer to fully load, though we are running against a non-minified server.

I ran some webpagetest tests, comparing the load time of the dev server (testpilot.dev.mozaws.net) and a similarly configured one-off cloudfront server running the remove-react-router branch (d1s1pwobne7vcc.cloudfront.net/). Neither server is minified--so there is room to improve size and overall render/load time. All tests were run with an empty cache, and the numbers below are from the median of 3 runs in each case. The old/new links are for the webpagetest results.

- homepage, desktop FF, cable connection (5/1 Mbps):
  - [old](https://www.webpagetest.org/result/170427_SA_1A4/): start render: 1.5 sec, load: 4.8 sec, size: 1.8 MB
  - [new](https://www.webpagetest.org/result/170427_3N_1A1/): start render: 1.2 sec, load: 6.9 sec, size: 3.2 MB
- experiment page, desktop FF:
  - [old](https://www.webpagetest.org/result/170427_VA_1BF/): start render: 2.7 sec, load: 5.8 sec, size: 2.5 MB
  - [new](http://www.webpagetest.org/result/170427_2Y_1B0/): start render: 1.1 sec, load: 8.4 sec, size: 4 MB
- homepage, mobile (Galaxy S7 / Chrome, mobile LTE connection):
  - [old](https://www.webpagetest.org/result/170427_4M_1MJ/): start render: 1.4 sec, load: 4.1 sec, size: 1.7 MB
  - [new](https://www.webpagetest.org/result/170427_22_1MH/): start render: 1.8 sec, load: 4.7 sec, size: 3.2 MB
- experiment page, mobile:
  - [old](https://www.webpagetest.org/result/170427_HJ_1MN/): start render: 2.9 sec, load: 5.1 sec, size: 2.4 MB
  - [new](https://www.webpagetest.org/result/170427_5R_1MK/): start render: 1.9 sec, load: 5.4 sec, size: 4 MB

What's a good target for load time?

Here are some production websites (note, our test servers were not production minified):

- mozilla.org homepage, desktop FF: start render 1.4 sec, load time 2.7 sec, 1.1 MB
- firefox download page, desktop FF: start render 2.0 sec, load time 3.4 sec, 620 KB


## noscript 

JS disabled: coherent but roadblocked on current; incompletely rendered in branch (but we could finish this before landing it and have a strong noscript option)

- homepage:
  - old: just shows a roadblock “js required, sorry”
  - new: doesn’t show the header, so oddly just shows a bunch of experiments
- experiment:
  - old: again, just shows a roadblock
  - new: shows the page content, but no header, and no privacy policy.
    - also shows the “this experiment is not enabled in your language (en)” warning that’s normally hidden by client-side JS.

fzzzy:
That reminds me, there's also a "this experiment is not enabled in your language" flash on the experiments page. I will fix this.

frontend/src/app/lib/inject.js
```diff
@@ -0,0 +1,37 @@
...
+      const staticNode = document.getElementById('static-root');
+      if (staticNode) {
+        staticNode.parentNode.removeChild(staticNode);
+      }
+    }
+  }
+  return provider;
+}
+
+export default inject;
```
lmorchard: I think you can just write `export default function inject` up top.

frontend/src/pages/privacy
lmorchard: Is this a symlink? Been trying hard not to use symlinks in the repo. Git doesn't seem to handle them well - I think mainly because they break on a Windows dev environment. (something I'll probably start trying soon if/when I get a new laptop) They also seem to be a bit inconvenient when searching commit history

frontend/src/pages/error.js
lmorchard: :+1: to simplified routes & de-duped info

frontend/tasks/scripts.js
```diff
@@ -64,15 +68,48 @@ gulp.task('scripts-generate-static-html', () => {
...
     debug: config.IS_DEBUG,
     fullPaths: config.IS_DEBUG,
-    transform: [babelify]
+    transform: [babelify],
+    standalone: 'bootstrap'
   }).external(vendorModules));
 });
 
+
+gulp.task('scripts-app-individual', (done) => {
```
lmorchard: Took a quick crack at simplifying / complicating this with Promises, since gulp supports Promises and `Promise.all()` does tracking of async queue completion:
```javascript
gulp.task('scripts-app-individual', () => {
  return Promise.all(
    fs.readdirSync(config.SRC_PATH + 'pages')
    .filter(entry => entry.endsWith('.js'))
    .map(entry => new Promise((resolve, reject) => {
      browserify({
        entries: [config.SRC_PATH + 'pages/' + entry],
        debug: config.IS_DEBUG,
        fullPaths: config.IS_DEBUG,
        transform: [babelify],
        standalone: entry.slice(0, entry.indexOf('.'))
      }).external(vendorModules).bundle().pipe(
        fs.createWriteStream(config.DEST_PATH + 'static/app/' + entry)
          .on('finish', resolve)
      );
    }))
  );
});

```

frontend/src/generate-static-html.js
```diff
@@ -1,24 +1,42 @@
...
+) {
   const head = ReactDOMServer.renderToStaticMarkup(headComponent);
-  const root = ReactDOMServer.renderToStaticMarkup(bodyComponent);
+  const staticRoot = ReactDOMServer.renderToStaticMarkup(staticBodyComponent);
+  let dynamicRoot;
+  let bootstrapScript;
+  if (prepareForClient) {
+    dynamicRoot = ReactDOMServer.renderToString(dynamicBodyComponent);
+    bootstrapScript = `<script src="/static/app/bootstrap.js"></script>
+<script>${componentName.slice(0, componentName.indexOf('.'))}.default(${dynamicBodyArgument});</script>`;
```
lmorchard: I think this is going to run afoul of [CSP rules](https://github.com/mozilla/testpilot/blob/456c7238f842c593cd6cf84068a0676b09da1747/bin/deploy.sh#L40): Inline JS source is disallowed.

Might work better to stick this info in a data-* attribute on `<body>` or something and handle it on load.

ghost:
Thanks for the updates here.  The new UX designs are being done as part of #2394 and, as mentioned, will be ready the sprint after next.  Getting rid of the loading animation will be a big win for loading time.

So, from here, I'd like to land the visual changes in #2394 and this rebased patch in the sprint after next.  If that's too far away for some of this, let's split those pieces into smaller PRs and land them sooner.

Thanks everyone!

frontend/src/app/actions/newsletter-form.js
lmorchard: FWIW, I think `createActions()` from redux-actions auto wrapped the undefined return value in a minimal action structure with a type name

frontend/src/bootstrap.js
```diff
@@ -0,0 +1,12 @@
+
```
lmorchard: This should probably live in `frontend/src/app/bootstrap.js`

frontend/src/generate-static-html.js
```diff
@@ -1,24 +1,42 @@
...
+    bootstrapScript = '';
+  }
   return `<!DOCTYPE html>
 <html>
   ${head}
   <body class="blue">
     <div class="stars"></div>
-    <div id="page-container">${root}</div>
+    <div id="page-container">${dynamicRoot}</div>
+    <div id="static-root">${staticRoot}</div>
```
lmorchard: Behavior of this static-root element is confusing me: I see that it's got the script includes for vendor JS and the individual page. But, it's also got a `<noscript>` block and the loading bars - both of which appear scrolled off the page because they're included *after* the server-rendered markup for the page content.

Maybe a) we don't need the loading bars after #2394; b) the noscript block should be at the top of the page; c) this div doesn't need to be removed?

frontend/src/generate-static-html.js
```diff
@@ -1,24 +1,42 @@
```
lmorchard, line 1: Figuring out how this module gets used twisted my melon a bit:

* [We generate `frontend/build/static/app/generate-static-html.js` via browserify](https://github.com/mozilla/testpilot/pull/2351/files#diff-2001630d3abeb76e05328a0b74f00a76R61)
* This bundle gets deployed to the servers - but we don't actually use it anywhere on the site
* Instead, this is [a library for *gulp tasks*](https://github.com/mozilla/testpilot/pull/2351/files#diff-1f5c2f0f3cd2018239cc9819551e5693R65) to do the static React generation things

I'd say this stuff more properly belongs in `frontend/tasks/util` except that it needs babel & browserify preprocessing to summon React powers into the gulp context. It probably shouldn't be shipped to the web along with the other app JS, though.

lmorchard (changes requested):
Spent today soaking my head in this and I think I understand what's going on. Overall, I think there's good stuff going on.

TL;DR: CSP will break things; UX from 2394 seems key; bundle size needs some work - all of which to say I think I verify @fzzzy's "need to fix items" and add a CSP item for good measure

I also had thought that maybe this could be spliced apart into simpler & smaller PRs, but I don't quite think that now after having gone through it.

A bit more summary of the comments to follow:

* This needs fixing - CSP on dev/stage/prod will deny execution of the inline JS in the `bootstrapScript` in `frontend/src/generate-static-html.js` that initializes the page
* Seems like we need #2394 to have a calmed-down UX that we can render server-side and not change drastically / noticeably on add-on detection
* The `static-root` element behavior seems odd / wrong - maybe related to #2394 
* `frontend/build/static/app/generate-static-html.js` probably shouldn't be deployed with the rest of the app JS
* symlinks seem problematic, but I could be out of date in my understanding
* Bundle size seems problematic

More on bundle size - we should try to split more out of the individual page bundles into a shared app bundle cached by the first page load. Like `vendor` except for our own code shared across pages.

I used [discify](https://github.com/hughsk/disc) to explore the bundles and noticed a few things:

* The bulk of most bundles comes from the same modules - dealing with the redux store, header / footer components, etc
* Page bundles each contain a copy of all the experiment content [because we started bundling in `experiments.json` at build time](https://github.com/fzzzy/testpilot/blame/456c7238f842c593cd6cf84068a0676b09da1747/frontend/src/app/store.js#L12). I didn't know we started doing that, but it seems like loading the nicely-cacheable JSON resource separately is more efficient by comparison in this new scheme
* Every bundle has a copy of React - which is weird, because I thought that was already in `vendor.js`

Going to keep poking some more, but I think there are a few actionable things I found so far.

lmorchard:
One more thought: It might turn out that a significant percentage of the site JS is shared and actual individual page bundles are super tiny - in which case, individual page bundles might not make as much pragmatic sense.

fzzzy:
Thanks so much for your comments les!

frontend/src/pages/privacy
fzzzy: Yes -- this was introduced by https://github.com/mozilla/testpilot/pull/2357 and I just moved the symlinks into a different directory. There was discussion about windows at the time, and I did some research which indicated git might not have as many problems with symlinks any more. Since testing these legal docs on windows isn't really that important, it didn't seem like a high priority. I'm still in favor of removing the symlinks if you have a better suggestion -- this just seemed easy. It should probably be done in another bug though.

frontend/src/app/components/ExperimentRowCard.js
fzzzy: You're right, a second prop would be much clearer. This started out as a quick hack at the beginning of the branch when I didn't think there was any possibility that this would land. I'll try to switch it to another property.

frontend/src/app/actions/browser.js
fzzzy: Good, I'm glad there wasn't some secret reason for doing it that way :)

frontend/tasks/scripts.js
fzzzy: Sweet, I didn't know gulp supported promises! Looks great, thanks.

frontend/src/generate-static-html.js
fzzzy: You're right, it shouldn't be shipped to the web. I wasn't really sure where to put it! I'll figure out where to put this and make sure it doesn't get shipped before this lands.

frontend/src/generate-static-html.js
fzzzy: Doh! Good catch. Good idea on the data-* attribute.

frontend/src/generate-static-html.js
fzzzy: Yeah, I don't think this makes sense going forward. I think it is left over from an intermediate state where I wasn't fully rendering the server-rendered markup. I think I can remove static-root entirely now that the server-rendered markup is fully implemented.

frontend/src/app/actions/newsletter-form.js
fzzzy: Great, that makes sense!

fzzzy:
About the individual bundles. I think you're right that most of the app is shared. At the moment the individual app bundles encode the component to be rendered at build time, and the component to hook up in the browser at runtime. I'll have to think about this more.

frontend/src/generate-static-html.js
fzzzy: I figured out why this gets shipped to the web. Because it calls commonBrowserify, it gets put into the static/app directory. I wasn't expecting that. I'll have to change it to not call commonBrowserify.

frontend/src/app/index.js
fzzzy: bootstrap.js has been renamed back to app.js since I switched back to having one big browserify bundle instead of multiple smaller ones. However the implementation is quite different.

frontend/src/generate-static-html.js
fzzzy: This is fixed now, generate-static-html.js gets put in build/ instead of build/static/app/ and does not get shipped to the site.

frontend/tasks/scripts.js
```diff
@@ -55,36 +61,72 @@ gulp.task('scripts-misc', () => {
...
     debug: config.IS_DEBUG,
     fullPaths: config.IS_DEBUG,
-    transform: [babelify]
+    transform: [babelify],
+    standalone: 'app',
+    bundleExternal: false
   }).external(vendorModules));
 });
 
+gulp.task('scripts-app-individual', () => {
```
lmorchard, line 83: Since [switching back to one app.js bundle](https://github.com/mozilla/testpilot/pull/2351/commits/9b339acb8a24b2923cee1e03d8bca2fd8a2c6f45), this task isn't needed anymore right?

frontend/tasks/scripts.js
```diff
@@ -61,14 +61,23 @@ gulp.task('scripts-misc', () => {
...
-  return commonBrowserify('generate-static-html.js', browserify({
+  return browserify({
     entries: [config.SRC_PATH + 'generate-static-html.js'],
     debug: config.IS_DEBUG,
     fullPaths: config.IS_DEBUG,
     transform: [babelify],
     standalone: 'generate-static-html',
     bundleExternal: false
-  }).external(vendorModules));
+  }).external(vendorModules)
```
lmorchard: Might be better to tweak `commonBrowserify` to accept an option for output path than copypasta most of it here.

frontend/tasks/scripts.js
```diff
@@ -55,36 +61,72 @@ gulp.task('scripts-misc', () => {
 });
 
 gulp.task('scripts-generate-static-html', () => {
-  return commonBrowserify('generate-static-html.js', browserify({
+  return browserify({
```
lmorchard: Might be better to tweak `commonBrowserify` to accept an option for output path than copypasta most of it here.

frontend/src/app.js
```diff
@@ -20,11 +20,24 @@ import share from './pages/share.js';
...
 
-error();
-experiment();
-experiments();
-home();
-legacy();
-onboarding();
-retire();
-share();
+const routes = {
```
lmorchard: He who removes react router is doomed to repeat it. 😀

frontend/src/app/components/MainInstallButton.js
```diff
@@ -54,6 +54,7 @@ export default class MainInstallButton extends React.Component {
       installButton = this.renderOneClickInstallButton(experimentTitle);
     } else {
       installButton = <span className="default-btn-msg" data-l10n-id="landingInstallButton">
+        Install the Test Pilot Add-on
```
lmorchard, line 61: Is this one of the bits where we removed the DOM text because it led to conflicts between React & L20n?

frontend/src/app.js
```diff
@@ -0,0 +1,43 @@
...
+  experiment,
+  experiments,
+  home,
+  legacy,
+  onboarding,
+  retire,
+  share
+};
+
+const spl = window.location.pathname.split('/');
```
lmorchard: We probably want to trim off & ignore trailing slashes here, too. (e.g. `/experiments` vs `/experiments/` and `/experiments/snooze-tabs` vs `/experiments/snooze-tabs/`

frontend/src/app/containers/HomePage.js
```diff
@@ -1,13 +1,9 @@
 import React from 'react';
 
 import HomePageNoAddon from './HomePageNoAddon';
-import HomePageWithAddon from './HomePageWithAddon';
 
 export default class HomePage extends React.Component {
   render() {
-    if (this.props.hasAddon) {
```
lmorchard: <strike>With this logic gone, is there a difference between HomePage and HomePageNoAddon anymore? Could just delete one or the other.</strike>

Wait, never mind: [happened in a later commit](https://github.com/mozilla/testpilot/pull/2351/commits/82f1b18ef5adf1ab6b282464c41faf27d3334a05)

lmorchard (approved):
Re-reviewed, have a couple of nits. Wouldn't mind taking another look through after a squash & rebase. Had trouble following a few things through the blizzard of tiny commits

frontend/tasks/scripts.js
fzzzy: It actually still is, because we use these scripts to do the server-side rendering. Not ideal, but maybe I can revisit this once I have more experience with the code in practice.

I did try to split things out so that all the app code was in a bundle and each of the scripts-app-individual bundles were very very tiny, but I had trouble getting the browserify external stuff to work the way I wanted. I think every single subpackage needs to be specified as being external, and not just a top level package? Hopefully I'm wrong about that.

frontend/tasks/scripts.js
fzzzy: Done.

frontend/src/app/components/MainInstallButton.js
fzzzy: Yes, it is, but now that we no longer have react-router trying to change the nodes that l20n touched, it isn't a problem any more.

Unless the page itself tries to change some nodes after l20n touches them, which isn't the case here.

frontend/src/app.js
fzzzy: Indeed. I would like to go back to the way it was before with one bundle per page so I don't have to reimplement client side routing, but only if I can solve the bundle bloat problem, and it is not a high priority.

frontend/src/app.js
```diff
@@ -0,0 +1,48 @@
+
+import 'babel-polyfill/browser';
+import es6Promise from 'es6-promise';
+import 'isomorphic-fetch';
+import 'l20n/dist/compat/web/l20n';
```
lmorchard: Looks like this needs to be `import 'l20n';` since [we upgraded L20n](https://github.com/mozilla/testpilot/commit/3681f7a5968785a3118250ff4a128eb38c511009#diff-9e31d7481520a2040f72bb71e10130f1R2)—causes [a CircleCI test run failure](https://circleci.com/gh/mozilla/testpilot/4087) that I can reproduce locally. This seems to fix it, after an `rm -rf node_modules && npm install` to make sure nothing's stale locally

fzzzy: Oh, I thought this was fixed during the rebase, but the fix must have gotten lost. Thanks for finding it!

fzzzy: Oops, I had it fixed locally but didn't commit it!

lmorchard:
That seems to have fixed it - things look better. Can you squash in that last new commit?

frontend/tasks/scripts.js
```diff
@@ -58,36 +64,66 @@ gulp.task('scripts-generate-static-html', () => {
...
+    standalone: 'generate-static-html',
+    bundleExternal: false
+  }).external(vendorModules), config.DEST_PATH);
 });
 
-gulp.task('scripts-app-main', () => {
+gulp.task('scripts-app', () => {
   return commonBrowserify('app.js', browserify({
-    entries: [config.SRC_PATH + 'app/index.js'],
+    entries: [config.SRC_PATH + 'app.js'],
```
lmorchard, line 74: Just realized this moved... I think it's because `bootstrap.js` came & went. This should probably be moved back to `frontend/src/app/index.js` rather than `frontend/src/app.js`

frontend/src/app.js
```diff
@@ -0,0 +1,48 @@
+
```
lmorchard, line 1: Yeah, kind of want to see this moved back to `frontend/src/app/index.js` - if only to see a clearer diff in changes with the old react-router powered version

frontend/src/app.js
```diff
@@ -0,0 +1,48 @@
...
+  experiment,
+  experiments,
+  home,
+  legacy,
+  onboarding,
+  retire,
+  share
+};
+
+let path = window.location.pathname;
```
lmorchard, line 34: Also, this might be a crazy idea / idea for a future PR—what if each static-rendered page had this:
```html
<body data-page-name="experiment" data-page-param="snooze-tabs">
```
The server naturally does the routing for us, and this block of code could be stripped down:
```javascript
const name = document.body.dataset.pageName;
const param = document.body.dataset.pageParam;
if (name in routes) {
  routes[name](param);
} else {
  routes.error();
}
```

frontend/tasks/scripts.js
lmorchard: Hmm, this makes me frown a little. But, I wonder if the real fix for this is to [switch to babel & ES6 in gulp itself](https://markgoodyear.com/2015/06/using-es6-with-gulp/) in a follow up PR? That way, we don't need to build bundles for use in the build process in the first place.

lmorchard:
Okay, gave this a looking over again again and my eyes are glazing over. I can't see anything particularly problematic except rough edges I think I'd like to see sanded off in follow up PRs. Main final things:

* Move `frontend/src/app.js` back to `frontend/src/app/index.js` so the diff is a bit clearer

* One final squash, maybe with [a cleaner / fleshed out commit message](https://chris.beams.io/posts/git-commit/) with more bullet points inventorying what all was done here. 

I feel particular about commit messages, mostly because they can be the best help or worse hindrance when looking through history & blame months later. And this is a biggun

lmorchard:
Alright, one last thing, since it was bugging me and I wondered how hard it would be - here's a pull request against your pull request branch: https://github.com/fzzzy/testpilot/pull/1

That switches to ES6 in Gulp. Doing that seems to have made getting rid of all the generated bundles easier than I expected. If that looks good, maybe merge into your branch, squash again - and except for [that last idea I had for routing](https://github.com/mozilla/testpilot/pull/2351#discussion_r115818908), I think we are basically ready to merge?

lmorchard:
Whoops - found a bug in my work. Forgot `frontend/tasks/util.js` is also imported by the frontend, so I have to yank the stuff that uses `react-dom/server` out of there and move it into the gulp task directly

lmorchard:
Okay, just updated https://github.com/fzzzy/testpilot/pull/1 with a small bugfix *and* tossed in my idea to excise window.location based routing altogether. And with that, I think I'm out of nits and ideas

lmorchard:
Actually, here, I'm going to be presumptive and close this well-trod PR in favor of this one: https://github.com/mozilla/testpilot/pull/2426

Addressed all my own nits, gave things one more squash & going over, tried to inventory the changes in more detail if only to make sure I had my head properly wrapped around things.

lmorchard:
(we can always re-open this one and nuke mine if it's a problem!)

fzzzy:
Thanks @lmorchard, #2426 looks good :) really looking forward to getting this landed. Will look closely on monday.

PR #2354, opened 2017-03-27
2350 reload window on disconnect

fzzzy:
This runs everything from scratch, preventing a situation where moving
the test pilot window to a new firefox window would double-poll the
test pilot addon for experiment state, causing infinite reloading


lmorchard (approved):
:+1: Also probably long since safe to be rid of this code branch for v0.8

PR #2355, opened 2017-03-28
2353 stop using unsafeWindow

fzzzy:


addon/src/lib/WebApp.js
```diff
@@ -55,13 +55,6 @@ export default class WebApp {
         worker.on('detach', () => this.hub.disconnect(worker.port));
       }
     });
-    this.beacon = new PageMod({
```
dannycoates, line 58: I don't know if anyone is using this or not (probably not) but we ought to check first before removing it and let them know. The whitelist has `'https://www.mozilla.org/*,about:home'`.

frontend/src/app/store.js
```diff
@@ -26,7 +26,7 @@ export const initialState = {
     data: experiments.results
   },
   addon: {
-    hasAddon: !!window.navigator.testpilotAddon,
+    hasAddon: null,
```
dannycoates, line 29: `false` maybe?

dannycoates: oic `null` is the uninitialized state. gross but whatevs

fzzzy: Yes, it's gross but it works. The only other way I can think of fixing the situation is to move the "has addon" pages to a different url tree than "doesn't have add-on" pages, which I think we need to do for other architectural reasons too. Then we can stop using the null hack.

fzzzy:
Ok, @lmorchard says about:home and mozilla.org do rely on this, so I will close this pr for now until we figure out what they use it for and if we can provide an async way for them to discover this information.

PR #2358, opened 2017-03-28
Fix #2162 big-border button focus state removes border.

### LINKED ISSUE #2162: "Install" button new thick white surrounding border is non persistent on Windows and Linux
[Notes]:
- This issue is only reproducible on stage server and only on Windows and Linux OSes.
 
[Affected versions]:
- Firefox 51.0.1 and above

[Affected Platforms]:
- All Windows
- All Linux

[Prerequisites]:
- Have a profile which contains the A/B testing experiment enabled.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites and navigate to "https://testpilot.stage.mozaws.net/" page.
2. Click the "Install the Test Pilot Add-on" button.
3. While observing the button thick white surrounding border, hover away from the button.
4. Click on a free area of the page and observe the button behavior. 

[Expected result]:
- The button surrounding white border remains unchanged no matter what actions are performed on the page.

[Actual result]:
- Step 3: The white surrounding border wrongly disappears when hovering away the button.
- Step 4: The surrounding white border reappears and persists even if button is hovered.

[Additional Notes]:
- While performing the STR we also noticed that the "Installing..." animation color is darker (like hovered state) after choosing to install TxP and becomes brighter after clicking on a free area of the page.
- Here is a screen recording of the issue: https://goo.gl/5HrULz.

ScottDowne:


ScottDowne:
Turns out it was a CSS issue.

The default test border state was higher CSS specificity while focused, thus removing the big-border test's border.

See: https://github.com/mozilla/testpilot/blob/master/frontend/src/styles/modules/_buttons.scss#L58-L59

fzzzy:
Thanks!

PR #2375, opened 2017-04-10
add og:url

### LINKED ISSUE #911: A few warnings when trying to validate Open Graph card support
Re: https://developers.facebook.com/tools/debug/sharing/?q=http%3A%2F%2Ftestpilot.dev.mozaws.net%2F

I'm getting a few warnings:

> ### Warnings That Should Be Fixed
> 
> | TYPE | MESSAGE |
> | --- | --- |
> | Extraneous Property | Objects of this type do not allow properties named 'twitter:title'. |
> | Extraneous Property | Objects of this type do not allow properties named 'twitter:description'. |
> | Extraneous Property | Objects of this type do not allow properties named 'twitter:url'. |
> | Extraneous Property | Objects of this type do not allow properties named 'twitter:image'. |
> | Share App ID Missing | The 'fb:app_id' property should be explicitly provided, Specify the app ID so that stories shared to Facebook will be properly attributed to the app. Alternatively, app_id can be set in url when open the share dialog. |
> | Parser Mismatched Metadata | The parser's result for this metadata did not match the input metadata. Likely, this was caused by the data being ordered in an unexpected way, multiple values being given for a property only expecting a single value, or property values for a given property being mismatched. Here are the input properties that were not seen in the parsed result: 'twitter:title, twitter:description, twitter:url, twitter:image' |

johngruen:
Fixes #911

### DIFF
--- a/frontend/src/generate-static-html.js
+++ b/frontend/src/generate-static-html.js
@@ -47,6 +47,7 @@ export function generateStaticPage({
     <meta name="twitter:card" content="summary" />
     <meta property="og:image" content={ image_facebook } />
     <meta name="twitter:image" content={ image_twitter } />
+    <meta property="og:url" content="https://testpilot.firefox.com" />
   </head>;
   const bodyComponent = <div>
     <noscript>


ScottDowne:
This successfully removed the og:url warning for me.

Can I give this an r+ :D

ScottDowne:
For what it's worth, if we really wanted to remove the app_id warning, we can put in the default app id.
 
`<meta property="fb:app_id" content="966242223397117" />`

This is what facebook puts in anyway if we don't have it, as per the link John found in #911 http://stackoverflow.com/questions/35066102/facebook-debugger-why-complaining-about-app-id-missing

This would probably just satisfy the warning if that's bugging people (I can see it being annoying), but wouldn't actually change anything useful. We can also just as functionally ignore it.

PR #2380, opened 2017-04-11
Updated tests to use tox, updated circleci config, changed shell script

b4handjr:
r? @fzzzy 

I switched from marionette to geckodriver for running these tests. They utilize tox, pytest, and selenium, as well as all current best practices.

I added them to the frontend folder as these are 'frontend' tests but they can be moved if need be.


b4handjr:
Should I keep trying with circle? I haven't heard anything from the team yet.

lmorchard:
Can someone summarize what the issues with CircleCI are? I heard there were issues in our standup today, but I'm not sure what they are exactly. Something about a port being taken?

fzzzy:
Yes, we just had a meeting about it and the consensus was that ops wants us to use circle because it produces better artifacts.

b4handjr:
Cool thanks for the update. I'll keep trying then.

fzzzy:
@lmorchard It doesn't happen if you run the integration tests locally. But you can see what happens in the circleci output. Even though the dev server seems to be starting, the firefox instance isn't able to connect to it for unknown reasons.

Is there a way we could get a screenshot of the result firefox sees or at least some more log messages, so we can figure out what the heck is going on?

lmorchard:
Ops started us on CircleCI because it was easier to generate Docker containers. We don't do that anymore since we switched to static site deployment, so that reason isn't relevant. 

However, we *do* deploy to our dev servers on every successful master branch test run, which seemed easier to set up on CircleCI than on Travis. We can switch as long as all that machinery keeps working

b4handjr:
> Can someone summarize what the issues with CircleCI are? I heard there were issues in our standup today, but I'm not sure what they are exactly. Something about a port being taken?

From how the tests run it seems to be that the server starts but the port isn't able to be accessed or something.

lmorchard:
I might try poking at this on my own copy of the branch, but FWIW you can "Rebuild with SSH" on the button in the results page to connect to test env and try commands in a shell there if that helps to diagnose

b4handjr:
> I might try poking at this on my own copy of the branch, but FWIW you can "Rebuild with SSH" on the button in the results page to connect to test env and try commands in a shell there if that helps to diagnose

Yeah I did this and there was nothing on the port. But something has to be there as the curl script that checks the address returns true before the server starts.

bin/circleci/run-integration-tests.sh
```diff
@@ -8,21 +8,18 @@
...
 #    date libraries (~25)
 #  * r=_6a68 and r=clouserw.  Conveniently discussed over the holiday break ;)
-exit 0
 
 npm start &
 STATIC_SERVER_PID=$!
 
 # Wait until the server is available...
-until $(curl --output /dev/null --silent --head --fail http://testpilot.dev:8000); do
+until $(curl --output /dev/null --silent --head --fail -k https://example.com:6678); do
```
lmorchard: What's expected to start on port 6678? The server started with gulp via `npm start` [is still on 8000](https://github.com/mozilla/testpilot/blob/ef97b71fbe47d93536529e9d6cdc59c1898de273/frontend/config.js#L4) - I don't see anything in this PR that changes it to 6678. And, I see this in my test output:
```
[18:10:00] Server started https://0.0.0.0:8000
```
Maybe changing this 6678 back to 8000 will fix? or vice versa, change `frontend/config.js` to accept an ENV var option if 8000 is taken by something else in circleci

b4handjr: Hmm maybe I forgot to change it back

b4handjr:
Still haven't really found a solution. I asked on the circleci help section: https://discuss.circleci.com/t/https-localhost-port-already-in-use/11968

lmorchard:
FWIW, poking a little more. I don't really understand the testing bits, yet, but I don't think the HTTPS server is the problem. I made a few manual tweaks to `run-integration-tests.sh` on a test box rebuilt with SSH - and as far as I can tell, the gulp web server is running before and after the `tox` command.

Both of these curl commands come up with the expected home page HTML just after server start is detected and just before the server is killed. The catch, though, is the SSL cert is self-signed (thus the `-k` option in curl). Could this be an issue for the tests?

```diff
diff --git a/bin/circleci/run-integration-tests.sh b/bin/circleci/run-integration-tests.sh
index f156ef4..6e5980f 100755
--- a/bin/circleci/run-integration-tests.sh
+++ b/bin/circleci/run-integration-tests.sh
@@ -17,11 +17,17 @@ until $(curl --output /dev/null --silent --head --fail -k https://example.com:80
     printf '.'; sleep 1
 done

+echo "Server started"
+curl -k https://example.com:8000/
+
 # Fire up the integration tests with Marionette
 tox
 TEST_STATUS=$?

-kill $STATIC_SERVER_PID
+curl -k https://example.com:8000/
+echo "Killing server"
+
+kill -9 $STATIC_SERVER_PID

 # Report what happened
 if [ $TEST_STATUS -eq 0 ]
```

b4handjr:
Well I have firefox accepting insecure certs so it shouldn't be a cert thing, unless that capability isn't working.

b4handjr:
@lmorchard Look like the server is working fine. Maybe it's something else. The test works fine on travis with the same settings so i really don't have a clue.

@davehunt do you have any idea what could be wrong here?

b4handjr:
Just wanted to update this. I haven't found a solution to get past the about:neterror. I am thinking it could be geckodriver, as version .16 is about to release but seeing that everything works on travis, it seems like a difference in something circle ci is doing and I can't pinpoint what. Any help would be appreciated.

frontend/test/ui/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.fixture
+def firefox_profile(firefox_profile):
+    firefox_profile.set_preference(
+        'extensions.install.requireBuiltInCerts', 'false')
+    firefox_profile.set_preference('xpinstall.signatures.required', 'false')
+    firefox_profile.set_preference('extensions.webapi.testing', 'true')
+    firefox_profile.set_preference('testpilot.env', 'local')
+    return firefox_profile
```
davehunt: Rather than passing an entire profile, you can now pass individual preferences. For example:

```python
@pytest.fixture
def firefox_options(firefox_options):
    firefox_options.set_preference('extensions.install.requireBuiltInCerts', False)
    firefox_options.set_preference('xpinstall.signatures.required', False)
    firefox_options.set_preference('extensions.webapi.testing', True)
    firefox_options.set_preference('testpilot.env', 'local')
    return firefox_options
```

The current pytest-selenium will still send a profile, but a future version will only do so when a profile object is used. By using `firefox_options` now, this will reduce the new session payload in the future.

frontend/test/ui/pages/desktop/base.py
```diff
@@ -0,0 +1,17 @@
+from pypom import Page
+
+from selenium.webdriver.common.by import By
+
+
+class Base(Page):
+
+    _copter_locator = (By.CLASS_NAME, 'copter')
```
davehunt, line 8: What's 'copter'? Class names can be obscure, but ideally page objects will be less so.

frontend/test/ui/pages/desktop/base.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self, selenium, base_url, locale='en-US', **kwargs):
```
davehunt: Are the pages localised, or is `locale` an artifact of copy/paste from another project? I don't see it used anywhere in this patch.

frontend/test/ui/pages/desktop/base.py
```diff
@@ -0,0 +1,17 @@
...
+        super(Base, self).__init__(
+            selenium, base_url, locale=locale, timeout=30, **kwargs)
```
davehunt: Do we need a 30 second timeout? The default of 10 should be more than enough for most applications.

frontend/test/ui/pages/desktop/home.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @property
+    def header(self):
+        return self.Header(self)
+
+    class Header(Region):
+
+        _root_locator = (By.CLASS_NAME, 'banner')
+
+        _copter_locator = (By.CLASS_NAME, 'copter')
```
davehunt: Duplicate from base.py?

frontend/test/ui/pages/desktop/home.py
```diff
@@ -0,0 +1,21 @@
...
+
+        @property
+        def copter(self):
+            return self.find_element(*self._copter_locator)
```
davehunt: We shouldn't return web elements from page objects. Instead we should provide a property such as `is_copter_displayed`. This way the test only has to be concerned with the API of the page object, and not that of Selenium, etc.

frontend/test/ui/requirements/requirements.txt
```diff
@@ -0,0 +1,10 @@
+foxpuppet==1.0a1
```
davehunt: As far as I can tell, FoxPuppet isn't being used by any tests.

frontend/test/ui/requirements/requirements.txt
```diff
@@ -0,0 +1,10 @@
...
+fxapom==1.10.0
```
davehunt: This is also unused.

frontend/test/ui/requirements/requirements.txt
```diff
@@ -0,0 +1,10 @@
...
+PyJWT==1.4.2
```
davehunt: Unused.

frontend/test/ui/requirements/requirements.txt
```diff
@@ -0,0 +1,10 @@
...
+PyPOM==1.1.1
+pytest==3.0.7
+pytest-instafail==0.3.0
```
davehunt: Unused.

frontend/test/ui/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[tool:pytest]
+base_url = https://example.com:8000/
```
davehunt: This configuration can also go into `tox.ini`.

tox.ini
```diff
@@ -0,0 +1,17 @@
+[tox]
+envlist = ui-tests, flake8
+skipsdist = True
+
+[testenv:ui-tests]
+passenv = DISPLAY
+deps = -rfrontend/test/ui/requirements/requirements.txt
+commands =
+    pytest -v --driver=Firefox --firefox-path=firefox/firefox frontend/test/ui {posargs}
```
davehunt: `frontend/test/ui` should be the default positional argument, to allow specifying an alternate path: `{posargs:frontend/test/ui}`.

davehunt: If you add pytest-html and `--html` you can generate a HTML report. CircleCI can then archive this as an artifact, which may help with investigating issues.

tox.ini
```diff
@@ -0,0 +1,17 @@
...
+
+[testenv:flake8]
+skip_install = true
+deps = -rfrontend/test/ui/requirements/flake8.txt
+commands = flake8 frontend/test/ui
```
davehunt: It's probably an idea to have one linting job for all Python code in the repository, rather than make this specific to the UI tests.

frontend/test/ui/pages/desktop/base.py
b4handjr: Copter is the name of the class in which the firefox is within the ship. It slides up from the bottom at https://testpilot.firefox.com

frontend/test/ui/pages/desktop/home.py
b4handjr: So this is for the header of the main home page. I guess since this is a Single Page Application, until you click on an experiment I didn't know if it should be called base, or home. But the ```copter``` is still something that should be checked as displayed in my opinion.

frontend/test/ui/requirements/requirements.txt
b4handjr: I will be adding an install experiment test so this was just jumping ahead a bit conforming to project goals.

frontend/test/ui/setup.cfg
b4handjr: I could call it ```--base-url``` correct?

davehunt: There's no need to change it, you can just move it into `tox.ini`.

frontend/test/ui/requirements/requirements.txt
davehunt: I'd add the additional dependencies in the patch that uses them.

frontend/test/ui/pages/desktop/home.py
davehunt: That's fine, but we can just use the one in `Base`, and `Home` can just be a simple subclass.

tox.ini
b4handjr: I suspect we will need a geckodriver log too.... *wink wink* 😉

frontend/test/ui/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@pytest.fixture
+def foxpuppet(selenium):
+    return FoxPuppet(selenium)
+
+
+@pytest.fixture
+def capabilities(capabilities):
+    capabilities['marionette'] = True
```
AutomatedTester, line 6: This is the default in Selenium, why are you setting it here?

davehunt: It's not the default when using pytest-selenium, which doesn't currently use `DesiredCapabilities.FIREFOX`.

b4handjr:
Using the curl script 
```
until $(curl -s --head http://example.com:8000 | head -n 1 | grep "HTTP/1.[01] [23].." > /dev/null); do
     printf '.'; sleep 1
 done
```

This returns as true before the server even starts.

b4handjr:
@pdehaan @fzzzy  Could I get this up in Travis for now?

fzzzy:
@jrbenny35 Yes, hook them up to travis and I'll talk to ops about why they prefer circle.

fzzzy:
@lmorchard would like to figure out why they are not working on circle, since circle also deploys to our dev servers.

b4handjr:
I am all ears as I have tried everything I can think of. Even having a chat with vladikoff on the fxa team as they have a similar setup.

lmorchard:
Well, this is bizarre: I run the test with ssh enabled, [it fails.](https://circleci.com/gh/lmorchard/testpilot/685) I ssh in to run the test script manually, [it passes](https://pastebin.mozilla.org/9021926). Trying to figure out what's different between the automated run and the manual run

lmorchard:
A little closer to... something? It looks like the web server starts up, lives long enough for the first request from curl, then dies or is killed before the tests run. This doesn't happen when I ssh in and run things manually. 

So, it seems like the automated test run is killing the background process, even with [the `background` flag in `circle.yml`](https://github.com/lmorchard/testpilot/blob/dc2ba2e52cac16b0de5fb2ab0326348514516e22/circle.yml#L34). Still weird, since this didn't happen back when we had integration tests using Marionette.

lmorchard:
Still fumbling forward: Looks like if I don't run `tox`, the gulp server never dies. So, I'm wondering what exactly `tox` is doing, and if any of that might kill processes

lmorchard:
Okay, so [I got the test run to pass](https://circleci.com/gh/lmorchard/testpilot/706) by [not using `tox` and running `flake` and `pytest` directly](https://github.com/jrbenny35/testpilot/compare/initial_ui_tests...lmorchard:jrbenny35-initial_ui_tests?expand=1).

Not sure what's going on with `tox` in CircleCI, but somehow the gulp server background process dies or gets killed as soon as `tox` starts. When I stop using `tox`, the server survives to the end of the integration test script.

Doesn't seem like an out-of-memory thing, and the test run works fine with `tox` if I ssh in and run it by hand. 

I guess my advice at this point would be to try not to use `tox`. ¯\\_(ツ)_/¯

b4handjr:
Tox huh. I will take a look into that.


@davehunt what do you think? Tox killing a gulp server? Maybe I need to start the gulp server within tox?

b4handjr:
I had to move the ```run-integration-tests.sh``` script to the root as tox couldn't find it. Maybe I am doing it wrong.

tox.ini
```diff
@@ -0,0 +1,22 @@
+[tox]
+envlist = ui-tests, flake8
+skipsdist = True
+
+[testenv:ui-tests]
+passenv = DISPLAY
+deps = -rfrontend/test/ui/requirements/requirements.txt
+commands =
+    ./run-integration-tests.sh
+    pytest -v --driver=Firefox --firefox-path=firefox/firefox \
```
davehunt: Instead of using `--firefox-path` can you just ensure the correct version is in the system path? This command line option is deprecated in pytest-selenium and will be removed in a future version.

davehunt:
> @davehunt what do you think? Tox killing a gulp server? Maybe I need to start the gulp server within tox?

I've not encountered tox killing any processes before... maybe we could ask CircleCI for some assistance?

davehunt:
> I had to move the `run-integration-tests.sh` script to the root as tox couldn't find it. Maybe I am doing it wrong.

You should be able to reference a sub-directory from Tox. What error were you seeing?

b4handjr:
> You should be able to reference a sub-directory from Tox. What error were you seeing?

It was an invocation error

circle.yml
```diff
@@ -18,18 +21,28 @@ dependencies:
...
+    - pip install tox mozdownload mozinstall
     - google-chrome --version
     - ./bin/circleci/install-node-dependencies.sh
-    - ./bin/circleci/setup-test-dependencies.sh
     - ./bin/circleci/build-addon.sh
 
 test:
+  pre:
+    - sudo apt-get upgrade firefox
+    - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.16.0/geckodriver-v0.16.0-linux64.tar.gz
```
davehunt: I recently updated the tests for pytest-selenium to always get the latest GeckoDriver release. You could potentially do something similar: https://github.com/pytest-dev/pytest-selenium/blob/master/.travis.yml#L30-L31

circle.yml
```diff
@@ -18,18 +21,28 @@ dependencies:
...
+    - firefox --version
+    - geckodriver --version
   override:
     - ./bin/circleci/test-addon.sh
     - ./bin/circleci/test-frontend.sh
-    - ./bin/circleci/run-integration-tests.sh
+    - tox
   post:
     - bash <(curl -s https://codecov.io/bash)
+    - cp geckodriver.log $CIRCLE_ARTIFACTS
```
davehunt: With the latest version of pytest-selenium this will be `driver.log`, however the upcoming release will change this by creating a log for each driver instance and storing them in tmp. It'll probably be safest to use the log integrated with the HTML report.

circle.yml
```diff
@@ -18,18 +21,28 @@ dependencies:
     # node_modules is included by default
 
   override:
+    - pip install tox mozdownload mozinstall
```
davehunt, line 23: You're not using these anymore.

tox.ini
```diff
@@ -0,0 +1,21 @@
+[tox]
+envlist = ui-tests, flake8
+skipsdist = True
+
+[testenv:ui-tests]
+passenv = DISPLAY
+deps = -rfrontend/test/ui/requirements/requirements.txt
+commands =
+    sh bin/circleci/run-integration-tests.sh
+    pytest -v --driver=Firefox --html=results.html frontend/test/ui {posargs}
```
davehunt: You should use `--self-contained-html` so you don't need to copy all the assets.

b4handjr:
It seems that circle ci doesn't let firefox get upgraded. running ```sudo apt-get update; sudo apt-get install firefox``` seems to download and install firefox 53. but when I call ```firefox --version``` It says 47.0.1

dannycoates:
In screenshots we use `get-firefox` to install the version we want and change the `PATH` env. I don't know if it's the "right way" but it seems to work ok. https://github.com/mozilla-services/screenshots/blob/master/circle.yml#L17-L18

b4handjr:
The test failed but it should pass.

b4handjr:
@fzzzy I think I got this going. Take a look please

fzzzy (approved):
Ok, this looks good now, if it is ready to merge change the title of the pr and remove the "WIP".

b4handjr:
I can squash into one commit if you'd like

fzzzy:
@jrbenny35 Good idea, you better squash it.

b4handjr:
Should be good now

PR #2395, opened 2017-04-24
updated

vschs007:
indentation updated


vschs007:
ok let me do some

lmorchard:
Not sure what you're working toward in this pull request, but I don't think we want to accept just white space changes.

PR #2399, opened 2017-04-26
Upgrade l20n to 5.0.0

stasm:
See https://discourse.mozilla-community.org/t/l20n-and-the-test-pilot-website/15148 for background.

This PR upgrades the `l20n` dependency which requires small changes in the syntax of the localization files. The PR includes the required changes for en-US.  Other locales will be converted in Pontoon and synchronized to Github.

When the PR is approved, @mathjazz will disable Pontoon's sync in order to safely migrate all other locales. Please wait for his explicit comment before merging this PR into master.

### DIFF
--- a/frontend/src/app/index.js
+++ b/frontend/src/app/index.js
@@ -1,5 +1,5 @@
 import 'babel-polyfill/browser';
-import 'l20n/dist/compat/web/l20n';
+import 'l20n';
 import es6Promise from 'es6-promise';
 import 'isomorphic-fetch';
 
--- a/frontend/src/generate-static-html.js
+++ b/frontend/src/generate-static-html.js
@@ -30,8 +30,6 @@ export function generateStaticPage({
     <meta name="availableLanguages" content={ available_locales } />
     <meta name="viewport" content="width=device-width" />
 
-    <link rel="localization" href="/static/locales/en-US/app.ftl" />
-    <link rel="localization" href="/static/locales/en-US/experiments.ftl" />
     <link rel="localization" href="/static/locales/{locale}/app.ftl" />
     <link rel="localization" href="/static/locales/{locale}/experiments.ftl" />
 
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -1,13 +1,13 @@
 siteName = Firefox Test Pilot
 
-# Page titles, put in the <title> HTML tag.
+// Page titles, put in the <title> HTML tag.
 [[pageTitle]]
 pageTitleDefault = Firefox Test Pilot
 pageTitleLandingPage = Firefox Test Pilot
 pageTitleExperimentListPage = Firefox Test Pilot - Experiments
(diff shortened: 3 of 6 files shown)


frontend/src/generate-static-html.js
```diff
@@ -30,8 +30,6 @@ export function generateStaticPage({
     <meta name="availableLanguages" content={ available_locales } />
     <meta name="viewport" content="width=device-width" />
 
-    <link rel="localization" href="/static/locales/en-US/app.ftl" />
-    <link rel="localization" href="/static/locales/en-US/experiments.ftl" />
```
stasm, line 34: This was a workaround for #2249 landed in #2287. I managed to track down the root cause. The bug was in the Babel setup of the `compat` build of l20n.js v4. By default Babel doesn't support extending native objects like Errors. L20n uses custom Error subclasses to decide what kind of fallback it should use. Without `instanceof` working correctly due to Babel, L20n never decided to fall back to the next language and instead immediately displayed the id of the translation.

locales/en-US/app.ftl
```diff
@@ -52,37 +52,37 @@ landingDownloadFirefoxTitle = Firefox
...
-# A modal prompt to sign up for the Test Pilot newsletter.
+// A modal prompt to sign up for the Test Pilot newsletter.
 [[emailOptIn]]
 emailOptInDialogTitle = Welcome to Test Pilot!
 emailOptInMessage = Find out about new experiments and see test results for experiments you've tried.
 emailValidationError = Please use a valid email address!
-# LOCALIZATION NOTE: The ':)' characters in the emailOptInInput placeholder are a smiley face emoticon.
-emailOptInInput =
-  [html/placeholder] email goes here :)
+// LOCALIZATION NOTE: The ':)' characters in the emailOptInInput placeholder are a smiley face emoticon.
```
flodolo: Can you get rid of "LOCALIZATION NOTE" since you're touching this file?

frontend/src/app/index.js
```diff
@@ -1,5 +1,5 @@
 import 'babel-polyfill/browser';
-import 'l20n/dist/compat/web/l20n';
```
stasm, line 2: In previous versions of l20n there were two directories, `dist/bundle/web` and `dist/compat/web` containing the original and the transpiled builds respectively. l20n is now transpiled using Babel's Latest preset by default and the result is `dist/web/l20n.js` which is also specified as the `main` field in `package.json`.

locales/en-US/app.ftl
```diff
@@ -1,21 +1,21 @@
 siteName = Firefox Test Pilot
 
-# Page titles, put in the <title> HTML tag.
+// Page titles, put in the <title> HTML tag.
```
stasm, line 3: The summary of syntax changes is available at https://github.com/l20n/l20n.js/blob/master/CHANGELOG.md.

fzzzy (approved):
Ok, I tested this locally and it looks good. Merge when ready.

PR #2402, opened 2017-04-28
Update activity stream privacy docs link

k88hudson:
We recently updated our docs to a different folder structure. Is this the right place to make the change?


chuckharmston (approved):

chuckharmston:
Thanks Kate! These should go out with the next production deployment on 5/9.

Issue #2392, opened 2017-04-20
Newsletter submission should include a "source_url" parameter


pmac:
The email marketing team would very much appreciate you adding this parameter. It'd be best to send the value as the current URL of the page, but if you'd rather hard-code "https://testpilot.firefox.com/" that works for us as well.

ScottDowne:
Hey all, figured I would pick up another few small bugs while I had some time between campaigns.

I used the hard coded url because that's what I've used with my previous basket interactions, but happy to change it to something along the lines of `window.location.origin + window.location.pathname` or something else.

Thoughts?

pmac:
👍 from me. It might be best and more future proof to use the `window.location.*` values, but this is better than what we have and is probably more robust. Are there multiple URLs at which the newsletter form currently lives on testpilot.f.c?

lmorchard:
The newsletter form lives in the footer on all pages of Test Pilot, but I'm not sure we're all that worried about getting granular

pmac:
The email marketing team is pretty worried about it from what I understand, but this is a step forward. Thanks!

PR #2404, opened 2017-05-01
Fixes: #2392 Adding source_url to basket newsletter submissions.

ScottDowne:


lmorchard (changes requested):
Because #1758 is not yet addressed, there are actually two places where this change needs to be made. One is `actions/newsletter-form.js`, and this PR covers that. 

The other is [the `subscribeToBasket()` function in `frontend/src/app/containers/App.js`](https://github.com/mozilla/testpilot/blob/7a6a26c024a88536359796ad395c7cbf0ab9b377/frontend/src/app/containers/App.js#L107). So, this tweak needs to be copied there, too.

lmorchard (approved):
Actually, you know what: Never mind my request for changes. This looks good, and addressing #1758 will fix the other bit. So I'll just approve & merge. Thanks for the PR!

PR #2411, opened 2017-05-03
Switched to shared NewsletterForm in EmailDialog

### LINKED ISSUE #1758: Use NewsletterForm component in EmailDialog

### LINKED ISSUE #2392: Newsletter submission should include a "source_url" parameter

### LINKED ISSUE #2403: Stop sending "lang" with newsletter subscription requests to basket
Our newsletter system only supports a few specific languages, and the `test-pilot` newsletter specifically only supports English and German. The test-pilot site appears to be sending whatever language is first in the user's `accept-language` header however. Basket can parse this header looking for one we support in the list if the `lang` parameter is not sent. `lang` is really meant to only support the small set of languages we support and that a user can choose via a pick list. So please either:

1. Stop sending the `lang` parameter and let Basket work out the best language, or
2. Add a drop down containing "English" and "German" only so that the user can decide for themselves.

Thanks!

lmorchard:
- Rework `frontend/src/app/components/EmailDialog` to use
  `NewsletterForm` instead of local form implementation

- Add failure state to `EmailDialog`

- CSS tweaks for `NewsletterForm` when used in a dialog

- Drop lang parameter from Basket POST (#2403)

- Use window.location as source parameter for Basket POST (#2392)

- Move `subscribeToBasket` to shared `app/lib/utils`

- Deleted some now unused strings

- Updated tests to reflect all the above

Fixes #1758.
Fixes #2392.
Fixes #2403.

### DIFF
--- a/frontend/src/app/actions/newsletter-form.js
+++ b/frontend/src/app/actions/newsletter-form.js
@@ -1,7 +1,4 @@
-
-
-export const basketUrl = 'https://basket.mozilla.org/news/subscribe/';
-export const sourceUrl = 'https://testpilot.firefox.com/';
+import { subscribeToBasket } from '../lib/utils';
 
 function makeSimpleActionCreator(type) {
   return (payload) => ({ type, payload });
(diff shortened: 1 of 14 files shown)


fzzzy:
Looks really good.

PR #2412, opened 2017-05-04
Remove the en and de-DE locales

### LINKED ISSUE #2017: If browser language is set to German/Germany [de-de], Cliqz experiment is available but the Test Pilot pages are displayed in English

stasm:
The en and de-DE locales were added in #2149 in order to fix #2017.  The issue was related to the language negotiation done by l20n. It didn't pick en-US nor de for some users.

In #2399 l20n was upgraded to v5.  It uses [fluent-langneg][1] under the hood to conduct the language negotiation.  It now properly selects the best available locale when the user's browser is configured to request de-DE or en.

[1]: https://www.npmjs.com/package/fluent-langneg

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,8 +22,6 @@ media/
 node_modules/
 static/
 tmp/
-locales/en/
-locales/de-DE/
 
 # Files to ignore within specific directories
 addon/*.rdf
--- a/frontend/config.js
+++ b/frontend/config.js
@@ -8,10 +8,9 @@ module.exports = {
   ENABLE_DEV_CONTENT: (process.env.ENABLE_DEV_CONTENT === '1'),
   AVAILABLE_LOCALES: (process.env.ENABLE_DEV_LOCALES === '1') ?
     // All locales on Pontoon for local & dev
-    // en is copied from en-US at build time, so we have to add it special
-    'en,' + fs.readdirSync('./locales').join(',')  :
+    fs.readdirSync('./locales').join(',')  :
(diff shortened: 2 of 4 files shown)


flodolo (approved):
It would be great to have another pair of eyes, but tested locally and it work as expected.

stasm:
@fzzzy - Can you take a look at this? @flodolo realized that the new version of l20n fixes an old bug and this workaround is no longer needed.

fzzzy:
Tested locally, works.

PR #2425, opened 2017-05-10
Remove import-api-content gulp task

lmorchard:
We finished the transition from Django long ago. This gulp task is just
noise, at this point.


fzzzy:
Looks good!

PR #2426, opened 2017-05-11
Remove react-router; add server-side rendering

### LINKED ISSUE #1306: Implement static page rendering with progressive enhancement

### LINKED ISSUE #2260: Remove react-router dependency.
We should instead generate one html file for each route. This will not only be faster, it will avoid the continuous issues with the localization library clashing with react-router partial page refreshes. It should also make the codebase easier to read.

### LINKED ISSUE #2383: URL on site - testpilot.firefox.com/experiments - pinwheels forever
https://testpilot.firefox.com/experiments  

... is one of the first links Google offers up on a search for "firefox testpilot". On my machine it pinwheels forever with an animated loading thing, and never takes me to a usable page.

I'm using release Firefox on Windows 10.

lmorchard:
Remove react-router; add server-side rendering

- Rename gulpfile.js -> gulpfile.babel.js to enable ES6 via Babel

- Remove routes and usage of react-router, switch to server-side
  rendering of pages defined under `frontend/pages` (#2260, #1306)

- Inject data-page-name & data-page-param attributes into `<body>` to
  drive app bootstrap as a replacement for window.location routing.

- Move Markdown pages from `compiled-templates` to `frontend/pages`
  alongside React-base pages

- Merge `generate-static-html.js` into `frontend/tasks/pages.js` now that
  Gulp runs under Babel + ES6

- Merge template for markdown-based pages into `frontend/tasks/pages.js`

- Delete Gulp task `scripts-generate-static-html` since we don't need to
  pre-process ES6 to use it in the build process.

- Remove logic that adapts home page to state of add-on installation in
  order to minimize content flashes on load, split into separate pages /
  and /experiments

- Restore /experiments page for users with add-on installed (#2383)

- Add-on install button switches to link to /experiments page if add-on
  already installed (#2383)

- Switch to normal <a> links, since we don't need <Link> for routing

- Tweaks to adapt Redux state to server-side rendering without browser
  context or add-on installation status

- Remove `history`, `react-router`, & `react-router-redux` dependencies

- Add `react-markdown` dependency to better integrate Markdown content
  in pages gulp task.

- Misc tweaks to clean up awkward module imports & exports

Fixes #2260.
Fixes #2383.
Fixes #1306.


frontend/src/app/index.js
```diff
@@ -2,47 +2,38 @@ import 'babel-polyfill/browser';
...
+  experiment,
+  experiments,
+  home,
+  legacy,
+  onboarding,
+  retire,
+  share
+};
+
+const name = document.body.dataset.pageName;
```
fzzzy, line 33: 👍

fzzzy:
Looks fantastic!

Issue #2434, opened 2017-05-16
We need to bring back loading states while pages complete rendering


johngruen:
There may be some down-the-road workarounds here, but there appear to be several issues related to removing the Loading state as page components get built. To name a few:

* Colors and icons do not seem to be fetched in time making experiment cards not correctly render color or icon
* The card button state swaps subsequent to page load for installed experiments
* The experiment counts getting fetched causes card heights to change
* the Install CTA button flashes
* On experiment details page does not reflect the correct state when loaded; state changes after you see the content.

There are a lot of small cuts here but the sum total is that the site w/o loading components has significant jank associated with toggling state that damages the overall experience of the site. 
Until we can resolve these conflicts without loaders we should bring them back.

fzzzy:
Thanks for such detailed feedback.

lmorchard:
> Colors and icons do not seem to be fetched in time making experiment cards not correctly render color or icon

Peeking at this, and I think this is actually a separate issue. I'm seeing these CSP errors on dev rejecting the background colors & images for experiment cards

<img width="963" alt="screenshot 2017-05-16 15 09 00" src="https://cloud.githubusercontent.com/assets/21687/26123724/ea15ad4a-3a49-11e7-86f5-9f056090e90d.png">

And I see this on dev:

<img width="1050" alt="screenshot 2017-05-16 15 13 14" src="https://cloud.githubusercontent.com/assets/21687/26123869/827a78c2-3a4a-11e7-8061-9879d286c3b5.png">

lmorchard:
Just submitted a PR to reproduce my last comment on local dev, and change CSP to allow inline styles because ugh.

lmorchard:
Ah, looks like that was #2432 - which turns out to be CSP related, not time related, and won't be fixed by a loading delay

PR #2436, opened 2017-05-16
Add 'unsafe-inline' to style-src in CSP to enable server-rendered inline React styles

lmorchard:
- Also inject content-security-policy header in the local dev server

- See also: https://github.com/facebook/react/issues/5878#issuecomment-172964375

Issue #2434


lmorchard:
Not 100% happy to just copypasta the CSP rules between shell & JS files, but couldn't think of an easy way to DRY that out.

fzzzy:
Let's go ahead and merge this now, but try to move the inline styles out into css files and remove unsafe-inline styles before the sprint is over.

elimisteve:
Doesn't adding `unsafe-inline` allow for attacks like the CSS-only keylogger found at https://no-csp-css-keylogger.badsite.io/ ?

In an attempt to help decrease the maximum severity of the problems caused by injections, keeping the CSS-related CSP settings safe is pretty important.

chuckharmston:
CC @april for the above comment.

april:
There is a particular pattern that can cause that issue. Are you setting `value` directly on form fields with each keypress?

elimisteve:
@april If one sets the `value` of a form field `onChange` while CSP allows for `unsafe-inline`, is that dangerous?

april:
Yes, that is exactly the pattern that can be dangerous. I would recommend not doing that, especially for sensitive fields like credit cards or passwords. For those, I recommend letting them be uncontrolled.

Note that you can still do things with them `onSubmit` and you can check and validate their values `onChange`, just don't send the current value back into the field's `value`.

(sorry for not responding for a while, I was out of town)

lmorchard:
FWIW, the only form submission we have on the site is for an email newsletter signup. Not entirely harmless, but also not a credit card or password field

april:
In that case, you can probably just do your normal controlled inputs and not worry about it too much.  :)

PR #2437, opened 2017-05-16
2434 reinstate loading indicator

### LINKED ISSUE #2434: We need to bring back loading states while pages complete rendering

fzzzy:
Fix #2434 Reinstate the loading indicator for now.

  - Remove the inline styles on the experiment cards, by creating experiments.css ahead of time from the yaml files

  - Remove the inline styles from the experiment details pages as well.

  - Remove unsafe-inline from the CSP again.


lmorchard (approved):
LGTM

PR #2438, opened 2017-05-16
Add optional check and notice for max version

### LINKED ISSUE #2424: Add and implement max_release field for experiments
We currently have a [`min_release`](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#min_release) field for experiments to require a minimum Firefox version for installation. We have a new need for a `max_release` field that behaves similarly, but prevents installation for Firefox versions beyond a maximum.

meandavejustice:
- fixes #2424


frontend/src/app/containers/ExperimentPage.js
```diff
@@ -588,11 +599,23 @@ export class ExperimentDetail extends React.Component {
     return null;
   }
 
+  renderMaximumVersionNotice(title, hasAddon, max_release) {
+    if (hasAddon && !this.maxVersionCheck(max_release)) {
+      return (
+        <div className="upgrade-notice">
+          <div data-l10n-id="versionChangeNotice">This experiment doesn't support your FireFox version, try downloading the stable release.</div>
+          <a onClick={e => this.clickUpgradeNotice(e)} data-l10n-id="versionChangeNoticeLink" href="https://www.mozilla.org/en-US/firefox/" target="_blank">Firefox Stable Release</a>
```
pdehaan: Does this need /en-US/ or can we omit that and just let mozilla.org handle locales?

locales/en-US/app.ftl
```diff
@@ -150,6 +150,8 @@ highlightPrivacy = Your privacy
 [[upgradeNotice]]
 upgradeNoticeTitle = {$title} requires Firefox {$min_release} or later.
 upgradeNoticeLink = How to update Firefox.
+versionChangeNotice = Your Firefox version isn't supported, try downloading the stable release.
```
pdehaan: This string isn't the same as the one above, which is:

```html
<div data-l10n-id="versionChangeNotice">This experiment doesn't support your FireFox version, try downloading the stable release.</div>
```

frontend/src/app/containers/ExperimentPage.js
meandavejustice: Oops you're right, I'll remove the `en-US`

meandavejustice:
Gotta get these tests passing and coverage fixed up before merging

locales/en-US/app.ftl
meandavejustice: Nice catch

meandavejustice:
more coverage yay!

PR #2460, opened 2017-05-19
Disable Redux logger middleware

lmorchard:
This thing is noisy on both client-side and during server-side build
process. Since [Redux DevTools][1] is a thing, let's turn off this logger.

[1]: https://github.com/gaearon/redux-devtools


fzzzy:
👍

PR #2466, opened 2017-05-24
simplify loading behavior

### LINKED ISSUE #2465: Clean up loading logic
Playing around on dev and local I noticed a few things about loading

* some form of loading gets displayed in `static-root` and in `page-container`; we could call it once in static root since that gets rendered first no matter what.
* these two elements don't get erased and injected in the same frame causing some odd scrollbar jumping
* the loading container feels like an extra react abstraction; seems like we can kill it, and just use the component

johngruen:
Fixes #2465


frontend/src/app/components/Loading.js
```diff
@@ -1,14 +1,16 @@
 import React from 'react';
 
 const Loading = () => {
-  return <div className="full-page-wrapper centered overflow-hidden">
-    <div className="loading">
-      <div className="loading-bar"></div>
-      <div className="loading-bar"></div>
-      <div className="loading-bar"></div>
-      <div className="loading-bar"></div>
+  return (
```
johngruen, line 4: wrapped return in parens

frontend/src/app/components/ExperimentCardList.js
```diff
@@ -39,7 +30,7 @@ export default class ExperimentCardList extends React.Component {
 
   render() {
     if (this.props.experiments.length === 0) {
-      return this.renderLoading();
+      return null;
```
johngruen, line 33: using this return null pattern throughout this commit in containers that used to return a loading state

frontend/src/pages/error.js
```diff
@@ -1,7 +1,7 @@
 
 import React from 'react';
 
-import inject from '../app/lib/inject'
+import inject from '../app/lib/inject';
```
johngruen, line 4: Meanwhile i noticed that all of these inject imports were missing semi-colons!

frontend/src/styles/modules/_experiment-list.scss
```diff
@@ -186,8 +186,8 @@
 }
 
 .past-experiments {
-  margin: 0 auto;
```
johngruen, line 189: Also, somebody left a non-alphabetized Sass rule. Shame!

johngruen:
Updated Tests

Issue #2423, opened 2017-05-05
Stop using the add-on SDK


ghost:
This is going to be a tracker for moving us away from the SDK.  If you make PRs, use the "connected to" keywords please.

Most of the content here comes from [our API spreadsheet](https://docs.google.com/spreadsheets/d/1r0EekWPF_v427dt0P_gs5MAEWHZDC78d7k1KWZNNfrs/edit#gid=0) and [the documentation comparing the SDK with webextensions](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Comparison_with_the_Add-on_SDK)

We don't actually know if these will easily convert to their webextension counterparts easily so if we run into any issues just bring them up and we'll figure them out.

Please note that this isn't a "make test pilot a web extension" issue -- we're just getting rid of the SDK for now.  One step at a time.

Replace these APIs:
- [ ] self
- [ ] page-mod
- [ ] ui/action/button
- [ ] timers
- [ ] L10n
- [ ] request
- [ ] tabs
- [ ] querystring
- [ ] simple-storage
- [x] notifications  (checking this one off, because deleting this code is already in #2417)
- [ ] util/uuid

lmorchard:
Think I'll officially put myself down for taking a swing at this

lmorchard:
Hmm, running into one wall so far: 

We've been using the SDK's `page-mod` module to inject `window.navigator` properties to inform [the web frontend](https://github.com/mozilla/testpilot/blob/master/addon/data/message-bridge.js) and [a couple of Mozilla sites & `about:home`](https://github.com/mozilla/testpilot/blob/master/addon/data/set-installed-flag.js) about Test Pilot installation state, version, and client UUID.

WebExtension content scripts looked like the easiest port-over approach. But, [WebExtension scripts are disallowed](https://bugzilla.mozilla.org/show_bug.cgi?id=1310082) on [Mozilla hosts that can use `mozAddonManager`. ](https://dxr.mozilla.org/mozilla-central/source/toolkit/mozapps/extensions/AddonManagerWebAPI.cpp#39)

So, I need to figure out how to do the equivalent from bootstrapped add-on code. 😞  Could use some pointers, if anyone knows how to easily do that and save me some research.

chuckharmston:
We could copy/paste [the `page-mod` source](https://dxr.mozilla.org/mozilla-central/source/addon-sdk/source/lib/sdk/content/page-mod.js) as a starting point?

[...]

PR #2475, opened 2017-05-25
Remove Notifications

### LINKED ISSUE #2417: whack all the notification code and docs
We're going to shut down notifications for a while.  Let's get rid of the old code and docs while we still can!

meandavejustice:
- remove all references to notification feature from addon, yaml
files, locales, flow-type declarations, and docs.
- fixes #2417 
- refs #2423


meandavejustice:
added both @dannycoates and @lmorchard for their familiarity with notifications(les wrote it initially I believe) and the addon(danny)

dannycoates (approved):
LGTM 💯

PR #2481, opened 2017-05-26
Fix #2478 Remove usage of dangerouslySetInnerHTML and replace with html-react-parser

### LINKED ISSUE #2478: Remove use of dangerouslySetInnerHTML
In preparation for fluent-react https://github.com/mozilla/testpilot/issues/1714 we need to remove our usage of dangerouslySetInnerHTML. I think we could use https://www.npmjs.com/package/html-react-parser to do this.

Might as well do this work in a different bug than fluent-react, to keep the patch sizes down.

fzzzy:


pdehaan:
I think you can also add the ESLint `react/no-danger` and/or `react/no-danger-with-children` rules in the base .eslintrc file to make sure we don't inject any future `dangerouslySetInnerHTML` instances:

```
rules:
  react/no-danger: error
  react/no-danger-with-children: error
```

meandavejustice:
@fzzzy would you like to add those eslint rules before merging?

fzzzy:
@meandavejustice Sure, I'll do that.

PR #2483, opened 2017-05-27
Get system info from a less expensive Telemetry API

### LINKED ISSUE #2422: Test Pilot can cause huge performance issues in Firefox (especially on Linux) with Telemetry turned on
See this Mozilla bug https://bugzilla.mozilla.org/show_bug.cgi?id=1360702 and this profile: https://perfht.ml/2pHdFkY

Test pilot is calling `getCurrentPingData` for some reason with a very high frequency.  That function is generally very expensive and on Linux it can be *extremely* expensive.  This made my real browsing profile so bad that I had to turn off telemetry for a few days in order to get a usable web browser.

Test pilot should only call `getCurrentPingData` a few times per browser session at maximum.

(Please note that I rarely look at github notifications unfortunately so sadly if you have any questions from me asking on github is probably not a great way to get in touch.  Sorry for this...)

jaredhirsch:
Fixes #2422.

See #2422 and the linked [bugzilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1360702) for background info.

Per [the last comment](https://github.com/mozilla/testpilot/issues/2422#issuecomment-301232382) in 2422, I'm not caching the value of `TelemetryEnvironment.currentEnvironment`, but just re-fetching it with each ping.

I haven't yet figured out how to A/B test performance profiles to verify that this fixes the bug, but the commenters in the bug seemed confident that this change would greatly reduce the amount of work done with each ping submission.

### DIFF
--- a/addon/src/lib/Telemetry.js
+++ b/addon/src/lib/Telemetry.js
@@ -15,6 +15,9 @@ import { storage } from 'sdk/simple-storage';
 import {
   TelemetryController
 } from 'resource://gre/modules/TelemetryController.jsm';
+import {
+  TelemetryEnvironment
+} from 'resource://gre/modules/TelemetryEnvironment.jsm';
 
 import type { ReduxStore } from 'testpilot/types';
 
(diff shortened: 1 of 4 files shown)


meandavejustice:
Failing because of a lint error:

``` 94:5  error  'payload' is defined but never used  no-unused-vars```

jaredhirsch:
closing in favor of #2519

Issue #2477, opened 2017-05-26
If there is a lint error, the integration tests never complete.


fzzzy:
To reproduce, introduce a lint error on a branch (for example, use a javascript string with double quotes " instead of apostrophe ') and then push it to github. You can verify that a lint error occurs locally by using `npm run test:all`.

When circle tries to run the tests, the lint fails, and then the script which waits for the integration tests to complete runs forever.

b4handjr:
Do you have example I could look at?

b4handjr:
Fixed in #2484

PR #2484, opened 2017-05-27
Change bash script to timeout if the server fails to start

b4handjr:
Fix for #2477


b4handjr:
This will be hard to prove but I have tested it locally and it should work.

PR #2485, opened 2017-05-30
Fix for addon bundle under windows

lmorchard:
Turns out that `&&` works in npm scripts under Windows, but `; ` doesn't


meandavejustice:
@lmorchard was there an issue for this one?

lmorchard:
Nope, just kind of a thing I noticed and found faster to fix than to report :)

PR #2493, opened 2017-06-02
Add annotation to indicate CSRF token isnt needed

psiinon:
This is to indicate to the ZAP baseline scan (and any other tools) that a CSRF token is not needed on this form. This is an approach we've used in AMO (https://github.com/mozilla/addons-server/pull/4235/files)
r? @relud


relud:
this looks good to me, based on how the ZAP failure was explained. r? @clouserw about actually merging this.

ghost (approved):

Issue #1781, opened 2016-11-08
Ensure that experiments locale file stays in sync with content.


chuckharmston:
In #1778 we added `experiments.ftl` to the repository, as a short-term step to start the flow of translations. However, the live file is still being generated in the build step, which introduces the possibility of it coming out of sync with the content.

We should consider remediation steps against this.

My suggestion:

1. Stop generating that build file in the build process.
1. Generate it manually with an npm command
1. Fail builds if the output of that npm command doesn't match the file in the repo.

lmorchard:
Actually peeking at this issue a bit as part of the news feed stuff, since we're adding another category of content that needs to be shoehorned into FTL - and we'd like to get string reviews of this stuff.

The suggested solution sounds good. Thinking about doing it as part of #2418 - or as a follow-up shortly thereafter.

PR #2495, opened 2017-06-02
News updates on /experiments

lmorchard:
Opening up a WIP PR for folks to look at news feed stuff so far. Some things are still TODO:

- [x] Split out to Issue #2521 <strike>Update doco to describe how news updates work</strike>
- [x] Finish polishing news item styles (stole [meandavejustice's commit as a start](https://github.com/mozilla/testpilot/commit/d2548d731132be87ec92ad24c41f1ba132a7ae6b))
- [x] Remove garbage dev-only news updates from YAML content, maybe replace with an example or two
- [x] Moved to PR #2508 <strike>Maybe also address #1781 because working through this exposed some issues with FTL string extraction</strike>
- [x] Split out to Issue #2520 <strike>Maybe add an RSS/Atom feed as part of the output 😀 </strike>
- Figure out what else this is missing... (need help with this)


frontend/tasks/content.js
```diff
@@ -98,9 +111,27 @@ function buildExperimentsData() {
...
+
   function endStream(cb) {
+    // One last sweep through updates to exclude dev content where not wanted
+    newsUpdates = newsUpdates
+      .filter(update => config.ENABLE_DEV_CONTENT || !update.dev);
+    extractNewsUpdateStrings();
+
     this.push(new gutil.File({
-      path: 'static/locales/en-US/experiments.ftl',
+      path: '../../locales/en-US/experiments.ftl',
```
lmorchard: This is a not-great hack in lieu of a real solution for #1781 - it spits the FTL outside the build directory and into `locales/en-US/experiments.ftl` because the [`assets-locales` task](https://github.com/mozilla/testpilot/blob/c1f139a2346b08b3b839520f7523cf4f0acee903/frontend/tasks/assets.js#L4) competes with this one and clobbers the fresh strings extracted from news updates and experiments with the old checked-in version. (Which is quite visible when strings are brand new.)

frontend/src/app/components/UpdateList.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import classnames from 'classnames';
+import moment from 'moment';
+import LayoutWrapper from './LayoutWrapper';
+import { newsUpdateL10nId } from '../lib/utils';
+
+export function prettyDate(date) {
+  return moment(date).format('MMMM Do, YYYY');
```
lmorchard: Hmm, I think we're going to have to tweak this format() to be more i18n friendly

meandavejustice: Based on [moment l18n docs](https://momentjs.com/docs/#/i18n/) It looked like all we needed to do was set `moment.locale(String);` which I did [here](https://github.com/mozilla/testpilot/pull/2495/files/c1f139a2346b08b3b839520f7523cf4f0acee903#diff-9e31d7481520a2040f72bb71e10130f1R23), though we'll need to test it out to be sure

lmorchard: Yeah, but I think that only applies to the [generic localeData API calls](https://momentjs.com/docs/#/i18n/locale-data/). This specific `format()` overrides whatever would come from a selected locale

lmorchard: Depending on how picky we want to get here, we could steal some formatting stuff from [this date-formats module in Snooze Tabs](https://github.com/bwinton/SnoozeTabs/blob/master/src/lib/time-formats.js). It had a lot of back & forth with l10n folks to get it just right

lmorchard:
Polished the CSS a bit, moved temporary YAML content into its own easily reverted commit

frontend/src/app/components/UpdateList.js
```diff
@@ -0,0 +1,65 @@
...
+          {newsUpdates.map((update, index) =>
+            <Update key={index} update={update}
+                    experiment={experimentsBySlug[update.experimentSlug]} />)}
+        </LayoutWrapper>
+      </div>
+    );
+  }
+}
+
+UpdateList.propTypes = {
```
fzzzy: This will have to change to use flow to merge cleanly with head. You might want to rebase.

locales/en-US/experiments.ftl
```diff
@@ -35,7 +34,7 @@ cliqzToursteps0Copy = Start typing in the URL bar to see real-time results.
 cliqzToursteps1Copy = Open a new tab and add your favorite websites for one-click access.
 cliqzToursteps2Copy = You can always give us feedback or disable Cliqz from Test Pilot.
 containersDescription = Contain yourselves! Create different containers for each of your online lives - your work self, your social self, your personal self - to stay organized and maintain privacy.
-containersIntroduction = Containers let you create profiles in Firefox for all of your online lives. Custom labels and color-coded tabs help keep different activities — like online shopping, travel planning, or checking work email — separate. Because Containers store cookies separately, you can log into the same site with a different account in each Container, and online trackers can’t connect your browsing in one container to another. So you can keep your shopping self separate from your social self from your work self, without worrying about being followed around the Web.
```
fzzzy: If you change the content of a localized string, aren't you supposed to change the l10n id too, like 'containersIntroduction2? Otherwise I don't think the localizers know they need to re-localize them.

fzzzy (dismissed):
I tested this out locally and read through the code. This looks really great! Just had a few small comments.

Also, about rss/atom, I absolutely think we should have feeds. Maybe it should be done in another pull request after this lands? It doesn't matter either way though if it is too much hassle.

locales/en-US/experiments.ftl
lmorchard: Yeah, this is accidental & should be snipped out of this PR. #2508 handles this more effectively

lmorchard:
On PTO, so will probably make corrections to this PR on monday.

Oh, and [this chevron image](https://github.com/mozilla/testpilot/pull/2495/files#diff-0d6e75148962e9141826d8fb548e3803) should ideally be an SVG if we can get one. I just kind of fudged it by snipping it out of the comp with gimp

johngruen:
Hey @lmorchard this looks good. Think there may be a few little pieces of polish we could add, but it'd probably be easier after merging as a follow up issue/

lmorchard:
Pulled in @johngruen's UI tweaks, reverted unintentional string changes, added flow types, rebased & squashed. Snipped out the example YAML content, but you should be able to get it back for local testing with a `git cherry-pick 597f196517e8c517f9efec5e784d5152defce4a7`

lmorchard:
Also snipped out my hack to FTL generation, since PR #2508 handles that more effectively

lmorchard:
Weird: Getting a flow type failure on CircleCI that I'm not getting locally with `npm run flow`.

fzzzy:
@lmorchard I think that's because the flow circleci checks that the flow coverage is greater than 60%, while just running 'flow' doesn't do that.

lmorchard:
I *was* able to reproduce the flow error - IIRC had v0.41 locally and CircleCI had 0.47 of flow-bin. And for some reason my local copy was ignoring the lack of `flow-typed/moment.js`. I added that and the error went away.

But, still seeing [test failures](https://circleci.com/gh/lmorchard/testpilot/747) and I'm rather baffled what's causing them. Going to try poking with some ssh-enable builds. <strike>Looks like frontend tests are just not running</strike> (hmm, looks like they *are* running, just with the results at the top), and integration test fails reliably with an error about tox not being installed.

lmorchard:
Reproduced locally & fixed the frontend mocha test failure. Still not sure why tox is failing, though

lmorchard:
Did some tweaks to the integration tests that seem totally unrelated to this PR, but looks like they passed. Not sure why they failed here and not elsewhere

PR #2501, opened 2017-06-06
WIP: Added 3 new UI Integration tests

b4handjr:
Adding a few tests at a time. 

The three new tests are:
1. Checking that the Testpilot addon install button loads
2. Checks the number of experiments matches the ENV variable(would like some feedback on this one).
3. Checking to make sure that the page that loads when you click an experiment is correct.

@fzzzy @davehunt @krupa

### DIFF
--- a/bin/circleci/run-integration-tests.sh
+++ b/bin/circleci/run-integration-tests.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 end=$((SECONDS+60))
 
+npm run content
 npm start &
 STATIC_SERVER_PID=$!
 
--- a/frontend/test/ui/pages/desktop/detail.py
+++ b/frontend/test/ui/pages/desktop/detail.py
@@ -1,0 +1,15 @@
+from selenium.webdriver.common.by import By
+
+from base import Base
+
+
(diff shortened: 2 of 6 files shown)


chuckharmston:
> Checks the number of experiments matches the ENV variable(would like some feedback on this one).

You may want to look at `/api/experiments.json`, which will contain an array of all the experiments for any given build. You'll need to compare with the `completed` and `launch_date` for each experiment to see if it should be active, but it's better than an environment variable.

b4handjr:
> You may want to look at /api/experiments.json, which will contain an array of all the experiments for any given build. You'll need to compare with the completed and launch_date for each experiment to see if it should be active, but it's better than an environment variable.

Is that within this repo?

chuckharmston:
> Is that within this repo?

No, it's built with `npm run content`.

b4handjr:
@chuckharmston how does my implementation look?

fzzzy (changes requested):
This looks good to me. Please squash it for merge.

b4handjr:
@fzzzy should be good

fzzzy (approved):

Issue #2219, opened 2017-02-21
Privacy_Preamble not localizing correctly


johngruen:
for some reason the privacy_preamble (See cliqz) is not getting passed in the en_us build experiments.fte file. I tried adding the field name to the utils.js file but the changes didn't get picked up. Right now we're relying on copy pasta to keep everything together.

lmorchard:
Confused by the description of this issue, need some clarification...

> for some reason the privacy_preamble (See cliqz) is not getting passed in the en_us build experiments.fte file

Trying to understand what you're doing & not seeing. 

What command are you running to build build and what experiments.ftl file are you looking at? Did you [copy the build version from frontend/build/static/locales/en-US/experiments.ftl to locales/en-US/experiments.ftl, and include it in a pull request](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization)?

> I tried adding the field name to the utils.js file but the changes didn't get picked up

Do you mean [`frontend/tasks/util.js`](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/util.js)? I don't see `privacy_preamble` in there on master.

I tried adding `privacy_preamble` to the `LOCALIZABLE_FIELDS` list locally and ran `gulp content-experiments-data`. After that, the `cliqzPrivacypreamble` string showed up in `frontend/build/static/locales/en-US/experiments.ftl`. (That `content-experiments-data` task should get run as a part of the usual `npm start` in local dev or `npm static` in the production build.)

> Right now we're relying on copy pasta to keep everything together.

What copy pasta in what files? [I see `cliqzPrivacypreamble` in `locales/en-US/experiments.ftl`](https://github.com/mozilla/testpilot/blame/master/locales/en-US/experiments.ftl#L27). Also see it in the prod locale:
```
➜  testpilot git:(master) ✗ curl -s https://testpilot.firefox.com/static/locales/en-US/experiments.ftl | grep cliqzPrivacypreamble
cliqzPrivacypreamble = The Cliqz Test Pilot experiment is brought to you through collaboration between Cliqz GmbH, and Mozilla. Cliqz is designed to protect your privacy and avoid collecting sensitive information that can be used to create user profiles. However, anonymizing data like this is difficult, and it may still be possible to identify specific individuals.
```
Basically, trying to understand exactly what you did that didn't work - and what exactly the current workaround is.

johngruen:
@lmorchard to confirm this is fixed

lmorchard:
[...]

PR #2508, opened 2017-06-08
Commands to extract strings, check for unextracted strings on CircleCI

lmorchard:
FWIW, this PR will conflict with PR #2495 - and vice versa.

This makes the FTL extraction process for experiment YAML content more explicit. Should run automatically in dev and change `locales/en-US/experiments.ftl` automatically. That, in turn, should make it appear changed when we assemble git commits. In CircleCI, uncommitted string changes should cause test failures.

And, as it turns out, I discovered uncommitted string changes while putting this together. So, those are in a separate commit. We might want someone like @flodolo to review those here (and in future PRs).


lmorchard:
To try this locally, `npm run l10n:check` should report "No missing strings". But, if you roll back one commit with `git checkout HEAD^`, another `npm run l10n:check` should fail and leave `locales/en-US/experiments.ftl` updated.

locales/en-US/experiments.ftl
```diff
@@ -35,7 +34,7 @@ cliqzToursteps0Copy = Start typing in the URL bar to see real-time results.
...
+containersIntroduction = <div class="warning"><strong>Note: Because of the security properties of Containers, Test Pilot experiments such as Page Shot, Snooze Tabs, Pulse, and Activity Stream may not work from within Container tabs. Remember, this is experimental technology!</strong></div> Containers let you create profiles in Firefox for all of your online lives. Custom labels and color-coded tabs help keep different activities — like online shopping, travel planning, or checking work email — separate. Because Containers store cookies separately, you can log into the same site with a different account in each Container, and online trackers can’t connect your browsing in one container to another. So you can keep your shopping self separate from your social self from your work self, without worrying about being followed around the Web.
```
flodolo: The first part of this string should be split into a different one, without the div markup. That would leave containersIntroduction unchanged.

Also, it's not OK to change existing strings without a new ID.

locales/en-US/experiments.ftl
```diff
@@ -45,7 +44,7 @@ containersDetails2Copy = Use color-coded labels to see different Containers at a
 containersToursteps0Copy = Start a Container to label and organize a set of tabs.
 containersToursteps1Copy = Hover over the New Tab button to select a container.
 containersToursteps2Copy = Use the toolbar menu to open Container tabs, create new Containers and organize, edit or delete Containers.
-containersToursteps3Copy = Hide all tabs in a Container or open them in a new window from the toolbar menu, too.
+containersToursteps3Copy = You can hide all open tabs in a Container to keep your browser tidy.
```
flodolo: Need a new string ID

locales/en-US/experiments.ftl
```diff
@@ -58,13 +57,13 @@ containersContributors7Title = Firefox UX
...
+minvidIntroduction = <p>Love multitasking? Try Min Vid. It’s like picture-in-picture where one picture is a media player and the other is the entire Web.</p> <ul> <li><strong>Click and go:</strong> Pop up the Min Vid frame to keep videos playing in the foreground while you browse.</li> <li><strong>Put it where you want it:</strong> Min Vid lets you drag the video frame anywhere on your desktop.</li> <li><strong>Controls where you need them:</strong> Min Vid includes controls to let you pause/play, scrub, drag, minimize, and adjust volume right in the frame.</li> <li><strong>Now supporting audio playback:</strong> Min Vid now supports playback from SoundCloud and direct mp3 links.</li> <li><strong>Create a queue:</strong> Min Vid now supports queues, just start up a video and add tracks from any supported domains (YouTube, Vimeo, SoundCloud, and direct links).</li> <li><strong>View your history:</strong> Min Vid keeps track of your history, just pop open the queue panel from the top right of the player and you can access all the media you've been playing.</li></ul>
```
flodolo, line 62: Looks like the only thing that changed is spacing between markups, so this one should be OK.

locales/en-US/experiments.ftl
```diff
@@ -58,13 +57,13 @@ containersContributors7Title = Firefox UX
...
 minvidMeasurements0 = We collect usage data when you engage with the context menu, experiment icon and player controls.
 minvidMeasurements1 = We also collect data on the number of times you encounter a playable video, the number of times you played the video and the video service that provided the video. This helps us understand how useful our users are finding the experiment.
 minvidMeasurements2 = We do not collect information about the specific videos you encounter.
 minvidDetails0Copy = Access Min Vid from YouTube, Vimeo, and SoundCloud players.
 minvidDetails1Copy = Watch video in the foreground while you do other things on the Web.
-minvidDetails2Copy = Right click on links to video to find Min Vid in the in-context controls.
+minvidDetails2Copy = Right click on links to media to find Min Vid in the in-context controls.
```
flodolo: Need a new string ID.

locales/en-US/experiments.ftl
```diff
@@ -143,7 +142,6 @@ snoozetabsContributors5Title = Firefox QA
 snoozetabsContributors6Title = Softvision QA
 snoozetabsContributors7Title = Softvision QA
 tabcenterDescription = What would it be like to move tabs from the top of the browser to the side? We wanted to find out!
-tabcenterWarning = Firefox is changing fast. Tab Center may have bugs in Firefox 55 and will be disabled completely in Firefox 56+.
```
flodolo: We just added this one…
https://github.com/lmorchard/testpilot/commit/842fe486715b9c0c6b56d3451856d6f3aa64dbac

Issue with the extraction?

locales/en-US/experiments.ftl
```diff
@@ -24,7 +24,6 @@ activitystreamContributors11Title = Software Engineer
 activitystreamContributors12Title = Senior UX Designer
 cliqzDescription = Get search results faster. Cliqz shows suggestions right below the URL bar as you type. All while maintaining your privacy.
 cliqzIntroduction = Get search results faster, directly in the browser. Start typing and see suggestions from around the Web in real time, right below the URL bar. And because Cliqz avoids collecting personally identifiable data, or building user profiles, your results are more private.
-cliqzPrivacypreamble = The Cliqz Test Pilot experiment is brought to you through collaboration between Cliqz GmbH, and Mozilla. Cliqz is designed to protect your privacy and avoid collecting sensitive information that can be used to create user profiles. However, anonymizing data like this is difficult, and it may still be possible to identify specific individuals.
```
flodolo: Isn't this still there? https://github.com/mozilla/testpilot/blob/ffd69e09aeb046d226ddf68ff0c01e00617d402e/content-src/experiments/cliqz.yaml#L28-L33

flodolo (changes requested):

locales/en-US/experiments.ftl
lmorchard: Looks like this one was added manually to the automatically generated FTL - it was never added to the YAML.

locales/en-US/experiments.ftl
lmorchard: Looks like we made string changes in YAML without ever extracting them to FTL (and changing the ID along the way) - just now exposed by making the extraction more explicit

locales/en-US/experiments.ftl
lmorchard: Looks like another manually added string in FTL

content-src/experiments/containers.yaml
```diff
@@ -9,16 +9,17 @@ incompatible:
   'blok@mozilla.org': 'Tracking Protection'
 thumbnail: /static/images/experiments/containers/icon/thumbnail.png
 description: >
-  Contain yourselves! Create different containers for each of your online lives -  
+  Contain yourselves! Create different containers for each of your online lives -
   your work self, your social self, your personal self - to stay organized and maintain privacy.
+warning: >
```
lmorchard, line 14: @johngruen - Need an r+ here, since it looks like [you were the last to touch the `introduction` ](https://github.com/mozilla/testpilot/blame/master/content-src/experiments/containers.yaml#L15). The HTML here is no bueno for L10n

lmorchard:
Okay, did some more work to ensure the extracted FTL syncs up with YAML: 

- added missing fields to the `LOCALIZED_FIELDS` list
- added some l10nsuffix fields where copy changed
- moved some HTML out of the containers `introduction` and into `warning`
- updates docs to highlight the above

Ran another extraction, which seems to have improved things.

lmorchard:
FWIW, this may also fix #2219 because `privacy_preamble` was missing from `LOCALIZED_FIELDS` until now

flodolo (approved):
For the strings part it looks definitely better, thanks.

content-src/experiments/containers.yaml
meandavejustice: confirmed this looks the same locally after the change

PR #2519, opened 2017-06-13
Stop calling getCurrentPingData when sending pings (#2422)

mystor:
I ran into this again due to [bug 1369594](https://bugzilla.mozilla.org/show_bug.cgi?id=1369594), so I figured I'd write a patch to switch the code away from calling getCurrentPingData() to instead directly fetch the environment from TelemetryEnvironment.

I've never used flow before, so I just guessed at what the definition for TelemetryEnvironment should look like.

### DIFF
--- a/addon/src/lib/Telemetry.js
+++ b/addon/src/lib/Telemetry.js
@@ -15,6 +15,9 @@ import { storage } from 'sdk/simple-storage';
 import {
   TelemetryController
 } from 'resource://gre/modules/TelemetryController.jsm';
+import {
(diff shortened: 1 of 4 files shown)


meandavejustice:
cc/ @6a68

lmorchard:
I think this might be a dupe of PR #2483 - looks nearly identical :)

jaredhirsch:
Nearly identical, but it pleases the linter. @mystor owe you a beverage in SF 👍

Issue #2521, opened 2017-06-13
Improve documentation about authoring news updates


lmorchard:
Splitting this out of PR #2495: Added `content-src/news_updates.yaml` and a `news_updates` property to experiment YAML files. But some things could bear explicit documentation:

* The [`content-src/news_updates.yaml` file](https://github.com/lmorchard/testpilot/blob/b8b2d67d1a2f648f9e6584ab7b87921bd3fa786b/content-src/news_updates.yaml) has some comments describing the schema of a news update 
* The [`UpdateList` component determines how they're displayed](https://github.com/mozilla/testpilot/pull/2495/files#diff-678689405eeae88dbc2b281198ec2ddbR21)
* The [`newsUpdateSelector` determines how they're filtered by experiment & `published` date and then sorted](https://github.com/lmorchard/testpilot/blob/b8b2d67d1a2f648f9e6584ab7b87921bd3fa786b/frontend/src/app/selectors/news.js#L7).
* And finally, [`frontend/tasks/content.js`](https://github.com/lmorchard/testpilot/blob/b8b2d67d1a2f648f9e6584ab7b87921bd3fa786b/frontend/tasks/content.js) handles how updates are pulled together and how strings are extracted for L10n.

PR #2524, opened 2017-06-15
News updates feeds

### LINKED ISSUE #2520: Publish an RSS or Atom feed for news update summaries
Splitting this TODO out of PR #2495: Would be nice to publish an RSS or Atom feed of the news updates we show on the site for interested parties to subscribe to. Some notes off the top of my head:

* Should be a gulp task in `frontend/tasks/content.js` - could just be a part of `content-experiments-data` which already assembles all the news updates for `api/news_updates.json`
* Here's a module for generating a feed that I was looking at: https://github.com/jpmonette/feed
* For [feed discovery](https://blog.whatwg.org/feed-autodiscovery), be sure to link the feed into [the `<head>` of pages on the site](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/pages.js#L192) like so: 
  ```
  <link rel="alternate" type="application/atom+xml"
      href="/feed.atom" title="Atom Feed">
  <link rel="alternate" type="application/rss+xml"
      href="/feed.rss" title="RSS Feed">    
  ```

meandavejustice:
fixes #2520

### DIFF
--- a/frontend/tasks/content.js
+++ b/frontend/tasks/content.js
@@ -7,6 +7,7 @@ const mkdirp = require('mkdirp');
 const through = require('through2');
 const gutil = require('gulp-util');
 const YAML = require('yamljs');
+const Feed = require('feed');
 
 const util = require('./util');
 
@@ -31,6 +32,34 @@ gulp.task('content-extract-experiment-strings', () =>
   gulp.src(config.CONTENT_SRC_PATH + 'experiments/*.yaml')
     .pipe(buildExperimentsFTL())
     .pipe(gulp.dest('./locales/en-US')));
+
+function populateRSSFeed(newsUpdates) {
+  return new Feed({
+    title: 'Test Pilot News Updates',
+    description: 'News Updates for Test Pilot experiments',
+    id: 'https://blog.mozilla.org/testpilot',
+    link: 'https://blog.mozilla.org/testpilot',
+    favicon: 'https://testpilot.firefox.com/static/images/favicon.ico',
+    feedLinks: {
+      rss: 'https://testpilot.firefox.com/feed.rss',
+      atom: 'https://testpilot.firefox.com/feed.atom',
+      json: 'https://testpilot.firefox.com/feed.json'
+    },
+    author: {
+      name: 'Mozilla',
+      link: 'https://testpilot.firefox.com'
+    },
+    feed: newsUpdates.map(post => {
+      return {
+        title: post.title,
+        id: post.link,
+        link: post.link,
+        date: post.published,
+        content: post.content
+      }
+    })
+  });
+}
 
 function buildExperimentsFTL() {
   const strings = [];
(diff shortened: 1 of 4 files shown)


meandavejustice:
commit history looks a little weird, gunna get to the bottom of that in the am

frontend/tasks/content.js
```diff
@@ -196,6 +245,7 @@ function extractNewsUpdatesFromExperiment(experiment) {
   return news_updates.map(update => ({
     ...update, experimentSlug: slug
   }));
+  delete experiment.news_updates;
```
lmorchard: This line will never be reached, since the previous line is `return`. I'd had this in this function earlier, but had to [move it into the gulp tasks](https://github.com/mozilla/testpilot/commit/cc33d43006c9b5b53e9e84a65028c89ee89c66a9#diff-52fa3e5c3e785506b7cc22f3df8cc3fbR47) because modifying these parameters in place ended up not working after Babel ran over the code.

frontend/tasks/content.js
```diff
@@ -32,6 +39,34 @@ gulp.task('content-extract-experiment-strings', () =>
...
+      name: 'Mozilla',
+      link: 'https://testpilot.firefox.com'
+    },
+    feed: newsUpdates.map(post => {
+      return {
+        title: post.title,
+        id: post.link,
+        link: post.link,
+        date: post.published,
+        content: marked(post.content)
```
lmorchard: Despite my earlier example content mentioning Markdown - the news update component doesn't actually render the content using Markdown. [It's just text.](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/UpdateList.js#L39) If we do this in the feeds here, we should add it to the `Update` component [using the `ReactMarkdown` component we use in the `pages` tasks](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/pages.js#L256)

package.json
```diff
@@ -80,6 +81,7 @@
     "gulp-uglify": "3.0.0",
     "gulp-util": "3.0.8",
     "jsdom": "11.0.0",
+    "marked": "0.3.6",
```
lmorchard: Also, the `ReactMarkdown` component [uses the `commonmark` module](https://github.com/rexxars/react-markdown/blob/master/package.json#L57) to render Markdown - might be a better idea to use that instead as a top-level dependency so we only have one Markdown implementation in the project

lmorchard (changes requested):
A few tweaks I'd like to see, but looks good overall!

pdehaan:
Semi-related, but do any experiments have any live news feed items currently? Or do we need to build locally to test news feeds? I tried trolling through a bunch of experiments, but didn't see anything news feedy in the YAMLs.

lmorchard:
> Semi-related, but do any experiments have any live news feed items currently?

No, this feature's brand-new and there's no content yet.

But, it looks like [this orphaned commit with example content](https://github.com/lmorchard/testpilot/commit/597f196517e8c517f9efec5e784d5152defce4a7) still exists in my repo, so that could be handy

lmorchard:
Also have issue #2521 to write better doco about this stuff

frontend/tasks/content.js
meandavejustice: whoops, think this got messed up in my rebase

lmorchard (approved):

lmorchard:
Looks like you pulled out the markdown stuff altogether, which is cool... we can add that in later pretty easily

PR #2526, opened 2017-06-15
Do not re-build static site for every integration test run

### LINKED ISSUE #2525: Build site only once to speed up integration tests
With PR #2518, we now test the site against Firefox release, dev-edition, and nightly. But, for each of those test runs, we rebuild the site's static content every time with `npm start`. 

I *think* we could speed this up by just doing `npm run server` instead. Assuming it re-uses the site built earlier in [the `build-frontend` step](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-frontend.sh), this might be an easy win. But, it could use some investigation

lmorchard:
- Build the site earlier, reuse the result in tests

- Add `set -ex` to test scripts to echo commands & fail fast

Fixes #2525

### DIFF
--- a/bin/circleci/run-integration-tests.sh
+++ b/bin/circleci/run-integration-tests.sh
@@ -1,8 +1,7 @@
 #!/bin/bash
 end=$((SECONDS+60))
 
-npm run content
-npm start &
+npm run server &
 STATIC_SERVER_PID=$!
 # Wait until the server is available...
 while [ $SECONDS != $end ]; do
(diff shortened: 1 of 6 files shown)


lmorchard:
This doesn't speed things up *nearly* as fast as I'd hoped. Sigh.

b4handjr:
We could start the server once and just install the firefox versions after each test

b4handjr:
Or only test all 3 versions on a deployment

lmorchard:
The server itself doesn't really take much time to start up. And I hate to exclude tests until deployment, because that's the worst time to get surprises. (i.e. when we're trying to hustle something out the door) Probably good to just consider this PR an incremental improvement as-is, and we can keep poking at it.

b4handjr:
Makes sense. R+

b4handjr (approved):

Issue #2536, opened 2017-06-20
Improve integration test suite speed.


chuckharmston:
Our build times have doubled in the last week:

- https://circleci.com/gh/mozilla/testpilot/4473
- https://circleci.com/gh/mozilla/testpilot/4503

And we even have builds that are occasionally timing out: https://circleci.com/gh/mozilla/testpilot/4539

This appears to be due to the addition of integration tests that test across three versions of Firefox. While good to have, I don't think that this is acceptable; I'd be in favor of backing these tests out until we can get builds to a more reasonable time.

b4handjr:
So for the most part they aren't timing out but it seems every now and then the browser does not load. My current config adds 2 more workers to a total of 4. And helps increase the build time just a hair. 

I have been looking into circle's parallel builds but I have not really done much with them as of yet. I think the only quickest way of reducing times would be to disable all 3 versions of firefox from being tested in serial, and maybe figure out a way to run them in parallel.

lmorchard:
Copying over some further scrollback from IRC:
```
<fzzzy> maybe we should just test nightly and release?
<lorchard> Well, that only saves about a minute & a half
<fzzzy> ah bummer
<lorchard> Not nothing, but they each take about 90 seconds between downloading the browser binary and firing everything up
<chuck> We should cache the binaries if we aren't doing that already, that will help a lot.
<lorchard> Probably some things we could do with caching the binaries, not sure about the rest
<chuck> Might make sense to run them as a separate nonblocking job.
<lorchard> we switched to mozdownload, but that doesn't seem to have caching out of the box
<chuck> Circle should be able to cache them.
<lorchard> But we have to know when to download new versions & cache invalidate
<lorchard> not simple file caching
<lorchard> I don't imagine it's impossible, just needs to be investigated
<lorchard> Can probably also run the tests in parallel, but not sure how off the top of my head
<lorchard> Could also just run them on master branch and speed up PR checks, though that kind of defeats the purpose
<fzzzy> lots of builds are timing out
<lorchard> Reticent to just back out the testing, because we've had a bad history of just turning off and never getting back around to dealing with integration test
<lorchard> But maybe we need to? meh
<_6a68> iafter working on Gecko for a while, where running automated tests takes like half a day, 8 minutes seems like, ¯\_(ツ)_/¯
<_6a68> *after
<lorchard> We should also run tests a lot less often if/when we move pontoon commits to an l10n branch
[...]

PR #2541, opened 2017-06-20
Temporarily remove integration tests until stabilized/sped up (connected to #2536).

chuckharmston:
I just had [another build](https://circleci.com/gh/mozilla/testpilot/4543) take 20 minutes before timing out. Let's pull back these tests until they're quicker and more stable.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -37,9 +37,6 @@ test:
   override:
     - ./bin/circleci/test-addon.sh
     - ./bin/circleci/test-frontend.sh
-    - sh ./bin/circleci/test-firefox-release.sh
-    - sh ./bin/circleci/test-firefox-dev.sh
-    - sh ./bin/circleci/test-firefox-nightly.sh
   post:
     - bash <(curl -s https://codecov.io/bash)
 


circle.yml
```diff
@@ -37,9 +37,6 @@ test:
   override:
     - ./bin/circleci/test-addon.sh
     - ./bin/circleci/test-frontend.sh
-    - sh ./bin/circleci/test-firefox-release.sh
```
lmorchard, line 40: We have a history of turning integration tests off and forgetting about them. Should at least leave the tests for Firefox release enabled and plan to improve things to the point where we can turn the other two back on.

lmorchard (changes requested):

circle.yml
chuckharmston: I'd agree with you if build time was the only concern, but these are causing builds to intermittently fail, and idempotency is an important quality of a good test suite.

b4handjr:
IDK if you saw my comment on the issue but I have parallel tests working. I just need to figure out the impact on the entire mozilla organizations use of containers and such.

https://circleci.com/gh/jrbenny35/testpilot/200#tests/containers/0

lmorchard:
Alright, let's do this: I just started a branch [restore-integration-tests](https://github.com/mozilla/testpilot/tree/restore-integration-tests). Going to merge this PR to try to resolve the timeouts we're seeing, but then start a PR immediately from that branch so we're nagged to fix this. And CircleCI tests will run from that branch as we hopefully throw work at it to improve things per issue #2536

Issue #2542, opened 2017-06-20
Skip tests in Circle CI for commits from Pontoon on master branch?


lmorchard:
Hacky idea in advance of moving l10n to a branch in #2430: 

What if we looked for `/^Pontoon: Update/` in the commit message on CircleCI and use that as a signal to just skip tests altogether? Chances are low that any particular Pontoon commit will break the site, and such commits are very frequent because our localizer community is prolific. So, we get a lot of mostly useless test runs.

lmorchard:
<strike>Alternatively, if Pontoon supported it, we could add `[skip ci]` to commit titles or descriptions - [because CircleCI supports this as a conventional signal to skip test runs](https://circleci.com/docs/1.0/skip-a-build/).</strike> Pontoon does not support this, and it sounds like it would take significant work to support custom commit templates per-project.

ghost:
Let's focus on the L10n branch

PR #2543, opened 2017-06-21
Restore integration tests in CircleCI

lmorchard:
This PR is on a branch of the main Test Pilot repo, so it should be able to accept commits & PRs from a bunch of us. Let's see if we can work here and improve things for issue #2541

### DIFF
--- a/bin/circleci/install-test-dependencies.sh
+++ b/bin/circleci/install-test-dependencies.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -ex
+
+GECKODRIVER_URL=$(
+  curl -s 'https://api.github.com/repos/mozilla/geckodriver/releases/latest' |
+  python -c "import sys, json; r = json.load(sys.stdin); print([a for a in r['assets'] if 'linux64' in a['name']][0]['browser_download_url']);"
+);
+
+curl -L -o geckodriver.tar.gz $GECKODRIVER_URL
+gunzip -c geckodriver.tar.gz | tar xopf -
+chmod +x geckodriver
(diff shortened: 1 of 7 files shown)


b4handjr:
So what is going to be judge of when we can turn these back on? Looking through all the builds, out of 100 builds there was a timeout error twice, and ENOSPC errors like 10 times or so.

lmorchard:
> So what is going to be judge of when we can turn these back on?

That's a good question. We've got a status meeting in Test Pilot Vidyo in about 20 minutes, where I'll bring this up. I'm hoping we can re-enable for at least release Firefox, and maybe throw in #2542 to skip tests for Pontoon commits so we hit CircleCI less often overall. There are more improvements for #2536 but I don't want to have us leave them disabled for long

b4handjr:
Speaking with another QA dev, it seems that running the tests in parallel can sometimes cause a race condition between geckodriver/firefox starting and the test harness. I have a PR almost ready to remove parallel testing and to enable a html report to always be collected.

lmorchard:
Tweaking this PR to re-enable for release Firefox. Let's see how long the test run takes.

b4handjr:
Before we merge I want to add some extra things. I can just push to your branch correct?

lmorchard:
> Before we merge I want to add some extra things. I can just push to your branch correct?

I think so... if not, say because you don't have permission to push to test pilot branches, I think you can submit a pull request to this branch instead of master. Gets tangly & confusing, but should be okay

b4handjr:
That failure is because the github api only allows for 50 requests an hour unless you config it to use oauth or something like that.

PR #2544, opened 2017-06-21
Do not copy ui-tests.html to CIRCLE_ARTIFACTS if it has not been generated

lmorchard:
Looks like this little bit breaks master branch tests after integration tests were disabled


lmorchard:
(self-merged to get tests fixed asap)

PR #2548, opened 2017-06-23
[TXP-27] Add feedback buttons to experiment cards

lmorchard:
Think I have a good first stab at adding the feedback button, including metrics and l10n and all that fun stuff. Some remaining questions:

- [x] Needs a hover state? (/cc @johngruen)

Also added a React Storybook install here. Not perfect, but you should be able to see what it's about like so:

* Start a dev server like usual with `npm start`
* In a second shell, start Storybook with `npm run storybook`
* Visit http://localhost:6006/ to hopefully see this:

capture

Issue #2546, opened 2017-06-22
TXP-48: Test Pilot Web Experiments


ChenMorpheus:
@johngruen 
In order to accommodate web experiments on Test Pilot, I created the proposal and took some future possibilities into consideration as well to ensure the flexibility, such as mobile experiments. However, for the current implementation, please feel free to implement only what we need for now. Besides, for web experiments, perhaps we can display how many users visit the web experiment instead of removing the user counts section. What do you think? 

![web experiments](https://user-images.githubusercontent.com/18019946/27419581-525ad138-5754-11e7-8c1c-92710308dbd0.png)

Please find assets here.
web experiment: https://github.com/mozilla/testpilot-assets/blob/master/Test_Pilot_Website/00_Assets/webexperiment.svg
add-on: http://design.firefox.com/icons/icons/addons/extensions.svg
mobile: http://design.firefox.com/icons/icons/new/device-mobile-16.svg

lmorchard:
Think I can start taking a look at this, might be a good thing to set up along with a React Storybook demo

lmorchard:
(meant to post this on #2462 but can come back to this one too)

ghost:
Les accidently clicked the take button!

lmorchard:
Hmm, well I have a PR for #2462, so maybe I *will* start on this. Sorry for the confusion

lmorchard:
One gotcha: Getting a visit count for a web experiment will be a small project itself. We have a JSON resource we load from Telemetry for add-on install counts. We'd need something similar for experiment sites - e.g. from Google Analytics or similar.

Otherwise, I should be able to build the rest of this on our side

lmorchard:
Also, it looks like platform isn't a single type - i.e. it's any combination of web, mobile, and add-on? Some questions / thinking out loud:

- Is it really any combination? If so, we'll need 6 strings for localization to cover the permutations: web; add-on; mobile; web + mobile; mobile + add-on; web + mobile + add-on.
- If there will be a web + add-on experiment, what do we do with the "Enable" / "Go to" button on the experiment detail page? Show both enable add-on and go-to site? Similar concern for displaying install count vs visit count

lmorchard:
Oh yeah, this also occurred to me just now: I see the concept of "enabled" for mobile & web experiments. That could be really tricky to track & implement. We only know if an add-on has been installed, not whether the user has visited a site or installed a mobile app

SoftVision-PaulOiegas:
[...]

PR #2549, opened 2017-06-23
Enable React Storybook

lmorchard:
Since I've started on two feature branches (#2548, #2546) based on these Storybook commits for the ExperimentRowCard component, maybe it would be interesting to consider this separately. Same instructions to see a base story for the component:

* Start a dev server like usual with `npm start`
* In a second shell, start Storybook with `npm run storybook`
* Visit http://localhost:6006/

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,0 +1,4 @@
+/* eslint-disable import/no-extraneous-dependencies, import/no-unresolved, import/extensions */
+
+import '@storybook/addon-actions/register';
+import '@storybook/addon-links/register';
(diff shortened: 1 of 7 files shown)


chuckharmston:
Could we also have a [Storybook dev site](https://github.com/storybooks/storybook-deployer)?

lmorchard:
> Could we also have a Storybook dev site?

Yes, eventually that would be great. Needs more work to resolve some issues I hit with trying to get the sass to build under webpack. Also depends on L20n assets for locales, so we'd need to tweak the build to include those. Punted on all of that for now by just pointing the storybook server at `frontend/build` for static files.

Nothing insurmountable, but wanted to get the ball rolling in time for SF

PR #2553, opened 2017-06-24
update to photon loading

### LINKED ISSUE #2552: update to photon loader
Eric pang made us a sick photon loader...will drop it in

johngruen:
Fixes #2552


frontend/src/app/components/Header.js
```diff
@@ -38,6 +38,10 @@ export default class Header extends React.Component {
     return this.props.hasAddon;
   }
 
+  setHeaderLinkPath() {
```
johngruen, line 41: because of the design of the new loader, i noticed a little double-tap when the user clicks on the TXP wordmark if the add-on is installed. This happens b/c the site has to step through a redirect which requires the react state to load before it starts. Adding this method to filter the URL pre-click should make the experience snappier

frontend/test/app/components/Header-test.js
```diff
@@ -35,6 +35,9 @@ describe('app/components/Header', () => {
     it('should not show the settings button', () => {
       expect(subject.find('.settings-button')).to.have.property('length', 0);
     });
+    it('should link to /', () => {
```
johngruen, line 38: Added tests for routing in the header component

PR #2558, opened 2017-06-27
graduate some experiments next week

clouserw:


pdehaan:
Now that we have newsfeeds (ref #2495), should we be doing some feed items about graduation?

pdehaan:
Also, these seem to be expiring next Wed (July 5). Is that enough time to write and test _four_ graduation reports for the site? Considering July 3rd (Mon) is a belated Canadian holiday and July 4th (Tue) is a US holiday?

ghost:
@pdehaan yeah, reports are already in the works.  We've been planning the 28th push / 5th graduate stuff for a while.  Not to say it isn't a tight squeeze though.

The newsfeeds bit is a good point though.  /cc @johngruen

jaredhirsch (approved):
🤠  yee haw

PR #2559, opened 2017-06-28
WIP 1714 fluent react

fzzzy:


frontend/src/app/components/EmailDialog.js
```diff
@@ -49,11 +50,15 @@ export default class EmailDialog extends React.Component {
...
-          <h3 className="modal-header" data-l10n-id="emailOptInDialogTitle">Welcome to Test Pilot!</h3>
+          <Localized id="emailOptInDialogTitle">
+            <h3 className="modal-header">Welcome to Test Pilot!</h3>
+          </Localized>
           <div className="modal-cancel" onClick={e => this.skip(e)}/>
         </header>
         <div className="modal-content modal-form centered">
-          <p data-l10n-id="emailOptInMessage" className="">Find out about new experiments and see test results for experiments you&apos;ve tried.</p>
+          <Localized id="emailOptInMessage">
+            <p className="">Find out about new experiments and see test results for experiments you&apos;ve tried.</p>
```
pdehaan: Does this need the empty `className` attribute?

frontend/src/app/components/EmailDialog.js
```diff
@@ -68,15 +73,21 @@ export default class EmailDialog extends React.Component {
...
           <div className="envelope" />
-          <p data-l10n-id="emailOptInSuccessMessage2">Thank you!</p>
+          <Localized id="emailOptInSuccessMessage2">
+            <p>Thank you!</p>
+          </Localized>
         </div>
         <div className="modal-actions">
-          <button id="email-success-continue" onClick={e => this.continue(e)} className="button default large" data-l10n-id="emailOptInConfirmationClose">On to the experiments...</button>
+          <Localized id="emailOptInConfirmationClose">
+            <button id="email-success-continue" onClick={e => this.continue(e)} className="button default large">On to the experiments...</button>
```
pdehaan: I think @flodolo generally recommends the `&hellip;` "&hellip;" over the 3 separate dots "...".

frontend/src/app/components/MainInstallButton.js
```diff
@@ -88,16 +102,22 @@ export default class MainInstallButton extends React.Component {
...
       <div>
           {!isFirefox ? (
-              <span data-l10n-id="landingDownloadFirefoxDesc" className="parens">(Test Pilot is available for Firefox on Windows, OS X and Linux)</span>
+              <Localized id="landingDownloadFirefoxDesc">
+                <span className="parens">(Test Pilot is available for Firefox on Windows, OS X and Linux)</span>
+              </Localized>
             ) : (
-              <span className="parens" data-l10n-id="landingUpgradeDesc2" data-l10n-args={JSON.stringify({ version: config.minFirefoxVersion })}>Test Pilot requires Firefox { config.minFirefoxVersion } or higher.</span>
+              <Localized id="landingUpgradeDesc2" $version={config.minFirefoxVersion}>
+                <span className="parens">Test Pilot requires Firefox { config.minFirefoxVersion } or higher.</span>
```
stasm, line 125: There are two ways to approach what should go between into the `<span>` here and I'd like to learn your take on which one is more desirable.

The first one is what you did here: put the valid JSX into the `<span>`.  It can serve as a fallback for when the `landingUpgradeDesc2` translation is missing. The drawback is that React will create the virtual DOM nodes for the content which `Localized` will then remove in favor of the translated string.

The other approach is to use valid FTL syntax as `span`'s content:

```
<Localized id="landingUpgradeDesc2" $version={config.minFirefoxVersion}>
  <span className="parens">
    {'Test Pilot requires Firefox { $version } or higher.'}
  </span>
</Localized>
```

This has the benefit of being exactly the string that needs to be translated which in the future could be used for extracting English strings into an FTL file. On the other hand, it doesn't work as well as the first approach when it comes to fallback.  it's just a static string and in particular, `$version` wouldn't get replaced if `landingUpgradeDesc2` is missing from all languages (English included).

frontend/src/app/components/NewsletterForm.js
```diff
@@ -75,36 +76,44 @@ export default class NewsletterForm extends React.Component {
...
-            this Privacy Notice
-          </a>.
-        </span>
+        { this.state.privacyNote ? <Localized id="newsletterFormPrivacyAgreementRequired">
+          <span style={{ color: 'red', marginRight: '0.5em' }}></span>
+        </Localized> : null }
+        <Localized id="newsletterFormPrivacyNotice">
+          <span>
+            I&apos;m okay with Mozilla handling by info as explained in
+            <a target="_blank" rel="noopener noreferrer"
```
stasm: It's currently not possible to have nested elements inside of `Localized` components. Instead you may try passing the `<a>` element as a `$`-prefixed argument to `Localized`. See the bottom of https://github.com/projectfluent/fluent.js/wiki/Localized for an example of this approach.

Adding support for nested elements is definitely planned, although I haven't made much research into it and don't know how hard it's going to be.

frontend/src/app/containers/App.js
```diff
@@ -100,14 +104,51 @@ class App extends Component {
...
+    function addLang(lang, response) {
+      if (response.ok) {
+        response.text().then(data => {
+          langs[lang] = `${langs[lang] || ''}${data}
+`;
+        });
+      }
+    }
+
+    Promise.all(navigator.languages.map(language =>
```
stasm: `navigator.languages` might include languages that Test Pilot doesn't support. You can use [fluent-langneg](https://github.com/projectfluent/fluent.js/tree/master/fluent-langneg) to make sure you only fetch files that are known to exist.

frontend/src/app/containers/App.js
```diff
@@ -100,14 +104,51 @@ class App extends Component {
...
+      Promise.all(
+        [
+          fetch(`/static/locales/${language}/app.ftl`).then(response => addLang(language, response)),
+          fetch(`/static/locales/${language}/experiments.ftl`).then(response => addLang(language, response))
+        ]
+      )
+    )).then(() => this.props.setLocalizations(langs));
```
stasm, line 133: You could possibly store the entire `messages` iterator in the state, instead of raw localization files, which would make the code in `render` a bit simpler.  You also wouldn't need the `addLang` function; you can use `cx.addMessages(data);` multiple times on a single `MessageContext` instance.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -146,12 +147,16 @@ export class ExperimentDetail extends React.Component {
...
-            We recommend <a href={helpUrl}>disabling these add-ons</a> before activating this experiment:
-          </p>
+          <Localized id="incompatibleHeader">
+            <h3>
+              This experiment may not be compatible with add-ons you have installed.
+            </h3>
+          </Localized>
+          <Localized id="incompatibleSubheader">
+            <p>
+              We recommend <a href={helpUrl}>disabling these add-ons</a> before activating this experiment:
```
stasm: Similar to my comment above, the workaround to `fluent-react` not supporting nested elements is to pass `<a>` as an argument to `Localized`. See the bottom of https://github.com/projectfluent/fluent.js/wiki/Localized.

frontend/src/app/containers/ExperimentPage.js
```diff
@@ -307,13 +319,19 @@ export class ExperimentDetail extends React.Component {
             })
             }>
             <div className={classnames('status-bar', statusType)}>
-              {(statusType === 'enabled') && <span data-l10n-id="isEnabledStatusMessage" data-l10n-args={JSON.stringify({ title })}><span></span></span>}
-              {(statusType === 'error') && <span data-l10n-id="installErrorMessage" data-l10n-args={JSON.stringify({ title })}><span></span></span>}
+              {(statusType === 'enabled') && <Localized id="isEnabledStatusMessage" $title={ title }>
+                <span><span></span></span>
```
stasm: What is the reason for having two `spans` here? You might need to put the outer one outside of the `statusType` check.

frontend/src/app/reducers/localizations.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+function setLocalizations(
+  state: Localizations,
+  action: SetLocalizationsAction
+): Localizations {
+  return action.payload;
+}
+
+export default function experimentsReducer(
```
stasm: Rename this to `localizationsReducer`?

stasm (commented):
Whoa, that's one huge PR.  It looks great too. Let me know if you have any questions about `fluent-react`.

frontend/src/app/reducers/localizations.js
fzzzy: Good catch :)

frontend/src/app/containers/ExperimentPage.js
fzzzy: I'm not sure, probably something having to do with CSS. I'll ask John about it.

frontend/src/app/containers/ExperimentPage.js
fzzzy: Great, thanks a lot. However what if the contents of the anchor needs to be localized too?

frontend/src/app/components/MainInstallButton.js
fzzzy: I don't have much of an opinion on this. My gut feeling is to go with the valid jsx. Even though the nodes will just be tossed, it doesn't seem like that big of an issue.

frontend/src/app/containers/ExperimentPage.js
stasm: You can pass `<Localized>` elements as arguments, too :) It's not ideal, but it's also a common practice in other React localization libraries from what I've seen.

frontend/src/app/containers/App.js
fzzzy: Awesome! Thanks!

frontend/src/app/containers/ExperimentPage.js
fzzzy: Ok, I wondered if that would work. 👍

fzzzy: I'm making some progress removing the html from localized strings. The described technique works fine, and having nested Localized components works fine too.

frontend/src/app/containers/App.js
fzzzy: I've decided to do this in another PR since the current behavior works and this PR is getting huge. I opened a bug for it: https://github.com/mozilla/testpilot/issues/2609

fzzzy:
This pr has been replaced by a new, rebased pr:

https://github.com/mozilla/testpilot/pull/2611

PR #2567, opened 2017-06-29
Deploy storybook builds to S3 bucket website

lmorchard:
Alright. So, this is kind of wonky, but it seems to work for now. The `deploy-storybook.sh` script does a little bit of hacky magic to get a snapshot of the `/static/` directory from a Test Pilot site build and pushes it up to a subdirectory of an S3 website. If all goes well with this PR's first commit, a Storybook build should appear here:

http://storybook.testpilot.dev.mozaws.net/b6a09bd227d601fb59648e62f6d66b6a2099406b/

The URL is per-commit, so this will change if the PR changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ build/
 dist/
 docs/_build
 frontend/build/
+frontend/storybook/
 media/
 node_modules/
 static/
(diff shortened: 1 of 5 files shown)


lmorchard:
One more tweak - added a bit to publish per-PR redirect pages. So, we can do this:

http://storybook.testpilot.dev.mozaws.net/pull/2567/

And that should redirect to the current commit for this PR

PR #2581, opened 2017-07-03
Ensure page content even without grad report

### LINKED ISSUE #2580: If there's no Grad Report at graduation, render normal content with a warning
Doesn't look like we're going to have grad reports quite ready for wednesday. Right now the whole left column of the page is blank. We should at least show the intro text with a notice that the grad report is coming soon.

johngruen:
Fixes #2580

### DIFF
--- a/frontend/src/app/containers/ExperimentPage.js
+++ b/frontend/src/app/containers/ExperimentPage.js
@@ -470,7 +470,36 @@ export class ExperimentDetail extends React.Component {
                 {graduated &&
                   <div className="details-description">
                     <section className="graduation-report">
-                      {parser(graduation_report || '')}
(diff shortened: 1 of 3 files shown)


johngruen:
#2580

ghost:
I don't think this is critical since we haven't pushed any notices out yet.  I think at this point we can just push the graduation reports and the notices and stuff all at once.  Our graudation alerts aren't very effective as-is anyway.

/cc @flodolo for new l10n

fzzzy:
@flodolo There are two new strings here.

flodolo:
> @flodolo There are two new strings here.

Can't spot anything wrong with them, thanks.

PR #2589, opened 2017-07-05
Add back python 3.6 for html reporting

b4handjr:
This was removed due to some circleci wonkiness but it is needed to correctly create an html report.

I will run this a few times to make sure it passes.


b4handjr:
Closing this.

PR #2615, opened 2017-07-14
update hosting documentation for Mozilla Extensions

ghost:
r? @relud 

These are the changes @milescrabill and I came up with for the new pipeline

### DIFF
--- a/docs/development/hosting.md
+++ b/docs/development/hosting.md
@@ -5,36 +5,17 @@
 Test Pilot experiments are hosted on testpilot.firefox.com which lets them get
 installed in Firefox with a minimal amount of user prompting and warning
 messages.  The process of getting add-ons on testpilot.firefox.com is
-documented below -- basically a series of shell scripts glued together with
-environment variables.
+documented below.
 
 ### tl;dr The Experiment Requirements
 
 Your repository must use a branch called `production` to hold the code you want
 to deploy.
 
-You must give the *Mozilla Test Pilot* user ownership access to your add-on on
-AMO (so it can sign new copies).
+Your repository must include a `npm run package` script which, when run, will
+build the add-on and leave it in a file named `addon.xpi` in the root
+directory.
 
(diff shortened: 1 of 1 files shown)


docs/development/hosting.md
```diff
@@ -46,11 +27,9 @@ The full publishing flow is as follows:
    does any internal checks.  These checks may include branch checks,
    verification that an authorized user made the commit, checks for passing
    tests, or other checks.
-0. Jenkins will set the `TESTPILOT_AMO_USER` and `TESTPILOT_AMO_SECRET`
-   environment variables
-0. Jenkins will run `bin/build-addon.sh`
-0. Jenkins will prepare appropriate S3 files (the signed .xpi file, /latest/
-   redirect, a new updates.json file)
+0. Jenkins will run `npm install` followed by `npm run package`
```
lmorchard, line 30: FWIW, I'd be a *little* hesitant to require an npm script versus a plain shell script. Mozilla Extensions (nee bootstrapped extensions) don't really need node & npm & package.json to produce an .xpi - just zip. 

Maybe practically speaking this is fine, though.  While rewriting the Test Pilot add-on for #2423, I ended up backing into a webpack build anyway - mainly because JavaScript Module imports are clunky IMO

ghost: I was a little worried about that too.  I don't have a strong opinion on it.  Do you think it's worth wrapping this in a shell script?

lmorchard: <strike>Yeah, I'd rather the "API" here be a shell script than an npm script, just for flexibility's sake</strike>

On second thought, never mind: npm is fine :) I don't feel that strongly about it, and personally am using node & npm on the Test Pilot add-on. Mainly playing devil's advocate for a dev case that I don't actually need

relud (approved):

docs/development/hosting.md
ghost: thx! 🌵

PR #2623, opened 2017-07-18
Add comment with more description of the new experiment platform labels

lmorchard:
Adding some comments to app.ftl, let's see if this pings anyone via CODEOWNERS


lmorchard:
Seems not to have worked :/

PR #2624, opened 2017-07-18
Add comment with more description of the new experiment platform labels

lmorchard:
Trying this again to trigger a review auto-request

### DIFF
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -150,14 +150,17 @@ userCountContainerAlt = Just launched!
 highlightPrivacy = Your privacy
 experimentGradReportPendingTitle = This experiment has ended
 experimentGradReportPendingCopy = We are working on a full report. Check back soon for the details.
+experimentGoToLink = Go to { $title }
+
+// Label shown next to a series of icons indicating whether an experiment is available as an add-on, mobile app, and/or web site
+[[experimentPlatform]]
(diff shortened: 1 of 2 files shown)


lmorchard:
Ah hah! Looks like it worked this time

locales/en-US/app.ftl
```diff
@@ -150,14 +150,17 @@ userCountContainerAlt = Just launched!
...
+[[experimentPlatform]]
 experimentPlatformWebAddonMobile = Firefox / web / mobile experiment
 experimentPlatformWebAddon = Firefox / web experiment
 experimentPlatformWebMobile = web / mobile experiment
 experimentPlatformAddonMobile = Firefox / mobile experiment
 experimentPlatformWeb = web experiment
 experimentPlatformAddon = Firefox experiment
-experimentPlatformMobile = Available on mobile
-experimentGoToLink = Go to { $title }
+experimentPlatformMobile = mobile experiment
```
flodolo: At this point we already have 7/8 languages in. Are there technical limitations preventing to change this string ID? If there aren't, we sadly need to.

flodolo (changes requested):

locales/en-US/app.ftl
lmorchard: Just pushed an update that changes the string ID. Looks like that was part of the copy/paste from the string changes pre-merge that I had uncommitted :/

flodolo (approved):
Thanks, this works.

Issue #2109, opened 2017-01-31
Make Experiment details content the same if TXP is installed or not


SoftVision-CosminMuntean:
[Notes]:
- Now that we have one click install, the user should also see the incompatible add-ons massage before installing Test Pilot together with an experiment that has incompatibilities.

[Affected versions]:
- Firefox 51.0.1

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Tabs Group add-on was previously installed on a new clean profile.

[Steps to reproduce]:
1. Open the profile from prerequisites and navigate to https://testpilot.stage.mozaws.net.com.
2. Click on Tab Center experiment card.
3. Observe the Tab Center experiment page.

[Expected result]:
- "This experiment may not be compatible with add-ons you have installed" warning message is displayed on the page.

[Actual result]:
- No warning message is displayed on the page.

![incompatible add-ons massage](https://cloud.githubusercontent.com/assets/19226534/22464569/30cd1e4e-e7c0-11e6-8080-1089f0a30d98.png)

dannycoates:
If we do this we should use mozAddonManager to check which ones are installed, not just show the full list.

johngruen:
We should just make the `details-content` the same for users with TXP not installed as it is for users with TXP installed

ckprice:
Could be a dup of #2394 or something else @johngruen to search.

SoftVision-PaulOiegas:
Blasphemy, our issue is older ! Just kidding. 😄 Right now, the incompatible message is only displayed if you have Test Pilot add-on installed. 
@ckprice Not sure if it's a dupe, but I think that what you guys are trying to do on #2394, will probably cover our case too. But maybe we should leave this open as a "depends on" so we won't forget to check this scenario when you guys implement it ?

fzzzy:
From looking at the code, I believe this issue has already been fixed. It is hard to test since I don't think we have any experiments which set the incompatible addons field right now. I think we should close this.

PR #2628, opened 2017-07-19
make some ui updates for next launch

johngruen:
This PR does a few small things:
* condenses the newsfeed slightly to move experiments up the page
* changes the layout of the experiment types notice on experiment pages to better match spec
* removes 'install test pilot' banner from web experiments
* gives the user a button to check out a web experiment even if they do not have test pilot installed
* makes it so experiment page layout is the same regardless of whether user has add-on


johngruen:
Fixes #2109

fzzzy:
Fantastic!

SoftVision-PaulOiegas:
Checked and confirmed:
- condenses the newsfeed slightly to move experiments up the page
- makes it so experiment page layout is the same regardless of whether user has add-on

The other 3 are conditioned by the new experiments. We'll test those after the new batch lands.

PR #2631, opened 2017-07-20
Skip flake8 on Pontoon commit

b4handjr:
This should skip flake8 on pontoon commits and hopefully solve the issue when tox isn't cached.


fzzzy (approved):
👍👍

PR #2632, opened 2017-07-20
Fix #2605 Allow youtube embeds by adding child-src https://www.youtube.com; to the CSP.

### LINKED ISSUE #2605: Allow youtube embeds
Per Sharon's user research, we should allow Youtube embeds.

fzzzy:


bin/deploy.sh
```diff
@@ -39,7 +39,7 @@ ONE_YEAR="31536000"
...
-CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://sentry.prod.mozaws.net https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self'; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__;\""
+CSP="\"content-security-policy\": \"default-src 'self'; connect-src 'self' https://sentry.prod.mozaws.net https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self'; img-src 'self' https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://ssl.google-analytics.com; style-src 'self' code.cdn.mozilla.net; report-uri /__cspreport__; child-src https://www.youtube.com;\""
```
lmorchard: Hmm, maybe this should be [`frame-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src)? [Looks like `child-src` is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/child-src). (only noticed because I had to look up what `child-src` does 😜 )

fzzzy:
I switched to frame-src, and verified that the cross-origin policy does not allow the frame to access the parent object.

PR #2636, opened 2017-07-21
give videos their own space

johngruen:
@fzzzy this isolates the youtube video from the introduction so that localizers don't have to contend with it


johngruen:
Fixes: #2637

Issue #2229, opened 2017-02-22
Broken experiment descriptions shown during prod deploy


jaredhirsch:
Opened the prod site during a deploy and saw some broken descriptions for the new experiments:

<img width="1009" alt="screen shot 2017-02-22 at 11 10 59 am" src="https://cloud.githubusercontent.com/assets/96396/23228132/5bbab0c6-f8f0-11e6-893c-938e4986ba56.png">

Need to investigate, but seems like a stale cached .ftl file (no strings for the new experiments).

A hard refresh fixed the issue.

@lmorchard suggested putting assets in a hashed directory name, so that pages load a correct set of assets. Sounds good to me, might be an opportunity to take a cleanup pass at the build/deploy tools.

SoftVision-CosminMuntean:
I think this issue is related to #2126. 
I have managed to reproduce this on Test Pilot website using a old Firefox profile with Test Pilot( v0.9.4-dev-045fca1) add-on installed. 
The new experiments pages are also affected, but after I refreshed the page the strings was correctly displayed.

![sonooze tabs](https://cloud.githubusercontent.com/assets/19226534/23249840/02f0a45e-f9af-11e6-98d9-11fcd64805e6.png)

Also, I noticed that the "New" badge from Test Pilot icon is red, but I am not sure why. Maybe this happens only on old profiles, I will investigate this.

chuckharmston:
This was fixed by #2644.

PR #2644, opened 2017-07-21
Fix #1714 Switch localization from l20n to fluent-react for better react integration.

### LINKED ISSUE #1714: investigate l20n/react integration
In #1712 we fixed a conflict in how l20n and react both modify the dom which could easily happen again elsewhere. We ought to look into a more integrated approach. 

This thread might have some leads: https://groups.google.com/forum/#!topic/mozilla.tools.l10n/XtxHgBEokCA

fzzzy:
Rebase number 2


frontend/tasks/pages.js
```diff
@@ -201,16 +201,13 @@ function generateStaticPage(prepareForClient, pageName, pageParam, component, {
     <meta name="availableLanguages" content={ available_locales } />
     <meta name="viewport" content="width=device-width" />
 
-    <link rel="localization" href="/static/locales/{locale}/app.ftl" />
```
lmorchard, line 204: Still haven't quite gotten through skimming this stuff to the point that I understand how localizations are loaded. But, if we're loading them dynamically in JS, I wonder if we can possibly solve #2229 by hashing the resource URL filenames and loading the right ones to bust cache? Maybe need a manifest of locale -> hashed URL

frontend/test/app/components/ExperimentDisableDialog-test.js
```diff
@@ -27,8 +28,8 @@ describe('app/components/ExperimentDisableDialog', () => {
 
   it('should render a modal container', () => {
     expect(subject.find('.modal-container')).to.have.property('length', 1);
-    expect(subject.find('.modal-header').props()['data-l10n-args'])
-      .to.equal(JSON.stringify({ title: experiment.title }));
+    expect(findLocalizedById(subject, 'feedbackUninstallTitle').props()['$title'])
```
lmorchard, line 31: Lots of test updates to account for the change in finding localized nodes... I wonder if in the future (i.e. not retroactively as a nit on this PR) we should just switch to known class names as handles for tests?

frontend/test/app/components/ExperimentEolDialog-test.js
```diff
@@ -18,8 +19,6 @@ describe('app/components/ExperimentEolDialog', () => {
     subject = shallow(<ExperimentEolDialog {...props} />);
   });
 
-  const findByL10nID = id => subject.findWhere(el => id === el.props()['data-l10n-id']);
```
lmorchard, line 21: man, i copypasta'd this everywhere 😞

gulpfile.babel.js
```diff
@@ -1,4 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies*/
+require('babel-polyfill');
```
lmorchard, line 2: Huh, how'd we get away with not having this before?

package.json
```diff
@@ -123,7 +126,7 @@
     "l10n:check": "npm run l10n:extract && git diff-files --quiet --ignore-all-space locales && echo 'No missing strings'",
     "test:ci": "cross-env NODE_ENV=test nyc mocha --recursive --reporter mocha-junit-reporter --require frontend/test-setup.js frontend/test",
     "test:watch": "npm test -- --watch",
-    "test:all": "npm run lint && npm run flow && npm run test && npm run flow-coverage && cd addon && npm run lint && npm run coverage && npm run test",
+    "test:all": "npm run lint && npm run flow && npm run test && npm run flow-coverage && cd addon && npm run lint && npm run coverage && npm run test && tox -e ui-tests",
```
lmorchard, line 129: Hmm, I haven't tried `test:all` in awhile. Up til now, it worked with just an `npm install`. This requires some python / pip installation steps to use tox, now, right?

frontend/src/app/containers/App.js
```diff
@@ -93,21 +98,72 @@ class App extends Component {
...
+    const negotiated = negotiateLanguages(
+      navigator.languages,
+      config.AVAILABLE_LOCALES,
+      { defaultLocale: 'en-US' }
+    );
+
+    const promises = negotiated.map(language =>
+      Promise.all(
+        [
+          fetch(`/static/locales/${language}/app.ftl`).then(response => addLang(language, response)),
```
lmorchard, line 125: Yeah, this bit where the .ftl resources are loaded... #2229 could be fixable by fetching another resource first, say `languages.json` that has contents like this generated in gulp:
```json
{
"en-US": ["/static/locales/en-US/app.08675309.ftl", "/static/locales/en-US/experiments.decafbad.ftl"],
"fr":  ["/static/locales/fr/app.cafebabe.ftl", "/static/locales/en-US/experiments.deadf00d.ftl"]
}
```
Then, do a fetch on those resources. Maybe a crazy idea.

lmorchard:
Overall, looks :100: to me. Cleaner l10n in general. My eyes glazed over at all the mechanical transforms in the markup, so I could have missed something, but test suite updated & passing feels like a good safety belt for that. Don't really have any nits, just ideas for future dev avenues

frontend/src/app/components/LocalizedHtml.js
```diff
@@ -0,0 +1,80 @@
+
```
lmorchard, line 1: Might need to soak my head in this component for a bit longer to understand the why & how

lmorchard:
<strike>One weird thing: Noticed the `installation_count` placeholders aren't getting filled in, though I *do* see the telemetry JSON getting loaded. Haven't dug into possibly why, but I don't see anything obvious in the diff that would break this</strike>

Looks like I was a rebase behind the times. Force-updated my checkout and this looks fixed.

gulpfile.babel.js
fzzzy: I think it was because we didn't have anything that used generators. I added a generator in this pr, and since pages are rendered at build time, there was an error because regeneratorRuntime wasn't available. This fixes that.

frontend/src/app/containers/App.js
fzzzy: 👍

frontend/test/app/components/ExperimentDisableDialog-test.js
fzzzy: Yes, we could do that, but for some reason looking for localization ids just seems like the right way to do it to me. It's already there, it's important because it determines the text the user sees there, makes sense to use it.

package.json
fzzzy: Hmm, good question. I think you're probably right. I added both test:all and bootstrap. Maybe we could add the python / pip steps to bootstrap. I'd be fine with simplifying all the npm scripts in another way too. I'm not really sure I understand them all.

frontend/src/app/containers/App.js
fzzzy: If we do this, let's require() the languages.json file so that it is included in the bundle and doesn't add another round trip.

frontend/src/app/components/LocalizedHtml.js
fzzzy: Yeah, it took me weeks to figure out how to do this properly. I should write some docs. Hopefully this component can get merged back into fluent-react.

frontend/test/app/components/ExperimentDisableDialog-test.js
lmorchard: I'm also kind of wondering if someday there's a way we can detect strings that have changed without a corresponding change in l10n ID. Way beyond scope for this PR, but might be part of what bugs me about using l10n IDs as hooks in tests. That is, changing the IDs breaks tests, but changing strings doesn't set any alarms off - feels backward

fzzzy: Yes, that would be very good.

fzzzy:
All right. Let's merge this. Any issues that are found can be handled in a followup. Also, I will write some comments for LocalizedHtml in a followup.

PR #2645, opened 2017-07-22
Fix CoC link and new container tab tour

hoelzro:
Hello there!

I noticed while reading the recent changes for testpilot-containers that the way to open a new tab in a container changed from hover to long press and that the tour didn't reflect that, so I fixed the tour.  I also noticed while going over the contributing document that the link to the CoC was broken, so I fixed that as well.


content-src/experiments/containers.yaml
```diff
@@ -58,7 +58,7 @@ tour_steps:
     copy: Start a Container to label and organize a set of tabs.
   -
     image: /static/images/experiments/containers/tour/tour2.jpg
-    copy: Hover over the New Tab button to select a container.
+    copy: Long press on the New Tab button to select a container.
```
chuckharmston, line 61: The build is failing because you changed the meaning of this string without changing the L10n ID, so translations won't be updated to reflect the new meaning. More info about [how we do localization](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization).

You'll want to add something like:

```yaml
copy_l10nsuffix: 'longpress'
```

Then run `npm run l10n:extract` so that [`experiments.ftl`](https://github.com/mozilla/testpilot/blob/master/locales/en-US/experiments.ftl) has the new string and ID.

chuckharmston (changes requested):
Thanks so much for the contribution! One small tweak and this will be good to merge.

chuckharmston:
CC @flodolo for a string change.

content-src/experiments/containers.yaml
chuckharmston: You can see what that might look like in #2646

hoelzro: Thanks for taking a look so quickly!  I'll make that tweak right away!

hoelzro: @chuckharmston The string change has been made - let me know if I got it right!

chuckharmston: This looks exactly right! r+ from me, we'll wait to merge this until it's reviewed by our head localizer, who reviews all string changes.

chuckharmston (approved):

flodolo (approved):

PR #2655, opened 2017-07-24
2382 show graduated to every locale

fzzzy:


lmorchard (approved):
Hmm, would like a small test, but otherwise LGTM

fzzzy:
Since it involves changing the browser locale, I think this is a little more involved to test. Let's rely on manual testing for this one.

lmorchard:
Not an integration test involving the browser & locale - just a unit test for the redux selector in isolation (e.g. [like these](https://github.com/mozilla/testpilot/blob/master/frontend/test/app/selectors/experiment-test.js))

PR #2670, opened 2017-07-26
update launch date and fix notes id

### LINKED ISSUE #2669: If one of the "Voice Fill" or "Notes" experiment page is visited, the "Just launched" tag dispersers from both cards
[Affected versions]:
- Firefox 54.0.1 and above

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a new Firefox clean profile.
- The "testpilot.env" string type pref was manually added in "about:config" page with "dev" value.
- The "extensions.webapi.testing" boolean type pref was manually added in "about:config" page with "true" value.

[Steps to reproduce]:
1. Open the Firefox with the profile from prerequisite and navigate to "https://testpilot.dev.mozaws.net/experiments" page.
2. Click one of the "Voice Fill" or "Notes" experiment.
3. Return to the experiments page and observe the cards.

[Expected result]:
- The "Just Launched" tag disappears only for the visited experiment card.

[Actual results]:
- The "Just launched" tag disappears from both cards even if only one was visited.

[Notes]:
- Here is a screen recording with the issue:
tag 2

johngruen:
Update launch dates. also...
Fixes #2669

### DIFF
--- a/content-src/experiments/notes.yaml
+++ b/content-src/experiments/notes.yaml
@@ -1,4 +1,4 @@
-id: 14
+id: 15
 title: 'Notes'
 slug: notes
 platforms: ['addon']
@@ -75,5 +75,5 @@ contributors:
     title: 'Softvision QA'
     avatar: /static/images/experiments/avatars/emil.jpg
 installation_count: 1
-created: '2017-07-27T14:00:00Z'
+created: '2017-08-01T14:00:00Z'
 order: 1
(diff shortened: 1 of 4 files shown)


content-src/experiments/notes.yaml
```diff
@@ -1,6 +1,6 @@
-id: 14
+id: 15
 title: 'Notes'
-slug: notes
+slug: note
```
pdehaan: Why did you change the slug?

johngruen: that is a GREAT question

PR #2672, opened 2017-07-27
reset old loading bars

johngruen:
New loader is janky. let's ditch it!


fzzzy:
This really does look much better.

PR #2693, opened 2017-07-31
Fix #2667 Fix the privacy link for the Send experiment.

### LINKED ISSUE #2667: Firefox Send privacy statement's "websites privacy notice" link doesn't work
Pretty sure this sounds vaguely familiar, but I couldn't find an open issue for it...

Browsing https://testpilot.dev.mozaws.net/experiments/send and noticed that the "websites privacy notice" text (in the "Your privacy" block) looks like a link, but doesn't have an `href=""` so clicking the text does nothing, which is unfortunate since it looks like a legal privacy link.

firefox_test_pilot_-_send

fzzzy:
The privacy link has to be in the 4th bullet point (#3 if counting from
0) in order to work properly. See the following commit for details:

https://github.com/mozilla/testpilot/commit/6d8dac23e745c6a7d2af43ae8168
56c75cb43e48


lmorchard (approved):
r+ from me since we're just shuffling bullet order and don't need new translations. Hoping we can get the tests green and merge before any further diligent l10n folks roll in with more translations

johngruen:
@flodolo i'm going to merge this since it's a hard requirement for us to ship experiments and I do not believe it will affect any ongoing l10n work. Just calling out that we're touching ftl files here.

flodolo:
This works, the only downside is history in Pontoon will look funky, e.g. https://pontoon.mozilla.org/pt-PT/test-pilot-website/all-resources/?search=sendMeasurements3&string=169082

PR #2705, opened 2017-08-01
add file hash to privacy section of Send

dannycoates:
The file hash is also included. We should call it out specifically.


ghost:
I'm ok calling this out, or not, since we have the "such as"

@flodolo - would we need to redo l10n here since "such as" is in there?

@eleemoz - any objections one way or the other?

learnedhoof:
no objection

content-src/experiments/send.yaml
```diff
@@ -19,7 +19,7 @@ discourse_url: 'https://discourse.mozilla-community.org/c/test-pilot/send'
 privacy_notice_url: 'https://github.com/mozilla/send/blob/master/docs/metrics.md'
 measurements:
   - >
-    When you use Send, Mozilla receives an encrypted copy of the file you upload, and basic information about the file, such as filename and file size. Mozilla does not have the ability to access the content of your encrypted file, and only keeps it for the time or number of downloads indicated.
+    When you use Send, Mozilla receives an encrypted copy of the file you upload, and basic information about the file, such as filename, file hash and file size. Mozilla does not have the ability to access the content of your encrypted file, and only keeps it for the time or number of downloads indicated.
```
chuckharmston, line 22: This needs an [`_l10nsuffix`](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization)

chuckharmston (changes requested):

content-src/experiments/send.yaml
lmorchard: Also needs an `npm run l10n:extract` run and the changes to en-US FTL added to the commit. (That's the failing test in CircleCI)

flodolo: If you feel the need to update en-US, that's not a trivial change. So +1 to both comments up here.

dannycoates:
I think this needs #2722 before this will work

content-src/experiments/send.yaml
```diff
@@ -28,7 +28,7 @@ measurements:
     Send is also subject to our <a>websites privacy notice</a>. When you visit the Send website, information such as your IP address is temporarily retained as part of a standard server log.
   - >
     Send will also collect information about the performance and your use of the service, such as how often you upload files, how long your files remain with Mozilla before they expire, any errors related to file transfers, and what cryptographic protocols your browser supports.
-
+measurements_l10nsuffix: '_v2_'
```
lmorchard: Don't need underscores here, since it just gets camel-cased when it's turned into an ID. Should also probably be more like "filehash" to be more descriptive.

locales/en-US/experiments.ftl
```diff
@@ -144,11 +144,11 @@ pulseContributors3Title = Firefox QA
...
+sendMeasurementsV21 = To allow you to see the status of your previously uploaded files, or delete them, basic information about your uploaded files are stored on your local device, such as Send’s identifier for the file, the filename, and the file’s unique download link. This is cleared if you delete your uploaded file or upon visiting Send after the file expires.
+sendMeasurementsV22 = Anyone you provide with the unique link (including the encryption key) to your encrypted file will be able to download and access that file. You should not provide the link to anyone you do not want to have access to your encrypted file.
+sendMeasurementsV23 = Send is also subject to our <a>websites privacy notice</a>. When you visit the Send website, information such as your IP address is temporarily retained as part of a standard server log.
+sendMeasurementsV24 = Send will also collect information about the performance and your use of the service, such as how often you upload files, how long your files remain with Mozilla before they expire, any errors related to file transfers, and what cryptographic protocols your browser supports.
```
flodolo: Why are all these IDs changing, when only one string has changed?

lmorchard: Because we don't have a mechanism for changing the ID for just one string in a number-indexed array of strings like this. And changing to a representation that *does* allow for changing individual IDs will require revision of all locales to match a new schema. [More comments on it over here](https://github.com/mozilla/testpilot/pull/2722#issuecomment-320068704)

lmorchard:
Looks like we're down to just the one string changing, which is cool by me. 

If there are no objections to that string needing new translations in the next hour or so, I might just go ahead and merge this to get it unstuck. @flodolo - any objections?

flodolo (approved):
Makes absolutely sense, thanks

PR #2711, opened 2017-08-02
Update notesMeasurements1

yfdyh000:
I guess it means statistics instead of predictions.


flodolo:
Changing the FTL file is not enough, you also need to change the yaml file where this string appears.

lmorchard:
Actually, the YAML has to be changed, an l10n ID suffix added to indicate the string has been changed, and then the FTL needs to be regenerated - per instructions here:

https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization

lmorchard:
I think we've decided not to accept this change, for now. It's a small rewording, but will generate an ask for re-translation across all locales. So, I think we'd like to hold off on that

PR #2714, opened 2017-08-02
Fix #2699 Double Back tap required from some links.

### LINKED ISSUE #2699: Double Back tap required from some links.
Can't repro this everywhere buy on HN i have to hit back twice to get to the original view:

https://news.ycombinator.com/item?id=14901234

Note: I have Test Pilot Installed, not sure if that matters.

fzzzy:


package.json
```diff
@@ -127,7 +127,7 @@
     "l10n:check": "npm run l10n:extract && git diff-files --quiet --ignore-all-space locales && echo 'No missing strings'",
     "test:ci": "cross-env NODE_ENV=test nyc mocha --recursive --reporter mocha-junit-reporter --require frontend/test-setup.js frontend/test",
     "test:watch": "npm test -- --watch",
-    "test:all": "npm run lint && npm run flow && npm run test && npm run flow-coverage && cd addon && npm run lint && npm run coverage && npm run test && tox -e ui-tests",
+    "test:all": "npm run lint && npm run flow && npm run test && npm run flow-coverage && cd addon && npm run lint && npm run test && tox -e ui-tests",
```
fzzzy, line 130: This is unrelated, it just fixes test:all now that the addon doesn't have an npm run coverage command any more, but I thought I'd sneak it into this pr.

frontend/src/app/containers/HomePage.js
```diff
@@ -0,0 +1,15 @@
+
+import React from 'react';
+
+import HomePageNoAddon from './HomePageNoAddon';
+import HomePageWithAddon from './HomePageWithAddon';
+
+export default class HomePage extends React.Component {
+  render() {
+    if (this.props.hasAddon) {
+      window.history.replaceState({}, '', '/experiments');
```
lmorchard: Would be nice to have a test associated with this new logic. Also, `window.history.replaceState()` should be wrapped up in the App component and passed in as a prop [like other uses of `window`](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App.js#L259). We've tried to keep direct access to browser APIs out of individual components, which should also make writing a test easier.

lmorchard (changes requested):

johngruen:
@fzzzy verified that this fixes the issue R+ in terms of functionality pending resolution of @lmorchard's comments.

frontend/src/app/containers/HomePage.js
fzzzy: Sounds good! Thanks!

lmorchard (approved):

lmorchard:
Yay coverage goes up

Issue #2627, opened 2017-07-19
Get Test Pilot add-on signed as a Mozilla Extension


ghost:
All the extensions are going through our new signing pipeline (and being hosted on testpilot.f.c/files/) which means they are good to go for when non-webextension restrictions hit Firefox.

However, Test Pilot still does its own thing (signing on AMO?  I thought that was disabled.  Maybe that's only for new add-ons).  We should have Test Pilot run through the same pipeline as everything else.

These days it's a matter of supporting `npm run package` (docs being updated in #2615)

lmorchard:
And now after saying `npm run package` was fine in #2615, I just remembered what the gotcha was here: 

We have two separate `package.json` files - one in the repo root for the website, and one under `addon/` for the add-on. Separate sets of NPM scripts and dependencies. (This is what Issue #2113 was about, but we punted on it.) So, just running `npm install && npm run package` in the root directory won't work - because that hits the website `package.json`. 

Maybe we can stick something like `"package": "cd addon && npm install && npm run package"` in the website `package.json`? That will still mean Jenkins installs all the website dependencies and *then* installs the separate set of add-on dependencies. 

Or, maybe we can make a special case where Jenkins does a `cd addon` first if it detects that an `addon` directory exists?

If we supported `bin/build-addon.sh` in Jenkins, this might be simpler to abstract away

lmorchard:
Another thought: If we change the URL for the main Test Pilot add-on along with Mozilla Extension signing - and I'm assuming we will - [we'll have to change the URL in the frontend code for installation](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/lib/addon.js#L18).

lmorchard:
Not sure I want to call it a complete "fix" for this issue, but we just merged PR #2652 - which implements `npm run package` in the root directory of the Test Pilot repo. That does this:
```
cd addon && npm install && npm run package && cp addon.xpi ..
```
Would be nice to see if jenkins could do that itself if an `addon` directory is found. But, in the meantime, this *should* at least work and result in a built addon.

ghost:
thanks @lmorchard 

@relud - please set test pilot up on /files/ so we can start QAing versions that go there

relud:
another thing: the addon version is extracted from the package.json in the root of the repository. will that be acceptable here, or do we need to implement getting version and addon name from `addon.xpi` first?

relud:
[...]

PR #2715, opened 2017-08-02
Updates for new Mozilla Extension signing infrastructure

### LINKED ISSUE #2651: Stop signing the main add-on in CircleCI
Related to #2643 & #2627 - we should stop signing the add-on via AMO in CircleCI runs on master branch.

This means that the add-on will no longer be installable from Dev & Stage environments via the big green button. The add-on would need to be installed in one of three ways:

1) From the prod site, where it's a release version signed as a Mozilla Extension
2) Via about:debugging as a temporary addon
3) Using an unbranded Firefox build - which I *think* could still run through the usual UX on Dev & Stage, but I might be wrong about that after 57 stops supporting legacy add-ons

lmorchard:
- Bump add-on version to 2.0.1
- Remove AMO-based signing from add-on build
- Change updateURL and updateLink in addon/package.json
- Stop attempting to sign the add-on on master branch CircleCI runs
- Exclude add-on from S3 upload in bin/deploy.sh
- Make addon URL path configurable across hosts for installation process
- Remove /static/addon/latest redirect in local dev server

Issue #2627.
Fixes #2651.


addon/bin/sign
```diff
@@ -1,80 +0,0 @@
...
-
-function checkExistsAndMv(fromFilePath, toFilePath, cb) {
-  fs.stat(fromFilePath, function(err) {
-    if (err) return cb(err);
-    fs.rename(fromFilePath, toFilePath, function(err) {
-      if (err) return cb(err);
-      else cb();
-    });
-  });
-}
```
fzzzy, line 80: Yay for removing code!

fzzzy (approved):
Nice to remove so much code related to this. It gets moved into ops repos, but hey, that's Somebody Else's Problem ;-)

PR #2722, opened 2017-08-03
Allow _l10nsuffix in YAML for parent properties

lmorchard:
This allows us to apply an L10n ID suffix to lists of strings, where the individual strings do not have named properties. For example:
```yaml
measurements:
  - foo
  - bar
measurements_l10nsuffix: updated
```

This results in L10N IDs like the following:
- measurementsUpdated0
- measurementsUpdated1

### DIFF
--- a/frontend/tasks/util.js
+++ b/frontend/tasks/util.js
@@ -49,11 +49,15 @@ function experimentL10nId(experiment, pieces) {
   if (typeof pieces === 'string') {
     pieces = [pieces];
   }
-  const suffix = lookup(experiment, `${pieces.join('.')}_l10nsuffix`);
-  if (suffix) {
-    pieces.push(suffix);
+  const piecesOut = [];
+  for (let i=0; i<pieces.length; i++) {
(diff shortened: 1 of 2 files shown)


lmorchard:
Not super happy with this change, but it seems to address a problem I just noticed in reviewing PR #2711:

We have no way to change the L10N ID for the list of strings under `measurements`, because each item in that list is a bare string. 

This PR lets us set a suffix for the parent property. That lets us change the ID (yay) - but it changes it for all strings in that list (boo).

The only other idea I had, was to change the structure of the `measurements` list so that each item was itself a structure (e.g. - text: blah blah). But, that would require altering every YAML file and every locale's `experiments.ftl` to match the new structure - as well as updating the experiment details page container

fzzzy (approved):
Excellent. Sure, all the string names changing is a "Boo", but I think it has to be that way -- the only other thing I could think is adding a suffix to a specific element in the list, but that might mean you have to have intimate knowledge of how the strings get munged.

PR #2743, opened 2017-08-08
Ensure proper add-on path is used while installing via mozAddonManager

lmorchard:
Quick fix to hardcoded path to old add-on.


lmorchard:
Also filed #2744 to follow up with removing the other spot where this add-on path was hardcoded - the defunct add-on based experiment install code.

ghost (approved):

Issue #2574, opened 2017-06-30
Please stop merging PRs involving strings without l10n review


flodolo:
@clouserw @johngruen 
I've tried multiple times to raise this issue, and failed. Let's try with a public issue, because I'm running out of options.

Here's what happened last night. This landed
https://github.com/mozilla/testpilot/commit/04c94f896f11f85f13748f9e65921f4e4a8cff24

And then this landed
https://github.com/mozilla/testpilot/commit/9f228465c97e7fb79804907dd9a1f5b4f84d4583

This morning I wake up, find suggestions on Pontoon from one of my fellow translators (@gialloporpora), and they make no sense, because they don't match the English text. So I go in, check the history, and find this mess.

Now we need to:
* Change the string ID for those 3 messages, to make sure people translate the right content.
* Send a message out to explain what happened.

I think the only way out of this is to make sure any string change lands reviewed beforehand by me, and that's not great, because Test Pilot is not the only thing I'm following, and I'm 9 hours away from most of the devs.

If you have better ideas, I'm all hears.

lmorchard:
We haven't gotten down to executing on this yet, but we have a notion of a plan to change how things work here: https://github.com/mozilla/testpilot/issues/2430#issuecomment-302809115

TL;DR: Move any changes involving strings to an l10n branch. Make PRs into that branch for l10n review. Once strings are localized, make a PR from l10n to master to land the feature. What we might be missing, though, is who else to ping for l10n reviews so we're not always pestering you (@flodolo)

I think this week has been a bit chaotic because we're all at the SF work week and trying to make some things happen quickly

johngruen:
+1 @lmorchard.  We could also look at setting flags in circle CI to warn us if we change values in one of the yaml files so that we make sure to appropriately modify keys.

flodolo:
> TL;DR: Move any changes involving strings to an l10n branch. Make PRs into that branch for l10n review. Once strings are localized, make a PR from l10n to master to land the feature. What we might be missing, though, is who else to ping for l10n reviews so we're not always pestering you (@flodolo)

OK, that would solve a lot of issues. The challenges are:
* To make sure we merge changes frequently from the l10n branch into master
* Figure out a way to have a testing server that has the latest strings translated

lmorchard:
[...]

PR #2759, opened 2017-08-10
update news feed8/10

johngruen:


content-src/news_updates.yaml
```diff
@@ -8,25 +8,6 @@
 #   dev: false
 #   content: >
 #     Example content goes here
--
-  slug: general-update-3
```
chuckharmston, line 12: Did you mean to delete these?

johngruen: yeah, they don't expire on their own ATM

chuckharmston (approved):
I'll let you self-merge after you've confirmed that you wanted to delete the other news updates.

(And after CI passes.)

content-src/news_updates.yaml
chuckharmston: I'm not sure we should do this; there will be l10n conflicts in the future when we add new ones.

locales/en-US/experiments.ftl
```diff
@@ -252,7 +252,5 @@ voicefillContributors4Title = Firefox UX
 voicefillContributors5Title = Softvision QA
 voicefillContributors6Title = Softvision QA
 voicefillContributors7Title = Advanced Development, Emerging Technologies
-testpilotNewsupdatesGeneralupdate3Title = New experiments launched
-testpilotNewsupdatesGeneralupdate3Content = Introducing Send, Notes, and Voice Fill. Click on the cards below to try them out.
-testpilotNewsupdatesGeneralupdate4Title = We started a blog
-testpilotNewsupdatesGeneralupdate4Content = Check out our brand new blog for in-depth discussion about Firefox Test Pilot.
\ No newline at end of file
+sendNewsupdatesSendupdate1Title = Send with Safari and Edge!
```
lmorchard, line 255: FWIW, [we need to wait to get some l10n review whenever changing or adding strings](https://github.com/mozilla/testpilot/issues/2574), so this shouldn't be just a quick merge. That's why we have the auto-request for review from @flodolo up there

content-src/news_updates.yaml
lmorchard: It's always been the plan to remove news updates from the YAML. [The slug (which should be unique) is a part of the l10n IDs](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/util.js#L63) - that should help prevent conflicts

content-src/experiments/send.yaml
```diff
@@ -82,3 +82,13 @@ installation_count: 1
 created: '2017-08-01T14:00:00Z'
 launch_date: '2017-08-01T14:00:00Z'
 order: 2
+news_updates:
+  -
+    slug: send-update-1
```
lmorchard, line 87: In fact, to keep these things extra unique, maybe stick a date in there - like `send-update-2017-08-10-1`. These slugs need to never be accidentally re-used with new updates later on when old ones have been long since deleted.

flodolo (approved):
I'm OK with these changes, but I think we need to start a larger discussion on how to manage this section.

They're not going to be localized, so I wonder if we could find a less prominent space to show them for language != en-US (is it just me or they're not displayed right now on TP?), and only localize a section title (TP Latest News) + blurb explaining that news are in English, and only show the title for each news in English.

lmorchard:
> They're not going to be localized

@flodolo <strike>Why aren't these strings going to be localized?</strike> Oh! I see what you mean: The *strings* in the news items on the site are localized, but the Medium blog posts aren't. Good point.

lmorchard:
Submitted issue #2762, but otherwise this looks like it's got all the r+'s

PR #2771, opened 2017-08-11
Use negotiated languages when selecting localizations

### LINKED ISSUE #2753: Localization issue reported on Twitter
@fzzzy

User reports: `Accept-Language: en,ru;q=0.7,ja;q=0.3`

Looks like they're seeing a mix of accepted languages :(

screen shot 2017-08-09 at 6 37 25 pm

### LINKED ISSUE #2770: Blank strings when using en-CA locale
Looks like we're falling back to in-DOM strings on the en-CA locale (maybe others too). Related to #2753 maybe?

As an example, the "Just Launched" string ends up blank - because that DOM node is blank. The platform indicator string, however, falls back to [DOM text](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/ExperimentPlatforms.js#L29) - which differs from [the final en-US strings that we shipped](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl#L156) with the feature.

If I have en-US in my list of languages in browser preferences, I see this:
capture

If I remove en-US and add en-CA, I see this:
capture

lmorchard:
Also use likelySubtags for better negotiation via fluent-langneg.

See also:
https://github.com/projectfluent/fluent.js/tree/master/fluent-langneg#likely-subtags

Fixes #2770
Fixes #2753

### DIFF
--- a/frontend/src/app/reducers/localizations.js
+++ b/frontend/src/app/reducers/localizations.js
@@ -1,34 +1,53 @@
 // @flow
 
+import type {
+  SetLocalizationsAction,
+  SetNegotiatedLanguagesAction,
+  LocalizationsAction
+} from '../actions/localizations';
+
 export type Localizations = {
   [language: string]: string
 };
 
-function defaultState(): Localizations {
+export type NegotiatedLanguages = Array<string>;
+
+export type LocalizationsState = {
+  localizations: Localizations,
+  negotiatedLanguages: NegotiatedLanguages
+};
+
+function defaultState(): LocalizationsState {
   return {
+    localizations: {},
+    negotiatedLanguages: []
   };
 }
 
-export type SetLocalizationsAction = {
-  type: 'SET_LOCALIZATIONS',
-  payload: Localizations
-};
-
 function setLocalizations(
-  state: Localizations,
+  state: LocalizationsState,
   action: SetLocalizationsAction
-): Localizations {
-  return action.payload;
+): LocalizationsState {
+  return { ...state, localizations: action.payload };
+}
+
+function setNegotiatedLanguages(
+  state: LocalizationsState,
+  action: SetNegotiatedLanguagesAction
+): LocalizationsState {
+  return { ...state, negotiatedLanguages: action.payload };
 }
 
 export default function localizationsReducer(
-  state: ?Localizations,
-  action: SetLocalizationsAction
-): Localizations {
+  state: ?LocalizationsState,
+  action: LocalizationsAction
+): LocalizationsState {
   if (!state) {
     return defaultState();
   }
   switch (action.type) {
+    case 'SET_NEGOTIATED_LANGUAGES':
+      return setNegotiatedLanguages(state, action);
     case 'SET_LOCALIZATIONS':
       return setLocalizations(state, action);
     default:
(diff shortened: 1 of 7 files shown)


lmorchard:
Also, this is kind of quick & dirty - the prioritized list of negotiated languages should probably be stashed & fetched from the redux store along with the localizations themselves

lmorchard:
[Wherein flow types + redux boilerplate drive me to drink](https://github.com/mozilla/testpilot/pull/2771/commits/e1a86bbee74392c0b86c92e5f3ae26bfa1107fa1#diff-e94fde84a3b827ee9417e6e58c9f070bR50) 🍸 

(edit: deleting my rambling that was here, and moving it to a separate issue)

frontend/src/app/reducers/localizations.js
```diff
@@ -1,34 +1,60 @@
...
+): LocalizationState {
+  return { ...state, negotiatedLanguages: action.payload };
 }
 
 export default function localizationsReducer(
-  state: ?Localizations,
-  action: SetLocalizationsAction
-): Localizations {
+  state: ?LocalizationState,
+  action: any // TODO: figure out how to correctly assert any Action here
```
lmorchard: The longer explanation, for completeness' sake:

I noticed that flow-type-annotated reducers in our code assert this action should be the set of actions notionally associated with the reducer.

The problem, generally speaking, is that *any* action is valid here. *Any* reducer can handle *any* action - and in fact *all* reducers see *all* actions passing through `dispatch()`. It's the switch statement that dynamically decides when to do something. This doesn't fail flow type linting, because the chain of code which makes that happen is buried inside vendor code in Redux.

But, when I try to assert the more generic `Action` type here, I don't know how to type-case it into the specific action types for the other sub-functions. I get errors when I try.

frontend/src/app/reducers/localizations.js
```diff
@@ -1,34 +1,60 @@
...
+): LocalizationState {
+  return { ...state, negotiatedLanguages: action.payload };
 }
 
 export default function localizationsReducer(
-  state: ?Localizations,
-  action: SetLocalizationsAction
-): Localizations {
+  state: ?LocalizationState,
+  action: Object // TODO: figure out how to correctly assert any Action here
```
kumar303: `action: SetLocalizationsAction | SetNegotiatedLanguagesAction` should work, no? Since those are the only two types your reducer will operate on, those are the only ones you need to declare.

lmorchard:
Got hung up on #2772 but thinking over the weekend about [Flow docs](https://flow.org/en/docs/frameworks/redux/#toc-typing-redux-actions) and what @kumar303 mentioned has turned me around.

lmorchard:
Made one more fussy change I probably should have left alone: Moved action types out of the reducer and into the action module because they seem more at home there

chuckharmston (commented):
I've tested this and can confirm that it fixes the cases in #2770 and #2753. I don't have any particular nits with the code, though admittedly I only understand about 80% of the problem and only some of the flowtypes bits. Tentatively approving, but I'll let @lmorchard hit the big green button when he's ready.

chuckharmston (approved):

Issue #2602, opened 2017-07-07
Icon in extended tabbar looks ugly


Tobias-B-Besemer:
The Test-Pilot-Icon in the extended tabbar (the arrow for the expanded list) looks ugly.
![2017-07-07 19_03_38-add-ons manager - nightly build 20170707030206](https://user-images.githubusercontent.com/11881073/27970364-d8904070-634e-11e7-9632-a68f03aef870.png)

johngruen:
ewww

SoftVision-PaulOiegas:
Maybe a dupe of #2585? I know it's on another menu but most likely it has the same root cause.

Tobias-B-Besemer:
Yes, seems to be the same bug! Just a other view of it...

nchevobbe:
Not sure if this is the same issue since I have different result, but still looking not that good :
 
<img width="342" alt="screen shot 2017-08-11 at 15 18 42" src="https://user-images.githubusercontent.com/578107/29214696-a19ea76c-7ea8-11e7-93f4-1ba1928ee615.png">

Tobias-B-Besemer:
And this one...
![2017-08-11 20_11_42-icon in extended tabbar looks ugly issue 2602 mozilla_testpilot - nightly](https://user-images.githubusercontent.com/11881073/29226210-73924dbe-7ed1-11e7-89ff-d6777206ee47.png)

Tobias-B-Besemer:
And 'Testpilot' should IMHO be written with a big 'T'.

Tobias-B-Besemer:
'Test Pilot' is now written in Add-ons with two words, but as tooltip in one word.
![2017-08-12 17_44_19-add-ons manager - nightly](https://user-images.githubusercontent.com/11881073/29242065-45058690-7f86-11e7-8e1e-d00016e87268.png)

lmorchard:
https://github.com/mozilla/testpilot/commit/619fb50bed94e9d1fe156384423d0ca92967a408

Tobias-B-Besemer:
What's about the wrong tooltip ???

lmorchard:
Got a fix for that too, I think https://github.com/mozilla/testpilot/pull/2794

SoftVision-PaulOiegas:
[...]

PR #2776, opened 2017-08-14
use context fill for webext icon

johngruen:
@lmorchard I adjusted the webextension manifest so we can use `context-fill` for the test pilot icon and inherit the standard toolbar icon color for a given theme. 

I basically copied what we did here: 
https://github.com/mozilla-services/screenshots/blob/master/addon/webextension/manifest.json.template

Note that this change is only supported in nightly ATM, though we're requesting uplift to Beta:
https://bugzilla.mozilla.org/show_bug.cgi?id=1388907


johngruen:
Also i tested this by using  `web-ext run --firefox=/Applications/FirefoxNightly.app/Contents/MacOS/firefox-bin` from the `addon/build/webextnsion` directory after running `npm run package:dev`. I couldn't figure out how to build and run the bootstrapped add-on on Nightly, which is the only place the necessary in-tree changes have landed thus far.

johngruen:
Fixes #2602 as well

johngruen:
Also, not sure what our versioning scheme is on the add-on, but this may require a version bump.

lmorchard:
If you ran `npm run package:dev`, that should give you an XPI that you can try [with about:debugging](https://blog.mozilla.org/addons/2015/12/23/loading-temporary-add-ons/). That, or [following the dev setup here](https://github.com/mozilla/testpilot/tree/master/addon#building-and-development) can get you running with a file watcher.

It's not a webextension, so loading it with `web-ext run` won't work.

And yeah, every change that we want to go out as an update requires a version bump. Generally trying to follow the semver pattern of major.minor.patch versioning, so it just needs a change from 2.0.2 to 2.0.3 in package.json

addon/src/webextension/icons/icon-16.svg
```diff
@@ -0,0 +1 @@
+<svg width="16" height="14" viewBox="0 0 16 14" xmlns="http://www.w3.org/2000/svg"><path fill="context-fill" d="M14.962 6.169c.51.253.865.738.954 1.3.092.83-.476 1.884-1.51 2.85.48 1.481 1.027 3.681.031 3.681-.724 0-1.68-1.05-2.42-2.079a10.922 10.922 0 0 1-3.534 1.019 10.847 10.847 0 0 1-4.886-.59C2.916 13.223 2.123 14 1.5 14c-.832 0-.587-1.538-.204-2.908a3.074 3.074 0 0 1-1.26-1.874c-.124-1.132 1.05-2.069 2.94-2.747C3.07 3.95 4.372 0 8 0a4.365 4.365 0 0 1 3.867 2.35c1.2-.207 2.415-.324 3.633-.35.5 0 .5 0 .5.5a12.08 12.08 0 0 1-1.038 3.669zM8 1.45c-2.212 0-3 2.05-3.433 4.65C4.42 6.99 4 8 7 8c3.5 0 4.487-1.609 4.43-1.928C11 3.696 10.2 1.45 8 1.45z"/></svg>
```
lmorchard, line 1: I'm not really sure what context-fill does, but it seems to be working in reverse from what's intended:

* Default theme: 
  capture2
* Light compact theme: 
  capture3
* Dark compact theme: 
  capture4

lmorchard (changes requested):
Hmm looks like [we need a new pref to get the unsigned add-on installed on Nightly](https://github.com/mozilla/testpilot/pull/2779): extensions.legacy.enabled = true

Did that, ran `npm run package:dev` and installed the XPI with `about:debugging`. Looks like the icon is the inverse of what we want it to be on light & dark themes

addon/src/webextension/icons/icon-16.svg
johngruen: @lmorchard i don't think the svg would have that drop shadow, are you sure you're looking at the version w/ the svg icon? I can verify later

lmorchard: Hmm, not sure what happened... I restarted and now it works fine when I install via about:debugging. So, mainly just needs a version bump I think

addon/package.json
```diff
@@ -2,7 +2,7 @@
   "title": "Test Pilot",
   "name": "testpilot-addon",
   "id": "@testpilot-addon",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
johngruen, line 5: @lmorchard i bumped the version in three places which seems not good, but maybe that's how it's done

lmorchard: No, you only change it in package.json. The webpack build copies that version over to install.rdf and manifest.json

lmorchard (approved):

lmorchard:
Doesn't hurt anything to change the version in all three spots, but we don't need to do that in the future. Will just merge though

PR #2779, opened 2017-08-14
Add mention of extensions.legacy.enabled to addon README

lmorchard:


addon/README.md
```diff
@@ -1,134 +1,138 @@
...
+```
+npm package      # Production mode without logging
+npm package:dev  # Development mode with lots of Browser Console logging
+```
+
+If you'd like to actively work on the add-on, here are some additional steps to set up a more convenient workflow:
+
+1. Install [Firefox Developer Edition][devedition]. (Nightly should work, too, but Dev Edition is preferred.)
+
+1. Set `extensions.legacy.enabled` to `true` [using `about:config`][aboutconfig],
```
lmorchard, line 32: Huh. Not sure why the diff looks like I re-wrote the whole file. Adding this bullet about extensions.legacy.enabled is all I changed.

lmorchard:
Ah hah. I guess my new windows setup stuck some DOS-style line endings in there when I switched from Atom to Vim (or vice versa). Split the line endings change out into a separate commit from the actual text change.

fzzzy:
Ahh DOS line endings... At least mac-specific \r line endings are no longer a thing.

Issue #1845, opened 2016-11-18
Audit existing avatar image sizes


pdehaan:
Ref: https://github.com/mozilla/testpilot/pull/1843#pullrequestreview-9301574

> @chuckharmston: Can you resize the avatar to 64x64 please?
> https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#contributors

| WxH | SIZE | PATH |
|:---:|-----:|:-----|
| 64x64 | 2.3KB | /frontend/src/images/experiments/activity-stream/avatars/kate-hudson.jpg
| 64x64 | 9.9KB | /frontend/src/images/experiments/activity-stream/avatars/ricky-rosario.jpg
| 400x400 | 34.8KB | /frontend/src/images/experiments/activity-stream/avatars/nan-jiang.jpg
| 500x500 | 128.2KB | /frontend/src/images/experiments/activity-stream/avatars/ursula-sarracini.jpg
| 371x341 | 74.4KB | /frontend/src/images/experiments/activity-stream/avatars/marina-samuel.jpg
| 64x64 | 2.2KB | /frontend/src/images/experiments/activity-stream/avatars/edwin-wong.jpg
| 64x64 | 2.2KB | /frontend/src/images/experiments/activity-stream/avatars/jared-kerim.jpg
| 64x64 | 2.1KB | /frontend/src/images/experiments/activity-stream/avatars/andrei-oprea.jpg
| 64x64 | 1.6KB | /frontend/src/images/experiments/activity-stream/avatars/bernardo-rittmeyer.jpg
| 64x64 | 3.0KB | /frontend/src/images/experiments/activity-stream/avatars/nick-chapman.jpg
| 460x460 | 132.2KB | /frontend/src/images/experiments/activity-stream/avatars/tim-spurway.jpg
| 125x125 | 8.2KB | /frontend/src/images/experiments/activity-stream/avatars/olivier-yiptong.jpg
| 64x64 | 7.1KB | /frontend/src/images/experiments/activity-stream/avatars/bryan-bell.jpg
| 112x112 | 2.9KB | /frontend/src/images/experiments/cliqz/avatars/cliqz-avatar.jpg
| 112x112 | 26.5KB | /frontend/src/images/experiments/avatars/tvyas.png
| 112x112 | 19.4KB | /frontend/src/images/experiments/avatars/amarchesini.png
| 112x112 | 27.6KB | /frontend/src/images/experiments/avatars/lcrouch.png
[...]

PR #2780, opened 2017-08-14
CHORE: resize avatars

CHBaker:
Resize all avatars from the experiments directory for consistency.

- Size: 64x64 pixels
- Passes all tests

addresses: #1845


johngruen:
@CHBaker hey, thanks for knocking this out...i'm guessing you did it with a script?

If so, I think we should probably do everything at 112px x 112px and then use width and height attrs in HTML to set images to 64 x 64 and browsers handle downsampling the images. This should work better for Retina/HiDPI displays.

CHBaker:
Yes @johngruen I can change them to 112x112 px, and I only just started contributing here so I'm not sure what files the images are sized to the html, if you point that out i'll get it taken care of

johngruen:
@CHBaker I commented on the original bug, sorry @pdehaan and I have been giving you conflicting info here :(

lmorchard:
Are there pending next steps to wrap up this PR? Or can we merge it?

CHBaker:
@lmorchard  These are good to merge, all images are 112px by 112px

meandavejustice:
Pulled this down _just to be sure_ and it's looking good. Can you squash these commits into one @CHBaker, then we should be good to merge

chuckharmston:
@meandavejustice we can squash from here! On it. Thanks for the contribution @CHBaker!

PR #2790, opened 2017-08-17
Added install and enable tests - part 2

b4handjr:
Found a problem in circleci that caused some weird hangs during the CI runs. It appeared that sometimes the ```@pytest.mark.skipif``` decorator would cause a problem. It could be that the ```set -ex``` bash script setting was causing the pytest script to exit when it parsed the decorator, if the arguments within ```@pytest.mark.skipif``` were incorrect.

See: https://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,6 +7,7 @@ machine:
   environment:
     DISPLAY: :99
     PATH: "/home/ubuntu/testpilot/firefox:$PATH"
+    MOZ_HEADLESS: 1
   hosts:
     example.com: 127.0.0.1
   node:
@@ -38,8 +39,7 @@ test:
     - ./bin/circleci/flake8.sh
     - mkdir integration-test-results
     - ./bin/circleci/test-firefox-release.sh
-    - ./bin/circleci/test-firefox-dev.sh
-    - # ./bin/circleci/test-firefox-nightly.sh
+    - ./bin/circleci/test-firefox-nightly.sh
   post:
     - bash <(curl -s https://codecov.io/bash)
     - cp -r integration-test-results $CIRCLE_ARTIFACTS
--- a/frontend/test/ui/conftest.py
+++ b/frontend/test/ui/conftest.py
@@ -1,5 +1,9 @@
+import os
+import shutil
+
 import pytest
 import requests
+import ruamel.yaml
 from requests.packages.urllib3.util.retry import Retry
 from requests.adapters import HTTPAdapter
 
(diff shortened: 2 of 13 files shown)


chuckharmston:
@jrbenny35 would you mind elaborating on what the bug was and why this fixed it?

b4handjr:
> would you mind adding info about what the bug was and why this fixed it?

Sure, I didn't track it down too much but I think it was circle ci caching. I don't think it was a problem with the script files. That's why there were so many commits. 

I guess saying there was a bug with the test setup would be incorrect, more so, a quirk of circleci and how the new tests were being set up.

circle.yml
```diff
@@ -17,7 +18,7 @@ machine:
 dependencies:
   cache_directories:
     - "addon/node_modules"
-    - "/home/ubuntu/firefox-downloads"
+    - # "/home/ubuntu/firefox-downloads"
```
fzzzy: I'm not super familiar with yaml -- what does this do? Exclude it from the list? Can we just delete the line instead?

circle.yml
```diff
@@ -34,8 +35,8 @@ test:
     - ./bin/circleci/flake8.sh
     - mkdir integration-test-results
     - ./bin/circleci/test-firefox-release.sh
-    - ./bin/circleci/test-firefox-dev.sh
-    - # ./bin/circleci/test-firefox-nightly.sh
+    - # ./bin/circleci/test-firefox-dev.sh
```
fzzzy: Same here, can we just delete this?

frontend/test/ui/conftest.py
```diff
@@ -17,11 +21,35 @@ def firefox_options(firefox_options):
...
     firefox_options.set_preference('extensions.webapi.testing', True)
-    firefox_options.set_preference('example.com', 'local')
+    firefox_options.set_preference('testpilot.env', 'local')
+    firefox_options.add_argument('-foreground')
     firefox_options.log.level = 'trace'
     return firefox_options
 
 
+@pytest.yield_fixture(scope='session', autouse=True)
+def enable_dev_experiment():
```
fzzzy, line 31: This isn't called anywhere. I can kind of guess that the decorator automatically applies it, but can you clarify?

scope='session' means that it gets run once per session, and autouse means it gets used?

Does the code before the yield run at the beginning of the session, and the code after the yield run at the end of the session?

frontend/test/ui/requirements/requirements.txt
```diff
@@ -1,7 +1,10 @@
 PyPOM==1.2.0
-pytest==3.2.1
+pytest==3.2.0
```
fzzzy: Is this intentional? What's the reason for the downgrade?

frontend/test/ui/test_install.py
```diff
@@ -0,0 +1,37 @@
+import os
+
+import pytest
+
+from pages.desktop.home import Home
+from pages.desktop.detail import Detail
+
+
+@pytest.mark.nondestructive
+@pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
```
fzzzy, line 10: Is there a technical reason for this skip?

fzzzy (changes requested):
I just have some questions about the diff to clarify my own understanding. Thanks.

circle.yml
b4handjr: I can remove this back to how it was. Nightly had a bug last week that caused it to not read the capabilities given to it by geckodriver.

circle.yml
b4handjr: So just leave release and nightly?

frontend/test/ui/conftest.py
b4handjr: You are 100% correct.

frontend/test/ui/requirements/requirements.txt
b4handjr: This wasn't intentional.

frontend/test/ui/test_install.py
b4handjr: Yes so the preference ```xpinstall.signatures.required', False``` cannot be set in firefox release or beta, but only dev and nightly. Currently dev, is being shipped as beta it seems, so this skips the install tests on release but runs them on nightly.

circle.yml
fzzzy: Please do.

circle.yml
fzzzy: I think release and nightly are good enough.

frontend/test/ui/test_install.py
fzzzy: Ok, great info, thanks.

b4handjr:
@fzzzy Am I missing something?

fzzzy:
@jrbenny35 Please squash it. Also it looks like the tests failed on the last commit.

fzzzy:
Oh, it has conflicts, so it needs rebasing.

b4handjr:
Should be good now!

fzzzy:
Although the tests pass on this, once it is merged to master, the script which deploys to dev breaks in a strange way. I'm going to revert this until we figure out what is going on there.

PR #2794, opened 2017-08-17
Change webextension name to Test Pilot so the tooltip & sidebar titles are correct on button

lmorchard:


SoftVision-PaulOiegas:
@lmorchard @fzzzy Not sure what this fixed, but I still see "testpilot" tooltip on all 3 servers. So something is wrong...
The strange part is that latest dev add-on version is 2.0.2, while in #2602 you guys bumped the version to 2.0.3.

tooltip

Issue #2430, opened 2017-05-16
Move pontoon commits to a branch


fzzzy:
We should tell pontoon to put commits on a branch, then only open a pull request for all the pontoon stuff once a sprint. This will cut down a lot on both #testpilot-bots spam and on circleci load.

Remaining steps (per @lmorchard):

- [x] Get testpilot-l10n.dev.mozaws.net working as an approved host name
- [x] Add l10n environment to add-on
- [x] Create an l10n branch on this repo
- [x] Update CircleCI to auto-deploy the site to the new dev server on l10n branch commits
- [x] [Bug 1393106](https://bugzilla.mozilla.org/show_bug.cgi?id=1393106) - Ping pontoon folks to move commits over to l10n branch, after all the above is done
- [ ] Consider any git automation that might help keeping master & l10n branches in sync (might need to be a manual post-deploy process?)
- [ ] Consider any test automation steps that might help make sure we run PRs through l10n branch before master
- [ ] Document the new l10n process

ghost:
@flod if we do this we lose the string preview on -dev (until we merge it in).  Is that important to have?

ghost:
er, I mean @flodolo ^

flodolo:
Yes, it's important to have a dev server that works against that branch (also cc @mathjazz).

I'm fine in having Pontoon committing to the branch, but we owe localizer an easy and automated way to check their work.

lmorchard:
Just had a brief thinking-out-loud about this on IRC... what if in addition to a branch and a second l10n-specific dev instance, we had a little more process too:

1. When there are new strings, we submit a pull request from `master` branch to an `l10n` branch.
1. Strings in that PR get reviewed / revised as needed, then merged into `l10n`
1. Commits from localizers on Pontoon flow into `l10n` branch.
1. When we decide we're ready to ship the locales, we submit a PR with a single squashed commit back to `master` - this will cut down on the number of commits in `master` history.

The wrinkle here, though, is squashing commits destroys [contributor attribution](https://github.com/mozilla/testpilot/graphs/contributors) when we have it - because git only supports a single author & committer per commit. 

So, if we did squash masses of Pontoon commits down to one, can we find another way to attribute the work of l10n contributors? (e.g. an API we can hit on Pontoon to get a contributor list and publish it on the site?)

mathjazz:
[...]

PR #2796, opened 2017-08-17
Changes to enable l10n dev server

lmorchard:
- CircleCI deployment to l10n server

- Tweaks to startup warning dialog to allow l10n server

- Add l10n environment to add-on

Issue #2430

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -2,7 +2,7 @@
   "title": "Test Pilot",
   "name": "testpilot-addon",
   "id": "@testpilot-addon",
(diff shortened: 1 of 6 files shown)


lmorchard:
Still some funky things happening when mozAddonManager is unavailable, but kind of want to tackle that as a follow up issue with some consideration toward how that should work

lmorchard:
Also, I've been running this through my own CircleCI account, and it's been successfully deploying:

https://testpilot-l10n.dev.mozaws.net/

Issue #2421, opened 2017-05-05
Add link to news feed in header


johngruen:
pretty basic, add link to the news feed in the `<Header>` component

meandavejustice:
This should link to the blog now (getting blog link/access tomorrow!)

PR #2799, opened 2017-08-18
Add blog link

johngruen:
Fixes #2121


flodolo (approved):
No issue for l01n (for clarity, I only look at strings in the FTL, not at the code)

flodolo:
> No issue for l01n (for clarity, I only look at strings in the FTL, not at the code)

I wonder if I should avoid reviewing the strings explicitly, since they make the PR reviewed and it might be confusing.

lmorchard (approved):
:+1: - with one nit: Would like to see the two commits squashed and a "Fixes #2421" added to the comment

lmorchard:
> I wonder if I should avoid reviewing the strings explicitly, since they make the PR reviewed and it might be confusing.

I think it's fine... just having a single approval doesn't mean a PR shouldn't have more

johngruen:
Squashed...will merge when tests finish

SoftVision-PaulOiegas:
@johngruen If I understood well from the PR code, the blog link should be displayed only on top of the page, right? You haven't mentioned any details about it. 
Adding an "M" button on the footer beside GitHub and Twitter ones would be to much?

PR #2802, opened 2017-08-18
Update README.md

ghost:


lmorchard:
I don't think we want to accept these edits, and there are several of these similar PRs.

PR #2803, opened 2017-08-18
Serialize dev server deployments from master branch

### LINKED ISSUE #2634: Collisions in parallel dev deployments sometimes break the site (loader never stops spinning)
**[Notes]:**
- Might be again something that will resolve by itself, but better safe than sorry. 

**[Affected versions]:**
- Firefox 54.0.1 and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- The "testpilot.env" string type pref was manually added in "about:config" page with "dev" value.
- The "extensions.webapi.testing" boolean type pref was manually added in "about:config" page with "true" value.

**[Steps to reproduce]:**
1. Open the Firefox with the profile from prerequisite and navigate to https://testpilot.dev.mozaws.net page.
2. Observe the page behavior.

**[Expected result]:**
- Test Pilot dev server Homepage is correctly loaded.

**[Actual results]:**
- The page blue background and loading animation appear but the page remains in infinite loading.
 
**[Additional Notes]:**
- Here is a console screenshot:

dev errors for not loading

lmorchard:
- do-exclusively script that watches CircleCI status API to ensure all
  builds that arrived earlier finish before this one deploys

Fixes #2634

### DIFF
--- a/bin/circleci/do-exclusively.sh
+++ b/bin/circleci/do-exclusively.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# https://github.com/bellkev/circle-lock-test/blob/master/do-exclusively
+
+# sets $branch, $tag, $rest
+parse_args() {
+    while [[ $# -gt 0 ]]; do
+        case $1 in
+            -b|--branch) branch="$2" ;;
+            -t|--tag) tag="$2" ;;
+            *) break ;;
+        esac
+        shift 2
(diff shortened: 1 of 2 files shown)


lmorchard:
Seems to work on my personal circleci account, hard to test otherwise. We can revert it right away if it breaks dev deployments

PR #2810, opened 2017-08-22
Fix #2736 Remove download counts; also remove the loading indicator

### LINKED ISSUE #2736: Remove download counts
The download counts have never been very accurate, they slow the site down considerably, and user testing has shown that users react in odd ways to them. We're just going to remove them entirely.

fzzzy:
The user counts were the only async thing that caused the loading
indicator to be required due to reflowing the page after the data came
in.


content-src/experiments/activity-stream.yaml
```diff
@@ -116,7 +116,6 @@ contributors:
     display_name: 'Bryan Bell'
     title: 'Senior UX Designer'
     avatar: /static/images/experiments/activity-stream/avatars/bryan-bell.jpg
-installation_count: 46732
```
lmorchard, line 119: :+1: These are an unused artifact since the days of exporting this data from the Django database

frontend/test/app/actions/experiments-test.js
```diff
@@ -1,36 +0,0 @@
-
```
lmorchard, line 1: Hmm, I'm surprised there aren't more actions & tests in here. But 🤷‍♂️

lmorchard (approved):
Looks good to me, code-wise.

fzzzy:
Thanks @lmorchard. I'd like @johngruen to take a look at this before it lands.

johngruen:
checking

johngruen:
@fzzzy this is erroring when it try to build it:

```
/Users/jgruen/Repos/testpilot/frontend/src/app/components/View.js:166
      var hasAddonManager = typeof navigator.mozAddonManager !== 'undefined';
                                   ^

ReferenceError: navigator is not defined
```

johngruen:
@fzzzy, looks fine to me, but you need to rebase before merging. Taking out the loaders seems okay, but let's see what it looks like when the site's not running off my local machine.

johngruen (approved):

PR #2811, opened 2017-08-22
Tweaks to improve Storybook deploys

### LINKED ISSUE #2784: Get a storybook deployment service working
[Storybook works in a local dev environment.](https://github.com/mozilla/testpilot/issues/2783) 

But, it would be nice to have a site somewhere with auto-built deployments of the storybook for pull requests and commits. That way, we can just paste an URL into a PR to demo the UI. Even better, have a github bot auto-comment when a storybook deployment happens.

At the SF work week, I wrote [a quick & dirty storybook deployment script that runs on CircleCI](https://github.com/mozilla/testpilot/blob/master/bin/deploy-storybook.sh). But, there are problems:

1. If we restrict storybook deployments to branches on the Mozilla repo, that means only contributors with access to create branches on the Mozilla repo can cause Storybook deployments. But, that might be a security benefit, because we don't really want arbitrary contributors to be able to get arbitrary code deployed to a Mozilla server (even on dev.mozaws.net) <strike>The script only works for CircleCI builds that use mine or Mozilla's CircleCI accounts. It requires credentials to upload to an S3 bucket, and only those two accounts have them. But, many PRs use CircleCI accounts that belong to other members of the dev team - and those accounts are missing credentials, causing the deployment to fail.</strike>

    * Maybe we can fix this by granting IAM accounts for core dev team members write access to the bucket and making sure those CircleCI accounts have credentials? 

    * There is also a possible security issue: Do we really want arbitrary PRs publishing to a website? It's at a mozaws.net domain, so maybe not that big a problem.

1. Moving to a different bucket & URL seems to have fixed this: <strike>The non-SSL URLs seem not to work in Firefox. They do work in Google Chrome. No idea why - maybe Firefox doesn't like that this is a subdomain of a site that is usually SSL?</strike> 

1. Doubled down on the hacks, which seems to have worked. <strike>The stories try to re-use the static assets and CSS from the local Test Pilot site. The CSS for experiments assumes the site is at the root of the domain. However, I want URLs at sub-paths like `/{git hash}/` and `/pull/2765/`. [I have some hacks to try to address this](https://github.com/mozilla/testpilot/blob/master/bin/deploy-storybook.sh#L16), but they're not great.</strike>

lmorchard:
- Rename the bucket for storybook deploys - fixes SSL/non-SSL problem

- Replace absolute paths with relative paths for experiment images in CSS

Fixes #2784

### DIFF
--- a/bin/deploy-storybook.sh
+++ b/bin/deploy-storybook.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-STORYBOOK_BUCKET=${STORYBOOK_BUCKET:-storybook.testpilot.dev.mozaws.net}
+STORYBOOK_BUCKET=${STORYBOOK_BUCKET:-testpilot-storybook.dev.mozaws.net}
 if [ -z "$STORYBOOK_BUCKET" ]; then
   echo "The S3 bucket is not set. Failing."
   exit 1;
@@ -13,10 +13,18 @@ if [[ $IS_PONTOON ]]; then
   exit 0;
 fi
 
+# Skip builds for non-Mozilla users
+if [ "$CIRCLE_PROJECT_USERNAME" != "mozilla" ]; then
+  echo "Skipping Storybook deploy for non-Mozilla CircleCI";
+  exit 0;
+fi
+
 # HACK: Build static storybook with URL paths edited to include git hash
 HASH=$(git --no-pager log --format=format:"%H" -1)
 cp -r .storybook .storybook-$HASH
 sed -i "s#href=\"/static#href=\"/$HASH/static#g" .storybook-$HASH/preview-head.html
+# HACK: correct paths to experiment images in existing build
+sed -i "s#url(/static/#url(../#g" ./frontend/build/static/styles/experiments.css
 ./node_modules/.bin/build-storybook -c .storybook-$HASH -o ./frontend/storybook/$HASH
 rm -rf .storybook-$HASH
 cp -r ./frontend/build/static ./frontend/storybook/$HASH
@@ -33,6 +41,7 @@ echo "Deployed Storybook to $STORYBOOK_URL"
 # Deploy a client-side redirect page for a per-PR URL
 if [[ ! -z $CI_PULL_REQUEST ]]; then
   STORYBOOK_PR_PATH=$(echo $CI_PULL_REQUEST | cut -d/ -f6-7);
+  STORYBOOK_PR_NUMBER=$(echo $CI_PULL_REQUEST | cut -d/ -f7);
   STORYBOOK_PR_URL=http://${STORYBOOK_BUCKET}/${STORYBOOK_PR_PATH}/index.html
   REDIRECT_HTML=$(cat <<EOF
 <!DOCTYPE html>
@@ -52,4 +61,10 @@ EOF
       s3://${STORYBOOK_BUCKET}/${STORYBOOK_PR_PATH}/index.html
 
   echo "Deployed Storybook PR redirect to ${STORYBOOK_PR_URL}"
+
+  curl \
+    -H'Content-Type: application/json' \
+    -H"Authorization: token $STORYBOOK_GITHUB_ACCESS_TOKEN" \
+    --data "{\"body\":\"# Storybook Deployment\\nPull Request: $STORYBOOK_PR_URL\\nCommit: $STORYBOOK_URL\"}" \
+    https://api.github.com/repos/mozilla/testpilot/issues/${STORYBOOK_PR_NUMBER}/comments
 fi


lmorchard:
Okay, using this PR as both a PR and a proof-of-concept. Here's what it deployed:

http://testpilot-storybook.dev.mozaws.net/pull/2811/

This seems to fix 2 of 3 issues in #2784 - and I think I've come around to thinking the 3rd issue is actually not an issue to be fixed. 

That is, requiring PRs to use mozilla repo branches might be a security-enhancing feature and not a bug. It stops outside contributors from triggering Storybook deploys - which is bad for contribution, but good for preventing potentially malicious folks from making our systems deploy bad things.

lmorchard:
As a follow-up, I'm investigating if there's a simple way to use the GitHub API in this deployment script to auto-post a comment to a PR with the deployed Storybook URL. Would like to make sure it only does it once, so as not to spam a PR for every time the PR is updated.

lmorchard:
Testing 1 2 3 :+1:

lmorchard:
Storybook deployed:
http://testpilot-storybook.dev.mozaws.net/pull/2811/index.html
http://testpilot-storybook.dev.mozaws.net/6ae6f66d10d196c07222cc8e0b209c0d8fa47b65/

lmorchard:
[Storybook bot] Deployed to S3
PR: http://testpilot-storybook.dev.mozaws.net/pull/2811/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/6ae6f66d10d196c07222cc8e0b209c0d8fa47b65/

lmorchard:
I'll probably regret this, but I just created a limited-scope personal access token and figured out how to make the script post PR comments on Storybook deploy

lmorchard:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/2811/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/cbe0de7f2619e913b72c7a61d7160ca25a28a78f/

lmorchard:
...and there's the automated comment - confusingly enough, posted from my account because I used my personal access token.

bin/deploy-storybook.sh
```diff
@@ -52,4 +55,10 @@ EOF
       s3://${STORYBOOK_BUCKET}/${STORYBOOK_PR_PATH}/index.html
 
   echo "Deployed Storybook PR redirect to ${STORYBOOK_PR_URL}"
+
+  curl \
+    -H'Content-Type: application/json' \
+    -H"Authorization: token $STORYBOOK_GITHUB_ACCESS_TOKEN" \
```
lmorchard, line 67: FWIW, this access token is configured in the CircleCI settings and shouldn't be visible to public

lmorchard:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/2811/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/aa927564f0e522b436bb8848627e35734f5d8563/

lmorchard:
Alright, one last tweak to restrict Storybook deploys to the main mozilla repo. Still posts a comment for every commit, but that might not be a bad thing since we can view each commit's deployment to see compare changes

fzzzy:
I think restricting deploys to users that have access is the right thing to do.

bin/deploy-storybook.sh
```diff
@@ -13,10 +13,18 @@ if [[ $IS_PONTOON ]]; then
...
+  echo "Skipping Storybook deploy for non-Mozilla CircleCI";
+  exit 0;
+fi
+
 # HACK: Build static storybook with URL paths edited to include git hash
 HASH=$(git --no-pager log --format=format:"%H" -1)
 cp -r .storybook .storybook-$HASH
 sed -i "s#href=\"/static#href=\"/$HASH/static#g" .storybook-$HASH/preview-head.html
+# HACK: correct paths to experiment images in existing build
+sed -i "s#url(/static/#url(../#g" ./frontend/build/static/styles/experiments.css
```
fzzzy, line 27: Is it really not possible to make the css use a relative path?

lmorchard: Tried it, but it broke gulp-rev-all hashing for production deploys - [so we reverted it](https://github.com/mozilla/testpilot/commit/17470d6520d197fb14c3d35b510ebb94a37a9a3d). Figured it was easier to hack it here than try to change that again

fzzzy: Oh, I remember that...

meandavejustice:
@lmorchard This is great work! I think the restriction makes sense for us. If a contributor becomes very regularly committing to our frontend, and we wanna figure out a way to give them access/make that work we can figure it out.

lmorchard:
> If a contributor becomes very regularly committing to our frontend, and we wanna figure out a way to give them access/make that work we can figure it out.

Yeah, that's a good point... we can also "sponsor" external feature branches into mozilla repo branches as needed, if a contributor has a pull request that needs storybook support

Issue #2737, opened 2017-08-07
News feed improvements


johngruen:
The news feed could use a few optimizations:
1. Items should be hidden by default after two weeks.
2. All Items should be in reverse chronological order (currently, they are ordered by experiment)
3. There should be an option to load older posts (see secreenshot), clicking this should load the five previous news items below
4. If a user has loaded all historical news items the load more link should disappear.
5. If a news item is clickable, it should include a GA ping `ec: ExperimentsPage Interactions, ea: click, el: news-item-{id}`

SoftVision-PaulOiegas:
I was wondering when we'll get the enhanced version of the News feed. Glad to see this back on the table. 👍

lmorchard:
> All Items should be in reverse chronological order (currently, they are ordered by experiment)

[They *should* be currently in reverse chronological order](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/selectors/news.js#L25) ([even has a unit test](https://github.com/mozilla/testpilot/blob/master/frontend/test/app/selectors/news-test.js#L57)) - where are you seeing them ordered by experiment? Wondering what this implementation in missing to make that happen

lmorchard:
> There should be an option to load older posts (see secreenshot)

Do you have this screenshot?

lmorchard:
> Items should be hidden by default after two weeks.

What should things look like if there's no news to be shown? (i.e. everything's older than 2 weeks) Our current component just hides itself and the "Latest News" header if there's no news to be shown

SoftVision-PaulOiegas:
@johngruen The "Show Past News" button should be displayed even if there are no news in the last 2 weeks? I'm asking because right now, the last post has reached the 2 weeks time, it disappeared, but there is no button displayed.

To be honest, it would be awkward to have no news and a button hanging to nothing there. We don't have a permanent section for the news to include it... But again, having the button only when there is at least a single news (2 weeks), would be also awkward.
Maybe we should think about this before a stage/production deploy.

PR #2815, opened 2017-08-24
News feed follow-up tweaks

lmorchard:
Issue #2737


flodolo (approved):

lmorchard:
FWIW, now that we have split the l10n branch off from master, merging this PR does *not* immediately make strings appear in Pontoon. We'll need to merge from master to l10n to make that happen, ideally after we've stabilized string changes for a bit

pdehaan:
Do we have any experiments with live news items?

lmorchard:
> Do we have any experiments with live news items?

Not yet, no

Issue #2473, opened 2017-05-25
Experiments images are not displayed on Facebook and Activity Stream one is not displayed on Twitter


SoftVision-PaulOiegas:
**[Affected versions]:**
- Firefox 53.0.3 and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- A Facebook and a Twitter profile was already logged on the used profile. 

**[Steps to reproduce]:**
1. Open browser with the profile from prerequisites and navigate to https://testpilot.firefox.com page.
2. Click on one of the active experiments cards and after redirected, copy the page URL.
3. In two new tabs, navigate to "https://www.facebook.com" and "https://twitter.com" pages.
4. Paste the copied URL in the Facebook message box and "Post" it.
5. Paste the same URL in the Twitter message box and "Tweet" it.
6. Observe both the posted messages on the social platforms.

**[Expected result]:**
- Both the social platforms are displaying an image specific to the experiment in the message.

**[Actual result]:**
- Twitter correctly shows the experiment image (except Activity Stream), while Facebook displays the default Test Pilot logo image.

**[Notes]**
- This issue occurs for all the experiments URL's in Facebook posts, and even if Test Pilot is installed or not.
- Activity Stream URL is not displaying an image on Twitter message. 
- Here is a screen recording of the issue: https://goo.gl/jKsOEI.

johngruen:
Jeez, how do es this stuff keep breaking when:

* we're using a common template and
* we haven't updated experiment yaml in months

Computers are a pain, let's all move upstate and start a farm.

lmorchard:
Think I *might* have figured out the Facebook issue in #2821 - we'll see

As for images not showing up for some experiments on Twitter, while they show up for others: I think I've worked out this one also. 

[We *did* introduce a fix for this about 6 months ago](https://github.com/mozilla/testpilot/commit/1cc53082d333cb8b82049161c8fcf1a5142926d7) - but it requires that all of our images used for social media live in a folder path like this:

`frontend/src/images/experiments/voice-fill/social/`

However, Activity Stream uses images in this path:

`frontend/src/images/experiments/activity-stream/experiments_experiment/e/1/`

The key is the `experiments/{experiment name}/social/` path pattern, which is disabled for cache-busting filename hashing. Otherwise, the URL used in the meta headers for social media links to broken URLs, because those URLs don't get seen by gulp-rev-all, which does the filename hashing

[...]

PR #2821, opened 2017-08-25
Social media thumbnail fixes

lmorchard:
Issue #2473


lmorchard:
Not sure if this will actually fix anything, but I did notice two things [on Facebook's Open Graph debugger](https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Ftestpilot.firefox.com%2Fexperiments%2Fvoice-fill):

* A warning that we were missing og:type
* A "redirect path" that appeared to treat og:url as canonical and bounced over to the home page for metadata

capture

lmorchard:
Think I also figured out why thumbnails haven't been showing up for Activity Stream and No More 404s

PR #2828, opened 2017-08-29
Revert "Added install and enable tests - part 2"

fzzzy:
Reverts mozilla/testpilot#2790

Although the tests pass, this fails to deploy on dev because the `dist` directory is not there -- I can't understand what could cause this. Is it because it switches from running `npm run server` to `npm start`? Can't think of why that would cause problems or what else could have broken it.


b4handjr:
Before the revert did you try a build without cache? If you see my original PR I state I had some weird issues that was caused my caching.

fzzzy:
I did try a rebuild without cache; it did not help. I did not yet land the revert, but I opened this pr with the revert patch in case that is the direction we need to go.

lmorchard:
> Is it because it switches from running npm run server to npm start? Can't think of why that would cause problems or what else could have broken it.

`npm run server` and `npm start` are not interchangeable

[`npm run server`](https://github.com/mozilla/testpilot/blob/master/package.json#L121) just launches [a server](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/server.js) to host [the content built](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-frontend.sh) earlier [in the dependencies phase](https://github.com/mozilla/testpilot/blob/master/circle.yml#L33) of the CircleCI run. That includes the `dist` directory, which deployment expects to deploy.

[`npm start`](https://github.com/mozilla/testpilot/blob/master/package.json#L120) is meant for local development only. It does *not* build the `dist` directory for production deployment. Furthermore, part of [what it does](https://github.com/mozilla/testpilot/blob/master/gulpfile.babel.js#L55) is [run `gulp clean`](https://github.com/mozilla/testpilot/blob/master/gulpfile.babel.js#L55), which [deletes `dist`](https://github.com/mozilla/testpilot/blob/master/gulpfile.babel.js#L23). Thus, undoing the production build done earlier in the CircleCI run.

So, that would be why `dist` is missing

fzzzy:
@lmorchard That explains it. Thanks.

b4handjr:
@lmorchard what should be done? I need to run ```npm start``` to allow the dev experiment to show up for testing.

fzzzy:
Since no quick solution is forthcoming, I'm going to go ahead and merge this revert.

b4handjr:
Well since no one proposed a solution what step needs to be done from the dependencies stage to build the correct directory? I'm pretty sure I tagged both of you on the review of this patch.

fzzzy:
@jrbenny35 I think the correct thing is to change npm run server so that it shows the dev example, but I'm not 100% sure.

b4handjr:
If the dist directory is there I can just copy it and restore it.

lmorchard:
> what should be done? I need to run npm start to allow the dev experiment to show up for testing.

Sorry, missed the original PR on this... What [npm start](https://github.com/mozilla/testpilot/blob/master/package.json#L120) does to get dev content to show up is set these env vars: `ENABLE_DEV_LOCALES=1 ENABLE_DEV_CONTENT=1`

[`build-frontend.sh` does it, too](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-frontend.sh#L3), but only for `master` branch commits. I can't remember why, but [`git blame` tells me we did that as part of introducing the dev-only experiment](https://github.com/mozilla/testpilot/commit/cc87033ae3b8948a12037d5ddf8adde882972833).

Maybe change `build-frontend.sh` to remove the `master` branch conditional and enable those vars for all branches?

b4handjr:
@lmorchard Thanks for that information. 

Based on what you are saying. ```build-frontend.sh``` should be run instead of ```npm start``` ? Or, run this: ```NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 ENABLE_DEV_LOCALES=1 npm run static
else``` ?

lmorchard:
No, you should run `npm run server` - `build-frontend.sh` already happens early on in the CircleCI run. I'm saying modify `build-frontend.sh` to remove the if/else so that it does for all branches what it only does for `master` right now. So, basically:
```bash
#!/bin/bash
set -ex
NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 ENABLE_DEV_LOCALES=1 npm run static
zip -r $CIRCLE_ARTIFACTS/frontend.zip frontend/build
```

b4handjr:
Right I understand that much, but I am lost at what exactly that will accomplish.

lmorchard:
> Right I understand that much, but I am lost at what exactly that will accomplish.

The `ENABLE_DEV_CONTENT=1` makes it build the dev experiment content

b4handjr:
So run ```ENABLE_DEV_CONTENT=1 npm run server``` ?

lmorchard:
No. Modify `build-frontend.sh` like I pasted above so it runs with ENABLE_DEV_CONTENT for all branches. `npm run server` doesn't build any content - that happens early in the dependencies phase with `build-frontend.sh`

`run-integration-test.sh` should keep `npm run server &` like it originally had

b4handjr:
@lmorchard I see some problems here. 

The way I enable the dev experiment is here: https://github.com/mozilla/testpilot/pull/2828/files#diff-be3d7275aad82e2ae7e96ff8a4c00673L30

Maybe I could do it another way by enabling it all the time. However I thought that this would get deployed as being enabled all the time. Is that true or no. If not, I can simple re-enable the dev experiment permanently.

b4handjr:
If you look at the dev-experiment.yaml it has a completed and an uninstalled property.  With these enabled I can not get the experiment to show during the ```npm run static``` command.

https://github.com/mozilla/testpilot/blob/master/content-src/experiments/dev-example.yaml

b4handjr:
@lmorchard I could use your input on my thoughts.

lmorchard:
> Maybe I could do it another way by enabling it all the time. However I thought that this would get deployed as being enabled all the time. Is that true or no. If not, I can simple re-enable the dev experiment permanently.

It will get deployed - but only to the dev server. I think enabling the dev-experiment is fine

PR #2836, opened 2017-09-01
Use webpack instead of Browserify for building app JS bundles

lmorchard:
So, this is nifty / terrifying because it:
1. completely replaces browserify
1. still gives us the app / vendor bundle split
1. reduces the size of both bundles
1. gives us faster rebuilds on file changes
1. was way quicker to throw together than I expected
1. seems to work without issue
1. even improves the usefulness of source maps

To try local development: `npm start`
To try production deployment: `NODE_ENV=production npm run static && npm run server`


lmorchard:
@fzzzy So, I expected this to take way longer and be way more complex than it ended up being. Seems to get an equivalent bundle shrink as what you've been seeing with closure compiler. Can't see any issues with it, so I assume I've missed something

fzzzy:
This branch cuts the Lighthouse time to interactive from 22 seconds to 14.5 seconds, which is good enough for our OKR this quarter 🎉

I still think we should work on performance a bit, but this helps a lot!

fzzzy:
The only question I have about this branch is why we would want to keep the app/vendor split. Since webpack does an incremental build it wouldn't be a problem to have the whole thing in one bundle. Since we don't update our dependencies that often any more and do it manually, the vendor bundle theoretically might not change in every release. If we could figure out how to calculate the hash based on the content of the bundle instead of the commit, we might be able to reuse the vendor bundle across releases. So maybe we should aim to do that at some point, and keep the app/vendor split.

fzzzy (approved):
Fantastic!

lmorchard:
> Since we don't update our dependencies that often any more and do it manually, the vendor bundle theoretically might not change in every release. If we could figure out how to calculate the hash based on the content of the bundle instead of the commit, we might be able to reuse the vendor bundle across releases. So maybe we should aim to do that at some point, and keep the app/vendor split.

I think gulp-rev-all *does* calculate the hash based on content. So, that's why I carried the split over. What *should* happen is that deployments without vendor bundle changes should leave it cached in cloudfront and users' browsers.

PR #2842, opened 2017-09-05
Added install and enable tests Part 2

b4handjr:
This should fix the issue with ```npm start``` deleting the ```dist``` dir.


bin/circleci/run-integration-tests.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
+NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 ENABLE_DEV_LOCALES=1 npm run static
```
lmorchard: Don't run this here - run it in [`bin/circleci/build-frontend.sh`](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-frontend.sh) by removing [the conditional](https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-frontend.sh#L3) so this runs on all branches and not just master. We only want to build the site once overall, not once for every integration test variant.

lmorchard (changes requested):

bin/circleci/run-integration-tests.sh
lmorchard: You can basically copy & paste what I have [over in this comment from the previous PR](https://github.com/mozilla/testpilot/pull/2828#issuecomment-326405418) into `bin/circleci/build-frontend.sh`

b4handjr: Well I would like this to work locally, if a user wants to run tests against the site. What do you think?

lmorchard: You can run the same `npm run static` command from build-frontend locally before running this. That, or `npm start` like we usually do for local development.

lmorchard: I also just realized that this `run-integration-tests` script doesn't actually run integration tests. It just starts a server.

But it doesn't stop an existing server while being re-run [in tox.ini](https://github.com/mozilla/testpilot/pull/2842/files#diff-b91f3d5bd63fcd17221b267e851608e8R11) for every test run. That probably means there's an error starting up the server a 2nd or 3rd time since the port's already taken. (edit: Actually, looking at circleci, I don't see an error. So, I wonder if Tox kills all the processes between runs)

We could probably move this script out of tox.ini and just run it once in circle.yml before the test-firefox-*.sh scripts - and maybe rename it to `start-server.sh` or something

b4handjr: Will the server be accessible even throughout different subprocesses? With the problems we had with circle in the beginning, it is nice to keep everything within the same shell process, as it seems they are pretty aggressive when it comes to killing things running in different shells.

lmorchard: It's in a subprocess now - that's what the `&` in `npm run server &` does. Trying to remember what the issue we had in circleci earlier was... I want to say it was something other than the server process getting killed but I can't recall :/ (Because now that I think about it, that sounds super familiar)

b4handjr:
So I had to keep the ```npm run server &``` command in a separate sh file as starting the server from the ```circle.yml``` did not work.

jrbenny35:
So I had to keep the ```npm run server &``` command in a separate sh file as starting the server from the ```circle.yml``` did not work.

lmorchard:
> So I had to keep the npm run server & command in a separate sh file as starting the server from the circle.yml did not work.

Does running that `start-server.sh` script from circle.yml work? FWIW, most commands should be going into scripts rather than directly into circle.yml. CircleCI starts a fresh new shell for every line in the YAML config and it disallows some things like `&` for background scripts.

lmorchard (approved):

tox.ini
```diff
@@ -3,11 +3,13 @@ envlist = ui-tests, flake8
...
 [testenv:ui-tests]
-passenv = DISPLAY PYTEST_ADDOPTS
+passenv = DISPLAY PYTEST_ADDOPTS MOZ_HEADLESS SKIP_INSTALL_TEST
 basepython = python3.6
 deps = -rfrontend/test/ui/requirements/requirements.txt
+whitelist_externals =
+    bash
 commands =
-    bash bin/circleci/run-integration-tests.sh
+    bash bin/circleci/start-server.sh
```
lmorchard, line 12: Still would kind of like to see this run from circle.yml instead of tox.ini, but it might not work. Might be fine to merge this as-is and tinker with that nit-pick later

b4handjr: Me too, but when I tried it, I got a requests error, meaning the server was not found.

lmorchard: Ah, yeah, probably circleci killing the old shell where it was in the background - whereas tox is all in the same shell.

PR #2849, opened 2017-09-08
Wait for Cloudfront invalidation to complete after upload to S3 bucket in dev deployment

lmorchard:
Still seeing back-to-back dev deployments end up breaking the site on dev. 

This is even with [the `do-exclusively.sh` wrapper](https://github.com/mozilla/testpilot/commit/5bb405f97f8b2f505fb64235f301484b31f1202f) to serialize deployments to S3. That makes me think we need to bite the bullet and also wait for the CloudFront invalidation to complete before attempting the next deployment. This PR should do that.

### DIFF
--- a/bin/circleci/invalidate-cloudfront-cache.sh
+++ b/bin/circleci/invalidate-cloudfront-cache.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -ex
+
+DISTRIBUTION_ID=$1
+
+aws configure set preview.cloudfront true
+
+INVALIDATION_ID=$(aws cloudfront create-invalidation \
+  --distribution-id $DISTRIBUTION_ID \
+  --paths '/*' | jq -r '.Invalidation.Id');
+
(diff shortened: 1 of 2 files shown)


bin/circleci/invalidate-cloudfront-cache.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -ex
+
+DISTRIBUTION_ID=$1
+
+aws configure set preview.cloudfront true
+
+INVALIDATION_ID=$(aws cloudfront create-invalidation \
+  --distribution-id $DISTRIBUTION_ID \
+  --paths '/*' | jq -r '.Invalidation.Id');
```
fzzzy, line 10: Just for my own understanding, I looked up what this jq command does:

jq is 'a lightweight and flexible command-line JSON processor.'

Given input {"Invalidation": {"Id": "foo"}} jq '.Invalidation.Id returns "foo". Adding the -r makes it return foo without the quotes.

PR #2851, opened 2017-09-08
Add option to use webpack-bundle-analyzer in JS build

lmorchard:
With this PR, you can run the analyzer on port 9888 in two ways:

1. `ANALYZER=1 npm start` - this gives you the full site build & usual dev server along with the analyzer served up on the side at 9888
1. `npm run analyzer` - this just builds the JS and serves up the analysis on 9888, a bit faster than full site build but you can't actually see the site

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "uglifyjs-webpack-plugin": "^0.4.6",
     "vinyl-source-stream": "1.1.0",
     "webpack": "^3.5.5",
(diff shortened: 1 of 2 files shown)


lmorchard:
Here's what it should look like, making me wonder if we can somehow jettison most of moment.js for a start:

capture

lmorchard:
Actually, forget this PR - I just merged it into #2852

PR #2852, opened 2017-09-08
More tweaks to reduce JS bundle size

lmorchard:
Figured out that [the seedrandom module optionally requires node's crypto module](https://github.com/davidbau/seedrandom/blob/released/seedrandom.js#L237) for a feature that we don't seem to use. However, that seems to prompt Webpack to pull everything in just in case.

Turning off the node-browser polyfills appears to save us ~309kb in the vendor bundle without breaking anything.

Before:
```
[14:13:26] Version: webpack 3.5.6
                   Asset     Size  Chunks                    Chunk Names
       static/app/app.js   256 kB       0  [emitted]  [big]  static/app/app.js
    static/app/vendor.js  1.07 MB       1  [emitted]  [big]  static/app/vendor.js
   static/app/app.js.map   960 kB       0  [emitted]         static/app/app.js
static/app/vendor.js.map  6.82 MB       1  [emitted]         static/app/vendor.js
```
After:
```
[14:17:56] Version: webpack 3.5.6
                   Asset     Size  Chunks                    Chunk Names
       static/app/app.js   256 kB       0  [emitted]  [big]  static/app/app.js
    static/app/vendor.js   761 kB       1  [emitted]  [big]  static/app/vendor.js
   static/app/app.js.map   959 kB       0  [emitted]         static/app/app.js
static/app/vendor.js.map  5.11 MB       1  [emitted]         static/app/vendor.js

WARNING in ./node_modules/seedrandom/seedrandom.js
Module not found: Error: Can't resolve 'crypto' in '/mnt/c/Users/lmorchard/devel/testpilot/node_modules/seedrandom'
 @ ./node_modules/seedrandom/seedrandom.js 232:17-34
 @ ./node_modules/seedrandom/index.js
 @ multi classnames clipboard domready es-symbol es6-promise html-react-parser js-cookie micro-email-validator moment raven-js react react-dom react-markdown react-redux redux redux-logger redux-promise reselect seedrandom tos-pp babel-polyfill/browser fluent/compat fluent-langneg/compat fluent-react/compat cldr-core/supplemental/likelySubtags.json html-react-parser/lib/dom-to-react react/lib/ReactDOMFactories querystring
```


lmorchard:
Tweaked to be more specific to just node crypto, which adds a couple of kb, since something somewhere apparently uses *some* node polyfill:
```
[14:23:39] Version: webpack 3.5.6
                   Asset     Size  Chunks                    Chunk Names
       static/app/app.js   256 kB       0  [emitted]  [big]  static/app/app.js
    static/app/vendor.js   763 kB       1  [emitted]  [big]  static/app/vendor.js
   static/app/app.js.map   959 kB       0  [emitted]         static/app/app.js
static/app/vendor.js.map  5.13 MB       1  [emitted]         static/app/vendor.js
```

lmorchard:
We get a warning for the non-resolvable `require` in seedrandom, but that's wrapped in a try/catch anyway. Poking around to see if there's a way to suppress that warning, but probably good to go as-in on this PR.

lmorchard:
Oh and hey, while I'm in here: Figured out that `react-markdown` (and thus `commonmark`) isn't used anywhere client-side, only included in the vendor bundle because it's in the dependencies list. Also, `redux-logger` isn't used anywhere at all anymore.

lmorchard:
Okay, one more thing since I got on a roll: Figured out how to pare down the locale modules included from moment.js to just the ones we use for AVAILABLE_LANGUAGES.

Also just went ahead and pulled #2851 into this PR, since the moment.js change would have conflicted and I think we want the analyzer anyway

lmorchard:
All-in-all, I think I got the the bundles down to under 1MB:
```
Version: webpack 3.5.6
Time: 17615ms
                   Asset     Size  Chunks                    Chunk Names
       static/app/app.js   262 kB       0  [emitted]  [big]  static/app/app.js
    static/app/vendor.js   559 kB       1  [emitted]  [big]  static/app/vendor.js
   static/app/app.js.map   971 kB       0  [emitted]         static/app/app.js
static/app/vendor.js.map  3.71 MB       1  [emitted]         static/app/vendor.js
```

fzzzy:
Very nice. I'm ok with removing the variant test code because we never actually used it for anything, and it should probably be re-thought if we ever do want to do some ab testing.

lmorchard:
Okay, seriously, last tweak - looks like we only use one small polyfill from babel for async/await. At least, it seems like nothing breaks when I reduce it down to that. If something does break, we can revert that part easily. But that seems to get another big reduction in size:
```
[15:41:01] Version: webpack 3.5.6
                   Asset     Size  Chunks                    Chunk Names
       static/app/app.js   256 kB       0  [emitted]  [big]  static/app/app.js
    static/app/vendor.js   480 kB       1  [emitted]  [big]  static/app/vendor.js
   static/app/app.js.map   959 kB       0  [emitted]         static/app/app.js
static/app/vendor.js.map  3.67 MB       1  [emitted]         static/app/vendor.js
```

webpack.config.js
```diff
@@ -62,6 +62,9 @@ module.exports = {
     path: path.resolve(__dirname, 'frontend/build'),
     filename: '[name]'
   },
+  node: {
+    crypto: false
```
fzzzy, line 71: ♥️

meandavejustice:
This is great work @lmorchard thanks for being so thorough!

And of course @fzzzy for making so much progress on this already

fzzzy:
Been testing this a lot and it is super duper fast. Awesome!

fzzzy (changes requested):
There are a number of warnings in the console that I'd like to see if we can get rid of before this lands.

Warning: It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.

You are currently using minified code outside of NODE_ENV === 'production'. This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or DefinePlugin for webpack (http://stackoverflow.com/questions/30030031) to ensure you have the correct code for your production build.

 Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client)  href="/experiments/containers" class="e
 (server)  href="/experiments/min-vid" class="expe

The first two are obviously the same thing, shipping a minified build outside prod. The last one I am ok with shipping because usually there are changes in state between build time and render time that are sometimes hard to get rid of. I'll look into the last one, @lmorchard can you look into the minification warnings?

lmorchard:
> The first two are obviously the same thing, shipping a minified build outside prod. 

I figured the minification warnings are fine, because the source maps work well enough. We can turn off minification in local dev, which would make those go away. But, I liked having it closer to prod as long as it didn't drag down dev rebuild times

> The last one I am ok with shipping because usually there are changes in state between build time and render time that are sometimes hard to get rid of. I'll look into the last one, @lmorchard can you look into the minification warnings?

The state warnings have been around awhile, off & on, not really related to the changes in this PR

fzzzy (approved):
Ok, that's great news about the sourcemaps! I wonder if there is a way to squelch those warnings.

PR #2857, opened 2017-09-12
Turn off minification in development mode

lmorchard:
Easiest way to squelch the minification warnings in React is to just not minify on dev. 🤷‍♂️


webpack.config.js
```diff
@@ -35,7 +37,7 @@ const vendorModules = Object.keys(packageJSON.dependencies)
 
 const plugins = [
   new webpack.DefinePlugin({
-    'process.env.NODE_ENV': `"${process.env.NODE_ENV || 'dev'}"`,
```
lmorchard, line 38: Also, this was a bad default. We do use "development" as the value and not "dev"

fzzzy (approved):

Issue #2808, opened 2017-08-22
Rework file structure so that scss files live with their respective components/containers


johngruen:
We should divinde our Sass so that sass is applied to individual components. Recommend we use a structure similar to @chuckharmston's implementation on the contrib site.

fzzzy:
Let's kick this to 34, it is too big to try to squeeze into what is left of this sprint.

johngruen:
Added the Photonization label since this seems related.

Issue #2807, opened 2017-08-21
Break down ExperimentDetails container into (mostly) simple components


johngruen:
In order to support the ongoing storybook work on the site, we should break down `ExperimentPage` Container into simple components.

lmorchard:
PR #2990 has been merged, so let's call this one finally done and open new issues for anything further that needs doing

PR #2859, opened 2017-09-14
Initial work toward new component bundle directories

lmorchard:
Some remaining TODOs:
- [x] Tests for new IncompatibleAddons component
- [x] Fix / move a broken ExperimentPage test based on IncompatibleAddons render
- [x] Tweak eslint rules for test modules using devDependencies
- [x] #2808 <strike>Actually migrate over some styles for the new bundles components</strike>
- [x] Clean up some commented out bits
- [x] Fix flow type validation failures
- [x] Docs based on https://gist.github.com/chuckharmston/01e5b7a16ba9b771863cbb8b2de64138

Done so far:

- Moved ExperimentRowCard and UpdateList components into subdirectories

- Extracted IncompatibleAddons as a component from ExperimentPage

- Webpack config to extract static/app/app.js.css from component imports

- Link to static/app/app.js.css styles from the static page template

- Storybook config changes to use frontend/src/app/**/{ComponentName}/stories.jsx files

- Mocha test command changes use frontend/src/app/**/{ComponentName}/tests.js

- Hacks to frontend/tasks/pages.js to ignore scss imports

- Hacks to test-setup.js to ignore scss imports

See also:
https://gist.github.com/chuckharmston/01e5b7a16ba9b771863cbb8b2de64138

Issue #2807


lmorchard:
@fzzzy - Still got some work to do here, but I think this is doing the right things so far with styles, stories, & tests.

fzzzy:
What on earth is going on in this failure here? o.O

frontend/src/app/components/IncompatibleAddons/webpack.config.js
```diff
@@ -0,0 +1,13 @@
+const path = require('path');
+
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.(css|scss)$/,
+        loaders: ["style-loader", "css-loader", "sass-loader"],
```
fzzzy: The sass-loader docs gave something very different than this. I never would have figured this out...

lmorchard: Ignore that... it didn't work and shouldn't have been there - I was editing the wrong file in the wrong dir & accidentally committed it

lmorchard:
> What on earth is going on in this failure here? o.O

Looks like the test modules are getting caught up in the <strike>webpack bundling</strike> JS linting with browser rules. But, I've seen that before and can fix

frontend/src/app/components/IncompatibleAddons/webpack.config.js
lmorchard: FWIW, these are the changes to the main webpack config, basically from the sass-loader docs: https://github.com/mozilla/testpilot/pull/2859/files#diff-11e9f7f953edc64ba14b0cc350ae7b9d

fzzzy:
Looking at the diff, I'm guessing the hacks you have in test-setup were what I needed to fix the weird problem I had trying to load sass earlier today.

I like the look of the way webpack is set up here better than what I had, too.

frontend/src/app/components/IncompatibleAddons/webpack.config.js
fzzzy: Yeah that changeset looks really good. I think I can actually understand how it works!

lmorchard:
Okay. Reorg'd this into a few more self-contained commits and cleaned some things up. And yay the tests pass!

<strike>Thinking I might punt on actually migrating styles over for now, though: It's going to take a bit more SCSS hacking than just moving the style files</strike>

Looks like we have #2808 for porting over SCSS - I'll say this PR *enables* index.scss per component directory, but we still have to rework the existing SCSS a bit to *use* the capability.

fzzzy:
I think this is ready to land. I would be fine with just linking to the gist in the documentation for now.

fzzzy (approved):
This is going to be very nice.

lmorchard:
This could use some revision, [but I just copypasta'd most of that proposal into the Storybook docs](https://github.com/lmorchard/testpilot/blob/e443d7fe4aecad48153705bc157707033d4d803e/docs/development/storybook.md). Might just call that good enough and merge after tests go green

fzzzy:
Seems good enough for now!

PR #2871, opened 2017-09-19
Fix #2863 Instead of symlinking to the legal-docs repository in node_modules, use node-fetch to download the .md files into frontend/src/pages/[terms,privacy]

### LINKED ISSUE #2863: [Stage server] Terms of Use and Privacy Notice pages are wrongly displaying the Test Pilot 404 page
**[Affected versions]:**
- All browsers

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- The "testpilot.env" string pref was manually added with "stage" value on a new clean profile in "about:config" page.
- The "extensions.webapi.testing" boolean pref was manually added with "true" value on the crated profile.

**[Steps to reproduce]:**
1. Open Firefox browser with the profile from prerequisites and navigate to https://testpilot.dev.mozaws.net page.
2. Click on the "Terms of Use" and "Privacy Notice" text links from under the "Install the Test Pilot Add-on" button.

**[Expected result]:**
- The "https://testpilot.stage.mozaws.net/privacy" and "https://testpilot.stage.mozaws.net/terms" pages are correctly loaded.

**[Actual result]:**
- Both the pages are displaying the Test Pilot 404 page.

**[Notes]:**
- The same behavior occurs for the bottom install button links and for the "Footer" text links ("Privacy" and "Terms"). 
- The "data" text link from experiment details page in "Your Privacy" section is experiencing the same issue since it has the "Privacy" page URL underneath it.

fzzzy:
The way this patch accomplishes getting the .md files into the
frontend/src/pages directory is a little odd, but it does work. If a
better implementation can be devised somehow that might be good.


frontend/tasks/pages.js
```diff
@@ -32,7 +33,38 @@ gulp.task('pages-experiments', () => {
     .pipe(gulp.dest(config.DEST_PATH + 'experiments'));
 });
 
-gulp.task('pages-compiled', () => {
+gulp.task('pages-legal', () => {
+  return git.clone(
```
lmorchard: Do we need to use git in gulp here? Can't we just copy `node_modules/legal-docs/firefox_testpilot_PrivacyNotice` and `node_modules/legal-docs/firefox_testpilot_Terms` to the spots where they used to be linked from?

Or was the problem that these files didn't actually get installed by npm?

fzzzy: That won't help, because the problem is that under docker, node_modules/legal-docs doesn't exist... I suppose we could figure out why that is instead.

fzzzy: And note that it doesn't exist even though the log shows git does clone it. ¯\\\_(ツ)\_/¯

lmorchard: Might be easier to fetch just the documents from raw github URLs, e.g.:

https://raw.githubusercontent.com/mozilla/legal-docs/master/firefox_testpilot_PrivacyNotice/en-US.md
https://raw.githubusercontent.com/mozilla/legal-docs/master/firefox_testpilot_PrivacyNotice/de.md
https://raw.githubusercontent.com/mozilla/legal-docs/master/firefox_testpilot_Terms/en-US.md
https://raw.githubusercontent.com/mozilla/legal-docs/master/firefox_testpilot_Terms/de.md

Trying out the docker command on my machine, but it's going really slow on the npm install step.

lmorchard: A *little* wary to clone a git repo in the middle of the gulp build process, but don't have a good reason why. This task is also hooked up to file watchers, but looks like they watch the files that come from the git repo, so it's kind of a no-op I guess.

fzzzy: Just downloading the actual urls sounds good. I had considered that but didn't know how to do that with gulp.

fzzzy:
Let's abandon this in favor of https://github.com/mozilla/testpilot/pull/2872

PR #2874, opened 2017-09-20
Always upload the add-on, not just for dev & stage parameter

lmorchard:
Uploading the unsigned add-on to production should be harmless, since the prod signed add-on is at `/files/@testpilot-addon/latest` and dev & stage have it at `/static/addon/addon.xpi`


lmorchard:
Going to just self-merge this to expedite breaking or fixing staging

PR #2887, opened 2017-09-26
Fix markdown syntax for broken image embed

devinreams:


meandavejustice:
Thanks @devinreams :)

Issue #2791, opened 2017-08-17
Make sure all <Localized> instances have default content


fzzzy:
We used to have problems where l20n and react-router would fight if there was default content in a localized node. Now that we switched to fluent-react, we can have default content again. I tried to make sure every Localized had default content, but I missed some. For the sake of consistency, we should make sure every Localized and LocalizedHtml instance has default content.

meandavejustice:
output from `git grep -i -n -C 1 --header --break "<localized" `

(may help to widdle down the areas needing review)

```
frontend/src/app/components/EmailDialog.js
52-        <header className="modal-header-wrapper">
53:          <Localized id="emailOptInDialogTitle">
54-            <h3 className="modal-header">Welcome to Test Pilot!</h3>
--
58-        <div className="modal-content centered">
59:          <Localized id="emailOptInMessage">
60-            <p>Find out about new experiments and see test results for experiments you&apos;ve tried.</p>
--
75-        <header className="modal-header-wrapper">
76:          <Localized id="newsletterFooterSuccessHeader">
77-            <h3 className="modal-header">Thanks!</h3>
--
82-          <div className="envelope" />
83:          <Localized id="newsletterFooterSuccessBody">
84-            <p>Thank you!</p>
--
87-        <div className="modal-actions">
88:          <Localized id="emailOptInConfirmationClose">
89-            <button id="email-success-continue" onClick={e => this.continue(e)} className="button default large">On to the experiments&hellip;</button>
--
99-        <header className="modal-header-wrapper">
100:          <Localized id="emailOptInDialogErrorTitle">
101-            <h3 className="modal-header">Oh no!</h3>
--
106-          <div className="envelope" />
107:          <Localized id="newsletterFooterError">
108-            <p className="error">
--
113-        <div className="modal-actions">
114:          <Localized id="newsletterFormSubmitButton">
115-            <button id="email-success-continue" onClick={e => this.reset(e)} className="button default large">Sign Up Now</button>

frontend/src/app/components/ExperimentDisableDialog.js
29-          <header className="modal-header-wrapper">
30:            <Localized id="feedbackUninstallTitle" $title={ experiment.title }>
31-              <h3 className="modal-header" />
--
36-            <Copter small={true}/>
37:            <Localized id="feedbackUninstallCopy">
38-              <p className="centered">
--
44-          <div className="modal-actions">
45:            <Localized id="feedbackSubmitButton">
46-              <a

[...]

PR #2905, opened 2017-10-04
fix hardcoded strings

### LINKED ISSUE #2775: Two strings are hardcoded in the newsletter form
“Your email here” placeholder is hardcoded:
capture d ecran 2017-08-13 a 17 27 05

“this privacy notice” is hardcoded as well:
capture d ecran 2017-08-13 a 17 26 56

Verified on production using French.

cc @flodolo

johngruen:
Fixes #2775

### DIFF
--- a/frontend/src/app/components/MainInstallButton.js
+++ b/frontend/src/app/components/MainInstallButton.js
@@ -41,10 +41,10 @@ export default class MainInstallButton extends React.Component {
     const { isFirefox, isMinFirefox, isMobile, hasAddon } = this.props;
     const isInstalling = this.state.isInstalling && !hasAddon;
     const terms = <Localized id="landingLegalNoticeTermsOfUse">
-      <a href="/terms">Terms of Use</a>
+      <a href="/terms"/>
     </Localized>;
     const privacy = <Localized id="landingLegalNoticePrivacyNotice">
-      <a href="/privacy">Privacy Notice</a>
+      <a href="/privacy"/>
     </Localized>;
 
     return (
--- a/frontend/src/app/components/NewsletterForm.js
+++ b/frontend/src/app/components/NewsletterForm.js
@@ -54,13 +54,14 @@ export default class NewsletterForm extends React.Component {
 
   renderEmailField() {
     return (
-      <input
-        type='email'
-        required
-        placeholder='Your email here'
-        value={this.props.email}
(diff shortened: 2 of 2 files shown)


frontend/src/app/components/MainInstallButton.js
```diff
@@ -41,10 +41,10 @@ export default class MainInstallButton extends React.Component {
     const { isFirefox, isMinFirefox, isMobile, hasAddon } = this.props;
     const isInstalling = this.state.isInstalling && !hasAddon;
     const terms = <Localized id="landingLegalNoticeTermsOfUse">
-      <a href="/terms">Terms of Use</a>
+      <a href="/terms"/>
```
johngruen, line 44: While trying to debug this issue, i noticed that these links were not getting l10n-ized as well. Removing strings from src files seems to resolve this issue

frontend/src/app/components/NewsletterForm.js
```diff
@@ -54,13 +54,14 @@ export default class NewsletterForm extends React.Component {
...
-        onChange={this.handleEmailChange}
-      />
+      <Localized id="newsletterFormEmailPlaceholder">
+        <input
+          type='email'
+          required
+          value={this.props.email}
+          onChange={this.handleEmailChange}
+        />
+      </Localized>
```
johngruen, line 64: the l10n key was already lying around, we just needed to wrap the element

frontend/src/app/components/MainInstallButton.js
lmorchard: We need to have default strings in these elements going forward, though, per #2791. If having a string in there is causing a problem, then that might be an issue with localization that we need to fix.

johngruen: @lmorchard AFAIK these are the only two examples where we're passing in links like this. I'm not super familiar with fluent react, but we can ref this PR on that bug maybe?

lmorchard: Should probably file a new issue, since #2791 is closed. Is it that translations just don't show up at all for these elements?

lmorchard: Looks like #2911 is filed now. Will merge this and see if we can work things out to get the original text put back in

PR #2933, opened 2017-10-12
Fixes #2928

### LINKED ISSUE #2928: Useless Incompatible add-on error displayed for Tab Center when only Test Pilot add-on is installed on Beta or Nightly
**[Notes]:**
- The error last sentence "Get the current version of Firefox." seems useless on Firefox 57 and 58 since the link redirects you to a lower version. 
- Also, I would suggest that only the first sentence from the message should be displayed, and only when the graduated add-on is still installed, even if disabled in about:addons due to version.

**[Affected versions]:**
- Nightly 58.0a1
- Beta 57.0b7

**[Affected Platforms]:**
- All Windows
- All mac OS
- All Linux

**[Prerequisites]:**
- Test Pilot add-on was already installed on the same profile.

**[Steps to reproduce]:**
1. Open Firefox Beta 57 or Nightly 58.0a1 browser with the profile from prerequisites and navigate to https://testpilot.dev.mozaws.net page.
2. Scroll down on the Homepage and click the "View Past Experiments" button under the last active experiment.
3. After past experiments cards are displayed, click on the Tab Center one.
4. Look for any error messages displayed on page.

**[Expected result]:**
- There are no error messages on the page, just some alerts regarding the graduation report and supported Firefox versions.

**[Actual result]:**
- There is an error message displayed on red background on the top right side saying "Tab Center is not supported in this version of Firefox. Get the current version of Firefox." even if Tab Center is not installed and cannot be.

useless error

jk47:
If experiment has graduated, do not show the warning.

### DIFF
--- a/frontend/src/app/containers/ExperimentPage/index.js
+++ b/frontend/src/app/containers/ExperimentPage/index.js
@@ -281,7 +281,7 @@ export class ExperimentDetail extends React.Component {
               </header>
               { this.renderExperimentControls() }
               { this.renderMinimumVersionNotice(title, hasAddon, min_release) }
-              { this.renderMaximumVersionNotice(title, hasAddon, max_release) }
+              { this.renderMaximumVersionNotice(title, hasAddon, max_release, graduated) }
             </LayoutWrapper>
           </div>
           <div className="sticky-header-sibling" style={{ height: `${stickyHeaderSiblingHeight}px` }} ></div>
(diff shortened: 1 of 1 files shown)


frontend/src/app/containers/ExperimentPage/index.js
```diff
@@ -609,7 +609,7 @@ export class ExperimentDetail extends React.Component {
   }
 
   renderMaximumVersionNotice(title, hasAddon, max_release) {
```
fzzzy: You'll need to add graduated as a parameter to renderMaximumVersionNotice, and pass it where renderMaximumVersionNotice is called.

fzzzy (changes requested):

frontend/src/app/containers/ExperimentPage/index.js
jk47: 👍

fzzzy (approved):

PR #2935, opened 2017-10-12
fix #2934 update/fix containers gradudation date

### LINKED ISSUE #2934: Containers should follow the same 7 days pre-graduation flow as previous experiments since it's graduation date is 10/17/2017
**[Notes]:**
- There is a default graduation process in place that shows notifications on Test Pilot experiment ticket and experiment details page. First notification is showed starting with 7 days before graduation and says "Ending soon". The second is showed 1 day before graduation saying "Ending tomorrow".
- If possible, the graduation date should be updated for production server too this week, in order to make aware the Test Pilot and Containers users about this, since this will happen next Tuesday.

**[Affected versions]:**
- Firefox 56.0 and above

**[Affected Platforms]:**
- All Windows
- All mac OS
- All Linux

**[Prerequisites]:**

**[Steps to reproduce]:**
2. Scroll down on the page and look for Containers experiment card.

**[Expected result]:**
- The Containers experiment card is still displayed and a notification is displayed informing that it's "Ending soon" (10/17/2017). 

**[Actual result]:**
- The Containers experiment card is no longer displayed as active experiment. Instead you can already find it in "Past Experiments" section (10/12/2017).

groovecoder:


fzzzy:
I can't explain why this makes the tests fail.

fzzzy:
It looks like it is because containers.yaml does not have an eol_warning. Here is what an eol_warning looks like from another experiment:

eol_warning: The No More 404s experiment will not be removed from your browser and the results will be reported here when it ends. Future updates will be available on addons.mozilla.org.

fzzzy:
We're just going to abandon this change. The eol date is already in the past, so when we push on tuesday the eol message will not be shown.

Issue #2657, opened 2017-07-25
[l10n] Remove content from old experiments


flodolo:
A contributor starting to localizing the website correctly noted that there's a ton of obsolete content in experiments.ftl

Example: AS, cliqz, nomore404, paheshot, etc.

Why can't we get rid of this content?

CC @rprys

fzzzy:
We still show all the old experiment content, under "view past experiments". It seems fine if new languages don't want to do the work to localize old experiments though.

flodolo:
Good point, I completely missed that we still have that content. Is the plan to keep them on the website for the time being?

flodolo:
Correction: we still have the experiments, but there's a lot of content not available anymore (tour, privacy notes).
https://testpilot.firefox.com/experiments/activity-stream

So, there's a lot of content that could be removed when an experiment retires.

fzzzy:
Good call on removing content that actually isn't used any more. Let's do this.

flodolo:
Any chance of getting resources on this? I keep getting this question from locales that want to localize Test Pilot.

fzzzy:
I'll do this now.

PR #2937, opened 2017-10-12
add-tp-grad-report

johngruen:
@lmorchard i deleted the old grad report...not sure if that's overkill


fzzzy:
@johngruen I think it is worth doing because otherwise new languages will translate the old grad report. Also related: https://github.com/mozilla/testpilot/issues/2657

SoftVision-PaulOiegas:
The button and link to the Medium article was correctly added on dev and stage servers. 👍

PR #2938, opened 2017-10-13
change usearch grad report to medium

johngruen:


SoftVision-PaulOiegas:
The button and link to the Medium article was correctly added on dev and stage servers. 👍

PR #2965, opened 2017-10-17
add photon colors to sass path

johngruen:


johngruen:
@bwinton suggests a better way to do this...let's find out!

https://github.com/FirefoxUX/photon/blob/master/src/styles/main.scss#L5

johngruen:
Follow up on above comment...@bwinton's suggestion doesn't work!

```
    frontend/src/styles/_vendor.scss
Error: File to import not found or unreadable: ~photon-colors/colors.scss.
       Parent style sheet: /Users/jgruen/Repos/testpilot/frontend/src/styles/_vendor.scss
        on line 2 of frontend/src/styles/_vendor.scss
>> @import '~photon-colors/colors.scss';
   ^
```

Womp womp...looks like this will only work if we fully switch to webpack for our style build

johngruen:
this works in vendor.scss:
```
@import './node_modules/photon-colors/colors.scss';
```
but seems unstable.

johngruen:
https://github.com/FirefoxUX/design-tokens/issues/14

johngruen:
This is now in good condition for review.

frontend/tasks/styles.js
```diff
@@ -5,6 +5,7 @@ const autoprefixer = require('gulp-autoprefixer');
 const del = require('del');
 const gulpif = require('gulp-if');
 const minifycss = require('gulp-cssnano');
+const moduleImporter = require('sass-module-importer');
```
bwinton: Maybe name this "sassImporter" to explain it a little better? :smile:

lmorchard:
> Womp womp...looks like this will only work if we fully switch to webpack for our style build

Yeah, we want to get there sometime soon, but for now...

PR #2966, opened 2017-10-17
componentize copter

johngruen:


johngruen:
@lmorchard we haven't make components out of anything that includes images yet. I pulled apart the copter component as a first step here.

webpack.config.js
```diff
@@ -114,6 +114,13 @@ module.exports = {
           // use style-loader in development
           fallback: 'style-loader'
         })
+      },
+      {
+        test: /\.(jpe?g|gif|png|svg)$/i,
+        loaders: [
+          'file-loader?hash=sha512&digest=hex&name=/static/images/[name]-[hash].[ext]',
```
lmorchard: Haven't used image-webpack-loader before. But looking [at the docs](https://github.com/tcoopman/image-webpack-loader#usage),  I think it would be cleaner to have this expanded out into a JS object like the other module rules in our config so far. Something like:
```javascript
{
  test: /\.(jpe?g|gif|png|svg)$/i,
  use: [
    {
      loader: 'file-loader',
      options: {
        hash: 'sha512',
        digest: 'hex',
        name: 'static/images/[name]-[hash].[ext]'
      }
    },
    {
      loader: 'image-webpack-loader',
      options: {
        bypassOnDebug: true,
      },
    },
  ],
}
```

package.json
```diff
@@ -85,6 +85,7 @@
     "gulp-sourcemaps": "2.6.0",
     "gulp-uglify": "3.0.0",
     "gulp-util": "3.0.8",
+    "image-webpack-loader": "^3.4.2",
```
lmorchard, line 88: Might also want to delete the `^` here so we lock the exact package version

lmorchard (approved):
Only had a small nit about the webpack config and npm package version. Otherwise, looks good and nice to add image processing to the webpack side of things!

johngruen:
Thanks for the feedback! Will push these changes tomorrow!

johngruen:
well, monday....

lmorchard (approved):

Issue #2943, opened 2017-10-16
Test Pilot website is not completely loaded in Microsoft Edge and Internet Explorer browsers


SoftVision-CosminMuntean:
[Notes]:
- This issue is reproducible on all servers.

[Affected versions]:
- Microsoft Edge
- Internet Explorer

[Affected Platforms]:
- All Windows

Steps to reproduce]:
1. Open the Edge or IE browser and navigate to "https://testpilot.stage.mozaws.net/" page.
2. Observe the Test Pilot page.

[Expected result]:
- Test Pilot page is correctly loaded.

[Actual results]:
- Only the blue background with stars and constellations is loaded.

[Additional Notes]:
- Here is a screenshot with the issue:
![ie](https://user-images.githubusercontent.com/19226534/31608631-990f7112-b279-11e7-809e-d0ca8983f942.png)

fzzzy:
This is because we are using Array.includes, which is part of ES2016, not ES2015. For some reason our babel settings don't convert it; probably our babel settings only use ES2015.

Easiest fix is just to switch back to indexOf; alternatively we could upgrade babel to support ES2017 (which has async/await, which would be nice)

chuckharmston:
We should be using [`babel-preset-env`](https://github.com/babel/babel-preset-env), in favor of which `babel-preset-es####` has been deprecated. Rather than choosing a version of the ES spec to comply to, you define your browser spec, and it makes sure it works in each of those browsers.

Worth noting that we should choose our browser spec carefully, as it has the potential to bloat the bundle size with polyfills if you're too liberal.

fzzzy:
.includes( was a red herring. It seems that IE chokes on Symbol. Investigating.

fzzzy:
Symbol is being used by fluent-react, which is in the vendor bundle, which isn't getting transpiled by our babel toolchain. We are including the compat version of fluent-react which has already been transpiled by babel, but still includes references to Symbol (which it shouldn't as Symbol is defined in es6)

fzzzy:
https://github.com/projectfluent/fluent.js/issues/79

lmorchard:
[Back when I replaced `babel-polyfill` with `regenerator-runtime`](https://github.com/mozilla/testpilot/commit/4537c8ce8dde6482fd74afa026fc657519d373e0#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L16), it seemed like async/await was the only thing we needed polyfilled. 

[...]

PR #2972, opened 2017-10-18
Add whatwg-fetch polyfill for older browsers like IE11

lmorchard:
Not super happy with this, since it looks like the vendor bundle is starting to grow again. But, it fixes an error in IE11 (and I assume other older browsers). Issue #2943

### DIFF
--- a/frontend/src/app/index.js
+++ b/frontend/src/app/index.js
@@ -1,4 +1,5 @@
 import 'regenerator-runtime/runtime';
+import 'whatwg-fetch';
 import Raven from 'raven-js';
 import moment from 'moment';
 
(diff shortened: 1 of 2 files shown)


lmorchard:
But, looking at the analyzer, this polyfill only adds 12.38kb so 🤷‍♂️

fzzzy:
At some point soon, we could start using `script type="module"` for browsers that support it, and `script nomodule` for browsers that don't, meaning we could only ship polyfills and transpiled code to older browsers. It looks like Firefox is the last browser to enable it by default, although it is available behind a flag.

http://caniuse.com/#search=script%20type%3D%22module%22

PR #2973, opened 2017-10-18
Remove legacy add-on communication channel code

### LINKED ISSUE #2744: Remove remaining non-mozAddonManager experiment install code from site
We still have [a fork in the frontend](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/lib/InstallManager.js) to switch between [mozAddonManager-based](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/lib/moz.js) and [add-on-based](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/lib/addon.js) approaches to enabling/disabling experiments.

We've gone all-in on mozAddonManager and removed the support from the add-on. So, we should remove this code fork and only support mozAddonManager.

lmorchard:
- Merge frontend/src/app/lib/moz.js into InstallManager.js since we only use mozAddonManager now
- Tweak InstallManager functions to quietly do nothing if mozAddonManager is unavailable
- Delete old legacy add-on comms channel in frontend/src/app/lib/addon.js
- Also fixes an error with CustomEvent in IE11

Fixes #2744


fzzzy:
This is great. I assume most code was just copied from frontend/src/app/lib/moz.js into InstallManager.js unchanged? It is pretty hard to read the diff.

lmorchard:
Yeah, copied the code over and added `if (!mam) { return; }` to the start of functions

PR #2979, opened 2017-10-19
Fix #2657 Remove ftl strings from old experiments so localizers don't have to translate strings which will never be shown anywhere.

### LINKED ISSUE #2657: [l10n] Remove content from old experiments

fzzzy:
Once an experiment has graduated, the following strings should be
removed for the experiment:

- measurements
- prefeedbackcopy
- toursteps
- eolwarning


flodolo (approved):
Thanks for doing this! Do we need to add this information/guidelines somewhere?

lmorchard (changes requested):
[You can't manually edit `experiments.ftl`, it's automatically generated from YAML.](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization) The CircleCI failure reflects this, with a check to ensure that any strings extracted from YAML also appear in `experiments.ftl` as committed to git.

The only way to get these strings out of `experiments.ftl` is to remove them from YAML. That, or hack the extraction code to omit them.

fzzzy:
@flodolo Yes, I think we should add it to any docs we have on graduating an experiment. (Do we have any? I'll look)

fzzzy:
@lmorchard Oh, right. I always forget that :P Ok, I will remove them from the yaml.

lmorchard (approved):

lmorchard:
Thought this would take me longer to look through, but seems like all the pages are working fine

PR #2982, opened 2017-10-23
Use mzl.la links on sharing page

### LINKED ISSUE #1604: Use mzl.la links on sharing page.
https://testpilot.firefox.com/share has a few links to Test Pilot with various GA `utm_` params. We should consider swapping them out with mzl.la-shortened versions of the link, if only for aesthetic reasons.

All mozilla.org links shortened on bit.ly are automatically made with the mzl.la domain, but since we're on firefox.com, it's not automatic. I believe that a logged-in user of the account that controls mzl.la can handle the creation of these links for us. Might take some digging to figure out who that is.

CC @ckprice

epq:
Fixes #1604.

### DIFF
--- a/frontend/src/app/containers/SharePage.js
+++ b/frontend/src/app/containers/SharePage.js
@@ -54,11 +54,6 @@ export default class SharePage extends React.Component {
     );
   }
 
-  shareUrl(medium: string, urlencode: boolean) {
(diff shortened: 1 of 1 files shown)


meandavejustice:
Thanks @epq !

Issue #2888, opened 2017-09-26
<esc> to exit modals


meandavejustice:
I want to be able to hit <esc> and exit our modals

kgeorgiou:
Hello @meandavejustice - is someone working on this or shall I pick it up?

Also, how about:
- `ArrowLeft` and `ArrowRight` keys for going through the steps of tour dialogs?
- `Enter` key for done/continue/submit actions in modals?

SoftVision-PaulOiegas:
Hi @kgeorgiou, I don't think someone is working on this and don't see anyone assigned to it either. I think it's ok if you pick it up.
Regarding to your proposal of extended keyboard actions support, let's see what @johngruen  has to say about this,

johngruen:
@kgeorgiou love it!

PR #2983, opened 2017-10-23
Handle key events in modals.

kgeorgiou:
The scope of this PR is to enable modals to handle the following key events:
- `<Escape>` to dismiss modals (#2888)
- `<Enter>` for submit/done/continue actions
- `<ArrowLeft>` and `<ArrowRight>` for going through the steps in Experiment Tour modals

For the modals to listen on key events they first have to gain **focus**. This was accomplished by following React's documentation on managing focus with [Refs and the DOM](https://reactjs.org/docs/refs-and-the-dom.html). Also, since the modal elements are `<div>` elements, `tabIndex="0"` was used on them for them to be focusable.

| Dialog Name | Escape | Enter | ArrowLeft | ArrowRight |
| --- | --- | --- | --- | --- |
| Experiment Disable * | Dismiss | Open survey URL | - | - |
| Experiment EOL | Dismiss | Disable Experiment | - | - |
| Experiment PreFeedback * | Dismiss | Open survey URL | - | - |
| Experiment Tour | Dismiss | Complete Tour (if on final step) | Previous step | Next step  |
| Email ** | Dismiss | Sign Up \| Try Again \| Proceed | - | - |
| Retire Confirmation | Dismiss | Uninstall Testpilot | - | - |

*: The `Enter` key in these modals attempts to programmatically launch the survey in a new tab (i.e. the expected behavior when the button is clicked.) There are a couple of issues with that. For one, the new tab is prevented by the pop-up blocker. Another tricky part is trying to honor the `rel="noopener noreferrer"` attributes on the anchor element. 
For `noopener`, the following does the trick ([Source](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)): 
```
const newWindow = window.open();
newWindow.opener = null;
```
However, I couldn't come up with a clean solution for applying `noreferrer` programmatically. 
Also note that even when the anchors are clicked, the `noreferrer` attribute has no effect: `document.referrer` in the survey page shows the referrer URL. Maybe it's worth opening a new issue for this.

**: There seems to be an issue with focusing the Email Dialog. Right after Testpilot is installed the focus goes from the modal container to `<body>`. Ideally, for this modal the focus should automatically go in the email input field for the user to be able to type straightaway rather having to click on it first.

Regarding the event metrics, right now most of the modal related actions send the following: `eventAction: 'button click'`. Is it worth changing this to something like `eventAction: 'button key'` for key events?

### DIFF
--- a/frontend/src/app/components/EmailDialog.js
+++ b/frontend/src/app/components/EmailDialog.js
@@ -22,6 +22,8 @@ export default class EmailDialog extends React.Component {
   props: EmailDialogProps
   state: EmailDialogState
 
+  modalContainer: Object
+
   constructor(props: EmailDialogProps) {
     super(props);
     this.state = {
@@ -32,11 +34,17 @@ export default class EmailDialog extends React.Component {
     };
   }
 
+  componentDidMount() {
+    this.modalContainer.focus();
+  }
+
   render() {
     const { isSuccess, isError } = this.state;
 
     return (
-      <div className="modal-container">
+      <div className="modal-container" tabIndex="0"
+           ref={modalContainer => { this.modalContainer = modalContainer; }}
+           onKeyDown={e => this.handleKeyDown(e)}>
         {!isSuccess && !isError && this.renderForm()}
         {isSuccess && this.renderSuccess()}
         {isError && this.renderError()}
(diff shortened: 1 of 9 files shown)


lmorchard:
Sorry for the delay in reviewing, will get to this within the next day or so. Looks very good though - with tests, even!

kgeorgiou:
No worries! This looks like something to be looked at after the important milestones of November.

lmorchard (approved):
Looks good, works well, tests pass. I think we can iterate on the rough edges mentioned in future PRs.

Issue #2391, opened 2017-04-19
mozAddonManagerInstall can be simplified a lot


fzzzy:
return mam.createInstall({ url }).then(install => install.install()) should be sufficient.

lmorchard:
Might do with some more details in this issue - is `install.install()` sufficient because it returns a promise that resolves on install ended and rejects on download / install failed?

fzzzy:
Yes. There was a long discussion about it on irc around the time I filed the bug... I could dig out the logs if needed.

lmorchard:
> There was a long discussion about it on irc around the time I filed the bug... I could dig out the logs if needed.

Yeah, that, or I guess self-assign it because no one else has the context behind this issue :)

fzzzy:
@lmorchard Your previous comment was exactly correct -- it returns a promise that does the right thing.

fzzzy:
This was fixed by https://github.com/mozilla/testpilot/pull/2984

PR #2984, opened 2017-10-23
Handle web ext permission rejection on install

### LINKED ISSUE #2697: Handle Opt-out of webExtension Installation
Currently we do not handle the following case: 

### STR

1. Go to the Voice Fill Experiment Page
2. Click `Enable Voice Fill`
3. On the webExtension Install dialog, click `Cancel`
screen shot 2017-10-06 at 1 30 54 pm

### Expected

The `Enable Voice Fill` Button should revert to its pre-installed state and an event should be emitted indicating that the user opted out of experiment installation.

```
ec: ExperimentDetailsPage Interactions
ea: Cancel From Permission
el: {experiment title}
```

Additionally in order to understand how cancels compare to installations, we should also record each time a user accepts installation from the permission dialog.

```
ec: ExperimentDetailsPage Interactions
ea: Accept From Permission
el: {experiment title}
```

### Actual

The `Enable Voice Fill` button barberpolls until page reset and the test pilot icon sheds a tear.

meandavejustice:
- add error and success events
- fixes #2697

### DIFF
--- a/frontend/src/app/lib/InstallManager.js
+++ b/frontend/src/app/lib/InstallManager.js
@@ -15,13 +15,24 @@ if (typeof navigator !== 'undefined') {
   mam = navigator.mozAddonManager;
 }
 
-function mozAddonManagerInstall(url) {
+function mozAddonManagerInstall(url, sendToGA) {
+  const start = url.indexOf('files/') + 6;
+  const end = url.indexOf('@');
+  const experimentTitle = url.substring(start, end);
   return mam.createInstall({ url }).then(install => {
-    return new Promise((resolve, reject) => {
-      install.addEventListener('onInstallEnded', () => resolve());
-      install.addEventListener('onDownloadFailed', () => reject());
-      install.addEventListener('onInstallFailed', () => reject());
-      install.install();
+    return install.install().then(() => {
+      sendToGA('event', {
+        eventCategory: 'ExperimentDetailsPage Interactions',
+        eventAction: 'Accept From Permission',
+        eventLabel: experimentTitle
+      });
+    }).catch((err) => {
+      sendToGA('event', {
+        eventCategory: 'ExperimentDetailsPage Interactions',
+        eventAction: 'Cancel From Permission',
+        eventLabel: experimentTitle
+      });
+      throw err;
     });
   });
 }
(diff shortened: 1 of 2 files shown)


frontend/src/app/lib/InstallManager.js
```diff
@@ -15,13 +15,29 @@ if (typeof navigator !== 'undefined') {
   mam = navigator.mozAddonManager;
 }
 
-function mozAddonManagerInstall(url) {
+function mozAddonManagerInstall(url, sendToGA) {
   return mam.createInstall({ url }).then(install => {
     return new Promise((resolve, reject) => {
```
fzzzy: I think this should be removed, and the "start", "end", "experimentTitle" and "install.install().then(..." lines moved up a level of indentation.

frontend/src/app/lib/InstallManager.js
```diff
@@ -15,13 +15,29 @@ if (typeof navigator !== 'undefined') {
...
+      const start = url.indexOf('files/') + 6;
+      const end = url.indexOf('@');
+      const experimentTitle = url.substring(start, end);
       install.addEventListener('onInstallEnded', () => resolve());
```
fzzzy: These three lines can be removed as per https://github.com/mozilla/testpilot/issues/2391

frontend/src/app/lib/InstallManager.js
```diff
@@ -15,13 +15,29 @@ if (typeof navigator !== 'undefined') {
...
       install.addEventListener('onDownloadFailed', () => reject());
       install.addEventListener('onInstallFailed', () => reject());
-      install.install();
+      install.install().then(() => {
```
fzzzy: Then, this needs to be `return install.install().then(...`

frontend/src/app/lib/InstallManager.js
```diff
@@ -15,13 +15,29 @@ if (typeof navigator !== 'undefined') {
...
+          eventAction: 'Accept From Permission',
+          eventLabel: experimentTitle
+        });
+      }).catch(() => {
+        sendToGA('event', {
+          eventCategory: 'ExperimentDetailsPage Interactions',
+          eventAction: 'Cancel From Permission',
+          eventLabel: experimentTitle
+        });
+        reject();
```
fzzzy: And this can be removed.

fzzzy (changes requested):
There's an extra promise that can be removed.

fzzzy:
One more thing about this: The actual bug in https://github.com/mozilla/testpilot/issues/2697 is that when the promise is rejected, we don't actually dispatch an event to redux which then puts the ui into a state that indicates an error occurred. I'm not exactly sure where the right place to put that code is.

Edit: Never mind, bubbling the install.install() promise error up to the top level promise fixed it.

fzzzy (approved):

fzzzy:
Looks really great!

PR #2988, opened 2017-10-25
add no script locale string

### LINKED ISSUE #1859: Disabled JavaScript warning is not localized
Just noticed looking at the code, but you can verify it by disabling JavaScript in devtools.
https://testpilot.stage.mozaws.net/ and browser set to Italian.

schermata 2016-11-22 alle 15 19 16

johngruen:
Starts to resolve #1859 

We can merge this any time since it doesn't go into the app.


flodolo (approved):

lmorchard:
Good to get the string in, though we won't be able to use it without some big site changes

PR #2991, opened 2017-10-26
fix unlocalized data/privacy link

### LINKED ISSUE #2917: "data" link no longer localized
#2728 no update to https://github.com/mozilla/testpilot/blob/0d7f5653a479a46f0987cbabb658bacaa6ea9159/locales/en-US/app.ftl#L137 and other locales.

meandavejustice:
- fixes #2917

data-link-localized

### DIFF
--- a/frontend/src/app/containers/ExperimentPage/index.js
+++ b/frontend/src/app/containers/ExperimentPage/index.js
@@ -215,6 +215,10 @@ export class ExperimentDetail extends React.Component {
       statusType = 'enabled';
     }
 
+    const privacy = (<Localized id="experimentMeasurementIntroPrivacyLink">
+                       <a target="_blank" rel="noopener noreferrer" href="/privacy"/>
+                     </Localized>);
+
     return (
       <section id="experiment-page">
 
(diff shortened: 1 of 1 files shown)


fzzzy (changes requested):
This isn't going to work because although you changed the experimentMeasurementIntro placeholder, you didn't change the ftl string. Changing the ftl string will require changing the name `experimentMeasurementIntro` to something else so that the strings get re-localized.

However, since it is using LocalizedHtml, what is there now should be working. I think the correct fix is to figure out why LocalizedHtml is not working properly in this case.

meandavejustice:
@fzzzy  I don't think the `experimentMeasurementIntro` string needs to be updated in this case since it's already been localized, it just wasn't being rendered.

I went with the fix in https://github.com/mozilla/testpilot/issues/2917#issuecomment-338997265 to get this working. But I think it makes sense to investigate this further, maybe as a stretch milestone issue

fzzzy:
@meandavejustice The problem is that the string in the ftl is:

`In addition to the <a href="/privacy">data</a> collected by all Test Pilot experiments, here are the key things you should know about what is happening when you use {experiment.title}:`

If you change it to this in the jsx:

`In addition to the {privacy} collected by all Test Pilot experiments, here are the key things you should know about what is happening when you use {experiment.title}:`

That does not change the string in the ftl. If you want to change the string so that the substitution happens using an ftl placeholder instead of relying on the way LocalizedHtml handles <a> elements, you'll need to change the string in the ftl, which means you'll need to rename the string as well, so that the new string gets re-localized.

However, I don't think the fix should be done like that -- I think we should figure out why LocalizedHtml is not copying the href attribute from the jsx string to the output from the parsed ftl string.

johngruen:
Let's go ahead and merge this and @fzzzy can file a follow on bug for 36

fzzzy (approved):
Ok, we're going to merge this now and fix it the right way later

TonnesM:
Just to make sure things won’t break at the next update: I see this fixed on prod, but dev still shows "data". Is that correct?

PR #2994, opened 2017-10-26
created lockbox experiment template

sandysage:
Created lockbox.yaml as starting template @johngruen

### DIFF
--- a/content-src/experiments/lockbox.yaml
+++ b/content-src/experiments/lockbox.yaml
@@ -6,25 +6,26 @@ platforms: ['addon']
 min_release: 57
 thumbnail: /static/images/experiments/notes/experiments_experiment/thumbnail.png
 description: >
-  A free password manager to assure your online accounts are secure without having to remember passwords and security codes.
+  A free password manager to assure your online accounts are secure without
+  having to remember passwords and security codes.
 introduction: >
-  Lockbox lets you add and access your website credentials right from the browser. These are secure to your local Firefox browser. If you want to further secure your entries, lock those in a vault with your Firefox Account. 
+  Lockbox lets you add and access your website credentials right from the
+  browser. These are secure to your local Firefox browser. If you want to
+  further secure your entries, lock those in a vault with your Firefox Account. 
 image_twitter: /static/images/experiments/notes/social/notes-twitter.jpg
 image_facebook: /static/images/experiments/notes/social/notes-fb.jpg
 video_url: https://www.youtube.com/embed/vaFvr_70yxw
-changelog_url: 'https://github.com/mozilla/notes/releases'
-contribute_url: 'https://github.com/mozilla/notes'
-bug_report_url: 'https://github.com/mozilla/notes/issues'
-discourse_url: 'https://discourse.mozilla-community.org/c/test-pilot/notes'
-privacy_notice_url: 'https://github.com/mozilla/notes/blob/master/docs/metrics.md'
+changelog_url: 'https://mozilla-lockbox.github.io/lockbox-extension/release-notes/'
+contribute_url: 'https://github.com/mozilla-lockbox/lockbox-extension'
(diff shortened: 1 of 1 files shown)


content-src/experiments/lockbox.yaml
```diff
@@ -0,0 +1,81 @@
...
+  -
+    display_name: 'Peter Dehaan'
+    title: 'Firefox QA'
+    avatar: /static/images/experiments/avatars/peter.jpg
+  -
+    display_name: 'Emil Pasca'
+    title: 'Softvision QA'
+    avatar: /static/images/experiments/avatars/emil.jpg
+created: '2017-08-01T14:00:00Z'
+launch_date: '2017-08-01T14:00:00Z'
```
lmorchard: Need to bump this date into the future before merging. Otherwise, this experiment will be published at our next deploy, temporary starting content and all. 

Might be better to keep this on another branch while it's in progress and get a PR together once the content is finalized

pdehaan: @lmorchard: But this is set to `dev: true`, so it'd only be be published to testpilot.dev.mozaws.net, correct?

lmorchard: Oh yeah, I missed the `dev: true` part. Probably okay, then.

meandavejustice: probably still good to get in the habit of having a future publish date in these types of prs

devinreams:
Howdy! Thanks for the help thus far. I believe we've addressed the "future date" concern in a recent commit now.

After chatting with @clouserw, if I understood correctly, he suggested we want to get this (even if placeholder content and still in progress) merged ASAP so it's part of future cycles.

Two questions:

1. Is this the right way to target just en-US (intended audience for launch)
2. Any other concerns or requested changes on what you see here?

Thanks again for your help!

--- 

EDIT: just noticed the failing test for number of items, because of future dating I believe?

content-src/experiments/lockbox.yaml
```diff
@@ -0,0 +1,53 @@
...
+id: 999
+title: 'Lockbox'
+slug: lockbox
+platforms: ['addon']
+locales:
+- 'en-US'
+locale_grantlist:
+- 'en-US'
+min_release: 57
+thumbnail: /static/images/experiments/notes/experiments_experiment/thumbnail.png
```
pdehaan, line 11: Do we want to put in Lockbox specific assets now, or are we waiting for a future PR?

content-src/experiments/lockbox.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - >
+    Lockbox will collect information about your use of the feature, such as how
+    often you create entries.
+  - >
+    Lockbox will also collect information about your interaction with the add-on,
+    such as how you access the feature, and what buttons you may click.
+xpi_url: 'https://testpilot.firefox.com/files/lockbox@mozilla.com/latest'
+addon_id: 'lockbox@mozilla.com'
+gradient_start: '#A665B4'
+gradient_stop: '#CF5858'
```
pdehaan, line 36: @johngruen Can we get some fresh gradient colors? 🙏

devinreams:
>Do we want to put in Lockbox specific assets now, or are we waiting for a future PR?

I understood the goal was to get this merged in now and then fill in assets later.

@clouserw if there's no immediate crunch for time then would you rather we keep this open until we have _everything_ (assets, colors, messaging, etc.) ready?

ghost:
@devinreams no big preference.  as long as it stays flagged 'dev' it won't be in our way.  I'd say land the sooner the better. :)

pdehaan:
+1 to landing this with `dev:true` and adding `is_featured:true` so it will be a nice dev server test for the featured experiment #3114 PR.

johngruen:
Closing this for now

Issue #2563, opened 2017-06-28
Define browser support matrix


jaredhirsch:
Which browsers do we really, really care about? Which are "best effort"?

Related to #2298 (busted on firefox focus)

grahamperrin:
[Test Pilot site lacks a scroll bar in Firefox on Mac OS X 10.13.x (macOS High Sierra) · Issue #2827 · mozilla/testpilot](https://github.com/mozilla/testpilot/issues/2827)

fzzzy:
Related to this, the work to switch to babel-preset-env (https://github.com/mozilla/testpilot/pull/2987) does allow IE 11 to work without JS errors, but IE doesn't implement flexbox correctly so the pages don't render properly. If we omit IE from the babel-preset-env configs, our bundle gets about 200-300k smaller. It would be nice to not support IE :-)

PR #3004, opened 2017-10-30
add browser support matrix

johngruen:
#2563


johngruen:
Still don't understand codecov

PR #3011, opened 2017-11-01
Update path for Test Pilot logo and halve its size in README

kgeorgiou:
With the componentization of [Copter](https://github.com/mozilla/testpilot/tree/master/frontend/src/app/components/Copter), the Test Pilot logo has moved. 
It also doubled in size: 328x265 -> 656x530.  

Since 656x530 looks too big for the README, I tried to halve the displayed size. I'd prefer a pure markdown way to adjust the copter's size but couldn't find one. Maybe it's worth to restore and keep an extra copy of the 328x265 copter.


meandavejustice:
Thanks @kgeorgiou !

Issue #3015, opened 2017-11-01
Fresh News updates not being shown


meandavejustice:
issue has been narrowed down to the `seenSelector`

https://github.com/mozilla/testpilot/blob/master/frontend/src/app/selectors/news.js#L46

PR #3020, opened 2017-11-02
More news update selector fixes and cleanup

lmorchard:
- Implement distinction between fresh news items within the last 2 weeks
  and news items since last viewed. This allows the items on the home
  page to stay visible, while allowing the news alert dialog to not be
  shown again

- Cookie and date information should be supplied from outside to a
  selector creator, not generated from within the selector.

Issue #3015

### DIFF
--- a/frontend/src/app/selectors/news.js
+++ b/frontend/src/app/selectors/news.js
@@ -1,19 +1,21 @@
 import { createSelector } from 'reselect';
-import moment from 'moment';
-import cookies from 'js-cookie';
 import experimentSelector from './experiment';
 
 export function publishedFilter(update) {
-  const hasPublishedField = (typeof update.published !== 'undefined');
-  const isPublished = moment(moment.utc()).isAfter(update.published);
+  const hasPublishedField = typeof update.published !== 'undefined';
+  if (!hasPublishedField) {
+    return false;
+  }
 
-  return (hasPublishedField && isPublished);
+  const now = Date.now();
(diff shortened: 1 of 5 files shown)


lmorchard:
Noticed a couple further issues with news update handling: 

- Although we want the news alert dialog to not appear until there are news items since last viewed, the list of items on the home page should stick around for 2 weeks regardless of last dialog view.

- Selectors should not use information from outside the store or other selectors - i.e. cookies or date. Instead, these can be passed into a "factory" function that creates a selector based on those parameters.

frontend/src/app/selectors/news.js
```diff
@@ -58,26 +65,31 @@ const makeNewsAgeSelector = includeStale => (newsUpdates, twoWeeksAgo) => {
...
-  if (includeStale) return result;
-  // if includeStale return before filtering out updates which have
-  // been seen.
+  if (!lastViewedDate) {
+    return result;
+  }
+
+  // If lastViewedDate is supplied, further filter result by that cut-off date
   return result.filter(update => {
     const dt = new Date(update.published || update.created).getTime();
```
meandavejustice, line 74: This isn't your change, but I'm realizing that here and above on line 64 we shouldn't be checking for `update.created` it should only be `published`

meandavejustice: I forgot to update that in the last pr

meandavejustice (approved):

PR #3029, opened 2017-11-06
news alerts ui issues

### LINKED ISSUE #3016: Resolve UX bugs for news-related badging
The new work to resolve #2419 has introduced a few usability bugs. We need to resolve the following:
1. All news items should have an optional `major` flag in yaml (optionally this could be keyed off the image field, but that seems semantically inaccurate)
2. Only `major` items should trigger a badge in the add-on and a modal on the site
3. Any badging or modals form major updates should expire after two weeks or first viewing.
4. users should only see modals once.

meandavejustice:
fixes #3016

### DIFF
--- a/addon/src/webextension/lib/browserAction.js
+++ b/addon/src/webextension/lib/browserAction.js
@@ -61,10 +61,25 @@ async function updateBadgeTextOnNew(topic, { experiments, news_updates }) {
     return dt >= clicked;
   });
 
-  const newsUpdates = (news_updates || []).filter(update => {
-    const dt = new Date(update.published).getTime();
-    return dt >= clicked;
+  let { value: lastViewed } = await browser.cookies.get({
+    url: getCurrentEnv().baseUrl,
+    name: 'updates-last-viewed-date'
   });
+
+  if (!lastViewed) lastViewed = 0;
+
+  /* only show badge for news update if:
+   * - has the "major" key
+   * - update has been published in the past two weeks
+   * - update has not been "seen" by the addon (clicked)
+   * - update has not been "seen" by the frontend (lastViewed)
+   */
+  const TWO_WEEKS = 2 * 7 * 24 * 60 * 60 * 1000;
+  const twoWeeksAgo = Date.now() - TWO_WEEKS;
+  const newsUpdates = (news_updates || []).filter((u) => u.major)
+        .filter((u) => new Date(u.published).getTime() >= twoWeeksAgo)
+        .filter((u) => new Date(u.published).getTime() >= new Date(lastViewed))
+        .filter((u) => new Date(u.published).getTime() >= clicked);
 
   BROWSER_ACTION_LINK = (newExperiments.length || newsUpdates.length) > 0
     ? BROWSER_ACTION_LINK_BADGED
--- a/addon/src/webextension/lib/browserAction.js
+++ b/addon/src/webextension/lib/browserAction.js
@@ -17,6 +17,8 @@ const environmentTopics = (...args) =>
 
 const storage = browser.storage.local;
 
+const TWO_WEEKS = 2 * 7 * 24 * 60 * 60 * 1000;
+
 const BROWSER_ACTION_LINK_BASE = [
   '/experiments',
   '?utm_source=testpilot-addon',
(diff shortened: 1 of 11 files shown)


addon/src/webextension/lib/browserAction.js
```diff
@@ -61,11 +61,34 @@ async function updateBadgeTextOnNew(topic, { experiments, news_updates }) {
...
 
-  const newsUpdates = (news_updates || []).filter(update => {
-    const dt = new Date(update.published).getTime();
-    return dt >= clicked;
+  // check for port number on local, we need to strip it off
+  // to properly fetch cookies.
+  const portIndex = getCurrentEnv().baseUrl.indexOf(':8000');
+  const baseUrl = (portIndex > -1) ?
+        getCurrentEnv().baseUrl.substring(0, portIndex)
+        : getCurrentEnv().baseUrl.indexOf(':8000');
```
fzzzy: Don't call getCurrentEnv().baseUrl 3 times. Assign the result to a variable baseUrlWithPort, then assign baseUrl based on that.

It seems like this will set baseUrl to a string if the previous value of baseUrl contains :8000, but then it will set it to -1 if it does not. If baseUrl is 'https://example.com' in the else case, I think this should just be the old value of baseUrlWithPort.

addon/src/webextension/lib/browserAction.js
```diff
@@ -61,11 +61,34 @@ async function updateBadgeTextOnNew(topic, { experiments, news_updates }) {
...
 
+  if (cookie) lastViewed = cookie.value;
+
+  /* only show badge for news update if:
+   * - has the "major" key
+   * - update has been published in the past two weeks
+   * - update has not been "seen" by the frontend (lastViewed)
+   * - update has not been "seen" by the addon (clicked)
+   */
+  const TWO_WEEKS = 2 * 7 * 24 * 60 * 60 * 1000;
```
fzzzy: Move this to the top of the file.

addon/src/webextension/lib/browserAction.js
```diff
@@ -61,11 +61,34 @@ async function updateBadgeTextOnNew(topic, { experiments, news_updates }) {
...
+  const twoWeeksAgo = Date.now() - TWO_WEEKS;
+  const newsUpdates = (news_updates || []).filter((u) => u.major)
+        .filter((u) => new Date(u.published).getTime() >= twoWeeksAgo)
```
fzzzy: Don't call new Date(u.published).getTime() three times... Assign the result to a `const publishedTime` on the line above.

addon/src/webextension/manifest.json
```diff
@@ -18,7 +18,13 @@
   "permissions": [
     "notifications",
     "storage",
-    "tabs"
+    "cookies",
+    "tabs",
+    "*://example.com/",
```
fzzzy: I think these are all safe to be https:// instead of *://, we always redirect http to https

fzzzy (changes requested):
Ok, just one small bug and a few nits and I think this is looking good

addon/src/webextension/lib/browserAction.js
meandavejustice: They are different dates in each instance, this is in a `.filter` call

meandavejustice: between these two I think the existing version is more readable 
```js
const newsUpdates = (news_updates || []).filter((u) => u.major)                                                                                                                                               
         .filter((u) => new Date(u.published).getTime() >= twoWeeksAgo)                                                                                                                                          
         .filter((u) => new Date(u.published).getTime() >= new Date(lastViewed))                                                                                                                                 
        .filter((u) => new Date(u.published).getTime() >= clicked);                                                                                                                                             
 ```
```js                                                                                                                                                                                                                
   const newsUpdates = (news_updates || []).map(u => {                                                                                                                                                           
     const pubDate = new Date(u.published).getTime();                                                                                                                                                            
     if (u.major &&                                                                                                                                                                                              
         pubDate >= twoWeeksAgo &&                                                                                                                                                                               
         pubDate >= new Date(lastViewed) &&                                                                                                                                                                      
         pubDate >= clicked) return u;                                                                                                                                                                           
 });
```

rearranging the date call for this case seems like a premature optimization

fzzzy: You're literally calling that code 3 times for every element in the list... I don't think it is premature optimization. If you put const pubDate above const newsUpdates, you will only call it once instead of 3 * news_updates.length.

fzzzy (approved):

addon/src/webextension/lib/browserAction.js
fzzzy: Never mind, I thought published only had to be parsed once, but it has to be parsed for each element of news_updates. Carry on.

PR #3051, opened 2017-11-10
3041 fix email dialog enter key interaction when privacy checkbox is unchecked

### LINKED ISSUE #3041: Dialog Newsletter sign up is wrongly validated when "Enter" key is pressed without checking the privacy checkbox
**[Affected versions]:**
- Firefox 56.0.2 and above

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**

**[Steps to reproduce]:**
2. Click the "Install the Test Pilot Add-on" big green button.
3. After TxP "Newsletter sign up" dialog is displayed, click the text field in order to focus it.
4. Leave text box empty or enter a valid email address.
5. Leave the privacy checkbox unchecked and hit the keyboard "Enter" key.
5. Observe the dialog behavior. 

**[Expected result]:**
- The sign up process is not performed and you are notified to check the privacy notice checkbox.

**[Actual result]:**
- You are wrongly advanced to the sign up confirmation dialog if the entered email is valid or to the newsletter sign up error dialog if an invalid address was used.

**[Notes]:**
- If a valid email address is used, you even get the sign up newsletter email.
- This behavior does not occur when clicking the dialog "Sign Up Now" button.
- This behavior does not occur on the footer Newsletter form.
- Here is a screen recording of the behavior https://goo.gl/p7NFBa.

kgeorgiou:
This fixes #3041. 

One caveat when: 
1. Checkbox is unchecked
2. Focus is on modal-container (not on its child component, [NewsletterForm](testpilot/frontend/src/app/components/NewsletterForm/index.js)) 
3. \<Enter\> key is pressed

The modal will not proceed, but the notification prompting to check the checkbox will not show up. However, if the focus is on the input form, the event will be handled by the form and the notification will show up as expected.

That's because the state for the notification on the checkbox is handled internally by [NewsletterForm](testpilot/frontend/src/app/components/NewsletterForm/index.js) which is a child component of Email Dialog. This can be addressed by _slightly_ modifying [NewsletterForm](testpilot/frontend/src/app/components/NewsletterForm/index.js), but since it's a component used in several other places, I judged it's better to leave it as is, unless deemed necessary to handle the caveat mentioned above.

### DIFF
--- a/frontend/src/app/components/EmailDialog.js
+++ b/frontend/src/app/components/EmailDialog.js
@@ -211,7 +211,17 @@ export default class EmailDialog extends React.Component {
         break;
       case 'Enter':
         if (!isSuccess && !isError) {
-          this.handleSubscribe(this.state.email);
+          // If event is bubbled up from the input form,
+          // do nothing and let the HTML form handle it.
+          if (e.target !== e.currentTarget) {
+            return;
+          }
+          // Submit only if privacy checkbox is checked.
+          if (this.state.privacy) {
+            this.handleSubscribe(this.state.email);
+          }
+          // TODO: Else, show notification that the checkbox
+          // needs to be checked to proceed (#3042)
         } else if (isSuccess) {
           this.continue(e);
         } else if (isError) {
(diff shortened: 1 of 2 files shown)


lmorchard:
Looks good to me... I think the caveat is enough of an edge case to get the issue fixed for now

PR #3078, opened 2017-11-15
DO NOT MERGE Fix #2741 Rename stage-static and production-static to just stage and production

### LINKED ISSUE #2741: Rename branches from stage-static and production-static to stage and production
We used the names stage-static and production-static during the switchover from the dynamic site to the static site. We might as well switch back to the names stage and production to be more consistent with the rest of the experiments now.

fzzzy:
We should wait to merge this until ops has deployed https://github.com/mozilla-services/cloudops-deployment/pull/1359


fzzzy:
This is just trivial documentation changes, so I am just going to merge it myself.

PR #3081, opened 2017-11-17
Update many dependencies

fzzzy:
Updated all the deps except:

- eslint
- eslint-config-airbnb
- eslint-plugin-flowtype
- flow-bin
- gulp-eslint
- uglifyjs-webpack-plugin

And also did not update to react 16. Everything else is updated and works.

Also, I reduced the required flow coverage from 60 to 50 in npm run test:all, which was only run by hand and wasn't passing at 60 for the last while.


lmorchard:
Looks good, tests pass. Giving the site a quick look through and then will probably just merge

PR #3050, opened 2017-11-09
2956 photon modals

meandavejustice:
fixes #2956

PR #3082, opened 2017-11-19
3059 Email Dialog "Enter" key validation

kgeorgiou:
The "Enter" key now fires a click event on the submit button, instead of directly calling `handleSubscribe(...)`. This approach initially struck me as "hacky", but it follows the [WHATWG spec](https://html.spec.whatwg.org/#implicit-submission): 
> If the user agent supports letting the user submit a form implicitly (for example, on some platforms hitting the "enter" key [...], must cause the user agent to fire a click event at that default button.

Removed 2 of the tests that I previously wrote due to issues with Enzyme and ref callbacks.

### DIFF
--- a/frontend/src/app/components/EmailDialog/index.js
+++ b/frontend/src/app/components/EmailDialog/index.js
@@ -25,6 +25,7 @@ export default class EmailDialog extends React.Component {
   state: EmailDialogState
 
   modalContainer: Object
+  submitButton: Object
 
   constructor(props: EmailDialogProps) {
     super(props);
@@ -37,9 +38,7 @@ export default class EmailDialog extends React.Component {
   }
 
   componentDidMount() {
-    if (this.modalContainer !== undefined) {
-      this.modalContainer.focus();
-    }
+    this.focusModalContainer();
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


fzzzy:
@kgeorgiou Unfortunately I introduced some conflicts with this PR. Sorry about that.

Do you know why the changes checking for this.modalContainer being null are required? I had to do a similar thing and I don't understand why.

kgeorgiou:
Hey @fzzzy, no worries, I'll take a look later today. 

For refs resolving to null in Enzyme: https://github.com/airbnb/enzyme/issues/316#issuecomment-208521009

But either way, I think always ensuring a ref isn't null before performing an action on it is a good thing.

meandavejustice:
@fzzzy This will likely conflict with #3050

meandavejustice:
@kgeorgiou We can try and get this merged now that our updates to modals are complete. There is a conflict here, let me know if you have any questions with resolving it.

kgeorgiou:
@meandavejustice sounds good, I'll take a look and resolve the conflict later today.

kgeorgiou:
That review request from flodolo was uninteded.

fzzzy:
Looks great. Thanks.

Issue #3088, opened 2017-11-22
Write an integration test to cover uninstalling the addon (the retire page)


fzzzy:
We test installing the add-on in the integration tests, but we don't test uninstalling it. We should.

ghost:
Marking this as a good first issue because CI will run these tests for you, you'll just need to write the tests.

b4handjr:
I actually just wrote a test for this...

PR #3089, opened 2017-11-22
Fix #2894 Upgrade to React 16. Delete some tests that don't pass after upgrading react, but replace one of them with an integration test.

### LINKED ISSUE #2894: Upgrade to React 16
[React 16 was released](https://reactjs.org/blog/2017/09/26/react-v16.0.html?utm_source=reactnl&utm_medium=email), and it looks like a lot of good stuff. [I started a branch to try upgrading](https://github.com/lmorchard/testpilot/tree/react-16). 

But, I ran into at least one dependency issue - [`html-react-parser` seems to need an upstream fix](https://github.com/remarkablemark/html-react-parser/issues/43).

So, I'm filing this issue and deferring further work on an upgrade for now. There might be more dependencies that need upgrades.

fzzzy:
- One of the tests is to make sure the header is "sticky" on the
experiment detail page when the add-on is installed. Wrote an
integration test to cover this.
- Three of the tests are related to the email dialog which already has
coverage in the integration tests.
- One of the tests is related to the retire page, which doesn't have
coverage. I opened a new issue for this.
https://github.com/mozilla/testpilot/issues/3088


fzzzy:
This branch is based on https://github.com/mozilla/testpilot/pull/3081 so it would be good to merge that pr first.

frontend/src/app/components/EmailDialog.js
```diff
@@ -35,7 +35,9 @@ export default class EmailDialog extends React.Component {
   }
 
   componentDidMount() {
-    this.modalContainer.focus();
+    if (this.modalContainer !== undefined) {
```
lmorchard, line 38: Noticed this change in a bunch of spots - did we end up with a race condition here or something? This should always end up [with a reference](https://github.com/fzzzy/testpilot/blob/1910271403f75fca1a52cc4664b2ab6f239908e5/frontend/src/app/components/EmailDialog.js#L48) when the component has mounted. Wondering under what conditions it doesn't, and if that's a different problem possibly related to React 16

lmorchard (approved):
LGTM pending resolution of the conflict in package.json

frontend/src/app/components/EmailDialog.js
fzzzy: I, too, would like to know why this happens, but I couldn't figure it out, so this change was just making it work without really knowing why. Interestingly, @kgeorgiou seems to have made similar changes in https://github.com/mozilla/testpilot/pull/3082/files. I thought the behavior change was needed because of the update to enzyme but if @kgeorgiou needed to make the same change without updating react and enzyme, ... ?

fzzzy:
Conflicts resolved, tests pass. Merging.

frontend/src/app/components/EmailDialog.js
kgeorgiou: I recently found out that for refs to work in Enzyme the components have to be rendered with `mount`, instead of `shallow` (https://github.com/airbnb/enzyme/issues/316#issuecomment-208521009). 
So by using `subject = mount(<SomeComponent {...{ onDismiss, sendToGA, getWindowLocation }} />);` will ensure that refs are evaluated as expected.

PR #3103, opened 2017-11-28
2956 photon modals

fzzzy:
Attempting to debug the test failures on Circle. Do not merge.


fzzzy:
Ok, looks like this is good to go now!

PR #3104, opened 2017-11-28
Updated circleci to version 2.0

### LINKED ISSUE #3099: Update to circleci 2.0 config
Circle CI's new 2.0 config allows for parallel builds that would help speed up our build times. I think it would be a good idea to update to this version.

b4handjr:
Fixes #3099

I think I got everything to work. Total build time with the integration tests is about 8:30, down from about 11 currently. 

I don't think I can test a deploy on my fork though...that is my only concern.
You can view the builds here: https://circleci.com/gh/jrbenny35/testpilot/tree/update_circleci_2

### DIFF
--- a/bin/circleci/build-addon.sh
+++ b/bin/circleci/build-addon.sh
@@ -2,4 +2,3 @@
 set -ex
 cd addon/
 npm run package
-cp *.rdf *.xpi $CIRCLE_ARTIFACTS/
--- a/bin/circleci/build-frontend.sh
+++ b/bin/circleci/build-frontend.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -ex
 NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 ENABLE_DEV_LOCALES=1 npm run static
-zip -r $CIRCLE_ARTIFACTS/frontend.zip frontend/build
+zip -r frontend.zip frontend/build
--- a/bin/circleci/install-test-dependencies.sh
+++ b/bin/circleci/install-test-dependencies.sh
@@ -12,16 +12,22 @@ GECKODRIVER_URL=$(
   python -c "import sys, json; r = json.load(sys.stdin); print([a for a in r['assets'] if 'linux64' in a['name']][0]['browser_download_url']);"
 );
 
+
 curl -L -o geckodriver.tar.gz $GECKODRIVER_URL
 gunzip -c geckodriver.tar.gz | tar xopf -
 chmod +x geckodriver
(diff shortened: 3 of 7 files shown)


bin/circleci/build-addon.sh
```diff
@@ -2,4 +2,3 @@
 set -ex
 cd addon/
 npm run package
-cp *.rdf *.xpi $CIRCLE_ARTIFACTS/
```
fzzzy, line 5: Why did you remove this? It would be useful to have the addon in CIRCLE_ARTIFACTS for debugging stuff.

fzzzy: Is the default directory the artifacts dir now?

b4handjr: Yeah I added it back. Its in the ```circleci.yml```. If you look at the ```build``` step you can see the artifacts.

b4handjr: Artifacts: https://circleci.com/docs/2.0/artifacts/#uploading-artifacts

fzzzy:
I had hoped to try landing this yesterday, but I was unable to. Unfortunately I think this means this will have to wait until January to land, as next week is the Austin all-hands, and then the week after that is a half week, and then we are off until January. I have some other work that is related to an OKR that is higher priority.

lmorchard (approved):
This all looks pretty good to me, too. Tempted to just merge and see how the dev deployment goes, since it would be easy enough to revert. Also thinking that sounds like a thing worth waiting until after the new year just for sanity's sake

fzzzy:
Yeah, for sanity I thought it would be best to wait until next year in case something breaks.

fzzzy:
Ok, now that we did an uneventful release yesterday with all of the code from the end of last year, I think it is time to just try merging this. I will revert if anything weird happens.

fzzzy:
Ok, unfortunately it failed in a weird way, so I'm just going to revert for further investigation.

fzzzy:
@jrbenny35 Unfortunately this failed in a very strange way, in a different way each time I re-ran the tests:

The first run failed because of some missing .deb packages (???)

https://circleci.com/gh/mozilla/testpilot/6481

Then the rebuild failed because ./bin/circleci/install-test-dependencies.sh didn't exist (????)

https://circleci.com/gh/mozilla/testpilot/6484

Then a rebuild without cache failed in the same way:

https://circleci.com/gh/mozilla/testpilot/6484

Can you please make a new branch, squash the old branch into it, and then make a new pull request? Then we can see if the PR passes or fails. Something strange is going on, but I don't know what. Sorry for the extra work. If you don't have time, I can do the work of rebasing and squashing, just let me know.

Issue #3024, opened 2017-11-03
Upgrade uniformly to node LTS v8.9.0


lmorchard:
We [specify an old Node version (v4.4.0) in our `package.json`](https://github.com/mozilla/testpilot/blob/master/package.json#L13). 

We use [v6.9.1 in `circle.yml`](https://github.com/mozilla/testpilot/blob/master/circle.yml#L14) for build & test on our dev server.

We use [v6.2.0 in cloudops scripts](https://github.com/mozilla-services/cloudops-deployment/blob/5b80e655a43869dec2a0c9b8cb3bd2e54758939d/projects/testpilot/pipeline-prod.groovy#L52) for prod & stage (this link will 404 for anyone not invited to that repo).

Current LTS [is v8.9.0](https://github.com/nodejs/Release#release-schedule) - maybe we should get on that train? This will involve updating `package.json` and `circle.yml` and coordination with ops to upgrade.

Not sure if this will break any of our build scripts, but fixing such breakage should be done too.

fzzzy:
I'd really like to require 8.9. I think we can get rid of the babel part of gulpfile.babel.js if we do that, which should speed most operations up.

lmorchard:
Oh hey, this is done!

PR #3110, opened 2017-11-29
Update to Node 8.9.0, drop gulp-connect, use webpack-dev-server locally

### LINKED ISSUE #3098: Drop gulp-connect, move to standalone connect-based local dev server
We use gulp-connect with [a custom middleware](https://github.com/mozilla/testpilot/blob/master/frontend/tasks/server.js) for serving things up in local dev. We want to move away from gulp and gulp-connect is getting stale. A good interim step would be to move the middleware into a plain standalone server. Blue sky long-term might be a move to webpack-dev-server, but not sure that's doable yet.

lmorchard:
- Update to Node 8.9.0

- Remove gulp server tasks, drop gulp-connect dependency

- Add standalone static dev-server using HTTPS and our custom middleware

- Switch to webpack-dev-server for local development

- Add some response logging via morgan

- Run webpack independently of gulp, drop webpack-stream dependency

- Prevent gulp renaming images with filenames already hashed by webpack

- Simplify filenames produced by webpack

- Add dist/ directory to frontend.zip circleci build artifact

- Tweaks to npm scripts to reflect new server changes

Fixes #3098
Issue #3024


lmorchard:
We *might* want to wait a deployment before merging this one if it looks okay: Part of #3024 requires updating ops scripts to use node 8.9.0 in building for stage & prod. That'll need to be coordinated with @relud on the day of push, I think. Changing it before merging this will break things and not changing it after merging this will break things.

bin/circleci/build-frontend.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -ex
 NODE_ENV=production ENABLE_PONTOON=1 ENABLE_DEV_CONTENT=1 ENABLE_DEV_LOCALES=1 npm run static
-zip -r $CIRCLE_ARTIFACTS/frontend.zip frontend/build
+zip -r $CIRCLE_ARTIFACTS/frontend.zip frontend/build dist
```
lmorchard, line 4: Just a tweak here to include the production-style build in the artifact, after gulp-rev-all hashes filenames.

frontend/tasks/dist.js
```diff
@@ -38,16 +39,9 @@ gulp.task('dist-thumbnail-images', () => {
 });
 
 gulp.task('dist-build', done =>
-  runSequence('clean', 'build', 'dist-rev-assets', 'dist-thumbnail-images', done));
+  runSequence('dist-rev-assets', 'dist-thumbnail-images', done));
 
 gulp.task('dist-watch', ['watch'], () =>
   gulp.watch(config.DEST_PATH + '**/*', ['dist-rev-assets']));
 
-gulp.task('dist-server', () => connect.server({
```
lmorchard, line 46: FWIW, I don't think anyone was using this gulp task - and it was non-SSL anyway.

package.json
```diff
@@ -127,10 +129,17 @@
   },
   "scripts": {
     "bootstrap": "npm install && cd addon && npm install && cd .. && gulp",
-    "start": "cross-env USE_HTTPS=1 NODE_ENV=development ENABLE_DEV_LOCALES=1 ENABLE_DEV_CONTENT=1 gulp",
+    "start": "cross-env NODE_ENV=development ENABLE_DEV_LOCALES=1 ENABLE_DEV_CONTENT=1 npm-run-all build:gulp watch",
```
lmorchard, line 132: Having split webpack out into its own build step complicated things a tiny bit: Gulp having built `experiments.json` is a prerequisite for webpack, because we `require()` it from the build directory in `frontend/src/store.js`

So, we have to run the gulp part of the build through once before starting watchers.

frontend/lib/dev-server.js
```diff
@@ -0,0 +1,58 @@
...
+  // If no dot and trailing slash, try rewriting /path to /path/index.html
+  if (
+    pathname.indexOf('.') === -1 &&
+    pathname.substring(pathname.length - 1) !== '/'
+  ) {
+    parsed.pathname = pathname + '/index.html';
+    req.url = url.format(parsed);
+  }
+
+  if (pathname.indexOf('.json') !== -1) {
```
pdehaan: Instead of `.indexOf('.json') !== -1 {}`, should we use ES6'esque [`String#endsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith), or Node [`path.extname()`](https://nodejs.org/dist/latest-v8.x/docs/api/path.html#path_path_extname_path)?

frontend/lib/dev-server.js
```diff
@@ -0,0 +1,58 @@
...
+// TODO: CSP copied from bin/deploy.sh - find a way to extract automatically?
+const CSP = `default-src 'self'; connect-src 'self' https://sentry.prod.mozaws.net https://www.google-analytics.com https://ssl.google-analytics.com https://basket.mozilla.org https://analysis-output.telemetry.mozilla.org; font-src 'self' code.cdn.mozilla.net; form-action 'none'; frame-ancestors 'self' https://pontoon.mozilla.org; img-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com https://www.google-analytics.com; object-src 'none'; script-src 'self' https://pontoon.mozilla.org https://ssl.google-analytics.com; style-src 'self' https://pontoon.mozilla.org code.cdn.mozilla.net; report-uri /__cspreport__; frame-src https://www.youtube.com;`;
+
+const DevServerMiddleware = (req, res, next) => {
+  const parsed = url.parse(req.url);
+  const { pathname } = parsed;
+
+  // If no dot and trailing slash, try rewriting /path to /path/index.html
+  if (
+    pathname.indexOf('.') === -1 &&
```
pdehaan: `if (!pathname.includes('.') && !pathname.endsWith('/')) {...}`?

fzzzy:
This is awesome! Let's wait until the next sprint to land this.

lmorchard:
Rebased against master, addressed comments, DRYed out the HTTPs options a bit

lmorchard:
Going to just land this, so we can move forward with node 8 shenanigans on stage soon.

Edit: Well, maybe I'll try a rebase against master to re-run tests just to be sure

lmorchard:
Okay, build & tests still work after a rebase, I'm gonna hit the button

lmorchard:
Hmm, master build failed, will rebuild without cache

PR #3118, opened 2017-12-04
Fix #3096 Fix the broken email signup dialog by not checking cookies until componentDidMount.

### LINKED ISSUE #3096: Experiments page UI breaks when you return to it after enabling an experiment together with Test Pilot add-on
**[Affected versions]:**
- Firefox 57.0 and above

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**

**[Steps to reproduce]:**
2. Scroll down and click on any active experiments cards except Send.
3. Click the "Install Test Pilot & Enable <experiment_name>" button.
4. Click the "Add" button on the doorhanger displayed for experiment enable.
5. Dismiss the experiment tour and click the "Firefox" logo on top left.
6. Observe the page behavior. 

**[Expected result]:**
- You are redirected to the Test Pilot experiments page where all the page elements are correctly displayed and the welcome + newsletter sign up pop up is displayed over a page overlay.

**[Actual result]:**
- You are redirected to the Test Pilot experiments page, but almost all the page elements are wrongly displayed and the welcome + newsletter pop up is wrongly displayed in page. 

**[Notes]:**
- This also reproduces if the Test Pilot toolbar button is clicked instead of the Firefox logo.
- This issue does not reproduces on Production server so it's probably a recent regression.
- The issue does not reproduce if you use "Back" button for navigation.
- Screen recording of the issue: https://goo.gl/SLesYU.
- Screenshot: 

test pilot ui break

fzzzy:
Checking them in the constructor will always fail when rendering the page html at build time.

Honestly I have no idea why this fixes the problem, but it does and I had a hunch to try it because it is just cleaner to do things that require browser state in componentDidMount.


lmorchard:
Hmm, this makes me wonder if we had some kind of race condition and moving into componentDidMount just gave it a better delay. Can't think of why that would happen with cookies though :/

fzzzy:
@lmorchard I suspect what happens is this:

The html nodes are rendered into the body at build time, where cookies are not available and showEmailDialog will be false. When the client tries to hook up the html nodes to the live nodes, showEmailDialog was true, causing the problem. (It didn't happen before React 16, so something changed in the way react handles this case.) By moving the setState out of the constructor into componentDidMount, showEmailDialog will be false in the constructor, matching build time, and then componentDidMount will set showEmailDialog to true.

lmorchard:
Something that might be worth looking into as a future enhancement is pulling all this kind of stuff into a componentDidMount handler in the App container. That way, any browser state changes (i.e. from non-existent at build to available on client) can be discovered in one spot rather than in every page

PR #3123, opened 2017-12-04
Fix #3014 Clean up old addon polling code.

### LINKED ISSUE #3014: Rebase the patches cleaning up the addon polling code that was written while diagnosing bug 2997
The code in question is here:

https://github.com/mozilla/testpilot/pull/3007/files

There are a lot of good cleanups of old stuff here, but it doesn't pass the integration tests.

fzzzy:
Rescue some of the addon polling cleanup code from an old branch. This will most likely fail the integration tests, so it is a wip until they are fixed.

### DIFF
--- a/frontend/src/app/containers/ExperimentPage/index.js
+++ b/frontend/src/app/containers/ExperimentPage/index.js
@@ -396,23 +396,7 @@ export class ExperimentDetail extends React.Component {
       return;
     }
 
-    installAddonPromise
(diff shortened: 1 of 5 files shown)


lmorchard (approved):
LGTM :+1: The onEnabled / onDisabled event handling could use some DRYing up, but this PR doesn't itself introduce the similar code chunks

PR #3132, opened 2017-12-05
fix gradient

johngruen:


johngruen:
Self merging since this is changing a hex value

frontend/src/styles/_theme.scss
```diff
@@ -1,5 +1,5 @@
 $theme-dark: $blue-90;
-$theme-light: rgba(37, 0, 62, .8);
```
johngruen, line 2: @youwenliang changing this to the same value w/o an opacity layer since that made the gradient look stripe-y

youwenliang: sure! sounds good

SoftVision-PaulOiegas:
@johngruen @fzzzy It seems that this was merged on dev right after the stage server deploy. Can we cherry-pick this somehow so it reaches production also if we deploy today?

SoftVision-PaulOiegas:
Also, the stripes are still visible in some of the areas but it's clearly an improvement.

Issue #3001, opened 2017-10-30
Fix position of the "Try our other experiments" header


chuckharmston:
To make sure that the new featured experiment UI doesn't cannibalize installations from other experiments, we'd like to make the "Or try our other experiments" header sticky to the bottom of the page if it would be off the page.

![](https://user-images.githubusercontent.com/18019946/31928140-96400df6-b8c9-11e7-9f46-9d4a2bb4ffa4.png)

![](https://user-images.githubusercontent.com/18019946/31928179-b2a71ec6-b8c9-11e7-966e-da4f586efa9d.png)

This is blocked pending UX in #2976.

Pekochen:
Hi Chuck,
Please see the link as below for the measurement.
https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Featured_experiment/#artboard4

please use"arrowhead down 16" for arrow icon. you can download it from
http://design.firefox.com/icons/viewer/

chuckharmston:
No longer blocked on #2976; the work from it just needs to be updated to match the Photon styles, tracked in #3019.

lmorchard:
Totally missed that I got assigned to this - will try to get to this today or monday

lmorchard:
> No longer blocked on #2976; the work from it just needs to be updated to match the Photon styles, tracked in #3019.

@chuckharmston I see #3019 is closed, but I don't see an update for this issue in there. I can probably wing it, but a spec update would be good. The plain header is there in a few of the artboards, but the fixed position state seems missing.

lmorchard:
Also, could we avoid adding a new string here by reusing "Try out these experiments as well" from the bottom of the experiment detail pages?
![capture](https://user-images.githubusercontent.com/21687/33580458-4d2c770e-d91b-11e7-8c34-47c659f0fbed.PNG)

Might not be a great idea, the more I think about it 😞

lmorchard:
Submitted PR #3143 for this. Still looking for input on styles, transitions, l10n

marniepw:
@johngruen talk about at Design meeting.

johngruen:
@ChenMorpheus this design causes overlapping issues with the CTA button, we need to revise.

ChenMorpheus:
@johngruen We'll come up with a proposal which doesn't have a bar.

johngruen:
@ChenMorpheus any progress here?

Pekochen:
[...]

PR #3114, opened 2017-11-30
Featured experiments

meandavejustice:
Opening this up to get some eyes on it while I'm finishing up a couple of tasks

- fixes #3003 
- fixes #3000 
- fixes #3002 
- fixes #2999 
- fixes #2998 
- fixes #3072
- fixes #3149 
- fixes #2696

### Running
In order to run, you can just add `is_featured:true` to one of the current experiments in `content-src` and `npm start`

### Todo
- [x] get rest of the new tests passing
- [x] finish implementing legal modal now that photon modals pr has been merged ( might take some slight reworking of the Modal component)
- [x] tweaks to featured experiment and news feed page for small device views
- [ ] figure out why the icons from ExperimentPlatform component still aren't white cc @johngruen (I tried updating this in inkscape and they appear white for me in there, maybe this is some sort of firefox svg rendering issue ?)

Issue #3039, opened 2017-11-08
Animation for scroll hint(Sticky banner)


Pekochen:
Hi @chuckharmston 
I made an animation sample for scroll hint.
Please check here for the animation!
https://www.dropbox.com/sh/jv3fk7kg3ehxo99/AADJiUSQPvToyocey6l40jpaa?dl=0
I also write down some information you may need.
![animation preference](https://user-images.githubusercontent.com/8252151/32541282-2e94f350-c4aa-11e7-9481-725266157624.png)
Arrow keeps looping. When you roll over, the arrow will stop and then when the user scrolls down the page; the sticky bar is going to move down and fade out.
When you press the arrow, it will change color to #0094FB. The rest of page(or try other experiments that part) is going up, and the sticky bar is going to move down and fade out at the same time.
I am not sure if the explanation is clear enough, if you have any question, please let me know.
Thanks!

chuckharmston:
Fully blocked on #3001, unassigning until that lands.

johngruen:
I am gonna go ahead an close this, and let's use #3001 to track everything

PR #3143, opened 2017-12-07
[WIP] Add fixed subtitle to home page

lmorchard:
- New Visibility component that changes class name based on visibility

- Move HomePage components into bundle

- Wrap experiments list on home page in Visibility component, styles to
  make the section subtitle appear fixed when the section is out of view

Issue #3001


lmorchard:
Still might need some work on a couple of things:
- [ ] Couldn't find an exact spec for this, so I just guessed on styles
- [ ] Didn't yet change the header content to "Or try our other experiments" - but will need l10n when changed

frontend/src/app/components/Visibility/index.js
```diff
@@ -0,0 +1,116 @@
...
+  isVisible?: boolean,
+  className?: string,
+  children?: Array<any>
+};
+
+type BannerState = {
+  isVisible: boolean
+};
+
+export default class Visibility extends React.Component {
```
lmorchard, line 16: Also "Visibility" might be too general a name for this utility component, since so far it only checks to see if the element is vertically higher than current scroll position. Doesn't do the inverse - which could be handy for simplifying the sticky header on experiment pages. Figured that could be for a future enhancement though

lmorchard:
Also looks like this will collide with PR #3114 thanks to the HomePage components move. But the changes here to those components are small, so we could put this PR off until that one's merged

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3143/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/fa3eabbf8202f827184034b52ed549107511e4ad/

lmorchard:
[Here's a direct link to the feature in the last storybook build](http://testpilot-storybook.dev.mozaws.net/fa3eabbf8202f827184034b52ed549107511e4ad/?selectedKind=HomePage&selectedStory=Fixed%20off-screen%20subtitle&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel)

SoftVision-PaulOiegas:
Just an observation. After looking on storybook animation for the fixed subtitle I've noticed that the attach and detach part is not really smooth when slow scrolling. 
See this recording https://goo.gl/jAF5DY.

lmorchard:
Yeah, there is no animation or transition at all right now. Didn't want to put too much into that without an idea of how the styles should work overall

fzzzy:
I did a very similar thing for pageshot -- what I did was set visibility: hidden on the element which is in the scrolling part (so that the page would continue to lay out as if that element was there), and then wait until that element is exactly where the non-scrolling element is. Then set visibility: visible on the scrolling element and display: none on the fixed element.

frontend/src/app/containers/HomePage/index.scss
```diff
@@ -0,0 +1,27 @@
...
+  h1,
+  h2 {
+    cursor: pointer;
+    width: 100%;
+    margin: 0;
+    padding: $grid-unit * 2 0;
+    position: fixed;
+    left: 0;
+    bottom: 0;
+    background: linear-gradient($transparent-black-2 40px, $transparent-black-15);
```
fzzzy, line 13: This background being transparent doesn't really work; it can clash with whatever happens to be in the page.
screen shot 2017-12-18 at 11 26 32 am

fzzzy (changes requested):
At the very least I think the background of the banner needs to be opaque. If, once the background is opaque, the fixed element could be made to not disappear until the scrolling element is exactly in the same place, that would be good, but I tried to make that change to the logic and it did not appear to be straightforward to accomplish.

lmorchard:
Looks like this animation spec is related to this issue, but has a curve to the banner I haven't seen before. Thinking that's the UX spec I'm missing https://github.com/mozilla/testpilot/issues/3039

frontend/src/app/containers/HomePage/index.scss
lmorchard: Yeah, like I said in a previous comment, I didn't put much work into the styles because I don't have a UX spec for it. I don't think the styles I have are good at all. This is mainly just to get it "functional"

lmorchard:
Just added a "WIP" to the title - let's not merge this yet, it needs some UX work

frontend/src/app/components/Visibility/tests.js
```diff
@@ -0,0 +1,78 @@
+/* global describe, beforeEach, afterEach, it */
```
pdehaan, line 1: Instead of specifying the methods individually, we could just set the `mocha` ESLint environment: `/* eslint-env mocha */`

Or, possibly look into using ESLint config `overrides` with a glob to automatically set the `mocha` env for `frontend/src/**/tests.js` files:
https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns

johngruen:
Okay, finally looking at this

frontend/src/app/components/Visibility/tests.js
johngruen: Yeah, that's a good idea....could you file an issue to track?

johngruen: b/c we do this in all the places

lmorchard:
Going to close this for now, and resubmit with updated UX

PR #3145, opened 2017-12-08
finalize component structure

johngruen:
i finished moving all the component CSS and Tests. Nothing new here, just cleaning up before the holidays :)


fzzzy:
omg yesssss

Issue #2617, opened 2017-07-14
Don't render empty experiment status messages


pdehaan:
Found while doing view-source and throwing it through an HTML linter https://html5.validator.nu/.

Currently we have source like this:

```jsx
<header>
  <div>
    <h3>{title}</h3>
    {subtitle && <h4 data-l10n-id={this.l10nId('subtitle')} className="subtitle">{subtitle}</h4>}
    <h4>{this.statusMsg()}</h4>
  </div>
  {this.renderFeedbackButton()}
</header>
```

The problem with this is that if `{this.statusMsg()}` returns an empty string or whatever, we're left with an empty `<h4>` tag.

> ### Warning: Empty heading.
>
> 1. From line 6, column 2663; to line 6, column 2684
>     ```html
>     ... Tabs</h3><h4 data-reactid="41"></h4><
>     ```
>
> 1. From line 6, column 3651; to line 6, column 3672
>     ```html
>     ... iners</h3><h4 data-reactid="53"></h4><
>     ```
>
> 1. From line 6, column 4670; to line 6, column 4691
>     ```html
>     ...n Vid</h3><h4 data-reactid="65"></h4><
>     ```

It _may_ be better off checking if `this.statusMsg()` is truthy before rendering it in a parent `<h4>` tag, similar to how we render the `subtitle`

kovsk1:
Hi @pdehaan , I'm a student at Seneca college learning open source, and I was hoping to work on this bug for my course.  If no one else is currently working on it, I'd like to give it a try.

SoftVision-PaulOiegas:
Hi @naperkovskiy, not sure if Peter has seen your message. But go for it! Any contributions are welcomed.

kovsk1:
@SoftVision-PaulOiegas Thank's, I'l get started on it asap.

kovsk1:
Hi @pdehaan, I fixed the issue but Im having trouble committing it due to gitignore file is set to ignore frontend/buld directory. Can I modify gitignore or maybe there is another way to skip it?

meandavejustice:
@naperkovskiy The frontend/build directory is generated. You shouldn't need to commit any files into that directory. Wanna post a git diff or patch into a [gist](https://gist.github.com/) and we'll take a look?

kovsk1:
@meandavejustice Sure, here is a link to gist. I put the change that I did in the comment.
https://gist.github.com/naperkovskiy/cab36258200e32715bd5a249f1af3ec4

[...]

PR #3150, opened 2017-12-13
Fixing issue #2617 

kovsk1:
I have edited the the following file so that it checks if the statusMsg() is truth and if it is not it will not render it. You can test it by inspecting the elements or running it through [validator](https://html5.validator.nu/)


meandavejustice:
Thanks @naperkovskiy :) :+1:

PR #3155, opened 2017-12-21
fix all the styles

johngruen:
This touches a lot of files b/c i also linted all of the CSS.


johngruen:
@meandavejustice this should fix all of the styles for your pull request

frontend/src/app/components/NewsletterFooter/index.scss
```diff
@@ -4,41 +4,44 @@
...
-    max-width: 360px;
     @include respond-to ('big') {
       margin-right: 55px;
     }
+
+    max-width: 360px;
+    width: 45%;
   }
 
   header {
```
pdehaan, line 43: Should this be combined w/ the `header {}` block up on L26-32?

meandavejustice: I'll fix this in #3114

Issue #3013, opened 2017-11-01
Remove /legacy page and redirect to home page from URL


fzzzy:
We sometimes get errors from really old versions of firefox visiting /legacy. There is nothing dynamic on that page; we could simply make it a static page and not include any js.

ghost:
Let's just get rid of this page.  (redirect to home page off this URL).

ppatel221:
Hey, Im a student who is taking an Open source development class at Seneca College and I'm currently looking to contribute to a project. Can I get started on this bug?

meandavejustice:
@ppatel221 Go for it :) I've updated the title to indicate what needs to be done now, feel free to ping us if you have any questions.

(sorry for the delayed response)

PR #3167, opened 2017-12-29
Remove Legacy page issue 3013

marvinsanchez1:
Fix issue #3013 Remove /legacy page and redirect to home page from URL

Do you also want to remove the following folder path testpilot/frontend/src/app/containers/LegacyPage ??

### DIFF
--- a/frontend/src/pages/legacy.js
+++ b/frontend/src/pages/legacy.js
@@ -2,8 +2,8 @@
 import React from 'react';
 
 import inject from '../app/lib/inject';
-import LegacyPage from '../app/containers/LegacyPage';
+import HomePage from '../app/containers/HomePage';
 
 export default function create() {
-  return inject('legacy', <LegacyPage />);
+  return inject('home', <HomePage />);
 }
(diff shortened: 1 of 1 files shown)


lmorchard:
> Do you also want to remove the following folder path testpilot/frontend/src/app/containers/LegacyPage ??

Yeah, I think this is the only place where this container is used, so we should remove it. Also looks like there's a test for it that can be deleted. [Did a quick repository search for "LegacyPage" that seems to confirm](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=LegacyPage+&type=)

fzzzy (changes requested):
This isn't going to solve the underlying problem, which is that users that are sent to the /legacy page produce exceptions from running the site javascript that are recorded in Sentry. I would prefer if frontend/src/pages/legacy.js was removed entirely, and the /legacy url was created using the same technique as the other static pages, /about/, /privacy/, and /terms/. Simply copying the text content into a en-US.md file inside a /legacy/ directory might be enough to do it.

chuckharmston:
Closing this since we haven't heard back in a few weeks. @msanchez5 feel free to reopen if you ever return to it.

Issue #3180, opened 2018-01-04
Drop bootstrap prefs reading/writing code


ghost:
This is bootstrap-y:

> Prefs (read-only) - legacy method for configuring add-on for current environment (i.e. dev / stage / prod)

> Prefs (read/write) - Backing up and enabling prefs for telemetry, healthreport, beacon.enabled

lmorchard:
I think this is a thing we can just drop. If I understand recent changes to Telemetry correctly, these pref changes won't be respected for us anyway in the near future

lmorchard:
Oh, wait, there's a pref unrelated to Telemetry that we still use: `testpilot.env`

If / when we lose access to prefs, we need some other mechanism for deciding which environment the add-on should use. Maybe settings in about:addons? I think we avoided that at first because we didn't want to make the setting easily discoverable

fzzzy:
Would it be acceptable to have the pref in about:addons if we just label it as "for development of testpilot use only"?

meandavejustice:
bumping this to get discussion going again,

@fzzzy I think the pref in about:addons is reasonable but not preferable... I wonder if we could show it only on dev builds, check for an environment variable, though that could be a bit of a pain as well.

PR #3198, opened 2018-01-05
Work without testpilot.env

fzzzy:


lmorchard (approved):
Works for me. Moving forward, [we should look at dropping testpilot.env altogether](https://github.com/mozilla/testpilot/issues/3180#issuecomment-355647621).

PR #3204, opened 2018-01-08
Fix 404s for images on local dev server

lmorchard:
I think this should fix the issue with missing images.

Not sure why it needs both publicPath and outputPath, but that combo seemed the only thing to fix it.

You can test this in two ways:
1. `npm start` - for the usual local server with file watching
1. `npm run static && npm run server` - for a static rendering of the site like we do on testpilot.dev.mozaws.net

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -143,7 +143,9 @@ module.exports = {
             options: {
               hash: 'sha512',
               digest: 'hex',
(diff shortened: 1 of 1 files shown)


lmorchard:
(requested @johngruen as a reviewer, since that's who also just noticed the issue)

PR #3206, opened 2018-01-08
Remove server-generated page container before rendering client-side

### LINKED ISSUE #3172: Changing content requires killing server and restarting
This is probably a dev papercut only, but as i add new items to the news feed, i notice things like assets and URLs getting out of step with the news item and/or news items not showing up at all.

Not sure if this is to do with gulp/webpack getting into a fight, but i figured i'd file anyway since it's a bit of a nuisance. 

## STR
1. start the server locally
2. add a new item to the `news_updates.yaml` file
3. refresh the page

## Expected
The news feed page should reflect the new set of updates accurately

## Actual
Images and Icons do not seem to align with the text in news feed items

Again, this can be remediated by re-running npm start

lmorchard:
- [React 16 Performs Less Strict Client-Side Checking][ssr]. That means
  when client-side rendering differs from what the server offers, React
  can do unexpected things.

  The fix in this PR, albeit drastic, is to remove the server rendering
  from the DOM before applying client-side

- Use more unique key props for news items, based on slugs - which didn't
  fix the original issue as expected, but is a good idea anyway.

[ssr]: https://hackernoon.com/whats-new-with-server-side-rendering-in-react-16-9b0d78585d67?token=nVWMJUyuYkp0s6eL

Fixes #3172

### DIFF
--- a/frontend/src/app/components/UpdateList/index.js
+++ b/frontend/src/app/components/UpdateList/index.js
@@ -157,9 +157,9 @@ export default class UpdateList extends React.Component {
            <h1 className="update-list-heading">Latest Updates</h1>
          </Localized>}
         <LayoutWrapper flexModifier="column-center">
-          {shownNewsUpdates.map((update, index) =>
+          {shownNewsUpdates.map(update =>
             <Update
-              key={index}
+              key={update.slug}
               sendToGA={sendToGA}
               update={update}
               experiment={experimentsBySlug[update.experimentSlug]}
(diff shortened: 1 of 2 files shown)


lmorchard:
Man, this one took me awhile to sort out, but I think it's a potential gotcha waiting to break pages. 

For example, if the static server-side HTML has an experiment filtered out due to locale or graduation date - the client-side rendering may do odd things when it tries to apply on top. Links and images may appear alongside the wrong textual content, as one possible effect. Using more unique & specific key props in lists of components do not seem to mitigate this effect.

meandavejustice:
@lmorchard Does this affect perceived page load at all?

lmorchard:
> Does this affect perceived page load at all?

Not as far as *I* could perceive, though FWIW it turns out deleting the server content and re-rendering completely client-side what React 15 was doing already and React 16 stopped doing.

fzzzy:
This looks good, and we better get it merged before it conflicts with the lint patch.

PR #3208, opened 2018-01-09
Use mozilla-central lint config.

fzzzy:
Many .js files still need to be converted to the mozilla-central lint
config, so for now don't fail lint on lint errors.


chuckharmston (approved):
This seems great. Can I request some documentation describing:

- The linting rules we use (links out to plugin docs are fine)
- Info about the migration (link to bug)
- How to migrate a module (i.e. how to run eslint on a single file, etc)
- That we've disabled lint failures.

Otherwise :+1:

lmorchard (approved):
Worked for me locally, looks like it worked fine in CI

frontend/src/app/config.js
```diff
@@ -1,40 +1,40 @@
 const defaultConfig = {
   isDev: true,
   minFirefoxVersion: 54,
-  experimentsURL: '/api/experiments.json',
+  experimentsURL: "/api/experiments.json",
   usageCountsURL:
-    'https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json',
+    "https://analysis-output.telemetry.mozilla.org/testpilot/data/installation-counts/latest.json",
```
lmorchard, line 6: FWIW, since we've changed this rule from single- to double-quote, you might be able to fix up some files with prettier. That seemed to be the only rule that tripped against eslint for me up to now. If that's the bulk of the lint failures with the new rules, I'd guess prettier is not incompatible with them

fzzzy: It was super easy to do with just find/replace in my editor.

There were a few more buried in there that looked like these:

160:9 error use element.remove() instead of element.parentNode.removeChild(element) mozilla/avoid-removeChild
133:48 error 'module' is not defined no-undef
608:5 error 'global' is not defined no-undef
37:7 error unnecessary '.call()' no-useless-call

Issue #3111, opened 2017-11-29
Move everything to webpack, remove gulp


lmorchard:
Would be nice to remove gulp from our build tools altogether and just run everything through webpack. I think we're getting very close to being able to do that. 

Webpack now builds & minifies all of the main app JS, handles CSS & images used by components. PR #3110 pulls webpack out of the gulp tasks and moves over to using webpack-dev-server.

Wanted to get my head around just what this would entail and came up with these TODOs:
- [x] f9989215fb466adff54f217b8abaccf784650665 Move JS linting from [scripts](https://github.com/mozilla/testpilot/tree/master/frontend/tasks/scripts.js) gulp tasks into plain npm scripts
- [x] 3689114618cec0ac5afe5f4d5f4c2f0d08b2d4f7 Move CSS linting from [styles](https://github.com/mozilla/testpilot/tree/master/frontend/tasks/styles.js) gulp tasks into plain npm scripts
- [x] 986a3e2ca4905f0cfdde0100139bd03cab489fab Convert any gulp tasks that simply copy files into using [copy-webpack-plugin](https://github.com/webpack-contrib/copy-webpack-plugin) in webpack config. (e.g. [assets](https://github.com/mozilla/testpilot/tree/master/frontend/tasks/assets.js) and [images](https://github.com/mozilla/testpilot/tree/master/frontend/tasks/images.js) tasks)
- [x] 59793fe209e9fb605d407ae7d64d110307aa3f5d Delete the [scripts gulp task](https://github.com/mozilla/testpilot/tree/master/frontend/tasks/scripts.js), add webpack entry points for legal.js and locale.js in [frontend/src/scripts](https://github.com/mozilla/testpilot/tree/master/frontend/src/scripts)
- [x] d50eb10d64d5984bafadb446ac5cf8f754e4fca6 Finish migrating CSS into component bundles, `import` any standalone styles in App component or somewhere central so webpack picks them up as dependencies. Then, we can delete the [styles gulp task](https://github.com/mozilla/testpilot/tree/master/frontend/tasks/styles.js)
[...]

PR #3211, opened 2018-01-09
Move JS linting out of gulp and into npm scripts

lmorchard:
- Drop gulp-eslint dependency

- Drop self-lint and scripts-lint gulp tasks

- Add onchange dependency to do file watching from npm scripts

- Add/change npm scripts for watch:lint, watch:lint:js, lint,
  lint:styles, lint:js

- Tweak .eslintrc to use node & browser envs

Issue #3111


.eslintrc
```diff
@@ -11,6 +11,10 @@ plugins:
   - react
   - mozilla
 
+env:
```
lmorchard, line 14: Not *super* happy about turning on node & browser env for all files, but I think it's probably okay. 

Mainly needed for gulp & webpack config, could be useful for other JS like gulp tasks (until they're all gone). Could use some refinement maybe to use [file glob overrides in .eslintrc](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns), but that seemed not to work in the YAML format when I tried. Could also be I didn't have the syntax right.

pdehaan: What about just specifying the ESLint environment in the **gulpfile.babel.js** and **webpack.config.js**, a la `/* eslint-env node, browser, whatever */`.
Or if we want to set the node env for a directory, we could just use a nested .eslintrc file (for tests or gulp tasks or whatnot).

PR #3214, opened 2018-01-09
Fix privacy notice checkbox error in newletter form not showing

### LINKED ISSUE #3042: [Newsletter] Notification that informs about the need to check the privacy notice checkbox is no longer displayed
**[Affected versions]:**
- Firefox 56.0.2 and above

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**

**[Steps to reproduce]:**
2. Click the "Install the Test Pilot Add-on" big green button.
3. After TxP "Newsletter sign up" dialog is displayed, click the text field in order to focus it.
4. Enter a valid email address and leave the privacy checkbox unchecked.
5. Click the "Sign Up Now" button and observe the dialog behavior. 

**[Expected result]:**
- The sign up process is not performed and you are notified that you need to check the privacy notice checkbox in order to proceed.

**[Actual result]:**
- No notification is displayed that informs you about the need to check the privacy notice checkbox.

**[Notes]:**
- The same behavior occurs also for the footer Newsletter sign up form.
- Here is a screen recording of the behavior https://goo.gl/9Ussyv.

meandavejustice:
- fixes #3042

form-error

### DIFF
--- a/frontend/src/app/components/NewsletterForm/index.js
+++ b/frontend/src/app/components/NewsletterForm/index.js
@@ -87,7 +87,7 @@ export default class NewsletterForm extends React.Component {
       <label className={this.makeRevealedClassNames()} htmlFor={fieldName}>
         <span className="newsletter-form__check" />
         { this.state.privacyNote ? <Localized id="newsletterFormPrivacyAgreementRequired">
-          <span>
+          <span className="error-msg">
             Please check this box if you want to proceed.
           </span>
         </Localized> : null }
@@ -112,7 +112,7 @@ export default class NewsletterForm extends React.Component {
     }
     return <Localized id='newsletterFormSubmitButton'>
       <button className={classnames("button", "large", this.props.isModal ? "default" : "outline")}
-        ref={this.props.buttonRef}>Sign Up Now</button>
+        ref={this.props.buttonRef} onClick={this.checkPrivacy}>Sign Up Now</button>
     </Localized>;
   }
 
@@ -124,6 +124,12 @@ export default class NewsletterForm extends React.Component {
         </p>
       </Localized>
     );
+  }
+
+  checkPrivacy = () => {
+    if (!this.props.privacy) {
+      this.setState({ privacyNote: true });
+    }
   }
 
   handleSubmit(evt: Object) {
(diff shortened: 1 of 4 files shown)


SoftVision-PaulOiegas:
Shouldn't that error point to the checkbox? 😄 Was doing that before it disappeared.

meandavejustice:
@SoftVision-PaulOiegas I was going off of this, https://github.com/mozilla/testpilot/issues/3042#issuecomment-356171913, but I think you are right, it should point to the checkbox

meandavejustice:
updated position 
checkbox-error-position

chuckharmston:
Meta question: why are we using custom interface elements for things that the browser has defaults for? I think we should be using `required="required"` on the form element instead.

meandavejustice:
@chuckharmston I agree, @fzzzy also brought this up. Sounds like that would be a good issue for us to file with the `dev-papercut` label. Might also become a `good-first-issue`

chuckharmston (changes requested):
I'm not sure we should be adding code + complexity that we know can be done better, simpler. Can you refactor to use HTML form validation instead?

meandavejustice:
@chuckharmston ~~I'm happy to refactor but I'm not sure it makes sense for this issue. We aren't really adding any code in this issue, it's tweaking to fix the existing implementation. I would rather get this merged, as its a minor tweak that gets our signups working, and get the required field version in at a later date.~~

On second thought, it's not too big of a change to switch to the required fields. I'll just make the changes :)

fzzzy:
I know this code previously used required="required". Not sure when it changed.

meandavejustice:
@fzzzy I'll look through history and see if I can just do what we were doing before

chuckharmston:
@meandavejustice is this ready for re-review?

@flodolo looks like we need your review here as well; we're removing a string and are replacing it with native browser validation.

meandavejustice:
@chuckharmston indeed, it is ready for review

flodolo (approved):
In case I don't reply quickly, feel free to go ahead when you're only removing strings ;-)

flodolo:
This is also going to conflict, one way or the other, with #3349. Not sure if Git will be smart enough.

chuckharmston:
screen shot 2018-02-15 at 10 29 02

chuckharmston (approved):
Works great, looks great, and I love deleting code. Thanks for sticking with it!

Issue #3223, opened 2018-01-10
Update new get-started illustrations for non-testpilot user


youwenliang:
There are three new illustrations for the get started section

Current:
![image](https://user-images.githubusercontent.com/8919308/34758721-036b9630-f613-11e7-8be8-8639855cf4d7.png)

Spec:
![image](https://user-images.githubusercontent.com/8919308/34758724-09526880-f613-11e7-9a53-60c5a5aba75b.png)

Assets:
https://github.com/mozilla/testpilot-assets/blob/master/Test_Pilot_Website/01_Measurements/Photonized_Txp_Website_Final/assets/illustration_01.svg
https://github.com/mozilla/testpilot-assets/blob/master/Test_Pilot_Website/01_Measurements/Photonized_Txp_Website_Final/assets/illustration_02.svg
https://github.com/mozilla/testpilot-assets/blob/master/Test_Pilot_Website/01_Measurements/Photonized_Txp_Website_Final/assets/illustration_03.svg
  
Full Spec:
https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Photonized_Txp_Website_Final/#artboard3

SoftVision-PaulOiegas:
@youwenliang One suggestion here. If we are targeting especially the non-firefox users, why not moving the Firefox download button above the illustrations.  You cannot "Get Test Pilot add-on" until you have Firefox installed in the first place.

youwenliang:
Actually, the "get-started" message is also showing for non-testpilot users, not just non-firefox users. I agree with you but I don't think that's a big problem, the button here is just another reminder to download firefox since there's already a download button on the top of the page.

(This is the main download button)
![image](https://user-images.githubusercontent.com/8919308/34811918-64560b6c-f6de-11e7-80ac-e12d96d39438.png)

SoftVision-PaulOiegas:
Yeah, it has been there for 2 years now. 😃 Anyway, it was just a UX suggestion that would apply only for non-firefox users since we will be working on that section soon.

SoftVision-PaulOiegas:
The illustrations were correctly changed. 👍

Issue #3222, opened 2018-01-10
Experiment page button style & CTA bar color


youwenliang:
There are some new updates on the buttons style on the top right corner, and CTA bar color (Voice Fill is enabled)

Current:
![image](https://user-images.githubusercontent.com/8919308/34758645-8b5607c0-f612-11e7-9847-b682b45e9117.png)

Spec:
![image](https://user-images.githubusercontent.com/8919308/34758620-697cb75c-f612-11e7-8414-3c83b1f2ce67.png)

Full spec: https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Photonized_Txp_Website_Final/#artboard20

SoftVision-PaulOiegas:
The bar was green before Photonization. Why did we changed it in the first place? 😕

ChenMorpheus:
@SoftVision-PaulOiegas So sorry about changing it again. We changed the color to the defined green in the Design system and followed the CTA button color to make it consistent. Besides, the original teal color would confuse users with the color of Voice Fill. Please find the link for more details, thanks.
https://design.firefox.com/photon/visuals/color.html

SoftVision-PaulOiegas:
I see. Thanks for clarification. 

_Goes back to TestRail to update the testcase color._ 😄

SoftVision-PaulOiegas:
Both bar and buttons were changed to match the designs. However, the vertical size of the green bar seems to be off by 4px (32 vs 36).

@youwenliang is this ok with you or it should be 36px?

johngruen:
I'm going to assume that since mark both filed the issue and made the PR that this is cool

Issue #3220, opened 2018-01-10
Experiment onboarding modals should not have "enabled" label


youwenliang:
There's an enabled label on the top left corner of the modal, which should not be there.
![image](https://user-images.githubusercontent.com/8919308/34758193-f87dd5fa-f610-11e7-8bc2-8ea8624fec79.png)

SoftVision-PaulOiegas:
Hmmm, then we wrongly understood this storybook design (#3149)? 
There was a design that had no "Enabled" string in it [here](https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Component_20171003/#artboard0), which is mandatory. And there was a Storybook deploy for Tour that had the "Enabled" label. So combining the received design with the storybook deploy, I thought it should be there.

If not, then both the design and Stoybook were wrong because one was missing the word in title and one was having it without a specific design...

To be honest I like how it looks, and the resemblance with the experiment cards, and now also with the news modal that displays the same [label](https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Modal_Photon_Styles/#artboard1). But it's up to you guys if you want to change it back.

SoftVision-PaulOiegas:
The removal of the label was implemented but we forgot to add back the "enabled!" string. Filled separately #3254 for this.

ChenMorpheus:
@marniepw @SoftVision-PaulOiegas I checked several related tickets (#3149, #3220) and I guess there was a confusion here. There are two kinds of modal; one is the onboarding modal which appears after users enable an experiment(see img_1); the other one is the updates modal which emerges once there is an update to an experiment(see img_1). These two modals share the similar style, but there is a difference.

Onboarding tour doesn't need to have "enabled" label since it happens right after users enable experiment. As for updates modal, it needs the "enabled" label to inform users two states: default and enabled(see img_2 and img_3). However, they don't have "enabled" on the header regardless of different modals.

Please check the links below for difference. 
https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Modal_Photon_Styles/#artboard0

https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Component_20171003/#artboard0

![image](https://user-images.githubusercontent.com/18019946/35263252-fb8cde16-0052-11e8-8169-907f03c4e977.png)
img_1

[...]

PR #3249, opened 2018-01-16
Update #3220 #3222 #3223

youwenliang:
Updating UI based on #3220, #3222 and #3223.


johngruen:
@youwenliang this looks good to me!

PR #3258, opened 2018-01-18
Ensures that Firefox iOS is excluded from UA checks for Firefox (closes #2642).

### LINKED ISSUE #2642: Error message on Test Pilot homepage on Firefox for iOS
img_0113

chuckharmston:
This is a very small fix, but testing it is…arduous. To get the right code paths you need to access it off an example.com host, so you need to do local testing on a host where you have access to `/etc/hosts`, and then you need to run Firefox for iOS in the simulator from there. Since it’s a simulator and not an emulator (i.e. ARM vs x86), there’s no access to the App Store, so you need to build https://github.com/mozilla-mobile/firefox-ios from scratch to do that.

To save you all that work, I instead recommend that you use these screenshots and the new unit tests as verification of it working locally, then test in dev once it's merged.

### Before

screen shot 2018-01-17 at 17 43 06

### After

screen shot 2018-01-17 at 17 43 22

### DIFF
--- a/frontend/src/app/lib/utils.js
+++ b/frontend/src/app/lib/utils.js
@@ -40,7 +40,8 @@ export function buildSurveyURL(ref, title, installed, clientUUID, survey_url) {
 }
 
 export function isFirefox(ua) {
-  return ua.indexOf("firefox") > -1 || ua.indexOf("fxios") > -1;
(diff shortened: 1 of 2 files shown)


chuckharmston:
Closing to reopen off of a Mozilla branch.

PR #3271, opened 2018-01-22
Don't use sudo when building frontend & addon on dev

### LINKED ISSUE #3262: /__version__ on testpilot.dev.mozaws.net has blank properties
We use this script to produce `/__version__` on the CircleCI build that goes to dev:

https://github.com/mozilla/testpilot/blob/master/bin/circleci/build-version-json.sh

However, as of the last build to dev, the properties in that file came out blank (look under "Build addon and frontend" and search for `cat version.json`):

https://circleci.com/gh/mozilla/testpilot/7061

My guess is that the switch to CircleCI 2.0 broke this script. Probably changes to environment variables, [though they look the same in docs to me](https://circleci.com/docs/2.0/env-vars/). Will probably call for some time tinkering in a build with SSH enabled to see what's happening with the script in the new environment.

lmorchard:
Fixes #3262 

Turns out that the root user doesn't inherit the CircleCI env vars expected by the build scripts. I'm also guessing this could cause some issues in the frontend & addon builds if we're using env vars to control those too (e.g. NODE_ENV and friends)

Guessing this was just a copypasta artifact, since most of the rest of the shell commands using sudo are installing dependencies.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,9 +22,9 @@ jobs:
       - run:
           name: Build addon and frontend
           command: |
-            sudo ./bin/circleci/build-addon.sh
-            sudo ./bin/circleci/build-version-json.sh
-            sudo ./bin/circleci/build-frontend.sh
+            ./bin/circleci/build-addon.sh
+            ./bin/circleci/build-version-json.sh
+            ./bin/circleci/build-frontend.sh
       - store_artifacts:
           path: ~/testpilot/addon/update.rdf
       - store_artifacts:


lmorchard:
[Looks like this worked in CircleCI](https://circleci.com/gh/lmorchard/testpilot/1136?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) - scroll down to "Build addon and frontend", expand the section, search in page for "version.json"

b4handjr:
> Guessing this was just a copypasta artifact, since most of the rest of the shell commands using sudo are installing dependencies. 

100% what this was. There may have been some limitations with circleci 1.0 too but I am not sure.

b4handjr (approved):

PR #3276, opened 2018-01-23
Exclude featured experiment from home page list

### LINKED ISSUE #3244: Do not include featured experiment in homepage card lists
This should apply to Test Pilot both with and without add-ons installed.

meandavejustice:
- fixes #3244

### DIFF
--- a/frontend/test/ui/test_homepage.py
+++ b/frontend/test/ui/test_homepage.py
@@ -58,8 +58,12 @@ def test_number_of_experiments(base_url, selenium):
     completed_experiments = len(
         [value for value in data['results'] if 'completed' in value and
          value['completed'] < str(datetime.utcnow())]) + 1
-    assert len(page.body.experiments) == int(
-        len(data['results']) - completed_experiments)
+    if page.featured.is_displayed:
+        assert len(page.body.experiments) == int(
+            len(data['results']) - completed_experiments - 1)
+    else:
+        assert len(page.body.experiments) == int(
+            len(data['results']) - completed_experiments)
 
 
 @pytest.mark.nondestructive
--- a/frontend/src/app/containers/App/index.js
+++ b/frontend/src/app/containers/App/index.js
@@ -14,7 +14,7 @@ import { getInstalled, isExperimentEnabled, isAfterCompletedDate, isInstalledLoa
 import { setState as setBrowserState } from "../../actions/browser";
 import { getExperimentBySlug } from "../../reducers/experiments";
 import { getChosenTest } from "../../reducers/varianttests";
-import experimentSelector, { featuredExperimentsSelectorWithL10n } from "../../selectors/experiment";
+import experimentSelector, { featuredExperimentsSelectorWithL10n, experimentsWithoutFeaturedSelectorWithL10n } from "../../selectors/experiment";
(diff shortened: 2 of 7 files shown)


fzzzy:
I guess I didn't see this before opening my pr. Is there any reason to do it this way instead of just not featuring any experiment?

meandavejustice:
@fzzzy This is a different issue, this makes it so we don't duplicate the featured experiment in the experiment card list.

frontend/src/app/containers/HomePageNoAddon.js
```diff
@@ -30,7 +30,9 @@ export default class HomePageNoAddon extends React.Component {
   props: HomePageNoAddonProps
 
   render() {
-    const { experiments, isAfterCompletedDate, featuredExperiments } = this.props;
+    const { isAfterCompletedDate, featuredExperiments } = this.props;
+
+    const experiments = this.props.experiments.filter(x => !x.is_featured);
```
chuckharmston: This should be in a [selector](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/selectors/experiment.js) (see: [`reselect`](https://github.com/reactjs/reselect)) to centralize this type of logic and prevent unnecessary recalculation and rendering of the DOM.

frontend/test/ui/test_homepage.py
```diff
@@ -58,8 +58,12 @@ def test_number_of_experiments(base_url, selenium):
...
         [value for value in data['results'] if 'completed' in value and
          value['completed'] < str(datetime.utcnow())]) + 1
-    assert len(page.body.experiments) == int(
-        len(data['results']) - completed_experiments)
+    if page.featured.is_displayed:
+        assert len(page.body.experiments) == int(
+            len(data['results']) - completed_experiments - 1)
+    else:
+        assert len(page.body.experiments) == int(
+            len(data['results']) - completed_experiments)
```
chuckharmston, line 66: Can you add another assertion that the featured experiment specifically is not in the list of experiments?

chuckharmston (changes requested):

frontend/test/ui/test_homepage.py
meandavejustice: filed https://github.com/mozilla/testpilot/issues/3295, I still can't get integration tests running on my machine, it looks to be because of https://github.com/tox-dev/tox/issues/121. I can loop back around to get this test written when I'm back at my house and on my linux machine.

meandavejustice:
Moved filtering into selector, added unit test, and fixed exiting tests to work with change.

filed https://github.com/mozilla/testpilot/issues/3295 as a reminder to fix when I have access to my linux machine since https://github.com/tox-dev/tox/issues/121 is causing failures on my windows laptop.

SoftVision-PaulOiegas:
@meandavejustice Hmm, cloned this branch locally and did a bit of testing on it. Not sure this is working as intended since I'm seeing also the "Cliqz" experiment while my browser is [en-US] localized. Maybe the change affected the localization restrictions?

chuckharmston:
This generally looks good, but I'm sitting on approving until the issue Paul raised is addressed.

meandavejustice:
@chuckharmston I'll try and get to paul's changes tomorrow morning, I'll ping ya when they're in

Issue #3278, opened 2018-01-23
Fix Storybook deployments for pull requests


lmorchard:
Looks like we lost [Storybook deployments for PRs](https://medium.com/firefox-test-pilot/storybook-test-pilot-26d21fdb6c64) in the move to CircleCI 2.0. ([example](https://github.com/mozilla/testpilot/pull/3038#issuecomment-342645306)). We should get those working again

lmorchard:
There was [a line in the 1.0 post-test commands](https://github.com/mozilla/testpilot/commit/fb5a39447b802deb0d3f99098c78e23dbe41fb29#diff-29944324a3cbf9f4bd0162dfe3975d88L46) that ran [`bin/deploy-storybook.sh`](https://github.com/mozilla/testpilot/blob/master/bin/deploy-storybook.sh) to build & publish a Storybook static rendering to S3. We should get that going again.

<strike>One caveat is that it uses S3 & github API keys in the CircleCI project settings. If those are lost, we'll need to acquire new credentials from AWS & github</strike> Just double checked CircleCI settings, and it looks like all those env vars are still in settings, so 👍

lmorchard:
Looks like it worked!

https://circleci.com/gh/mozilla/testpilot/7149
http://testpilot-storybook.dev.mozaws.net/0624094b509e56785f6940957ca2e0c9ddd0f50c/

PR #3280, opened 2018-01-23
Fix storybook and codecov to run on every commit.

b4handjr:
Fixes: #3278


lmorchard:
Looks like Storybook deploy step ran and properly exited on a non-mozilla repo, so let's try it on the real thing

PR #3285, opened 2018-01-24
Show "enabling..." on featured experiment enable

### LINKED ISSUE #3267: [Featured Experiment] The "Enable Notes" green button animation when clicked, wrongly states "Installing..." instead of "Enabling..."
**[Affected versions]:**
- Firefox v57.0.4 and above

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**
- test Pilot add-on was already installed on the same profile.

**[Steps to reproduce]:**
1. Open Firefox browser with the profile from prerequisites and navigate to https://testpilot.dev.mozaws.net/experiments page.
2. Click the "Enable Notes" green button while observing its animation.

**[Expected result]:**
- A barberpole animation appears on the button with "Enabling..." string 

**[Actual result]:**
- A barberpole animation appears on the button but wrongly displays "Installing..." string.

**[Notes]**
- This issue occurs also on Stage server.
- Here is a screen recording of the issue: https://goo.gl/qUaT5o.

meandavejustice:
- fixes #3267

### DIFF
--- a/frontend/src/app/components/MainInstallButton/index.js
+++ b/frontend/src/app/components/MainInstallButton/index.js
@@ -109,6 +109,12 @@ export default class MainInstallButton extends React.Component {
   renderInstallButton(isInstalling: boolean, hasAddon: any) {
     const { experimentTitle, isExperimentEnabled, experiment } = this.props;
     let installButton = null;
+    const installingButton = (<Localized id="landingInstallingButton">
+      <span className="progress-btn-msg">Installing...</span>
+    </Localized>);
+    const enablingButton = (<Localized id="landingEnablingButton">
+      <span className="progress-btn-msg">Enabling...</span>
+    </Localized>);
 
     if (experimentTitle) {
       if (hasAddon && !isExperimentEnabled(experiment)) {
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/MainInstallButton/index.js
```diff
@@ -123,13 +129,11 @@ export default class MainInstallButton extends React.Component {
...
     const makeInstallButton = (extraClass = "") => {
+      const isEnabling = (experimentTitle && hasAddon && !isExperimentEnabled(experiment) && isInstalling);
       return <button onClick={e => this.install(e)}
         className={classnames(`button primary main-install__button ${extraClass}`, { "state-change": isInstalling })}>
         {!isInstalling && installButton}
-        {isInstalling &&
-          <Localized id="landingInstallingButton">
-            <span className="progress-btn-msg">Installing...</span>
-          </Localized>}
+        {isInstalling ? isEnabling ? enablingButton : installingButton : null} {/* eslint-disable-line no-nested-ternary */}
```
meandavejustice: This seems like a reasonable case to make an exception for the nested ternary rule.

locales/en-US/app.ftl
```diff
@@ -37,6 +37,7 @@ landingLegalNotice = By proceeding, you agree to the <a>Terms of Use</a> and <a>
 [[landingInstall]]
 landingInstallButton = Install the Test Pilot Add-on
 landingInstallingButton = Installing...
+landingEnablingButton = Enabling...
```
flodolo: Please use a single unicode `…` instead of `...`. Fixing the old string would be great

meandavejustice: @flodolo will I need to update the key for `landingInstallingButton` for this change?

flodolo: No need for new ID, each locale should have their own rules for punctuation signs (English should always use the single unicode character)

SoftVision-PaulOiegas:
Hi @meandavejustice, brought the branch with the fix locally and tried to test the fix, but something is quite strange. There is no "Featured Experiment" section on that branch.
Am I doing something wrong?

meandavejustice:
@SoftVision-PaulOiegas You can just manually set `is_featured` to true in notes.yaml for testing

SoftVision-PaulOiegas:
Uuu, nice! Didn't knew it's that easy to make one of the experiments featured. I've tried in the past but I've searched in totally other files than <experiment>.yaml ones.

The fix seems to be ok. Even tried with browser window resize to mobile view and other negative testing but all went well. 👍

PR #3286, opened 2018-01-24
Fix #3231 Properly localize the Notes featured experiment terms and privacy notice.

### LINKED ISSUE #3231: The "Notes" privacy link from "Featured experiment" section is not working if Notes is not enabled
**[Affected versions]:**
- Firefox 57.0.4 and above

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**

**[Steps to reproduce]:**
1. Open Firefox browser and navigate to https://testpilot.dev.mozaws.net/experiments page.
2. Click the "Notes" link from the featured experiment section.
3. Observe the browser behavior.

**[Expected result]:**
- The Notes Privacy modal is correctly displayed.

**[Actual result]:**
- Nothing happens.

**[Notes]:**
- Here is a screen recording with the issue:
notes link

fzzzy:


locales/en-US/experiments.ftl
```diff
@@ -71,6 +71,8 @@ nomore404sContributors2Title = Director, Wayback Machine, The Internet Archive
 nomore404sContributors3Title = Senior Crawl Engineer, The Internet Archive
 notesDescription = A simple notepad built into Firefox. Ever need to jot something down while you browse the web? This is the experiment for you.
 notesIntroduction = Notes adds a simple, one-page notepad to your browser’s sidebar. We’ve added a few basic formatting options to start, but we want your feedback to tell us what features you’d like to see us add.
+notesLegalnotice = By proceeding, you agree to the {$terms} and {$privacy} policy of Test Pilot and {$privacylink}.
```
flodolo: I'm confused: what replaces `$terms` and `$privacy` in this string?

Also, is there a chance to have a single string for this sentence, instead of 2? If not, please add a comment explaining what replaces what.

fzzzy: It's confusing: It would probably be better to use the LocalizedHtml component that I wrote instead of splitting this into multiple strings, but I wasn't able to get it to work.

$terms and $privacy are replaced by landingLegalNoticeTermsOfUse and landingLegalNoticePrivacyNotice, which are single-word strings we are borrowing from elsewhere on the site.

It might be best for me to try harder to get this working with LocalizedHtml, which would enable notesLegalnotice to be the only string used in this message. I'll spend a bit of time trying it.

flodolo: > $terms and $privacy are replaced by landingLegalNoticeTermsOfUse and landingLegalNoticePrivacyNotice, which are single-word strings we are borrowing from elsewhere on the site

Can you point me to the strings? Because I've searched, and failed to find them anywhere in l10n files.

flodolo (approved):
Thanks, strings-wise is much better.

locales/en-US/experiments.ftl
fzzzy: @flodolo I was able to replace that ugly implementation with a much simpler LocalizedHtml implementation. Please take a look at the new code instead.

Thank you for complaining about that!

Issue #3268, opened 2018-01-19
[Featured Experiment] Notes experiment "Tour" modal is wrongly displayed when cancelling the install from webextension doorhanger


SoftVision-PaulOiegas:
**[Affected versions]:**
- Firefox v57.0.4 and above

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**
- Following 5 prefs have been set or added on a new profile: "extensions.legacy.enabled = true" "xpinstall.signatures.required = false" "testpilot.env = dev" "extensions.webapi.testing = true" "extensions.install.requireBuiltInCerts = false".
- test Pilot add-on was already installed on the same profile.

**[Steps to reproduce]:**
1. Open Firefox browser with the profile from prerequisites and navigate to https://testpilot.dev.mozaws.net/experiments page.
2. Click the "Enable Notes" green button.
3. When the webextension install doorhanger appears, click the "Cancel" button.
4. Observe Test Pilot page behavior.

**[Expected result]:**
- The "Enabling..." animation stops and the button returns to "Enable Notes" default state.

**[Actual result]:**
- The animation continues to run and also the Notes "Tour" modal is wrongly displayed.

**[Notes]**
- This issue occurs also on Stage server.
- Here is a screen recording of the issue: https://goo.gl/SkuW1g.

meandavejustice:
Figured out what the issue was here over the weekend (not throwing an error in the right place). But fixing this exposed how we aren't handling the error case in the featured experiment/home page view. I'm going to implement the same error ui we show on the experiment detail page.

meandavejustice:
The error ui from experiment detail pages doesn't actually work for the featured experiment on home page.
I have a fix that simply reverts the install button on cancel, instead of hanging in the "enabling..." state. I'll open up a pr for that soon. But we will still need an informative ui for the error state.

SoftVision-PaulOiegas:
Verified this on both dev and stage servers and no longer reproduces. The button returns to the initial position when flow is canceled. 
However, I've seen your last message only after I logged an issue (#3317) to implement an error also for when the enable action is not performed (such as no internet). Added needs:UX label on it also.

Also, found a minor issue that appeared after this fix was merged and logged it separately (#3316).

PR #3312, opened 2018-02-01
Reset featured experiment enable button on cancel dialog chosen

meandavejustice:
- refs #3268


chuckharmston (approved):
This looks good, but the build is failing.

Issue #3182, opened 2018-01-04
Drop nsIObserver


ghost:
From bootstrap:

> nsIObserver - hacky way to receive messages from legacy extension based experiments, used as proxy to telemetry and to manage a/b testing (may be unused now, though I think)

lmorchard:
Not sure if anything is using this any more - the last experiment I remember using it was (I think) containers. I seem to remember Min Vid was using it, but stopped at some point?

meandavejustice:
~~This is also used in [lib/appStartup.js](https://github.com/mozilla/testpilot/blob/0eed4b5eaa10becf3fa05bdf396529e9f0e302ec/addon/src/lib/appStartup.js) which affects how we lead all of our modules in bootstrap.js. Going to loop back around to this one after I've taken care of more of the other bootstrap issues.~~ This is actually #3181

lmorchard:
FWIW, that `sessionstore-windows-restored` event observer is just a startup performance hack for a legacy add-on embedded webextension. Once we move to a pure webextension, we won't need it and can just rely on the default behavior of webextension loading.

Issue #3181, opened 2018-01-04
Drop startup deferral code


ghost:
In bootstrap:

> sessionstore-windows-restored event - defer init until after browser startup

With all the web extension improvements, hopefully this isn't a thing that we need to do manually anymore?

meandavejustice:
This looks like it should be the one of if not _the_ last issue we address to close out the `bootstrap` issues. I vote that we bump it to `Sprint 41` milestone.

What do yall think @chuckharmston @clouserw

ghost:
I don't know enough to talk about what order things should be done in.  Doing this last is fine with me.  I don't think we should kick it out of the sprint unless we have to though

meandavejustice:
fixed in https://github.com/meandavejustice/testpilot/commit/9c11090c64b0205d3d2965f025703221883b9777

Issue #3177, opened 2018-01-04
Figure out if we need install/uninstall tracking for test pilot add-ons


ghost:
This is still using bootstrap:

> Addon Manager - overall tracking install/uninstall of Test Pilot experiments for metrics

Do we still need this?

lmorchard:
[We send a count of enabled experiments in the daily Telemetry ping](https://github.com/mozilla/testpilot/blob/master/addon/src/webextension/lib/metrics.js#L116). If we stop needing that, we can stop tracking installs / uninstalls. Or, if there's a way to count installed experiments from a WebExtension, we could do that instead - though I don't know how to do this off the top of my head.

meandavejustice:
@lmorchard It looks like we can get our list of install experiments with the [management api](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/management)

lmorchard:
Yeah, I think at the point when we were coming up with these issues, that API only reported on themes - [and then only by using getAll, because oninstall / onuninstall weren't supported](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/management#Browser_compatibility). Probably want to verify it returns all the data we want

fzzzy:
Can't we just use mozAddonManager for this? We know it works, it's not bootstrap, and we're whitelisted for it.

fzzzy:
Actually, aren't we already using mozAddonManager for it? If so, it's not bootstrap, so maybe this can be closed as invalid.

meandavejustice:
@fzzzy We still need an answer for how we should get the list of enabled experiments for the daily telemetry ping. We may be able to use the webextension management api for that, but we haven't verified that yet (I'll check that this afternoon). So we can't close the issue until that is resolved. But it looks like that may be the only question left to answer before install/uninstall handling is removed.

lmorchard:
> Actually, aren't we already using mozAddonManager for it? If so, it's not bootstrap, so maybe this can be closed as invalid.

Yeah, no - this happens in the add-on, independent of website visits. It's for the daily telemetry ping that measures what experiments are active

johngruen:
I will reach out to Ilana and Sunah about this, I don't know if this ping is behind our redash dashboards or if that's just a filter on the installed add-ons field in telemetry. I think it's the latter, but I will confirm.

fzzzy:
@lmorchard the addon could load a hidden page which uses mozAddonManager to fetch the data, then sends it up to the addon.

[...]

PR #3325, opened 2018-02-06
Remove bootstrap event observers WIP DO NOT MERGE

meandavejustice:
- refs #3182, #3181 needs to be closed before this issue is truly
complete.
- removes addon/src/lib/events.js
- references in addon/README removed


lmorchard (changes requested):
This removes the bootstrap-side code that supports events from the webextension side - but the code that uses [observeEventTopic](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=observeeventtopic&type=) and [notifyEventTopic](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=notifyeventtopic&type=) is still in there. That stuff should get removed too

Looks like notifyEventTopic is only really used for the A/B testing feature that we never ended up using. The observeEventTopic side was used by other bootstrapped add-ons relaying telemetry through Test Pilot, but no one should be using that either.

addon/src/lib/events.js
```diff
@@ -1,87 +0,0 @@
...
-  observer.register();
-}
-
-export function unobserveEventTopic(eventTopic) {
-  observers[eventTopic].unregister();
-  delete observers[eventTopic];
-}
-
-function observeIdleDaily() {
-  PubSub.publishSync(eventsTopics("idle-daily"));
```
lmorchard, line 52: Worth noting here is that [code listening for this idle-daily event](https://github.com/mozilla/testpilot/search?utf8=%E2%9C%93&q=idle-daily&type=) is what [sends the daily telemetry ping](https://github.com/mozilla/testpilot/blob/6f67e2ad6a7d2b7e600bbefe60b98f1b4bf16494/addon/src/webextension/lib/metrics.js#L100). So, this is also somewhat related to #3177 in that it will stop the daily ping altogether

meandavejustice:
Closing for now, need to add back in the daily telemetry ping that @lmorchard pointed out. Will reopen when ready for review

lmorchard:
> need to add back in the daily telemetry ping that @lmorchard pointed out

Main thing there is to figure out (probably in #3177) whether we can do one of these:
1. Stop the daily ping because we don't need it? (might be a @johngruen question, because I'm fuzzy on exactly what we do with that data these days)
1. Move the daily ping to GA?
1. Switch to [a new Telemetry API whenever it's available](https://bugzilla.mozilla.org/show_bug.cgi?id=1280234#c31) - @6a68 will know more about that
1. Use a bundled WebExtension Experiment version of that new Telemetry API if necessary? (maybe not, if we can hold off fully going webextension until after that API ships)

PR #3332, opened 2018-02-07
Remove survey popup from addon, and deprecated addon examples

meandavejustice:


chuckharmston (approved):
RIP my beautiful variant testing framework. 😭 💀

This looks good. Are we sure that no active experiments are using anything that we're deleting here?

PR #3334, opened 2018-02-07
Add docs for developing an experiment with Test Pilot

ghost:
First version.  Any feedback?

### DIFF
--- a/docs/experiments/developing_an_experiment.md
+++ b/docs/experiments/developing_an_experiment.md
@@ -1,0 +1,112 @@
+[👈 Back to README](../../README.md)
+
+# Test Pilot Recommended Development Process for Experiments
+
+The following process has proven to be the most streamlined in the Test Pilot
+world.  You're not required to follow it, but if you do things will probably
+move more smoothly and it'll be easier to get help for your experiment because
+it's a familiar process.
+
+## Sprints
+We're fans of sprinting.  The standard two weeks is a good balance of feature
+development and shipping.
+
+## Communication
+Yay meetings. :)
+
+- Pick Slack or IRC.  Having a channel on each just makes it harder to decide
+  where to talk.  Now that contributors can be invited to Slack, most
+  experiments choose that.
+- Schedule stand-ups at least twice a week.  Most experiments can fit a new-bug
+  triage into the stand-up if they do it consistently so the bugs don't build
+  up.
+- Schedule a sprint planning meeting.
+- Schedule a product and analysis meeting.  This meeting is focused on
+  reviewing the analytics and talking about actions we should take based on
+  them.
+- Every Test Pilot experiment has a [forum on discourse](https://discourse.mozilla.org/c/test-pilot).
+  Be sure to visit regularly and engage with your users.
+
+
+## Development
+Using GitHub is the easiest way to get going and includes several useful tools
+for free.  Test Pilot already has teams organized on GitHub so adding them to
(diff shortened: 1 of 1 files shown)


docs/experiments/developing_an_experiment.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Sprints
+We're fans of sprinting.  The standard two weeks is a good balance of feature
+development and shipping.
+
+## Communication
+Yay meetings. :)
+
+- Pick Slack or IRC.  Having a channel on each just makes it harder to decide
+  where to talk.  Now that contributors can join Slack, most experiments choose
```
lmorchard: > Now that contributors can join Slack...

I haven't heard this before - is there a page somewhere we can point contributors at to get a sign up?

ghost: It's a manual add via service-now right now.  They are trialing it.  Some info on https://mana.mozilla.org/wiki/display/CCT/Slack

lmorchard: Might be splitting hairs, but this could be worth mentioning - contributors can't *join* but they can in some cases be *invited* on a case-by-case basis. So we'd have to connect with them through other channels like github issues and discourse threads.

docs/experiments/developing_an_experiment.md
```diff
@@ -0,0 +1,116 @@
...
+  up.
+- Schedule a sprint planning meeting.
+- Schedule a product and analysis meeting.  This meeting is focused on
+  reviewing the analytics and talking about actions we should take based on
+  them.
+- Every Test Pilot experiment has a [forum on
+  discourse](https://discourse.mozilla.org/c/test-pilot).  Be sure to visit
+  regularly and engage with your users.
+
+
```
chuckharmston: Nitpicking: newlines aren't consistent.

docs/experiments/developing_an_experiment.md
```diff
@@ -0,0 +1,116 @@
+# Test Pilot Recommended Development Process for Experiments
```
chuckharmston, line 3: Most of our docs have this at the top: `[👈 Back to README](../../README.md)`

chuckharmston (approved):
Looks great, thanks for taking my earlier feedback into consideration! I won't merge, but it's ready to merge as far as I'm concerned.

docs/experiments/developing_an_experiment.md
lmorchard: Looks like a plain old word-wrap since the markdown link has no spaces in it. Could neaten it up with a footnote-style link. e.g.:
```markdown
- Every Test Pilot experiment has a [forum on discourse][forum]. Be sure to visit
  regularly and engage with your users.

[forum]: https://discourse.mozilla.org/c/test-pilot
```

docs/experiments/developing_an_experiment.md
ghost: good call, thanks

docs/experiments/developing_an_experiment.md
ghost: I'll update

docs/experiments/developing_an_experiment.md
ghost: I'll update this and let the lines with links go past 80 chars.  I do prefer footnote links, but I don't think we use them elsewhere.

PR #3347, opened 2018-02-15
Remove Frame Scripts from addon DO NOT MERGE WITHOUT APPROVAL FROM JGRUEN

### LINKED ISSUE #3176: Do something with our Frame Scripts
As part of our departure from bootstrap, we need to do something with this:

> Frame Scripts to set window.navigator properties - our legacy way of informing about:home & mozilla.com pages about Test Pilot participation

Maybe we can just drop this?  Why do those pages need to know about Test Pilot?

meandavejustice:
- fixes #3176

Within the webextension api we are no longer able to modify the
navigator object on mozilla.com or about:home, so snippets will no
longer be able to identify whether or not a user is in testpilot.
Additionally, we won't be providing information to the frontend
through the addon, this isn't an issue as we now use
mozAddonManager on the frontend.


meandavejustice:
@johngruen I went ahead and pulled out the frame script code based on our discussion in the txp triage this morning. I wanted to double check with you before we remove this, since it is the component that interacted with snippets. As we discussed this morning, there doesn't appear to be a path forward for this feature, as we aren't able to modify about:home or mozilla.com from our webextension.

fzzzy:
For the record, if they still want to detect the presence of test pilot once this is removed, they could use mozAddonManager to check to see if the test pilot addon is installed (assuming they are whitelisted or could be whitelisted to use it).

meandavejustice:
opting for a bootstrapless-branch which will be merged when telmetry webextension support lands in firefox 60, branch here https://github.com/meandavejustice/testpilot/tree/bootstrapless-addon

PR #3349, opened 2018-02-15
Switch en-US FTL file to new syntax

flodolo:
* `#` for comment sigil
* no more sections,`##` for section comments
* equal sign on empty value
* also used spaces around equal sign consistently

CC @pike

### DIFF
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -1,14 +1,12 @@
 siteName = Firefox Test Pilot
 
-// Page titles, put in the <title> HTML tag.
-[[pageTitle]]
+## Page titles, put in the <title> HTML tag.
 pageTitleDefault = Firefox Test Pilot
 pageTitleLandingPage = Firefox Test Pilot
 pageTitleExperimentListPage = Firefox Test Pilot - Experiments
 pageTitleExperiment = Firefox Test Pilot - {$title}
 
-// Links in the footer.
-[[footerLink]]
+## Links in the footer.
 footerLinkCookies = Cookies
 footerLinkPrivacy = Privacy
 footerLinkTerms = Terms
@@ -16,8 +14,7 @@ footerLinkLegal = Legal
 footerLinkFeedback = Give Feedback
 footerLinkAbout = About Test Pilot
 
-// Items in the menu.
-[[menu]]
+## Items in the menu.
 home = Home
 menuTitle = Settings
 menuWiki = Test Pilot Wiki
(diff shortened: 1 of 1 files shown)


locales/en-US/app.ftl
```diff
@@ -1,23 +1,20 @@
 siteName = Firefox Test Pilot
 
-// Page titles, put in the <title> HTML tag.
-[[pageTitle]]
+## Page titles, put in the <title> HTML tag.
```
Pike: drive-by nit, put-put-put. `shown as title of the html page` is less techy, and doesn't sound like you should add <title> around your strings.

locales/en-US/app.ftl
```diff
@@ -1,23 +1,20 @@
...
 pageTitleDefault = Firefox Test Pilot
 pageTitleLandingPage = Firefox Test Pilot
 pageTitleExperimentListPage = Firefox Test Pilot - Experiments
 pageTitleExperiment = Firefox Test Pilot - {$title}
```
Pike: Should we add a Term for Firefox Test Pilot?

locales/en-US/app.ftl
```diff
@@ -26,27 +23,23 @@ menuFileIssue = File an Issue
...
+## Related to the installation of the Test Pilot add-on.
 landingInstallButton = Install the Test Pilot Add-on
 landingInstallingButton = Installing…
 landingEnablingButton = Enabling…
 
-// Related to a one click to install test pilot and an experiment.
-[[oneClickInstall]]
+## Related to a one click to install test pilot and an experiment.
 oneClickInstallMinorCta = Install Test Pilot &amp;
 oneClickInstallMajorCta = Enable {$title}
```
Pike: `title` could use a descriptive comment.

locales/en-US/app.ftl
```diff
@@ -55,41 +48,35 @@ landingDownloadFirefoxTitle = Firefox
...
-[[onboarding]]
+## Shown after the user installs the Test Pilot add-on.
 onboardingMessage = We put an icon in your toolbar so you can always find Test Pilot.
 
-// Error message pages.
-[[error]]
+## Error message pages.
 errorHeading = Whoops!
 errorMessage = Looks like we broke something. <br> Maybe try again later.
 notFoundHeader = Four Oh Four!
```
Pike: Should this have a comment? Looking at existing translations, it might.

locales/en-US/app.ftl
flodolo: It might make sense, but in case it should be in a different issue/discussion, given how many translations we would lose.

flodolo:
Thanks, I think I've fixed all the comments.

chuckharmston (approved):

Issue #3321, opened 2018-02-05
[Android][Mobile view] The "View Past Experiments" button breaks the UI on French locale


SoftVision-PaulOiegas:
**[Affected versions]:**
- Firefox 58.0.1 (mobile)

**[Affected Platforms]:**
- Android 7.0 (Samsung Galaxy S6 with 2K display)

**[Prerequisites]:**
- Latest Firefox browser version was installed on the mobile device.
- French language was selected from "Settings>General>Language" menu.

**[Steps to reproduce]:**
1. Start the Firefox Android app and navigate to https://testpilot.dev.mozaws.net page.
2. Scroll down until the "View Past Experiments" button is reached.
3. Observe how the button is rendered in French locale.

**[Expected result]:**
- The button is correctly rendered in French locale and fits the designed area.

**[Actual result]:**
- The button is correctly translated but it exceeds its designed area breaking the UI. 

**[Notes]:**
- This also reproduces on stage server.
- Here is a screenshot of the issue: 
![screenshot_20180205-162957](https://user-images.githubusercontent.com/17258026/35811771-fbef47e6-0a97-11e8-81f4-a8668f5080a4.png)

johngruen:
let's bump the l/r padding down by a few pixels on each side

divyabaid16:
@johngruen PR #3363 have failed two circleci builds. Can you please guide me for this?

meandavejustice:
Hey @divyabaid16 the failures aren't your fault. Some code was merged into master on friday that caused the failures. It should be fixed up today and then your tests should pass. Sorry for the confusion

chuckharmston:
Reviewing now!

SoftVision-PaulOiegas:
Looks ok on my Galaxy S6 using French locale. Marking as verified. 👍 
Thanks for the contribution @divyabaid16!

PR #3363, opened 2018-02-17
[Android][Mobile view] The "View Past Experiments" button works fine with the UI on French locale #3321

divyabaid16:
Fix : [Android][Mobile view] The "View Past Experiments" button breaks the UI on French locale #3321


chuckharmston (approved):
This works great at 360px wide:

screen shot 2018-02-20 at 10 08 12

I'd hope that it would at 320px as well:

screen shot 2018-02-20 at 09 55 03

But that's a sufficiently small proportion of our userbase that I don't think it's worth dealing with.

Thanks for the contribution @divyabaid16!

PR #3372, opened 2018-02-21
Fix race condition and refactor to simplify install method in mainInstallButton

### LINKED ISSUE #3371: [Featured Experiments] Loading indicator and onboarding not displayed
### STR

1. Make sure test pilot is uninstalled
2. click the `install test pilot and enable notes` button

### Results
Test Pilot and notes install but there is no barberpoll on the installing notes button and the experiment onboarding tour does not appear.

meandavejustice:
- fixes #3371

### DIFF
--- a/frontend/src/app/components/MainInstallButton/index.js
+++ b/frontend/src/app/components/MainInstallButton/index.js
@@ -31,7 +31,9 @@ export default class MainInstallButton extends React.Component {
 
     this.setState({ isInstalling: true }, () => enableExperiment(experiment)
       .then(() => {
-        if (postInstallCallback) postInstallCallback();
+        if (postInstallCallback) {
+          postInstallCallback();
+        }
       }).catch((err) => {
         this.setState({isInstalling: false});
       }));
--- a/frontend/src/app/components/MainInstallButton/index.js
+++ b/frontend/src/app/components/MainInstallButton/index.js
@@ -26,6 +26,17 @@ export default class MainInstallButton extends React.Component {
     };
   }
 
+  enableExperimentWrapped() {
+    const { experiment, enableExperiment, postInstallCallback } = this.props;
+
+    this.setState({ isInstalling: true }, () => enableExperiment(experiment)
+      .then(() => {
+        if (postInstallCallback) postInstallCallback();
+      }).catch((err) => {
+        this.setState({isInstalling: false});
+      }));
+  }
+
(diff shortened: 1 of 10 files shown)


frontend/src/app/components/MainInstallButton/index.js
```diff
@@ -26,6 +26,17 @@ export default class MainInstallButton extends React.Component {
     };
   }
 
+  enableExperimentWrapped() {
+    const { experiment, enableExperiment, postInstallCallback } = this.props;
+
+    this.setState({ isInstalling: true }, () => enableExperiment(experiment)
+      .then(() => {
+        if (postInstallCallback) postInstallCallback();
```
chuckharmston: Braces here, please.

chuckharmston (approved):
Other than one nit, it looks and works well. Thanks!

frontend/src/app/components/MainInstallButton/index.js
meandavejustice: You want inline or newlines?

chuckharmston: Newlines.

johngruen:
I am not seeing the fix working here.

feb-21-2018 09-57-02

meandavejustice:
@johngruen Anything special to reproduce?

meandavejustice:
@johngruen This is what I'm getting, maybe try pulling again? I just pushed another commit slightly simplifying the code to make it more readable.

one-click-install

chuckharmston:
I see what Dave is seeing here.

johngruen:
@chuckharmston @meandavejustice check the url, you guys are on /experiments installers will mostly be at / where i am still seeing the issue

meandavejustice:
@johngruen just pointed out to me that the behavior he recorded above happens when the you visit `/` and what me and @chuckharmston were seeing was when you start out on `experiments`. Which is caused by the page updating and switching views mid installAndEnable, once the TXP addon is installed. https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/HomePage/index.js#L10

Which led me to look at our home page views and wonder if `homePageWithAddon` and `homePageWithoutAddon` are separate for historical reasons. We may be able to combine them into one view, (separating out appropriate components) and avoid this rerender that messes up our one click installs, and avoids code duplication as well.

PR #3388, opened 2018-02-28
add new notes contributors

vladikoff:
@johngruen r?


chuckharmston (changes requested):
@vladikoff since you're changing strings, you'll need to update the English locale file. That's why the build is failing. More here:

https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization

Moving Udara down in the list in particular my have some deleterious effects to localization. I'd suggest not doing that.

vladikoff:
@chuckharmston by new strings do you mean titles such as `title: 'Community Contributor'`?

chuckharmston:
Indeed. Those need to be localized.

vladikoff:
@chuckharmston ok I removed the new "cool" titles and moved Udara back, so no new strings just new people :)

chuckharmston:
Unfortunately, that won't do it; even though the content of the strings may not be new, the string itself is. Sébastien's and Ethan's titles, for example, are considered different strings in the system.. 

[Activity Stream's](https://github.com/mozilla/testpilot/blob/master/locales/en-US/experiments.ftl#L6-L18) contributor titles might be a nice example (note that #7 and #8 are the same text). 

The documentation I linked in the original review explains the process of generating a new locale file well.

SoftVision-PaulOiegas:
@vladikoff Since we are modifying this section, would you guys mind to add Marius Coman too? He's doing a lot of QA on Notes side by side with Emil and he has the same "title" as him.
Would be much appreciated, but no hard feelings if you guys don't consider it ok. :)

vladikoff:
@SoftVision-PaulOiegas for sure! 👍

SoftVision-PaulOiegas:
@vladikoff Thanks a lot!I bet he will be quite happy to see his name there.

chuckharmston:
@vladikoff With this most recent commit, by changing the indexes of contributors (e.g. @rfeeley changes from notesContributors2 to notesContributors5) you're changing the meaning of some strings without invalidating existing translations using `_l10nsuffix`. You'll either need to invalidate all the affected titles (not preferred, because we don't want to waste localizers' time) or just add the new contributors to the bottom of list.

vladikoff:
> just add the new contributors to the bottom of list.

Okay :(

chuckharmston (approved):

chuckharmston:
@flodolo r?

flodolo (approved):

vladikoff:
Thank you all!! 🍰

PR #3391, opened 2018-02-28
add featured experiment table

### LINKED ISSUE #3234: Featured experiments UI lacks event reporting

teonbrooks:
Related to #3234.

These are the metrics to be included for the Test Pilot website. Since the featured experiment is in a section than the other experiments, I copied over those metrics, but is that necessary?

### DIFF
--- a/docs/metrics/ga.md
+++ b/docs/metrics/ga.md
@@ -109,20 +109,20 @@ Here are the current events on the website as of this writing, grouped by their 
 
 #### `Featured Experiment`
 
-| Description                      | `eventAction`         | `eventLabel`                  | Custom Dimensions |
-| -------------------------------- | --------------------- | ----------------------------- | ----------------- |
-| Click Featured Experiment Video  | video click           | Play Featured video           | [12]              |
-| Click View Details               | link click            | View Featured details         | [4]               |
-| Featured Experiment Button       | button click          | Manage Featured Button State  | [1, 4, 5]         |
-| Click Cancel on tour dialogue    | button click          | cancel tour                   |                   |
-| Complete the tour                | button click          | complete tour                 |                   |
-| Click Take Tour on tour dialogue | button click          | take tour                     |                   |
-| Click next during Tour           | button click          | forward to step `n`           |                   |
-| Click back during Tour           | button click          | back to step `n`              |                   |
-| Click pagination dot during Tour | button click          | dot to step `n`               |                   |
-| Click General Terms link         | link click            | Open general terms            |                   |
-| Click General Privacy link       | link click            | Open general privacy          |                   |
-| Click Featured Privacy link      | link click            | Popup Featured privacy        |                   |
+| Description                      | `eventAction`         | `eventLabel`                  | Custom Dimensions      |
+| -------------------------------- | --------------------- | ----------------------------- | ---------------------- |
+| Click Featured Experiment Video  | video click           | Play Featured video           | [1, 2, 3, 12, 10, 11]  |
+| Click View Details               | link click            | View Featured details         | [1, 2, 3, 4, 10, 11]   |
+| Featured Experiment Button       | button click          | Manage Featured Button State  | [1, 2, 3, 4, 5, 10, 11]|
+| Click Cancel on tour dialogue    | button click          | cancel tour                   | [1, 2, 3, 10, 11]      |
+| Complete the tour                | button click          | complete tour                 | [1, 2, 3, 10, 11]      |
+| Click Take Tour on tour dialogue | button click          | take tour                     | [1, 2, 3, 10, 11]      |
+| Click next during Tour           | button click          | forward to step `n`           | [1, 2, 3, 10, 11]      |
+| Click back during Tour           | button click          | back to step `n`              | [1, 2, 3, 10, 11]      |
+| Click pagination dot during Tour | button click          | dot to step `n`               | [1, 2, 3, 10, 11]      |
+| Click General Terms link         | link click            | Open general terms            | [1, 2, 3, 10, 11]      |
+| Click General Privacy link       | link click            | Open general privacy          | [1, 2, 3, 10, 11]      |
+| Click Featured Privacy link      | link click            | Popup Featured privacy        | [1, 2, 3, 10, 11]      |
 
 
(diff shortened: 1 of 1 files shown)


docs/metrics/ga.md
```diff
@@ -107,6 +107,24 @@ Here are the current events on the website as of this writing, grouped by their
...
+| Featured Experiment Button       | button click          | Manage Featured Button State  | [1, 4, 5]         |
+| Click Cancel on tour dialogue    | button click          | cancel tour                   |                   |
+| Complete the tour                | button click          | complete tour                 |                   |
+| Click Take Tour on tour dialogue | button click          | take tour                     |                   |
+| Click next during Tour           | button click          | forward to step `n`           |                   |
+| Click back during Tour           | button click          | back to step `n`              |                   |
+| Click pagination dot during Tour | button click          | dot to step `n`               |                   |
+| Click General Terms link         | link click            | Open general terms            |                   |
+| Click General Privacy link       | link click            | Open general privacy          |                   |
+| Click Featured Privacy link      | link click            | Popup Featured privacy        |                   |
```
chuckharmston: > Since the featured experiment is in a section than the other experiments, I copied over those metrics, but is that necessary?

Definitely makes sense, so long as we can segment those events in GA to only show data from featured experiments, something that I don't think is possible with this as specified. I'd suggest another custom dimension to indicate that.

docs/metrics/ga.md
```diff
@@ -185,7 +203,8 @@ Here is a list of dimensions we are currently using
 | Home Page, Experiment Detail Page, Share Page | Which test has this user been selected for?       | 8         | {'installButtonBorder', ''} |
 | Home Page, Experiment Detail Page, Share Page | Which cohort has this user been selected for?     | 9         | {'bigButton', ''} |
 | All                                           | In which responsive breakpoint is the user?       | 10        | {'mobile','small','medium','big'} |
-| All experiment-specific locations and events  | On which experiment did this action take?         | 11        | experiment slug | 
+| All experiment-specific locations and events  | On which experiment did this action take?         | 11        | experiment slug |
+| Featured Experiment Video Progress            | Duration of Featured YouTube video                | 12        | {n}
```
chuckharmston: In general, custom dimensions should be for categorical variables. Even the numeric-based custom dimensions above (e.g. cd3) have such a narrow range of possible values that they are effectively categorical. Depending on the questions you want this to answer, perhaps we could make it categorical by grouping users like this:

- Video not watched (<= 10% of duration)
- Video fully watched (>=90% of duration)
- Video partially watched (all others)

docs/metrics/ga.md
```diff
@@ -107,6 +107,24 @@ Here are the current events on the website as of this writing, grouped by their
 | Retire from Test Pilot | button click | Retire |
 | Shows upgrade notice | Upgrade Warning | upgrade notice shown |
 
+#### `Featured Experiment`
```
chuckharmston, line 110: Since this is always happening on the homepage, each of these should include `cd1`, `cd2`, and `cd3`. Everything should include `cd10`. Since each of these operate on a specific featured experiment, they should all include `cd11`.

chuckharmston (changes requested):
Looking good! Some comments included. Let's talk  through any outstanding questions in our 1:1 on Tuesday.

teonbrooks:
thanks for the suggestions @chuckharmston. i updated the table, and we can discuss it more in our 1:1

docs/metrics/ga.md
```diff
@@ -173,19 +191,21 @@ We use custom dimensions to refine our pageviews on Test Pilot ([docs from Googl
...
+| Experiment Detail Page                        | Experiment title                                        | 5         | "xyz"  |
+| Experiment Detail Page                        | Installation count                                      | 6         | {n}    |
+| Home Page, Experiment Detail Page             | Determine if installation will require a restart        | 7         | {'no restart','restart required'}   |
+| Home Page, Experiment Detail Page, Share Page | Which test has this user been selected for?             | 8         | {'installButtonBorder', ''} |
+| Home Page, Experiment Detail Page, Share Page | Which cohort has this user been selected for?           | 9         | {'bigButton', ''} |
+| All                                           | In which responsive breakpoint is the user?             | 10        | {'mobile','small','medium','big'} |
+| All experiment-specific locations and events  | On which experiment did this action take?               | 11        | experiment slug |
+| Featured Experiment Video Progress            | Amount of video watched (0:<=10%, 1:10%>x<90%, 2:>=90%) | 12        | {0,1,2}|
```
chuckharmston: My only suggestion here is to use names rather than numbers, which will make reporting in GA much more satisfying.

docs/metrics/ga.md
chuckharmston: :+1:

docs/metrics/ga.md
```diff
@@ -107,6 +107,24 @@ Here are the current events on the website as of this writing, grouped by their
...
+| Featured Experiment Button       | button click          | Manage Featured Button State  | [1, 2, 3, 4, 5, 10, 11]|
+| Click Cancel on tour dialogue    | button click          | cancel tour                   | [1, 2, 3, 10, 11]      |
+| Complete the tour                | button click          | complete tour                 | [1, 2, 3, 10, 11]      |
+| Click Take Tour on tour dialogue | button click          | take tour                     | [1, 2, 3, 10, 11]      |
+| Click next during Tour           | button click          | forward to step `n`           | [1, 2, 3, 10, 11]      |
+| Click back during Tour           | button click          | back to step `n`              | [1, 2, 3, 10, 11]      |
+| Click pagination dot during Tour | button click          | dot to step `n`               | [1, 2, 3, 10, 11]      |
+| Click General Terms link         | link click            | Open general terms            | [1, 2, 3, 10, 11]      |
+| Click General Privacy link       | link click            | Open general privacy          | [1, 2, 3, 10, 11]      |
+| Click Featured Privacy link      | link click            | Popup Featured privacy        | [1, 2, 3, 10, 11]      |
```
chuckharmston: I'd still like to see a way to segment the events here that take place on both featured and non-featured experiments.

chuckharmston (changes requested):

chuckharmston (approved):

PR #3410, opened 2018-03-06
Remove the /legacy page (closes #3013).

### LINKED ISSUE #3013: Remove /legacy page and redirect to home page from URL

chuckharmston:
There has been 2 visits of /legacy in this calendar year, one of which was me yesterday, so I'm skipping the redirect here.

>  1 addition and 38 deletions. 

😎


fzzzy:
I'm not so sure of that. We get a lot of hits to the legacy page in sentry. But, those clients can't run our js because they are old, malware infested browsers. So, it's probably fine to remove it still.

dannycoates (approved):
less is more

PR #3411, opened 2018-03-06
WIP Fix #3405 Use react-router-dom again.

### LINKED ISSUE #3405: Move back to react-router.
We removed `react-router` due to problems in our localization process that have since been resolved. At this point in time, I think it makes sense to restore it for a few reasons:

1. Being a single-page app will be a much better and smoother UX.
1. It's a de facto library, and will make it easier for contributors familiar with React to help us.
1. We store a bunch of state that we asynchronously retrieve from the browser on page load. When that state is used to vary UI, it feels janky. On react-router we only have to do that once for the initial  load.

For bonus points: whenever collected, store the pieces of state referenced in point 3 in localstorage, then attempt to rehydrate that on page load. That will help us render without jank even faster, and we can update asynchronously if things change.

@fzzzy has a partial branch to do this. Could you post that, Donovan?

fzzzy:


dannycoates:
Thanks @fzzzy this is a good starting reference. The complicated bit remaining is making this work for the static generation. Like, we'll need `StaticRouter` in here somewhere me thinks.

dannycoates:
closing in favor of #3416

PR #3414, opened 2018-03-09
Updated some tests.

b4handjr:
It seems some of the integration tests were failing. I went over and redid the failing ones just a bit.


b4handjr:
This looks good now. It should fix the failures.

PR #3416, opened 2018-03-12
use react-router

dannycoates:
Ok, so first stab at react-router...


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/6a336a322dd186d932b4fbef58c56f4d35b2279d/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/a533e9cb67cbf07e77a38cd275e2496975d98da8/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/c39544ef65d28538050adc0633c83a2b0a9b6cc1/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/12a5ed1de763977320af1f3efffd29afaf59fbc1/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/34da4cf234a7f1dcbb5a95069bd49f7fb98d55d9/

fzzzy:
Sweet. Can't wait to dig in and review this tomorrow.

chuckharmston:
Console warning: 

```
Warning: Failed prop type: The prop `context` is marked as required in `StaticRouter`, but its value is `undefined`.
    in StaticRouter
```

chuckharmston:
There's a pretty big delay between click and navigation (note the time between when the URL changes in the address bar and when the view is updated). Might be worth profiling this.

frontend/src/app/components/FeaturedExperiment/FeaturedButton.js
```diff
@@ -84,8 +85,7 @@ export default class FeaturedButton extends React.Component {
       eventCategory,
       eventAction: "Open detail page",
       eventLabel: title,
-      dimension11: slug,
-      outboundURL: evt.target.href
```
chuckharmston, line 88: Nice catching that these are no longer needed 👍

dannycoates:
> There's a pretty big delay

It's pretty slow for me too in Nightly. Chrome doesn't have the same delay

chuckharmston (changes requested):
I don't have many specific code comments. Other than the console warning and long delay, this seems very solid.

frontend/src/app/index.js
```diff
@@ -6,33 +6,73 @@ import moment from "moment";
...
-  home,
-  onboarding,
-  retire,
-  news
-};
-
-const name = document.body.dataset.pageName;
-const param = document.body.dataset.pageParam;
-if (name in routes) {
-  routes[name](param);
```
fzzzy, line 35: I totally forgot we had to do this this way. Nice that it can be removed.

frontend/src/app/index.js
```diff
@@ -6,33 +6,73 @@ import moment from "moment";
...
+    </Switch>
+  </BrowserRouter>;
+}
+
+const s = store();
+const provider = <Provider store={ s }>
+  { routes() }
+</Provider>;
+
+if (typeof document !== "undefined") {
```
fzzzy: I feel like all stuff that needs to only run in the browser context should be moved behind a single action that gets fired by a single place, but that's probably some cleanup for another day.

frontend/src/app/index.js
lmorchard: Yeah, this was a pretty trashy makeshift router

lmorchard:
This is kind of weird... My `testpilot.env` was incorrectly set to `production` and I see what looks like a flip in add-on detection with each navigation between home and experiment and back:



This flipping went away when I changed it to `local`. Other than the add-on detect flipping, I didn't really see any slow down in navigation

dannycoates:
> This is kind of weird...

So when `testpilot.env` is `local` the `navigator.testpilotAddon` flag gets set on example.com and picked up by the App container in [componentDidMount](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App/index.js#L109). Otherwise it sets it to false before InstallManager sets it to true. Nothing new there...

It all happens **again** between routes because each has it's own App container that gets mounted. It'd be nice if we only had one App container, but I haven't been able to make that work yet.

So it's not ideal how the App code runs, but it should only behave badly when the env and domain don't match up.

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/0e68f469bfeda27dcee4d77d67dd240e1a60f946/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/d57b54804b293bdc63bd44a10ef20424a7170ce7/

dannycoates:
Changing the `Route`s to use `render` instead of `component` help the perf a bit. See the [docs](https://reacttraining.com/react-router/web/api/Route/component) about using a function with component for why.

I don't see any easy wins from here. **Correct me if I'm wrong.** I've profiled in Firefox and Chrome. The long poles are the mounting of the ExperimentPage and painting.

For me on Nightly js takes about 400ms and paint around 280ms fullscreen.

screen shot 2018-03-14 at 4 19 39 pm

The Chrome profile gives more readable detail on what the js is doing, and happens to be faster. You can see how long individual components take.

screen shot 2018-03-14 at 3 48 03 pm

I bet we can speed that up, but it seems like it'll take more invasive measures.

chuckharmston (approved):

chuckharmston:
Copying a comment I made on IRC:

> now that we’ve tagged, I’d be cool shipping it and asking @SoftVision-PaulOiegas and team to help us pinpoint where the performance issues are.

fzzzy:
@dannycoates That's very useful info about component versus render, thanks.

frontend/src/app/lib/inject.js
```diff
@@ -1,34 +1,20 @@
...
-import ReactDOM from "react-dom";
 import { Provider } from "react-redux";
-
-import { setupAddonConnection } from "./InstallManager";
-import store from "../store";
+import { StaticRouter } from "react-router";
 import App from "../containers/App";
+import store from "../store";
 
 export default function inject(name, component, callback) {
```
fzzzy: This is no longer injecting the node into the dom, so it's kind of badly named now.

dannycoates: Yeah, I think we should do another pass of refactoring around this. I wanted to make this diff about as small as I could to start.

fzzzy: It might be better just to copy paste the implementation of this into every place that calls inject now. `name` seems to no longer be used and callback is now also no longer used.

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/65612b214f3de2c749f31a6ea67d398a2a331b57/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3416/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/277ef990d8c822370617104c1d732ec274eb877a/

PR #3426, opened 2018-03-16
Move some more things out of gulp into webpack

lmorchard:
This checks off 3 more boxes in #3111 - I think it all works fine, but could definitely use some poking and sanity checks.


lmorchard:
Looks like there might be a failed integration test counting number of experiments, which this shouldn't affect. Graduation dates maybe?

lmorchard:
Also, we lose source maps for CSS, but I'm pretty sure they were broken anyway: Looks like we never produced them from original CSS, only from CSS that had passed through autoprefixer. So, the maps wouldn't have pointed at the right source

dannycoates:
I might be missing something... When I run `./bin/circleci/build-frontend.sh` and `npm run server` I just get a blank blue page in the browser. The html is there. The css *seems* to all be there too. Weird.

lmorchard:
> I might be missing something... When I run ./bin/circleci/build-frontend.sh and npm run server I just get a blank blue page in the browser. The html is there. The css seems to all be there too. Weird.

Whoa, this is bizarre. I don't see it with just `npm start`, but I do with `npm run static && npm run server`. Everything on the page seems to be there in the source, but everything is invisible. Rolling over the page, I can still see the pointer change for links and I can navigate through the site. Videos even start playing

lmorchard:
Still poking at this - looks like there's an `opacity: 0` on `.view` that never gets transitioned to visibility when the site is run through gulp's prod optimizations. Trying to work out what has changed to break that, since the styles look identical in build & dist directories except for changes to image filenames

lmorchard:
Giving this a rebase and a tweak to disable cssnano in postcss - looks like it was conflicting with the gulp optimizations at the tail end, though still not sure exactly how. The CSS still gets minimized though

dannycoates:
It looks like the css order has changed, causing for example:

screen shot 2018-03-23 at 2 01 54 pm
screen shot 2018-03-23 at 2 01 43 pm

lmorchard:
> It looks like the css order has changed

Ah hah, turns out the import order in `frontend/src/app/index.js` matters. 😞 

Tweaked that so that the common styles now get imported first before all the sub-pages, and that seems to have moved the `normalize.css` stuff from the end of the final CSS and back to the beginning

lmorchard:
Hmm, looking at the integration test failure after rebasing against master... Seems to be failing on the sticky footer stuff, which should no longer be displayed since we no longer have a featured experiment.

Oh, and it only fails on Nightly and not Release. <strike>Weird.</strike> Oh, I see. We [skip those tests on Release](https://1188-40666253-gh.circle-artifacts.com/0/home/circleci/project/integration-test-results/ui-test-release.html) but [run them on Nightly](https://1188-40666253-gh.circle-artifacts.com/0/home/circleci/project/integration-test-results/ui-test-nightly.html). And somehow, the add-on install confirmation isn't appearing, which seems like a weird thing for this PR to cause :/

lmorchard:
FWIW, just noticed the integration tests fail in the same way on another PR https://circleci.com/gh/mozilla/testpilot/9067

PR #3435, opened 2018-03-21
Auditing metrics for featured experiment

meandavejustice:
filed https://github.com/mozilla/testpilot/issues/3436 as a follow up since the Featured Video metric opens up some security questions.

- refs #3234

### DIFF
--- a/docs/metrics/ga.md
+++ b/docs/metrics/ga.md
@@ -116,7 +116,6 @@ Here are the current events on the website as of this writing, grouped by their 
 | Featured Experiment Button       | button click          | Manage Featured Button State  | [1, 2, 3, 4, 5, 10, 11, 13]|
 | Click Cancel on tour dialogue    | button click          | cancel tour                   | [1, 2, 3, 10, 11, 13]      |
 | Complete the tour                | button click          | complete tour                 | [1, 2, 3, 10, 11, 13]      |
-| Click Take Tour on tour dialogue | button click          | take tour                     | [1, 2, 3, 10, 11, 13]      |
 | Click next during Tour           | button click          | forward to step `n`           | [1, 2, 3, 10, 11, 13]      |
 | Click back during Tour           | button click          | back to step `n`              | [1, 2, 3, 10, 11, 13]      |
 | Click pagination dot during Tour | button click          | dot to step `n`               | [1, 2, 3, 10, 11, 13]      |
--- a/frontend/src/app/components/ExperimentTourDialog/index.js
+++ b/frontend/src/app/components/ExperimentTourDialog/index.js
@@ -8,7 +8,6 @@ import StepModal from "../StepModal";
 
 import type { InstalledExperiments } from "../../reducers/addon";
 
-import { getBreakpoint } from "../../containers/App";
 import { experimentL10nId } from "../../lib/utils";
 
 type ExperimentTourDialogProps = {
@@ -31,7 +30,6 @@ export default class ExperimentTourDialog extends React.Component {
       dimension1: hasAddon,
       dimension2: installed ? Object.keys(installed).length > 0 : false,
       dimension3: installed ? Object.keys(installed).length : 0,
-      dimension10: getBreakpoint(window.innerWidth),
       dimension11: experiment.slug,
       dimension13: "Featured Experiment"
     });
--- a/frontend/src/app/components/ExperimentTourDialog/index.js
+++ b/frontend/src/app/components/ExperimentTourDialog/index.js
@@ -45,7 +45,7 @@ export default class ExperimentTourDialog extends React.Component {
     };
 
     if (isFeatured) {
-      gaArgs = this.applyAdditionalMetricsArgs;
+      gaArgs = this.applyAdditionalMetricsArgs(gaArgs);
     }
 
     sendToGA("event", gaArgs);
@@ -62,7 +62,7 @@ export default class ExperimentTourDialog extends React.Component {
     };
 
     if (isFeatured) {
-      gaArgs = this.applyAdditionalMetricsArgs;
+      gaArgs = this.applyAdditionalMetricsArgs(gaArgs);
     }
 
     sendToGA("event", gaArgs);
@@ -79,7 +79,7 @@ export default class ExperimentTourDialog extends React.Component {
     };
 
     if (isFeatured) {
-      gaArgs = this.applyAdditionalMetricsArgs;
+      gaArgs = this.applyAdditionalMetricsArgs(gaArgs);
     }
 
     sendToGA("event", gaArgs);
--- a/frontend/src/app/components/ExperimentTourDialog/index.js
+++ b/frontend/src/app/components/ExperimentTourDialog/index.js
@@ -25,23 +25,22 @@ export default class ExperimentTourDialog extends React.Component {
   props: ExperimentTourDialogProps
 
   applyAdditionalMetricsArgs(gaArgs: Object) {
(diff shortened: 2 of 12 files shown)


docs/metrics/ga.md
```diff
@@ -116,7 +116,6 @@ Here are the current events on the website as of this writing, grouped by their
 | Featured Experiment Button       | button click          | Manage Featured Button State  | [1, 2, 3, 4, 5, 10, 11, 13]|
 | Click Cancel on tour dialogue    | button click          | cancel tour                   | [1, 2, 3, 10, 11, 13]      |
 | Complete the tour                | button click          | complete tour                 | [1, 2, 3, 10, 11, 13]      |
-| Click Take Tour on tour dialogue | button click          | take tour                     | [1, 2, 3, 10, 11, 13]      |
```
meandavejustice, line 119: There isn't a take tour button on the tour dialogue, though I may be missing something here. cc/ @teonbrooks

frontend/src/app/components/ExperimentTourDialog/index.js
```diff
@@ -6,44 +6,85 @@ import React from "react";
...
 export default class ExperimentTourDialog extends React.Component {
   props: ExperimentTourDialogProps
 
+  applyAdditionalMetricsArgs(gaArgs: Object) {
+    const { experiment, installed, hasAddon } = this.props;
+    return Object.assign(gaArgs, {
+      dimension1: hasAddon,
+      dimension2: installed ? Object.keys(installed).length > 0 : false,
+      dimension3: installed ? Object.keys(installed).length : 0,
+      dimension10: getBreakpoint(window.innerWidth),
```
chuckharmston: `sendToGA` [already includes](https://github.com/mozilla/testpilot/blob/ee2d65de75f721022b77ee1c06fbd95b412bf7ae/frontend/src/app/containers/App/index.js#L249) `dimension10` for all requests.

frontend/src/app/components/ExperimentTourDialog/index.js
```diff
@@ -6,44 +6,85 @@ import React from "react";
...
+
   stepNextPing = (newStep: number) => {
-    this.props.sendToGA("event", {
+    const { sendToGA, isFeatured } = this.props;
+
+    let gaArgs = {
       eventCategory: "ExperimentDetailsPage Interactions",
       eventAction: "button click",
       eventLabel: `forward to step ${newStep}`,
-      dimension11: this.props.experiment.slug
```
chuckharmston, line 27: Why are we deleting `dimension11`? The docs certainly make it seem like it should be included: https://github.com/mozilla/testpilot/blob/master/docs/metrics/ga.md#custom-dimensions

chuckharmston:
I don't see any implementation where `experiment13` is `Experiment List`

meandavejustice:
@chuckharmston I was only including featured experiment metrics in the pull request. I assumed backfilling dimensions for other metric calls would be out of the scope of this issue

chuckharmston:
Not sure I'd characterize it exactly that way; I see it more as implementing the changes Teon proposed in #3391.

The other values of `dimension13` were introduced so that we can segment some of the events (like `button click / take tour`), so just having them for featured experiments and nowhere else only partially gets there.

chuckharmston (changes requested):

frontend/src/app/components/ExperimentTourDialog/index.js
meandavejustice: You're correct, I'll add that one back in

meandavejustice:
@chuckharmston 

> I see it more as implementing the changes Teon proposed in #3391.

That makes sense, I'll go ahead and apply all of those changes in this pr.

docs/metrics/ga.md
teonbrooks: hmm...yeah, I agree, there isn't a take tour button, I don't know how I came to have that then. that's fine to delete

teonbrooks:
related to this pr, just noticed that there is no longer a featured experiment section on the testpilot website. is that gone temporarily or is that the ui for the page?

chuckharmston:
> related to this pr, just noticed that there is no longer a featured experiment section on the testpilot website. is that gone temporarily or is that the ui for the page?

We un-featured Notes for the time being.

meandavejustice:
>I don't see any implementation where experiment13 is Experiment List

@chuckharmston After reviewing this again today, I'm not seeing which metrics these are to be added to. Nothing in the [featured experiment metrics](https://github.com/mozilla/testpilot/blob/master/docs/metrics/ga.md#featured-experiment) section applies to the experiment list. Unless I'm missing something, we should remove "Experiment LIst" from the possible values for dimension13

cc/ @teonbrooks

meandavejustice:
@chuckharmston can I get you to take another look at this when you have time?

frontend/src/app/components/ExperimentTourDialog/index.js
```diff
@@ -6,44 +6,83 @@ import React from "react";
...
-    });
+      dimension11: experiment.slug,
+      dimension13: "Experiment Detail"
+    };
+
+    if (isFeatured) {
+      gaArgs = this.applyAdditionalMetricsArgs;
```
pdehaan: I'm a bit confused here... It looks like `this.applyAdditionalMetricsArgs` is actually a method (defined above on L27), and takes a `gaArgs` object argument. Is this just calling `applyAdditionalMetricsArgs()` with a null/undefined `gaArgs` argument? Or should this be something like `gaArgs = this.applyAdditionalMetricsArgs(gaArgs)` which will inject additional properties into the `gaArgs` defined above on L39?

chuckharmston: Yeah, this is exactly right.

frontend/src/app/components/ExperimentTourDialog/index.js
```diff
@@ -88,25 +127,25 @@ export default class ExperimentTourDialog extends React.Component {
   }
 
   onCancel = (ev: Object) => {
-    const { sendToGA, onCancel } = this.props;
+    const { sendToGA, onCancel, experiment } = this.props;
```
chuckharmston, line 130: This soothes my elephant brain. 🐘

frontend/src/app/components/FeaturedExperiment/FeaturedButton.js
```diff
@@ -42,18 +42,56 @@ export default class FeaturedButton extends React.Component {
...
+      sendToGA("event", {
+        eventCategory,
+        eventAction: "link click",
+        eventLabel: "Open general privacy",
+        dimension1: hasAddon,
+        dimension2: Object.keys(installed).length > 0,
+        dimension3: Object.keys(installed).length,
+        dimension11: experiment.slug,
+        dimension13: "Featured Experiment"
+      }, ev);
```
chuckharmston: There's a lot of duplication in the above three methods. It might make sense to have a single currying function that does the majority of the work.

chuckharmston (changes requested):
This is looking very good. Few notes below, then we should be good to merge it. Very nice job! 👍

frontend/src/app/components/ExperimentTourDialog/index.js
meandavejustice: @pdehaan thanks for catching this, I accidently pulled off the actual calling of the function when I switched away from using `.bind(this)`

meandavejustice:
@chuckharmston Comments addressed, merging :)

PR #3484, opened 2018-04-11
put the loading spinner back

### LINKED ISSUE #3453: Don't render application until we have the correct application state (fix loading jank)
Test Pilot is looking pretty zippy these days, but experiments details pages still flash incorrect content for a quick second while rendering.

To view this, go to: 
1. https://testpilot.firefox.com/experiments/notes with test pilot installed
2. Note the flash of content at the top of the page before the site detects the add-on

Either we should avoid rendering the page content until state is resolved or we should bring back the loader. Loaders are better ui than janky loading.

IIRC this is an issue on the featured experiment UI as well, though we're not displaying a featured experiment at the moment.

dannycoates:
This is *sort of* quick and dirty as it doesn't wait for the addon manager to return the installed state, but in my experience the l10n part tends to be the longer wait anyway.

fixes #3453

### DIFF
--- a/frontend/src/app/components/Loading/index.scss
+++ b/frontend/src/app/components/Loading/index.scss
@@ -1,3 +1,5 @@
+@import '../../../styles/_utils';
+
 // HT: https://www.pexels.com/blog/css-only-loaders/
 .loading {
   @include flex-container(row, space-between, center);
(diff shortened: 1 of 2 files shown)


meandavejustice:
Looks like the unit tests are failing because of changes related to https://github.com/mozilla/testpilot/pull/3435 somehow. May have been a circle issue though cus everything was green on that pr. Investigating.

Pulled this down, tested. Everything else looks good to me. :+1:

meandavejustice:
Unit test issues are fixed in #3485. This pr has nothing to do with those errors, merging.

PR #3489, opened 2018-04-12
Progress toward moving content generation into Webpack

lmorchard:
This is a real big one, but it gets us a big step closer to entirely moving off gulp. This basically moves everything but static HTML creation and the gulp-rev-all step into Webpack.

Opening a WIP PR now to start getting some feedback on the new ContentTransformPlugin and restructuring the content generation


bin/build-content-json.js
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env node
```
lmorchard, line 1: This script is kind of a temporary hack: `frontend/src/app/store.js` imports `../../build/api/{experiments,news_updates}.json`. That causes chicken-and-egg problem, because those JSON assets need to be built before everything else can be bundled.

Running this script first takes care of that. Once HTML generation moves into webpack, we should be able to import YAML directly in store.js and no longer need this.

bin/extract-strings.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
```
lmorchard, line 1: This script exists to continue supporting the l10n flow and `npm run l10n:check` by generating just the FTL in isolation from the rest of the webpack build

frontend/lib/content-transformer-plugin.js
```diff
@@ -0,0 +1,134 @@
+const fs = require("fs");
```
lmorchard, line 1: This is the Webpack plugin I've been hacking on to generate static content from collections of files. It could use some documentation - but until then, the modules under `frontend/lib/content` are working examples of use.

This is a little over-engineered because it supports async/await all the way down. But, it handles standalone content generation in scripts. It also works as a Webpack plugin that sets up file watchers for all the source files to re-generate static content as needed.

frontend/tasks/content.js
```diff
@@ -1,244 +0,0 @@
-const gulp = require("gulp");
```
lmorchard, line 1: Everything in here has been scooped out and re-organized into simpler content transform functions in modules under `frontend/lib/content`

frontend/lib/content/css.js
```diff
@@ -0,0 +1,16 @@
+const cssForExperiment = experiment => `
```
lmorchard, line 1: This is probably the simplest example of a transform function. Just turns experiments into a single big CSS file.

lmorchard (commented):
Highlighting some points of interest

lmorchard:
Pushed an update to fix some issues in generating FTL

dannycoates:
Looks great to me 👍

meandavejustice:
Aiming to get this reviewed and merged tomorrow

PR #3500, opened 2018-04-13
changed addon update-version

### LINKED ISSUE #3459: review update-version script
This script is causing problems with web-ext. Do we still need it?

dannycoates:
I went with the more lax Firefox version requirements since it lets us have more entropy. With the goofball sorted sha we'll have a few more bits to avoid collisions than with the 15 bits using the Chrome requirements.

I didn't know what to do with the circle stuff so I took it out.

fixes #3459

### DIFF
--- a/addon/manifest.json
+++ b/addon/manifest.json
@@ -8,7 +8,7 @@
       "update_url": "https://testpilot.firefox.com/files/@testpilot-addon/updates.json"
     }
   },
-  "version": "3.0.0",
+  "version": "3.0.0vffeb6433",
   "icons": {
     "16": "icons/icon-16.png",
     "32": "icons/icon-32.png",
--- a/addon/package.json
+++ b/addon/package.json
@@ -2,7 +2,7 @@
   "title": "Test Pilot",
   "name": "testpilot-addon",
   "id": "@testpilot-addon",
-  "version": "3.0.0",
+  "version": "3.0.0vffeb6433",
   "private": true,
   "description": "Test Pilot is a privacy-sensitive user research program focused on getting new features into Firefox faster.",
   "repository": "mozilla/testpilot",
@@ -22,7 +22,7 @@
   },
   "scripts": {
     "start": "web-ext run",
-    "prepackage": "npm run locale && npm run lint && npm run lint:extension",
+    "prepackage": "npm run locale && npm run lint && npm run lint:extension && npm run update-version",
     "package": "shx rm -rf web-ext-artifacts && cross-env web-ext build --overwrite-dest && shx mv web-ext-artifacts/test_pilot-*.zip ./addon.xpi",
     "lint": "eslint .",
     "lint:extension": "web-ext lint --self-hosted",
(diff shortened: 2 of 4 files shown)


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3500/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/219c06a9051af6fde4bb9aa9b5eb819f95353419/

meandavejustice:
@dannycoates prepackage in circleci is the only place/reason we have the update-version script from my understanding. 

from https://github.com/mozilla/testpilot/issues/3459#issuecomment-378020870

> Yes, if we want a signed build of the add-on deployed to dev with every master commit without having to change the version number every time. (But, it's possible we stop caring about that at some point.)

.gitignore
```diff
@@ -30,6 +30,7 @@ tmp/
 addon/_locales/
 
 # Files to ignore within specific directories
+addon/update.rdf
```
meandavejustice: I don't think there should be an `update.rdf` generated now

meandavejustice (changes requested):
based on @lmorchard's comment on https://github.com/mozilla/testpilot/issues/3459#issuecomment-378020870
This should fix the circleci building on dev since you added it back into the `prepackage` script. We'll have to merge this to test on dev.

Only comment is on the addition to `.gitignore`

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3500/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/000a4a2c099a7c1cbfc0575885c0651c975e0548/

addon/manifest.json
```diff
@@ -8,7 +8,7 @@
       "update_url": "https://testpilot.firefox.com/files/@testpilot-addon/updates.json"
     }
   },
-  "version": "3.0.0",
+  "version": "3.0.0vffeb6433",
```
lmorchard, line 11: Looks like the altered version snuck into the commit - we should change this back to 3.0.0 so it doesn't end up on production

addon/package.json
```diff
@@ -2,7 +2,7 @@
   "title": "Test Pilot",
   "name": "testpilot-addon",
   "id": "@testpilot-addon",
-  "version": "3.0.0",
+  "version": "3.0.0vffeb6433",
```
lmorchard, line 5: Same here, RE: version

addon/manifest.json
dannycoates: > so it doesn't end up on production

you mean like production right now? ;)

screen shot 2018-04-16 at 10 49 10 am

We should have a build directory since we have build time assets like this

lmorchard: Yeah 😞 I forget when that broke, I think sometime when we moved to mozilla key signing or circle 2.0. Also, yeah, this script should probably be reworked to not alter the in-repo package or manifest

PR #3534, opened 2018-05-04
measure first featured video click

### LINKED ISSUE #3234: Featured experiments UI lacks event reporting

### LINKED ISSUE #3436: Add metric for featured experiment video
This is a follow up issue for https://github.com/mozilla/testpilot/pull/3435.

The [Click Featured Experiment Video](https://github.com/mozilla/testpilot/blob/master/docs/metrics/ga.md#featured-experiment) metric calls for dimension12, which asks "How much did the video did the user watch?". In order to attain this information we'll need to add youtube's Iframe api.

We'll probably need to get a sign off from security to add this exception to our CSP ruleset.

meandavejustice:
- fixes #3234
- fixes #3436

### DIFF
--- a/docs/metrics/ga.md
+++ b/docs/metrics/ga.md
@@ -109,17 +109,17 @@ Here are the current events on the website as of this writing, grouped by their 
 
 | Description                      | `eventAction`         | `eventLabel`                  | Custom Dimensions          |
 | -------------------------------- | --------------------- | ----------------------------- | -------------------------- |
-| Click Featured Experiment Video  | video click           | Play Featured video           | [1, 2, 3, 12, 10, 11, 13]  |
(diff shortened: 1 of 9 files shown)


dannycoates (changes requested):
I'm strongly against polling the page to see if the iframe gets clicked. I don't know of a less invasive way to do it that wouldn't require significant effort to do reliably though. (It's unfortunate that it's not just an event handler like most other things) 

Given that we can find how many views the video has by other means I think we should reconsider how important tracking this would be. @johngruen ?

johngruen:
We can get views by domain from youtube's analytics...this feels more stable overall even if it's less granular particularly if we get embedded on other firefox properties... +1 to @dannycoates.

image

PR #3540, opened 2018-05-09
Remove update list

### LINKED ISSUE #3527: review l10n ids for UpdateList component
Update List is rarely shown, only when:
* addon is installed
* no featured experiment
* there are updates to show
So it's easily overlooked.

Seems like some l10n ids were maybe wrongly deleted.

meandavejustice:
- fixes #3527


flodolo (approved):
No issues strings-wise

PR #3550, opened 2018-05-14
Photonize footer

### LINKED ISSUE #3225: New Photonized Footer design based on AMO
Should update the footer design.

Spec:
image

Link:
https://mozilla.github.io/testpilot-assets/Test_Pilot_Website/01_Measurements/Photonized_Txp_Website_Final/#artboard2

meandavejustice:
- fixes #3225


locales/en-US/app.ftl
```diff
@@ -12,9 +12,16 @@ footerLinkPrivacy = Privacy
...
-
-## Items in the menu.
+footerLinkAboutHeader = About
+footerLinkAboutUs = About us
+footerLinkSupportHeader = Support
+footerLinkDownload = Download Firefox
+footerLinkMobile = Mobile
+footerLinkFeatures = Features
+footerLinkBeta = Beta, Nightly, Developer Edition
+footerLinkRetire = <h3>Uninstall Test Pilot</h3>
```
flodolo: Do we need to expose the HTML tags here? Can we just have the string?

locales/en-US/app.ftl
```diff
@@ -12,9 +12,16 @@ footerLinkPrivacy = Privacy
...
 footerLinkLegal = Legal
 footerLinkFeedback = Give Feedback
-footerLinkAbout = About Test Pilot
-
-## Items in the menu.
+footerLinkAboutHeader = About
+footerLinkAboutUs = About us
+footerLinkSupportHeader = Support
+footerLinkDownload = Download Firefox
+footerLinkMobile = Mobile
```
flodolo: I assume this switches to the mobile version of the website? Can we add a comment to explain it?

meandavejustice: This links to the site where you can download mobile firefox, android or ios

locales/en-US/app.ftl
meandavejustice: This is a typo, thanks for catching that

locales/en-US/app.ftl
flodolo: OK, definitely worth a comment then.

locales/en-US/app.ftl
```diff
@@ -12,9 +12,17 @@ footerLinkPrivacy = Privacy
 footerLinkTerms = Terms
 footerLinkLegal = Legal
 footerLinkFeedback = Give Feedback
-footerLinkAbout = About Test Pilot
-
-## Items in the menu.
+footerLinkAboutHeader = About
```
flodolo: After looking at the code, I think it would be a good idea to add a comment to this one, e.g.

```
# This is a header displayed above a set of links about Mozilla and Test Pilot
```

locales/en-US/app.ftl
```diff
@@ -12,9 +12,17 @@ footerLinkPrivacy = Privacy
...
+footerLinkAboutUs = About us
```
flodolo: Should this be "About Mozilla"? I would expect *us* = Test Pilot

flodolo (approved):
In general it looks good, one question "About us".

locales/en-US/app.ftl
meandavejustice: us in this case is mozilla, but It could be test pilot I suppose

PR #3560, opened 2018-05-18
make tours optional for add-on experiments

johngruen:
Makes it so we don't need a tour for add-on experiments that handle their own onboarding


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3560/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/76552e5e873c90bd7ad7437212703565ee7b0767/

PR #3573, opened 2018-05-24
Fix #3494 Document what the toplevel App container does, and suggest …

### LINKED ISSUE #3494: Annotate App Container to Explain How It Works For Future Generations
The App container has some stuff that might be better in utils.js or elsewhere (@dannycoates : details needed).

fzzzy:
…some code that can be refactored.


dannycoates (approved):
LGTM 💯

PR #3587, opened 2018-05-29
Another fix for the storybook deploy script

lmorchard:
accidentally a semicolon


lmorchard:
Probably will just self-merge since the error is making tests fail on master :/

PR #3602, opened 2018-06-04
Adds content for Advance.

chuckharmston:
home

detail

example-screenshot


chuckharmston:
(@flodolo this is so WIP that there's no need for review; I'll ping you when we're ready to merge.)

pdehaan:
@chuckharmston Interesting! Can you ping me the repo URL for Laserlike? I couldn't find it in GitHub org (unless it's private or something sneaky sneaky).

content-src/experiments/laserlike.yaml
```diff
@@ -0,0 +1,78 @@
+id: 18
```
pdehaan: Do we need some `dev: true` style footgun protection so this doesn’t accidentally land on prod?

chuckharmston: Good call! ✅

PR #3623, opened 2018-06-14
Add Lockbox

nickbrandt:
- We still need to figure out how to get the Privacy section to link off to the Lockbox site for the full privacy content.
- Some other content may still need adjusted for other areas of the Lockbox Test Pilot instance.
- We want to ensure that this is set to not be live for now.


devinreams:
Moved to #3624 for simplicity and because I hate git sometimes

content-src/experiments/lockbox.yaml
```diff
@@ -0,0 +1,162 @@
+id: 16
+slug: firefox-lockbox
+title: 'Firefox Lockbox'
+# subtitle: 'Take your passwords everywhere'
+order: 3
+is_featured: true
+platforms: ['mobile']
+web_url: https://example.com/web-experiment
```
pdehaan, line 8: Should this be https://lockbox.firefox.com ?

PR #3624, opened 2018-06-14
Add Lockbox (DO NOT MERGE TILL AFTER 6/28 deploy and l10n merge)

devinreams:
Fixes https://github.com/mozilla-lockbox/mozilla-lockbox.github.io/issues/47

My bad, I gave bad advice so re-created @nickbrandt's branch here. Replacing #3623

- [x] We want to ensure that this is set to not be live for now.
- [x] We are unsure what ID # should be used for Lockbox in the YAML file
- [x] We need to get all the proper PI/QA folks included
- [x] Some other content may still need adjusted for other areas of the Lockbox Test Pilot instance.
- [x] but make sure it becomes "Featured" when ready to launch
- [x] Final items: https://github.com/mozilla-lockbox/mozilla-lockbox.github.io/issues/44

PR #3627, opened 2018-06-18
Add docs about experiments strings not included when dev is true

devinreams:
Follow up from https://github.com/mozilla/testpilot/pull/3624#issuecomment-397460789

We had an experiment set to "dev: false" which exported strings and then they got removed at "dev true" slightly unexpectedly. This is to help clarify why/when strings are not exported.

### DIFF
--- a/docs/content/reference.md
+++ b/docs/content/reference.md
@@ -21,6 +21,8 @@ An experiment's YAML file is considered to contain the canonical localization fo
 
 To manually extract strings for Pontoon, run the command `npm run l10n:extract`. If you have a dev server running via `npm start`, this should happen automatically when you change YAML files. In either case, the FTL file should *not* be edited manually.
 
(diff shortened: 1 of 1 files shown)


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3627/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/84f3ca4d44e0304a30944001bac8f775e6e9793b/

docs/content/reference.md
```diff
@@ -593,12 +595,14 @@ locale_grantlist:
 
 ## `dev`
 
-A boolean indicating whether this experiment should only appear in a dev environment, i.e. for testing or localization. Required.
+A boolean indicating whether this experiment should only appear in a dev environment, i.e. for testing or active development. Required.
```
lmorchard, line 598: Oof, yeah, good catch - that was totally wrong 😞

PR #3640, opened 2018-06-26
make experiment types more detailed: Merge 6/28 with Devin's patch

### LINKED ISSUE #3634: "mobile experiment" description should be ios or android unless both are present.

johngruen:
@devinreams @meandavejustice this PR makes the experiment type UI a little more expressive and standardized so that instead of saying "mobile experiment" we actually declare specific device types.  ~~Would be good to get this merged with the other mobile stuff today or tomorrow!~~ 

Err, actually since this has strings we should push it to l10n with devin's PR for lockbox after we push to stage.

fixes #3634

### DIFF
--- a/frontend/src/app/components/ExperimentPlatforms/index.js
+++ b/frontend/src/app/components/ExperimentPlatforms/index.js
@@ -4,22 +4,19 @@ import { Localized } from "fluent-react/compat";
 
 import "./index.scss";
 
-const AVAILABLE_PLATFORMS = ["web", "addon", "mobile"];
+const AVAILABLE_PLATFORMS = ["web", "addon", "ios", "android"];
 
 export default function ExperimentPlatforms({ experiment }) {
-  const platforms = experiment.platforms.map(p => (p === "android" || p === "ios") ? "mobile" : p) || [];
-  const enabledPlatforms = AVAILABLE_PLATFORMS
-    .filter(platform => platforms.indexOf(platform) !== -1);
+  let enabledPlatforms = AVAILABLE_PLATFORMS
+    .filter(platform => experiment.platforms.includes(platform));
   if (enabledPlatforms.length === 0) { return null; }
+
+  // sort list so that l10n ids are consistent regardless of array order in YAML
+  enabledPlatforms.sort();
 
   let l10nId = "experimentPlatform" + enabledPlatforms
     .map(platform => platform.charAt(0).toUpperCase() + platform.slice(1))
     .join("");
-
-  if (l10nId === "experimentPlatformMobile") {
-    // HACK: string changed after initial commit, so the ID had to change
-    l10nId = "experimentPlatformMobileApp";
-  }
 
   return (
     <h4 className="experiment-platform">
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -177,13 +177,17 @@ startedDateLabel = Experiment start date: <b>{$startedDate}</b>
 nonExperimentDialogHeaderLink = Test Pilot
 
 ## Label shown next to a series of icons indicating whether an experiment is available as an add-on, mobile app, and/or web site
-experimentPlatformWebAddonMobile = Firefox / web / mobile experiment
-experimentPlatformWebAddon = Firefox / web experiment
-experimentPlatformWebMobile = web / mobile experiment
-experimentPlatformAddonMobile = Firefox / mobile experiment
-experimentPlatformWeb = web experiment
 experimentPlatformAddon = Firefox experiment
-experimentPlatformMobileApp = mobile experiment
+experimentPlatformAndroid = Android experiment
+experimentPlatformIos = iOS experiment
+experimentPlatformWeb = web experiment
+experimentPlatformAddonWeb = Firefox / web experiment
+experimentPlatformAddonAndroid = Android / Firefox experiment
+experimentPlatformAddonIos = iOS / Firefox experiment
+experimentPlatformAddonAndroidWeb = Android / Firefox / web experiment
+experimentPlatformAddonAndroidIosWeb = Android / iOS / Firefox / web experiment
+experimentPlatformAndroidWeb = Android / web experiment
+experimentPlatformAndroidIos = Android / iOS experiment
 
 ## Shown when an experiment requires a version of Firefox newer than the user's.
 upgradeNoticeTitle = {$title} requires Firefox {$min_release} or later.
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -184,6 +184,7 @@ experimentPlatformWeb = web experiment
 experimentPlatformAddonWeb = Firefox / web experiment
 experimentPlatformAddonAndroid = Android / Firefox experiment
 experimentPlatformAddonIos = iOS / Firefox experiment
+experimentPlatformAddonAndroidIos = Android / iOS / Firefox experiment
 experimentPlatformAddonAndroidWeb = Android / Firefox / web experiment
 experimentPlatformAddonAndroidIosWeb = Android / iOS / Firefox / web experiment
 experimentPlatformAndroidWeb = Android / web experiment
--- a/content-src/experiments/dev-example.yaml
+++ b/content-src/experiments/dev-example.yaml
@@ -2,7 +2,7 @@ dev: true
 id: 999
 title: 'Dev Example'
 slug: dev-example
-platforms: ['addon', 'ios']
+platforms: ['addon', 'ios', 'android']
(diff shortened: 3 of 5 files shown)


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3640/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/4344c87d2fda696a6f9d3d42c1a1a2a933697156/

frontend/src/app/components/ExperimentPlatforms/index.js
```diff
@@ -4,23 +4,20 @@ import { Localized } from "fluent-react/compat";
...
 
 export default function ExperimentPlatforms({ experiment }) {
-  const platforms = experiment.platforms.map(p => (p === "android" || p === "ios") ? "mobile" : p) || [];
-  const enabledPlatforms = AVAILABLE_PLATFORMS
-    .filter(platform => platforms.indexOf(platform) !== -1);
+  let enabledPlatforms = AVAILABLE_PLATFORMS
+    .filter(platform => experiment.platforms.indexOf(platform) !== -1);
   if (enabledPlatforms.length === 0) { return null; }
 
+  // sort list so that l10n ids are cosnsitent regardless of array order in YAML
```
flodolo: nit: consistent

flodolo (approved):
No issue on the strings, but the labels are getting longer, and that moves the image. Considering a +30% in characters on average, that doesn't look great.

schermata 2018-06-26 alle 12 40 14

johngruen:
@flodolo good point. We don't have any experiments that will use the super long strings yet, but we should make the UI i little more robust before we do.

frontend/src/app/components/ExperimentPlatforms/index.js
```diff
@@ -4,23 +4,20 @@ import { Localized } from "fluent-react/compat";
...
 
-const AVAILABLE_PLATFORMS = ["web", "addon", "mobile"];
+const AVAILABLE_PLATFORMS = ["web", "addon", "ios", "android"];
 
 export default function ExperimentPlatforms({ experiment }) {
-  const platforms = experiment.platforms.map(p => (p === "android" || p === "ios") ? "mobile" : p) || [];
-  const enabledPlatforms = AVAILABLE_PLATFORMS
-    .filter(platform => platforms.indexOf(platform) !== -1);
+  let enabledPlatforms = AVAILABLE_PLATFORMS
+    .filter(platform => experiment.platforms.indexOf(platform) !== -1);
```
pdehaan: Easier to use `includes()` rather than `indexOf() !== -1`?

johngruen: c'mon man, i didn't even write that code :)

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3640/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/0ee0e09b2e92a3203a64d8e9c840b2ff45b4fc0e/

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3640/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/523e6b7ad8205b0423e1e080ebcf1dcc3e18f8c9/

devinreams:
@johngruen you may already be aware and trying to tackle this elsewhere, perhaps at #3643, but on the home page the display for for "Available on..." uses the platform names as-is (lowercase) so it says "ios" and "android" instead of "iOS" and "Android" (proper caps).

https://github.com/mozilla/testpilot/blob/5ce9cef92baf1c9398c28efa7fdc7d934c63715e/frontend/src/app/components/ExperimentPlatforms/index.js#L30

firefox_test_pilot

lmorchard:
> on the home page the display for for "Available on..." uses the platform names as-is (lowercase) so it says "ios" and "android" instead of "iOS" and "Android" (proper caps).

That code constructs a concatenated en-US string with all lowercase as a fallback / placeholder. 

But, [the `l10nId` we construct](https://github.com/mozilla/testpilot/pull/3640/files#diff-61892e5984a0031fb30379a9de834a0cR17) should always cause that fallback to be replaced with [the appropriate localizable strings from FTL](https://github.com/mozilla/testpilot/pull/3640/files#diff-fa0fe0311fc532fa1c84a7f9414af2eaR181). (i.e. those *do* have the proper case for en-US)

If that's not happening, then there's a bug here. We might need to peek at what that `l10nId` looks like and see if it's not matching the FTL and thus resulting in the fallback being displayed.

devinreams:
>That code constructs a concatenated en-US string with all lowercase as a fallback / placeholder.

@lmorchard Ah yeah, thanks for explaining.

Then I believe I found just a minor bug due to the inclusion of _all three_ platforms in `dev-example.yml`: `['addon', 'ios', 'android']`

If I just include `['ios', 'android']` or `['addon', 'android']` its OK. 

screen shot 2018-06-26 at 12 25 06 pm

screen shot 2018-06-26 at 12 24 29 pm

So just this combination (of all three) is not being handled, I think? And I assume Notes will fall into all three at some point in the future but not yet (so not an immediate concern)?

lmorchard:
Yeah, looks like we're missing the string `experimentPlatformAddonAndroidIos` here. It's annoying, but we have to spell out every possible permutation of the platforms because simple string concatenation is not localizable.

lmorchard:
Looks like I can commit to this branch, so I just tried pushing an update to add the string. This is an area that could really use some Storybook stories to flip through all the states and verify things

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3640/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/87ea6af6bb1fd3647af6f1c729ad6eb7afc42dab/

lmorchard (approved):
After fixing the missing string, this looks okay to me now. Wouldn't mind a second okay, since one of my commits is in there now :)

pdehaan:
> we have to spell out every possible permutation of the platforms because simple string concatenation is not localizable.

Instead of string concatenation and listing every combination, why not just treat the platforms a list so we'd show:

> ### Dev Example
> - [add-on icon] Firefox experiment
> - [mobile icon] Android experiment
> - [mobile icon] iOS experiment

That way the localizers would only need to translate each string once and we couldnt need to track every combination, and we wouldnt have a semi-confusing cluster of icons.

locales/en-US/app.ftl
```diff
@@ -177,13 +177,18 @@ startedDateLabel = Experiment start date: <b>{$startedDate}</b>
...
+experimentPlatformIos = iOS experiment
+experimentPlatformWeb = web experiment
+experimentPlatformAddonWeb = Firefox / web experiment
+experimentPlatformAddonAndroid = Android / Firefox experiment
+experimentPlatformAddonIos = iOS / Firefox experiment
+experimentPlatformAddonAndroidIos = Android / iOS / Firefox experiment
+experimentPlatformAddonAndroidWeb = Android / Firefox / web experiment
+experimentPlatformAddonAndroidIosWeb = Android / iOS / Firefox / web experiment
+experimentPlatformAndroidWeb = Android / web experiment
+experimentPlatformAndroidIos = Android / iOS experiment
```
pdehaan, line 191: I'd need to chart this out to be sure of all possible combinations (if we care right now), but i think we're missing at least: `"iOS / web experiment"`

flodolo:
Having a list of platforms sounds like a great idea to me.

johngruen:
This block shows up in a number of places, I don't think using lists as @pdehaan proposes would work on the experiment cards. The extraneous part to me seems to be the word "experiment"

We could just do a concatenated string so the format was just something like

```
[icon] [icon] Firefox / Android
```

Or similar. I'm happy to make this happen before tomorrow morning's l10n push. @flodolo would that work for you?

flodolo:
> We could just do a concatenated string so the format was just something like
> 
> [icon] [icon] Firefox / Android
> 
> Or similar. I'm happy to make this happen before tomorrow morning's l10n push. @flodolo would that work for you?

No, unfortunately concatenation is very brittle (spacing, character to use, etc. might change depending on the language).

We would need locale-sensitive [List Formatting](http://cldr.unicode.org/development/development-process/design-proposals/list-formatting), which is all but a proposal for now.

johngruen:
Okay, I'll file a bug to track a less brittle version than what appears here, but let's go ahead and merge this as is since it covers our needs for the current round of experiments.

johngruen:
@meandavejustice can u give this one more sign off

meandavejustice:
:+1:

johngruen:
Actually, there's no reason we can't just merge this. which i am doing.

PR #3649, opened 2018-06-27
Fix #3646 Allow the site to work when an experiment has no measuremen…

### LINKED ISSUE #3646: "Your privacy" section requires measurements
A quick discussion with legal, product, and Test Pilot engineering has led me to think a nice improvement would be for us to "make it possible to leave off the bullets sections" in the "Your privacy" section.

screen shot 2018-06-26 at 11 29 33 am

There are cases where the "measurements" in an experiment can't fully capture everything we want to say, and the `privacy_notice_url` and a blurb _should_ be sufficient to let users know where to get more information.

A rough proposal would be to:

- make sure the page doesn't crash if `measurements` is missing from experiment yaml
- check if `measurements` is empty and if `privacy_notice_url` is not, then modify the blurb and still link to the privacy URL
  >In addition to the data collected by all Test Pilot experiments, you can learn more about the data collection for {experiment} here.
- make sure the "featured experiment" links and home page modal render properly
  screen shot 2018-06-26 at 12 14 58 pm

...or something like that.

fzzzy:
…ts section. In this case the privacy_notice_url will serve the same function.

### DIFF
--- a/content-src/experiments/dev-example.yaml
+++ b/content-src/experiments/dev-example.yaml
@@ -25,16 +25,8 @@ changelog_url: 'https://github.com/meandavejustice/min-vid/blob/master/CHANGELOG
 contribute_url: 'https://github.com/meandavejustice/min-vid'
 bug_report_url: 'https://github.com/meandavejustice/min-vid/issues'
 discourse_url: 'https://discourse.mozilla-community.org/c/test-pilot/min-vid'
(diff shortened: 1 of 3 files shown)


devinreams (approved):
Smart 👍 from a UX perspective.

This way we could use the `privacy_preamble` to say anything that's needed before the link. 🙆‍♂️

PR #3663, opened 2018-06-29
Make title and icon of featured experiment link to detail page

meandavejustice:


SoftVision-PaulOiegas:
Verified on Dev server and everything looks and works ok. Also checked on local with moving several experiments to featured and haven't found any issues either. 👍

PR #3664, opened 2018-06-29
fix construction of download url

meandavejustice:


fzzzy (approved):
There’s no way to test this except to try it. It looks good to me.

PR #3666, opened 2018-06-29
Re-add Lockbox legal notice for featured experiment

### LINKED ISSUE #3665: Featured Lockbox content is missing words for privacy and terms
https://testpilot.dev.mozaws.net/__version__

```js
{
  commit: "0aa1d837ac973412e576caa32065d5221d48d79c",
  version: "",
  source: "https://github.com/mozilla/testpilot",
  build: "https://circleci.com/gh/mozilla/testpilot/11761"
}
```

firefox_test_pilot

devinreams:
Fixes #3665

## Before

firefox_test_pilot

## After
screen shot 2018-06-29 at 1 56 20 pm
screen shot 2018-06-29 at 1 56 25 pm


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3666/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/11976398368a5012e8c6d63909f5ce7185f6bdb3/

flodolo (approved):

content-src/experiments/lockbox.yaml
```diff
@@ -125,4 +125,4 @@ contributors:
 privacy_preamble: >
   Firefox Lockbox allows Firefox Accounts users to access the login and password information (which we call “credentials”) on their mobile phones, outside any browser, by using Firefox Sync. In addition to the data collected by all Test Pilot experiments...
```
flodolo, line 126: There is one other error here :-\

```
In addition to the data collected by all Test Pilot experiments...
```

Shouldn't really be there, and I noticed while reviewing the Italian translation. The safest way to avoid errors would be to use a new string ID for this one, so you'd ignore existing translation, which likely included that piece.

flodolo (changes requested):

flodolo:
P.S. If we could get this in, and exposed to the #l10n branch during the weekend, it would be great.

content-src/experiments/lockbox.yaml
devinreams: >>In addition to the data collected by all Test Pilot experiments...

>Shouldn't really be there

This was a creative solution to including required legal language while simultaneously removing it since we don't have "measurements" for this experiment:

screen shot 2018-06-30 at 7 39 25 am

So while its redundant, I don't have a good way to include that sentence other than adding it to this preamble. 😔

flodolo: > So while its redundant, I don't have a good way to include that sentence other than adding it to this preamble.

Why do you need a preamble at all? The link is visually separated.

devinreams: Yeah, Legal has asked to include a) that blurb that starts with "Firefox Lockbox allows..." and also b) the note about the other Test Pilot data collected and c) the link to the custom policy. 

So that satisfies all the above with just one preamble paragraph and policy link.

flodolo: OK, I think I'm starting to get it. The problem is stating somewhere that the experiment collects the "data collected by all Test Pilot Experiments".

I wonder why it can't be a simple sentence, but at this point it's not an error.

P.S. dev servers seem pretty dead :-\

flodolo (approved):

devinreams:
>If we could get this in, and exposed to the #l10n branch during the weekend, it would be great.

The `l10n` branch is up to date with master! 🔀

Issue #3625, opened 2018-06-14
[Storybook] You should not use Link outside a Router


pdehaan:
Spotted while trolling a Storybook build...

http://testpilot-storybook.dev.mozaws.net/b357c4942f09572d1ff88cc2ba0f4a2f9c57b3f9/?knob-Has%20Addon=false&knob-Enabled=false&knob-Experiment=%7B%22title%22%3A%22Sample%20experiment%22%2C%22description%22%3A%22This%20is%20an%20example%20experiment%22%2C%22subtitle%22%3A%22%22%2C%22slug%22%3A%22snooze-tabs%22%2C%22survey_url%22%3A%22https%3A%2F%2Fexample.com%22%2C%22created%22%3A%222010-06-21T12%3A12%3A12Z%22%2C%22modified%22%3A%222010-06-21T12%3A12%3A12Z%22%7D&selectedKind=ExperimentRowCard&selectedStory=base%20state%20with%20knobs&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

<img width="1437" alt="storybookerror" src="https://user-images.githubusercontent.com/557895/41441904-84dcc040-6fe9-11e8-85c6-a0381138863f.png">

PR #3669, opened 2018-07-02
fix all the buttons

johngruen:
This patch reworks the experiment control buttons in order to make multi-platform experiments look better.


frontend/src/app/containers/ExperimentPage/ExperimentButtons.js
```diff
@@ -0,0 +1,241 @@
+import React from "react";
```
johngruen, line 1: This entire file is basically just extracted button types from the `ExperimentControls` Page

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3669/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/41dfe96057ecf1d377666f107ce1f5eaca8098dc/

frontend/src/app/containers/ExperimentPage/ExperimentControls.js
```diff
@@ -1,76 +1,63 @@
...
+    document.querySelectorAll(".measurements").forEach(el => {
+      if (el.offsetTop) {
+        window.scrollTo(0, el.offsetTop);
+      }
+    });
     flashMeasurementPanel();
   };
 
-  const buttons = platforms.map(platform => createButton({
+  const buttons = createButtons({
```
johngruen, line 57: No longer creating an array here, all the return possibility are generated in a switch statement in the `createButtons` function

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3669/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/5e9754ee788c1221ea96dddf19562c69ea964e6f/

johngruen:
One difference in this PR is that I removed the os store icons from the experiment controls section unless the page is viewed on mobile. on iOS in particular, this link is a dead end and i don't want people hitting it and wandering off.

MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3669/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/3b4382f83fbe16d9ad6e3ad36ebea34922baba45/

content-src/experiments/dev-example.yaml
```diff
@@ -2,7 +2,8 @@ dev: true
 id: 999
 title: 'Dev Example'
 slug: dev-example
-platforms: ['addon']
+platforms: ['ios', 'web']
```
lmorchard, line 5: FWIW, we should probably create new Dev Experiment YAML files rather than keep tweaking this one, so we can cover multiple scenarios. I think our integration tests rely on certain bits of content in this original `dev-example.yaml`, so we could end up breaking those if we don't update / add tests to them.

frontend/src/app/containers/ExperimentPage/ExperimentButtons.js
lmorchard, line 1: These buttons should probably all have Storybook stories, but #3625 breaks most of the stories that involve these buttons right now. 😞

lmorchard (commented):
Already merged while I was looking, just had a couple nits / comments but nothing to block merging

content-src/experiments/dev-example.yaml
johngruen: Good call

PR #3688, opened 2018-07-03
match addon versions

meandavejustice:


SoftVision-PaulOiegas:
Dev and Stage add-ons versions match (3.0.4vd975220).

Issue #3571, opened 2018-05-24
Remove traces of restartRequired


fzzzy:
restartRequired hasn't been used in a very long time. It can be removed.

Vimal-Raghubir:
I am interested in working on this issue. Just to be clear the `restart.isRequired` portion is still required but the `restartRequired` isn't correct?

fzzzy:
@Vimal-Raghubir I'm pretty sure everything having to do with restarting is no longer needed.

PR #3691, opened 2018-07-03
Remove traces of restartRequired

Vimal-Raghubir:
As indicated in the commit messages, these portions of code were removed since restartRequired is not needed anymore. The issue that indicated this required fix is #3571.


frontend/src/app/containers/App/index.js
```diff
@@ -63,7 +59,6 @@ import { localizationsSelector, negotiatedLanguagesSelector } from "../../select
 import { chooseTests } from "../../actions/varianttests";
 import addonActions from "../../actions/addon";
 import newsletterFormActions from "../../actions/newsletter-form";
-import RestartPage from "../RestartPage";
```
lmorchard, line 66: Since the `RestartPage` container is no longer used, it should be removed from the project too

fzzzy:
I'll take a look at this tomorrow.

fzzzy:
@Vimal-Raghubir Thanks, this looks good, but there are conflicts now. Can you please pull from master into your branch, fix the conflicts, and push to your branch? You could also rebase the branch if you know how to do that and prefer that.

fzzzy (changes requested):
Conflicts need resolving.

PR #3700, opened 2018-07-05
Update Lockbox launch time to 6 AM Pacific

devinreams:
Per @sandysage @marniepw @johngruen this will go live at 6 AM Pacific, 7 AM Mountain, 1 PM GMT/UTC.

PR #3701, opened 2018-07-05
Merge latest l10n strings to master for latest release

devinreams:
This PR will merge the localized strings from Pontoon down into `master` branch (and this PR will stay up-to-date as new commits are added).

@meandavejustice assuming you're doing the release tag, you can just hit the big green merge button here right before the tag and everything should come over clean and easy without fussing with git on the command line, if you prefer.

(I would do this myself but may be away from keyboard during PTO so wanted to set this up so it's not forgotten and as easy as possible)


content-src/experiments/lockbox.yaml
```diff
@@ -23,8 +23,8 @@ thumbnail: /static/images/experiments/lockbox/icon/thumbnail.png
 gradient_start: '#00BCD8'
 gradient_stop: '#0086DC'
 
-created: '2018-07-10T14:00:00.000000Z' # 8 AM Mountain, 7 Pacific (3 PM GMT) on July 10th
-launch_date: '2018-07-10T14:00:00.000000Z'
+created: '2018-07-10T13:00:00.000000Z' # 6 AM Pacific, 1 PM GMT on July 10th
```
lmorchard, line 26: This isn't an l10n string - should probably go in a separate PR

meandavejustice: https://github.com/mozilla/testpilot/pull/3700#issuecomment-402774702

PR #3707, opened 2018-07-05
Add Lockbox availability disclaimer

devinreams:
Fixes mozilla-lockbox/lockbox-ios#532

Sorry for the last-minute change @flodolo. Product / Marketing asked we add this.


locales/en-US/experiments.ftl
```diff
@@ -61,7 +61,7 @@ containersContributors7Title = Firefox UX
...
+firefoxlockboxIntroduction = <p>Take your passwords everywhere with Firefox Lockbox. This Test Pilot experiment gives you access to all the logins you've saved to Firefox, in a secure app on your iPhone. Just log in to Lockbox with your Firefox Account, and your saved usernames and passwords will sync to the app using 256-bit encryption. So you’ll have easy access to online accounts, even on the go.</p> <p><strong>Firefox Lockbox requires a Firefox Account in order to access your previously <a href="chrome://passwordmgr/content/passwordManager.xul">saved logins</a>. If you do not have an account set up, get started <a href="https://lockbox.firefox.com/faq.html#how-do-i-get-my-firefox-saved-logins-into-firefox-lockbox">here</a>.</strong></p> <p><em>Want to be notified when our Android app launches? Sign up <a href="https://goo.gl/forms/ZwLIfHSGLrYcM6k83">here</a>.</em></p> <p>Please note, Firefox Lockbox is currently an English only app available in a few regions including Australia, New Zealand, United Kingdom, United States, and Canada.</p>
```
flodolo, line 64: Two issues here:

- You're adding text to an existing string, without changing the ID. Which means existing translations will remain the same, and nobody will know about the new text. You need to use a new ID.
- Once you add the new ID, we're losing the entire translation. That's one of the reasons I keep pushing for being able to split these strings up :-\

pdehaan, line 64: Sorry, probably unrelated to these changes, but... Should these `<a>` tags have an `href=""` attribute?
I notice we don't use the same pattern elsewhere in this file, such as:

```
 firefoxlockboxLegalnotice = By proceeding, you agree to the <a>terms</a> and <a>privacy</a> policies of Test Pilot and the <a>Firefox Lockbox privacy policy</a>.
```

flodolo: If, on the other hand, the goal is to add this text only to English, I guess it makes sense. Note that new translation will include this text, existing ones won't know it exists…

flodolo: The href is not needed, e.g.

```
notesLegalnotice = By proceeding, you agree to the <a>terms</a> and <a>privacy</a> policies of Test Pilot and the <a>Notes privacy policy</a>.
```

devinreams: @pdehaan Yeah, it confused me at first but seems the links are automagically populated on the front end. 😁

devinreams:
>You're adding text to an existing string, without changing the ID. Which means existing translations will remain the same, and nobody will know about the new text. You need to use a new ID.

Yup, but this also means we avoid your concern that only English strings will appear on production, right? 😔 

I don't actually know if its worth translating this text anyway since its a temporary disclaimer and will likely be removed as we open the experiment up to more locales.

>Once you add the new ID, we're losing the entire translation. That's one of the reasons I keep pushing for being able to split these strings up :-\

Understood, I don't think changing the ID is possible as its generated by the experiment YAML.

Also, splitting out into "multiple-line introductions" is also likely not something the team is able to tackle in the short term but may be a good consideration for a new issue in the "Stretch" backlog.

The other Test Pilot devs and @johngruen will have to correct me if wrong on either point.

devinreams:
>Note that new translation will include this text, existing ones won't know it exists…

Yeah, I couldn't think of anywhere better to add this for now, though.

This is all the more reason for us to launch, extract the strings from the app, and get it onto Pontoon ASAP! 😁 

- https://github.com/mozilla-lockbox/lockbox-ios/issues/440
- https://github.com/mozilla-lockbox/lockbox-ios/issues/207

johngruen:
Okay i am going to merge this guy.

locales/en-US/experiments.ftl
lmorchard: Yeah, if we revise an existing string, [we need to use the `<fieldname>_l10nsuffix` property in YAML to generate a new ID](https://github.com/mozilla/testpilot/blob/master/docs/content/reference.md#localization). That also invalidates the whole string for existing localizations too - because the ID changes - so we lose whatever we might have just merged in.

PR #3729, opened 2018-07-12
make txp default slug for mozAddonManager

johngruen:
It turns out that this is like the single most important GA event for detecting experiment installs...

if there's no slug, when this gets called, it's b/c the test pilot add-on is getting installed. We should pass that in as the default value here.

### DIFF
--- a/frontend/src/app/lib/InstallManager.js
+++ b/frontend/src/app/lib/InstallManager.js
@@ -19,7 +19,7 @@ function getAddonId() {
   }[window.location.host];
 }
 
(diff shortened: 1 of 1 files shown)


MozillaTestPilotStorybook:
# Storybook Deployment
Pull Request: http://testpilot-storybook.dev.mozaws.net/pull/3729/index.html
Commit: http://testpilot-storybook.dev.mozaws.net/335d5f339f19fabcbb7fb0ce7927e508d509ba71/

PR #3756, opened 2018-07-30
add some dimensions to sendToGa call

### LINKED ISSUE #3735: Add dimensions to mozAddonManagerEvent, SMS Events, Featured Experiment Events and Experiment Buttons
Needs to include: 
`dimension1`
`dimension2`
`dimension3`

meandavejustice:
fixes #3735

### DIFF
--- a/frontend/src/app/containers/App/index.js
+++ b/frontend/src/app/containers/App/index.js
@@ -310,9 +310,15 @@ function sendToGA(type, dataIn, evt = null) {
     }
   };
   if (window.ga && ga.loaded) {
+    let installed = {};
     const chosenTest = getChosenTest();
     data.hitType = type;
     data.hitCallback = hitCallback;
+    if (typeof window !== "undefined") {
+      data.dimension1 = (window.localStorage.getItem("hasAddon") === "true");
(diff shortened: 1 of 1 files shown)


frontend/src/app/containers/App/index.js
```diff
@@ -310,9 +310,13 @@ function sendToGA(type, dataIn, evt = null) {
     }
   };
   if (window.ga && ga.loaded) {
+    let installed = {};
     const chosenTest = getChosenTest();
     data.hitType = type;
     data.hitCallback = hitCallback;
+    data.dimension1 = (window.localStorage.getItem("hasAddon") === "true");
+    data.dimension2 = installed = JSON.parse(window.localStorage.installations);
```
fzzzy: I feel like the JSON.parse bit here should be moved up into the let installed line.

fzzzy (changes requested):
One small change request.

frontend/src/app/containers/App/index.js
meandavejustice: oh yeah, that was a remnant of some `window` variable checking I was doing but realized wasn't necessary, pushing a fix in 3, 2, 1...

fzzzy (approved):

PR #3774, opened 2018-08-07
Fixes location of Advance's metrics.md.

chuckharmston:
CC @SoftVision-CosminMuntean


chuckharmston:
Will self-merge when the build passes.

PR #3779, opened 2018-08-07
Fix #3777: Trap error from JSON.parse() on undefined value from localStorage

### LINKED ISSUE #3777: Error in sendToGA with JSON.parse on an undefined value
Seeing this on the dev site right now, goes away when I disable uBlock Origin. Others do not report the same thing happening, but [it does look like sendToGA will result in an error if localstorage produces an undefined value](https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/App/index.js#L304).

capture

lmorchard:
This seems to fix the immediate error for me

### DIFF
--- a/frontend/src/app/containers/App/index.js
+++ b/frontend/src/app/containers/App/index.js
@@ -301,7 +301,12 @@ function sendToGA(type, dataIn, evt = null) {
     }
   };
   if (window.ga && ga.loaded) {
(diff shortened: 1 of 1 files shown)


lmorchard:
Spent awhile tracking down what looks like additional weirdness with uBlock Origin, but couldn't come up with anything conclusive. Meanwhile, this seems to fix the JSON.parse() error and gets the site working again for me.

PR #3780, opened 2018-08-07
Enable polling in webpack file watchers

lmorchard:
So I noticed on my Windows machine that webpack wasn't rebuilding the site on file edits. I think that's because my editor locks files and webpack file watchers don't catch changes. 

This seems to fix it by changing to polling for changes rather than watching file system events. Appears to work without too much burden on my machine. I *think* this should work fine on OS X and Linux, but worth a check.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -102,6 +102,9 @@ module.exports = {
     // filename: "[name].[hash].js"
     filename: "[name].js"
   },
+  watchOptions: {
+    poll: 1000
+  },
   node: {
     crypto: false
   },


meandavejustice:
Looks good to me, going to wait until someone on mac or linux can take a look

devinreams (approved):
I made a change and webpack re-compiled automagically on macOS 10.14 ✨ 

testpilot_ _node_ _npm_rvm_bin_path__users_dreams__rvm_bin_term_program_apple_terminal_gem_home__users_dreams__rvm_gems_ruby-2_3_4_ _120x60

PR #3783, opened 2018-08-07
Fix #3631: Use withLocalization HOC rather than <Localized> for newsletter email input placeholder

### LINKED ISSUE #3631: fluent react error with email input wrapped in <Localized>in MobileDialog

lmorchard:
This is an ugly hack, but it fixes the error and still results in a localized input field placeholder.

It would be less ugly if I could figure out how to lookup a string attribute from `getString()`, but that seems not possible with the current `withLocalization` HOC

### DIFF
--- a/frontend/src/app/containers/App/index.js
+++ b/frontend/src/app/containers/App/index.js
@@ -226,6 +226,9 @@ class App extends Component {
         if (typeof localizations[lang] === "string") {
           const cx = new MessageContext(lang, {useIsolating: false});
           cx.addMessages(localizations[lang]);
+
+          // HACK: #3631 - hoist the attribute for this string up into its own separate string
+          cx.addMessages("newsletterFormEmailPlaceholderHackedString = {newsletterFormEmailPlaceholder.placeholder}");
           yield cx;
         }
       }
--- a/frontend/src/app/components/NewsletterForm/index.js
+++ b/frontend/src/app/components/NewsletterForm/index.js
@@ -1,7 +1,8 @@
 // @flow
 
 import classnames from "classnames";
-import { Localized } from "fluent-react/compat";
+// $FlowFixMe: Flow says withLocation isn't a named export, but we can still import it
+import { Localized, withLocalization } from "fluent-react/compat";
 import React from "react";
 
 import LocalizedHtml from "../LocalizedHtml";
@@ -18,6 +19,25 @@ type NewsletterFormProps = {
   setEmail?: Function,
   buttonRef?: Function
 }
+
+// HACK: #3631 - avoid an error with <Localized> by using withLocalization and
+// a hacked string from generateMessages in App container
+const EmailField = withLocalization(({ value, onChange, getString }) => (
+  <input
+    ref={el => {
+      if (el) {
+        el.setAttribute(
+          "placeholder",
+          getString("newsletterFormEmailPlaceholderHackedString")
+        );
+      }
+    }}
+    {...{ value, onChange }}
+    type='email'
+    placeholder="Your email here"
+    required
+  />
+));
 
 export default class NewsletterForm extends React.Component {
   props: NewsletterFormProps
@@ -46,15 +66,7 @@ export default class NewsletterForm extends React.Component {
 
   renderEmailField() {
     return (
-      <Localized id="newsletterFormEmailPlaceholder" attrs={{placeholder: true}}>
-        <input
-          type='email'
-          placeholder="Your email here"
-          required
-          value={this.props.email}
-          onChange={this.handleEmailChange}
-        />
-      </Localized>
+      <EmailField value={this.props.email} onChange={this.handleEmailChange} />
     );
   }
 
--- a/frontend/src/app/components/NewsletterForm/tests.js
+++ b/frontend/src/app/components/NewsletterForm/tests.js
@@ -31,7 +31,7 @@ describe("app/components/NewsletterForm", () => {
       expect(subject).to.have.length(1);
     });
 
-    it("should take its value from the store", () => {
+    it("should take its value from the prop", () => {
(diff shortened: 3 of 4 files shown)


lmorchard:
@stasm FWIW, would you happen to know of a non-ugly way to get access to an attribute of a string from `getString`? [This one, to be specific?](https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl#L226) Doing a slight hack here to work around a weird error we're seeing with `<Localized>`

stasm:
That's not currently supported by `getString` because my assumption was that `getString` would be used only with translations intended specifically for it. Hence, these translations would not need to have attributes. Compound translations, i.e. those with attributes, are usually intended to localize some kind of objects, elements or components—declaratively via `<Localized>`.

Do you think there are other use-cases for something like `getString("message.attribute")`? Being able to work around bugs in `<Localized>` counts as one :)

I should also fix the root cause here. I'll make looking into #3631 my priority this week. I'd like Test Pilot to upgrade to `fluent` 0.7 and I'll be happy to help to make it happen.

frontend/src/app/containers/App/index.js
```diff
@@ -226,6 +226,12 @@ class App extends Component {
         if (typeof localizations[lang] === "string") {
           const cx = new MessageContext(lang, {useIsolating: false});
           cx.addMessages(localizations[lang]);
+
+          // HACK: #3631 - host the attribute for this string up into its own separate string
+          const placeholderHackString = cx._messages
+            .get("newsletterFormEmailPlaceholder").attrs.placeholder;
+          cx.addMessages(`newsletterFormEmailPlaceholderHackedString = ${placeholderHackString}`);
```
stasm: Accessing `_messages` is risky; the internal API is subject to change. Instead, you could construct a new message which directly references the attribute (on the Fluent side). Note that that's a regular string, not a template literal, and the `{}` are Fluent syntax.

```js
cx.addMessages("newsletterFormEmailPlaceholderHackedString = {newsletterFormEmailPlaceholder.placeholder}");
```

In order to make it less hacky still, it might be a good idea to add that new message to `app.ftl` to avoid having to construct it programmatically on the runtime. Maybe in a separate file which isn't exposed to localizers?

```properties
newsletterFormEmailPlaceholderHackedString = {newsletterFormEmailPlaceholder.placeholder}
```

@flodolo, what's your take on adding such a message to Test Pilot's FTL?

lmorchard: FWIW, I'd like to make this slightly less hacky, but not so much less hacky that we're comfortable with forgetting it's there and leaving things this way. Ultimately, this hack should be stable until it can be replaced with something cleaner after a fluent upgrade

meandavejustice: I wonder if it makes sense to just wait on this til the new fluent release mentioned in https://github.com/mozilla/testpilot/issues/3631#issuecomment-411419408

lmorchard:
The failure in CircleCI is weird, too... Running lint & flow locally, I see almost none of the warnings or errors that show up in the test run:
```
C:\Users\lmorc\devel\testpilot [master ↑1]> npm run lint

> testpilot@1.0.0 lint C:\Users\lmorc\devel\testpilot
> npm-run-all lint:*


> testpilot@1.0.0 lint:styles C:\Users\lmorc\devel\testpilot
> sass-lint --verbose 'frontend/src/styles/**/*.scss' 'frontend/src/app/**/*.scss'


> testpilot@1.0.0 lint:js C:\Users\lmorc\devel\testpilot
> eslint ./gulpfile.babel.js ./webpack.config.js './frontend/{src,tests}/**/*.{js,jsx}'


C:\Users\lmorc\devel\testpilot\gulpfile.babel.js
  2:1  warning  This module could be parsed as a valid script  import/unambiguous

C:\Users\lmorc\devel\testpilot\webpack.config.js
  3:1  warning  This module could be parsed as a valid script  import/unambiguous

✖ 2 problems (0 errors, 2 warnings)

C:\Users\lmorc\devel\testpilot [master ↑1]> npm run flow

> testpilot@1.0.0 flow C:\Users\lmorc\devel\testpilot
> flow frontend

No errors!
C:\Users\lmorc\devel\testpilot [master ↑1]>
```

frontend/src/app/containers/App/index.js
lmorchard: Well, the reason I threw this together is that this error breaks the site for me right now. Without this fix, I can't actually view the site on my machine. I had to have this in the background - or delete the Localized component from NewsletterForm - while getting through to diagnosing other errors yesterday

stasm: > FWIW, I'd like to make this slightly less hacky, but not so much less hacky that we're comfortable with forgetting it's there and leaving things this way.

Makes sense to me. In this case I recommend the slightly less-ugly approach of

```js
cx.addMessages("newsletterFormEmailPlaceholderHackedString = {newsletterFormEmailPlaceholder.placeholder}");
```

:)

lmorchard:
Couldn't reproduce the flow error locally. Did a "rerun with ssh" and managed to "fix" the flow error by disabling for that one line. For some reason it failed on the `withLocalization` import - even though it looks like that import works fine. The "real" fix, I guess, would be to write a separate Flow description of Fluent - and I didn't really want to get into that

lmorchard:
<strike>Aaaand now it's failing a unit test because the email field doesn't render unless wrapped in an l10n context, which the test doesn't do 😞</strike>

Edit: Nope, that was a legit test failure actually - I named a prop `value` in the new component where other parts named it `email`. Hopefully this fixes it...

lmorchard:
One more tweak because I accidentally a console.log statement

meandavejustice:
closing for #3835

Issue #2577, opened 2017-06-30
Experiment specific news feed updates should appear on experiment page?


pdehaan:
I faked a couple news updates for min-vid and noticed that they only seem to appear on the /experiments page, and not on the /experiments/min-vid page like you'd expect.

<img width="1440" alt="firefox_test_pilot" src="https://user-images.githubusercontent.com/557895/27756507-02c1224e-5dad-11e7-8c38-b9f36c18ee87.png">

---

But I don't see updates when I go to /experiments/min-vid:

<img width="1440" alt="firefox_test_pilot_-_min_vid" src="https://user-images.githubusercontent.com/557895/27756562-82c67ce6-5dad-11e7-82e0-466daf09a7c5.png">

<img width="1440" alt="firefox_test_pilot_-_min_vid" src="https://user-images.githubusercontent.com/557895/27756568-8f6e9fa0-5dad-11e7-9a33-08455da9564c.png">

---

Where I added this to the content-src/experiments/min-vid.yaml file

```yaml
# ...
news_updates:
  -
    slug: min-vid-update-1
    title: My Update title
    link: https://blog.mozilla.org/testpilot/example
    created: 2017-06-01T12:00:00Z
    published: 2017-06-02T12:00:00Z
    dev: false
    content: >
      Example content goes here
  -
    slug: min-vid-update-2
    title: My Update title dos
    link: https://blog.mozilla.org/testpilot/example2
    created: 2017-06-01T12:00:00Z
    published: 2017-06-03T12:00:00Z
    dev: true
    content: >
      Example content goes here, homeskillet.
```

fzzzy:
Seems like a good idea.

ghost:
This is r+ by Product/UX.

chuckharmston:
This is a good idea, but has never been designed and won't be implemented in H1. Leaving it open for posterity.

PR #3785, opened 2018-08-07
news_updates are no longer displayed on the site

devinreams:
I may have missed something entirely, but as I travel around the code and issue history, we decided not to output news updates on the experiments themselves: #2577 

And since the `/experiments` landing page itself is gone there's nowhere that news updates appear, right?

So I'm thinking we shouldn't advertise them in the reference docs anymore, perhaps nor in the template doc. What do ya'll think?

### DIFF
--- a/docs/content/reference.md
+++ b/docs/content/reference.md
@@ -194,27 +194,6 @@ introduction: >
 
 ![Example use of introduction](img/introduction.png)
 
-## `news_updates`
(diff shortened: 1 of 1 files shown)


meandavejustice:
@devinreams We still show them in the news updates dialog

devinreams:
>We still show them in the news updates dialog

@meandavejustice Ah ok. Pardon my ignorance, how do I get to that dialog?

PR #3805, opened 2018-08-20
fix image loading issue

meandavejustice:
- seemed to be an issue with publicPath option for file-loader
- also removed carats from package.json for more consistent dep
versions


lmorchard:
When trying this PR, images are broken for me. The static path is duplicated:

capture

So, we can go ahead and merge this, but I'm still concerned: Since this seems to break / not break in different ways on different machines - which way will it break / not-break on stage & prod?

meandavejustice:
@lmorchard Weird, we should def not merge until we can get this to act the same on our machines.

lmorchard (commented):
I guess just to add my review: 🤷‍♂️ Breaks on my machines, not others - opposite situation before the patch. (And my machines include an Ubuntu 18.04 box in EC2, so I've at least worked out it's not literally my laptop)

Issue #3771, opened 2018-08-03
review dependencies


clouserw:
We have a lot of them!  https://github.com/mozilla/testpilot/network/dependencies

pdehaan:
Here's the always mildly-suspicious output from **depcheck**:

```sh
➜  testpilot git:(master) $ git rev-parse --short HEAD # a8682f5b

➜  testpilot git:(master) $ npx depcheck@latest

Unused dependencies
* domready
* legal-docs

Unused devDependencies
* autoprefixer
* babel-eslint
* babel-register
* cross-spawn
* cssnano
* empty
* eslint-config-react
* eslint-plugin-jsx-a11y
* flow-typed
* gulp-cache
* gulp-if
* gulp-multi-dest
* gulp-rename
* gulp-uglify
* merge-stream
* mocha-junit-reporter
* node-normalize-scss
* photon-colors
* react-addons-test-utils
* react-test-renderer
* require-globify
* url-loader
* vinyl-source-stream

Missing dependencies
* eslint-config-plugin:import
* eslint-config-plugin:flowtype
* eslint-config-plugin:react
* eslint-config-plugin:mozilla
* @storybook/addon-links
* globby
* l20n
* redux-actions
* isomorphic-fetch
* react-router-redux
* history
* redux-logger
* redux-thunk
* fbjs
* object-assign
* _process
* select
* lodash.isplainobject
* delegate
* warning
* invariant
* deep-equal
* query-string
* lodash._basefor
* lodash.isarguments
* lodash.keysin
* lodash.isarray
* strict-uri-encode
* json-stringify-safe
* lodash
* hoist-non-react-statics
* reduce-reducers
* deep-diff
* symbol-observable
* tiny-emitter
* good-listener
* flux-standard-action
```

meandavejustice:
Here is output for `npm outdated`:

```
meandave@LAPTOP-NA0Q0G86:/mnt/c/Users/Dave Justice/Code/testpilot$ npm outdated
Package                             Current         Wanted   Latest  Location
@storybook/addon-actions      3.4.0-alpha.1  3.4.0-alpha.1   3.4.10  testpilot
@storybook/addon-knobs        3.4.0-alpha.1  3.4.0-alpha.1   3.4.10  testpilot
@storybook/react              3.4.0-alpha.1  3.4.0-alpha.1   3.4.10  testpilot
autoprefixer                          8.1.0          8.6.5    9.1.0  testpilot
[...]

PR #3818, opened 2018-08-23
Upgrade to fluent 0.8 and fluent-react 0.8.

### LINKED ISSUE #3789: Update to a newer version of Fluent
The Test Pilot website uses two Fluent dependencies: the core `fluent` package as well as `fluent-react`, the bindings for React. In #3631 I mentioned that I'm working on a fix to `fluent-react`.

Independently of `fluent-react`, I'd like Test Pilot to update to a newer version of the core `fluent` package. You're currently o n 0.4.3 and the latest is [0.7.0](https://github.com/projectfluent/fluent.js/blob/master/fluent/CHANGELOG.md#fluent-070-july-24-2018).

I tried updating it in my local setup, and unfortunately I ran into a build error. One of the changes 0.7.0 shipped with is an updated compatibility matrix for the `fluent/compat` builds. And while this new matrix is aligned with your browser support documented in the [FAQ](https://github.com/mozilla/testpilot/blob/master/docs/faq.md#what-browsers-are-supported-by-test-pilot), it turns out it's not compatible with old versions of Webpack. More specifically, UglifyJS breaks on non-ES5 syntax 🤦‍♂️  (e.g. `class` or `async`).

I'm filing this issue to let you know of my intent to update `fluent` somehow and to track my progress on dealing with the build error. And also to ask if you have any plans to upgrade to a newer version of Webpack and Uglify? :)

stasm:
I've started working on upgrading the `fluent` dependencies to their most recent versions. Fixes #3789. Below is the summary of changes made in this PR.

#### [`fluent 0.8.0`](https://github.com/projectfluent/fluent.js/blob/master/fluent/CHANGELOG.md#fluent-080-august-20-2018)

- [x] Rename `MessageContext` to `FluentBundle`.

#### [`fluent-react 0.6.0`](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#fluent-react-060-february-1-2018)

This version introduced [React Overlays](https://github.com/projectfluent/fluent.js/wiki/React-Overlays) which allow translations to use some limited HTML markup.

- [x] Don't pass elements as `$props` to `<Localized>`. Instead, pass them as regular props. See [details](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#dont-pass-elements-as-arguments).
- [x] Add `attrs` to `Localized` if any attributes should be localized. See [details](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#add-attrs-to-localized). I didn't have to change anything because it looks like all impacted callsites already use `attrs`, which is great.
- [x] Don't use `LocalizedHtml` where a regular `Localized` would do. I was able to remove all but one uses of `LocalizedHtml`.

The one remaining use of `LocalizedHtml` is for experiments' introduction texts. They use arbitrary amount of markup which means that a single generic `Localized` cannot know which elements to pass as props.

Because all the other use-cases for using `LocalizedHtml` are gone, I was able to re-write it completely using the `withLocalization` HOC and tailor just to this one remaining use-case.

#### [`fluent-react 0.7.0`](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#fluent-react-070-may-18-2018)

- [x] Use regular `Localized` also when HTML entities are present in the translation.

#### [`fluent-react 0.8.0`](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#fluent-react-080-august-21-2018)

- [x] Rename the `messages` prop of `LocalizationProvider` to `bundles`.
- [x] Fix the build time error related to `document` being undefined.

#### To Do:

- [x] Rename affected message identifiers. In case of messages generated from yaml, do I use `_l10nsuffix`?
- [ ] Test SSR. So far I ran `npm run build:gulp-dist` and it completed successfully. Is there more that I should test?
- [x] Use Fluent's `DATETIME` to format dates; remove the `formatDate` utility.


frontend/src/app/components/FeaturedExperiment/FeaturedButton.js
```diff
@@ -78,14 +77,17 @@ export default class FeaturedButton extends Component<FeaturedButtonProps, Featu
...
-        <a href="#" onClick={launchLegalModal}></a>.
-      </p>
-    </LocalizedHtml>);
+    return (<Localized id={this.l10nId("legal-notice")}
+      $title={title}
+      terms-link={<a href="/terms" onClick={(ev) => this.sendMetric(ev, {eventLabel: "Open general terms"})}></a>}
+      privacy-link={<a href="/privacy" onClick={(ev) => this.sendMetric(ev, {eventLabel: "Open general privacy"})}></a>}
+      modal-link={<a href="#" onClick={launchLegalModal}></a>}>
+        <p className="main-install__legal">
+            By proceeding, you agree to the <terms-link>terms</terms-link>{" "}
```
stasm: I'm not quite sure if using custom elements like that is a good practice in React. The contents of `<p>` will be replaced anyways by the translation, but before it happens, this fragment of the render tree will be constructed by React, I believe.

Perhaps it would be better to use a regular string?

```jsx
        <p className="main-install__legal">
            {`By proceeding, you agree to the <terms-link>terms</terms-link>
            and <privacy-link>privacy</privacy-link> policies of Test Pilot and
            <modal-link>this experiment's privacy policy</modal-link>.`}
        </p>
```

In the future, `fluent-react` might be able to extract en-US string from such elements. Right now, it would be just easier on the CPU.

frontend/src/app/components/RetireConfirmationDialog/index.js
```diff
@@ -37,9 +35,9 @@ export class RetireConfirmationDialog extends Component<RetireConfirmationDialog
               <Localized id="retireMessageUpdate">
                 <p>As you wish. This will uninstall Test Pilot. You can disable individual experiments from the Firefox Add-ons Manager.</p>
               </Localized>
-              <LocalizedHtml id="retireEmailMessage">
+              <Localized id="retireEmailMessage" em={<em></em>}>
```
stasm, line 38: In future version of `fluent-react`, translations will be able to use text-level elements like `<em>` and `<br>` without requiring the code to pass them as props explicitly. See https://github.com/projectfluent/fluent.js/issues/186. Once this lands, I'll file another PR to remove all these `em={<em></em>}` et al.

frontend/src/app/containers/ExperimentPage/DetailsDescription.js
```diff
@@ -67,7 +67,7 @@ export default function DetailsDescription({
               </div>}
             <LocalizedHtml id={l10nId("introduction")}>
               <div>
-                {parser(introduction)}
+                {introduction}
```
stasm, line 70: IIUC, parsing the introduction text here uses up CPU cycles just for the localization to take over milliseconds later. This content will only be visible if the `introduction` id is missing from all locales, including English. I think this is unlikely enough to recognize this as a rare error scenario in which displaying unparsed HTML is acceptable.

stasm:
The CI error is caused by UglifyJS not parsing the `class` syntax. According to my investigation from https://github.com/mozilla/testpilot/issues/3771#issuecomment-413834142, `uglify-es` used by `uglifyjs-webpack-plugin` should parse this file fine. So I guess there's something which runs some other minification step using `uglify-js`. Looking at `package.json` I suspect `gulp-uglify`. Can you help me track down where this minification happens?

stasm:
> Looking at package.json I suspect gulp-uglify. Can you help me track down where this minification happens?

Correction; I must have been looking at an old dependency graph. `gulp-uglify` is no more! (60bdfd28868e13eead015a3e6fd79c639feb73ce). Which leaves me puzzled about the CI error even more :)

meandavejustice:
@stasm you should try pushing again now that @lmorchard has updated our webpack config to work with version 4

frontend/src/app/containers/ExperimentPage/DetailsDescription.js
```diff
@@ -85,11 +85,11 @@ export default function DetailsDescription({
                     </strong>
                   </Localized>}
                 {copy &&
-                  <Localized id={l10nId(["details", idx, "copy"])}>
+                  <LocalizedHtml id={l10nId(["details", idx, "copy"])}>
```
stasm, line 88: None of the current experiments use markup for the details copy. It would be safer to use a regular `<Localized>` here.

frontend/src/app/containers/ExperimentPage/DetailsDescription.js
```diff
@@ -129,20 +129,24 @@ export const LocaleWarning = ({
...
   return (
     <Warning
       titleL10nId="eolIntroMessage"
-      titleL10nArgs={JSON.stringify({ title: experiment.title, completedDate })}
-      title={title}
+      titleL10nArgs={{$title, $completedDate}}
+      title={`${experiment.title} is ending on ${experiment.completed}`}
     >
-      <Localized id={l10nId("eolWarning")}>
+      <LocalizedHtml id={l10nId("eolWarning")}>
```
stasm, line 145: None of the current experiments use markup for the EOL warning. It would be safer to use a regular `<Localized>` here.

frontend/src/app/components/FeaturedExperiment/FeaturedButton.js
```diff
@@ -78,14 +77,17 @@ export default class FeaturedButton extends Component<FeaturedButtonProps, Featu
       this.sendMetric(ev, {eventLabel: "Popup Featured privacy"});
     };
 
-    return (<LocalizedHtml id={this.l10nId("legal-notice")}
-      $title={title}>
+    return (<Localized id={this.l10nId("legal_notice")}
```
stasm, line 80: I had to change the id here from `legal-notice` to `legal_notice`. Even if there's some normalization going on in the `l10nId` util, the code which looks up a possible `_l10nsuffix` uses the id verbatim, which resulted in searching for `legal-notice_l10nsuffix`. It might be a good idea to fix the `experimentL10nId` util in a future PR.

stasm:
The CI error is related to Flow and my use of the `FluentDateTime` constructor from `fluent/compat`. I've fiddled with the `flow-typed` directory and with `frontend/.flowconfig` but without much success.

stasm (commented):
@flodolo The FTL files are ready for your review.

frontend/src/app/containers/ExperimentPage/DetailsOverview.js
```diff
@@ -84,24 +84,34 @@ export default function DetailsOverview({
...
-          Experiment End Date: <b>{completedDate}</b>
+          Experiment End Date: <b>{completed}</b>
         </span>
-      </LocalizedHtml>
+      </Localized>
     );
   }
 
-  const startedDate = formatDate(created);
+  const startedDate = new FluentDateTime(created, {
```
stasm, line 103: @flodolo Now that `startedDate` is a `FluentDateTime` object, do you think we should use the `DATETIME` builtin explicitly in the FTL? Like so:

```diff
-startedDateLabel = Experiment start date: <b>{$startedDate}</b>
+startedDateLabel = Experiment start date: <b>{DATETIME($startedDate)}</b>
```

I don't even think we need to rev the id for this, but I'll leave the decision up to you.

locales/en-US/app.ftl
```diff
@@ -36,7 +36,7 @@ headerLinkBlog = Blog
 landingIntroOne = Test new features.
 landingIntroTwo = Give your feedback.
 landingIntroThree = Help build Firefox.
-landingLegalNotice = By proceeding, you agree to the <a>Terms of Use</a> and <a>Privacy Notice</a> of Test Pilot.
+landingLegalNotice = By proceeding, you agree to the <terms-link>Terms of Use</terms-link> and <privacy-link>Privacy Notice</privacy-link> of Test Pilot.
```
flodolo: This one still needs a new ID.

locales/en-US/app.ftl
```diff
@@ -98,7 +98,7 @@ mobileDialogSuccessSecondary = Check your device for the email.
 mobileDialogAnotherDeviceLink = Send to another device
 mobileDialogError = Enter a valid email:
 mobileDialogErrorSMS = Enter a valid phone number or email:
-mobileDialogNoticeSMS = SMS service available in select countries only. SMS & data rates may apply. The intended recipient of the email or SMS must have consented. {$learnMore}
+mobileDialogNoticeSMS = SMS service available in select countries only. SMS & data rates may apply. The intended recipient of the email or SMS must have consented. <a>Learn more</a>.
```
flodolo: Same here: we need a new ID, and not much we can do to avoid losing content.

frontend/src/app/containers/ExperimentPage/DetailsOverview.js
flodolo: If we do, we either need to rev the ID, or change the existing translation via script.

flodolo (changes requested):
The number of changed strings grew significantly from the first time I checked the PR.

I wonder if we should try to write a script to udpate existing translations, when only the markup changes. It should be relatively straightforward to do for most of them, only one string, `mobileDialogNoticeSMS`, would require more work, and it would avoid losing a lot of translated content.

If we do that, we could even avoid using new string IDs.

Forgot one detail: translation via Pontoon happens in the #l10n branch. We need to make sure branches are in sync before this is merged, and to sync them back if we decide to change files via script.

locales/en-US/app.ftl
stasm: 🤦‍♂️ I think I accidentally didn't commit this. Sorry, of course it needs a new id.

frontend/src/app/containers/ExperimentPage/DetailsOverview.js
stasm: The current value of 

```properties
startedDateLabel = Experiment start date: <b>{$startedDate}</b>
```

will continue to work, because Fluent recognizes the argument as a `FluentDateTime` anyways.

flodolo: That's good, but nobody will realize they can actually customize the format if we don't add `DATETIME()`.

stasm: That's a good point. Regardless of whether we need to rev the id, or use a script, would you rather use the version with `DATETIME()` or without?

flodolo: I think it's good to have DATETIME(), it would be easy to help a locale customize that if needed.

stasm:
> The number of changed strings grew significantly from the first time I checked the PR.

Yeah, that's because the first few version of this PR didn't include changes to yaml files :(

stasm:
@flodolo I updated the strings using `DATETIME` together with their ids. Over to you for deciding if we need a script to update the translations or not. Thanks!

flodolo (approved):
FTL looks good

flodolo:
@stasm 
Thoughts? I don't want to have a full parser and serializer for this…
https://gist.github.com/flodolo/d4f45a6da82f1cbb4d99a077ab7a8228

[Test output](https://github.com/mozilla/testpilot/compare/l10n...flodolo:l10n_test?expand=1). The l10n branch is behind, not having the Advance strings and the updated Lockbox one.

stasm:
@flodolo The scrpit and the test output look good to me, thanks. In particular, I think it makes sense to consider `mobileDialogNoticeSMSWithLink` a new string.

stasm:
Hey @lmorchard, can you help me fix the Flow errors in this PR please?

I'm going to be offline next week and the week following it. What's your advice wrt. to when to land this PR? Assuming I can get a green build today/tomorrow, would you be OK merging it around Wednesday? I'd like to be around for a couple of days after it lands, just in case there are any issues. If we don't merge this by Wednesday, I'd suggest waiting until I'm back from my PTO mid-September.

Since this PR touches strings, I'm also not sure if there's a schedule to follow wrt. the localization process. @flodolo, do you have a recommendation about when to best merge this PR?

flodolo:
I don't have a strong preference, we just need to make sure l10n and master branches are in sync before merging, and sync back after merging this PR and running the script in a different PR. Doing it early this week is probably a good idea.

meandavejustice:
@stasm Just tested this locally.

A solution for now would be to create a file at `flow-typed/fluent.js` with the following contents:
``` js
declare module 'fluent/compat' {
  declare module.exports: any;
}
```
This gets the flow tool passing. It would be nice to have flowtype definitions bundled along with fluent, maybe as a longer term goal.

stasm:
@meandavejustice Thank you! I remember fiddling with the `flow-types` directory, but I didn't realize I could use `any` like you did.

The PR is now green. I've been testing it locally and things look good. In particular, I tested most of the strings which use markup, as well as the strings which display started and completed dates (which are now formatted by Fluent).

This PR is now ready for a review.

meandavejustice (approved):
Works great for me locally, not seeing any issues.

Thanks for taking care of this @stasm !

stasm:
Thanks. I've pushed a small update to `package.json` to use [`fluent-react` 0.8.1](https://github.com/projectfluent/fluent.js/releases/tag/fluent-react%400.8.1) which makes the peer dependency on `fluent` more explicit. Otherwise, the rest of the code is the same.

flodolo:
@meandavejustice 
We'll need to run [this script](https://gist.github.com/flodolo/d4f45a6da82f1cbb4d99a077ab7a8228) on localization.

Ideally:
* Sync l10n and master
* Run the script
* Sync l10n and master again

I can run the script and open a PR, but not sure I can help with the other two points.

PR #3823, opened 2018-08-23
Disable storybook builds for now

lmorchard:
Looks like the Webpack upgrade broke Storybook. Since we're not using it, and it'll probably take an upgrade / reconfiguration of storybook, let's just disable it for now


lmorchard:
Also opened an issue to fix and re-enable https://github.com/mozilla/testpilot/issues/3824

PR #3825, opened 2018-08-23
3792 web ext cross

### LINKED ISSUE #3792: [Featured experiment] Web experiments wrongly display "Download Firefox" button instead of "Go to <experiment_name>" button on 3rd party browsers
**[Notes]:**
- If Send experiment will get featured after the new UI for the website and the Android app will be released, this issue could surface. We should think at a way to manage this.
- If you run the local page on Firefox browser, the button becomes "Enable Send" so both the cases are something we haven't thought at since we never had a web experiment featured.

**[Affected versions]:**
- All 3rd party browsers

**[Affected Platforms]:**
- All platforms

**[Prerequisites]:**
- Test Pilot Master branch was locally cloned and setup was performed to run it locally. 
- Lockbox experiment was marked as not featured (false) in the `lockbox.yaml` file and Send experiment was marked as featured (true) in the `send.yaml` file.

**[Steps to reproduce]:**
1. Open Firefox browser and navigate to https://testpilot.firefox.com page.
2. Observe the big green button content displayed on the featured experiment section.

**[Expected result]:**
- The button shows the "Go to Send" string and allows you to navigate to its website.

**[Actual result]:**
- The button wrongly shows "Download Firefox" for a web experiment that can be used on all browsers and platforms.

**[Aditional Notes]:**
- Here is a screenshot of the issue: 

screen shot 2018-08-14 at 16 41 52

meandavejustice:
fixes #3792 (wrongly displaying firefox download button for web experiements on mobile and cross browser)

### DIFF
--- a/frontend/src/app/components/MainInstallButton/index.js
+++ b/frontend/src/app/components/MainInstallButton/index.js
@@ -101,7 +101,7 @@ export default class MainInstallButton extends Component<MainInstallButtonProps,
           </p>
         </LocalizedHtml>}
 
-        {!showWebButton &&isMinFirefox && !isMobile && experimentLegalLink && experimentLegalLink}
+        {!showWebButton && isMinFirefox && !isMobile && experimentLegalLink && experimentLegalLink}
       </LayoutWrapper>
     );
   }
--- a/frontend/src/app/components/MainInstallButton/index.js
+++ b/frontend/src/app/components/MainInstallButton/index.js
@@ -85,7 +85,7 @@ export default class MainInstallButton extends Component<MainInstallButtonProps,
 
     if (showWebButton) {
       return this.renderWebExperimentButton();
-    } else if (!!experiment && isMinFirefox && !isMobile) {
+    } else if (isMinFirefox && !isMobile) {
       return this.renderInstallButton(isInstalling, hasAddon);
     }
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/MainInstallButton/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-
+/* eslint no-nested-ternary: 0 */
```
lmorchard: nit: Might want to not turn this off. Instead break out that nested ternary into a separate render function so it reads better.

frontend/src/app/components/MainInstallButton/index.js
```diff
@@ -74,14 +96,15 @@ export default class MainInstallButton extends Component<MainInstallButtonProps,
       <LayoutWrapper flexModifier={layout} helperClass="main-install">
         <div className="main-install__spacer" />
         {/* $FlowFixMe */}
-        {(isMinFirefox && !isMobile) ? this.renderInstallButton(isInstalling, hasAddon) : this.renderAltButton(isFirefox, isMobile) }
+        {(showWebButton) ? this.renderWebExperimentButton() : (isMinFirefox && !isMobile) ? this.renderInstallButton(isInstalling, hasAddon) : this.renderAltButton(isFirefox, isMobile)
```
lmorchard: Yeah, this is getting hairy enough to break out into a more verbose if/then/else in like a separate `renderMainButton` function:
```
if (showWebButton) {
...
} else if (isMinFirefox && !isMobile) {
...
} else {
...
}
```

lmorchard (commented):
Basically just one nit on a visual review, will give it a run on my machine and see if anything breaks for me since apparently I'm a magnet for that

lmorchard (approved):
Gave it a run, works as expected! I'm r+ on merge, though would be r++ with the nit fixed

PR #3828, opened 2018-08-24
3660 integration tests

### LINKED ISSUE #3660: integration test breakage caused by now building the addon for dev
We may need to build 2 addons, one "local" for integration tests and one "dev" for pushing to the dev site

meandavejustice:
fixes #3660 

Attempting to fix integrations tests which were broken after addon started being built for the `dev` environment on circleci.

### DIFF
--- a/frontend/test/ui/test_install.py
+++ b/frontend/test/ui/test_install.py
@@ -77,7 +77,7 @@ def test_enable_and_disable_experiment(
     selenium.add_cookie({'name': 'updates-last-viewed-date',
                          'value': datetime.datetime.now().isoformat(),
                          'max_age': 120,
-                         'domain': 'example.com'})
+                         'domain': 'testpilot.dev.mozaws.net'})
     experiment = page.find_experiment(experiment='Dev Example')
     experiment.install_and_enable()
     firefox.browser.wait_for_notification(
--- a/frontend/test/ui/test_install.py
+++ b/frontend/test/ui/test_install.py
@@ -77,7 +77,7 @@ def test_enable_and_disable_experiment(
     selenium.add_cookie({'name': 'updates-last-viewed-date',
                          'value': datetime.datetime.now().isoformat(),
                          'max_age': 120,
-                         'domain': 'testpilot.dev.mozaws.net'})
+                         'domain': 'example.com'})
     experiment = page.find_experiment(experiment='Dev Example')
     experiment.install_and_enable()
     firefox.browser.wait_for_notification(
--- a/frontend/test/ui/test_install.py
+++ b/frontend/test/ui/test_install.py
@@ -7,32 +7,6 @@ from pages.desktop.home import Home
 from pages.desktop.detail import Detail
 
 
-@pytest.mark.nondestructive
-@pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
-                    reason='Skip install on Release and Beta Firefox.')
-def test_install_of_test_pilot_addon(
-        base_url, selenium, firefox, notifications):
-    """Check that the testpilot addon is installable and installs."""
-    page = Home(selenium, base_url).open()
-    selenium.execute_script(
-        "document.querySelector('.landing-experiments').scrollIntoView();"
-    )
-    if not page.featured.is_displayed:
-        page.header.click_install_button()
-        firefox.browser.wait_for_notification(
-            notifications.AddOnInstallConfirmation).install()
-        firefox.browser.wait_for_notification(
-            notifications.AddOnInstallComplete
-        ).close()
-    else:
-        page.featured.click_install_button()
-        firefox.browser.wait_for_notification(
-            notifications.AddOnInstallConfirmation).install()
-        firefox.browser.wait_for_notification(
-            notifications.AddOnInstallComplete
-        ).close()
-
-
 @pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
                     reason='Skip install on Release and Beta Firefox.')
 @pytest.mark.nondestructive
--- a/frontend/test/ui/test_install.py
+++ b/frontend/test/ui/test_install.py
@@ -1,36 +1,36 @@
 import os
-# import datetime
+import datetime
 
 import pytest
-# from pages.desktop.experiments import Experiments
+from pages.desktop.experiments import Experiments
 from pages.desktop.home import Home
-# from pages.desktop.detail import Detail
+from pages.desktop.detail import Detail
 
 
-# @pytest.mark.nondestructive
-# @pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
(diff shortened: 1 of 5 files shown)


frontend/test/ui/test_install.py
```diff
@@ -1,36 +1,36 @@
...
-#         page.featured.click_install_button()
-#         firefox.browser.wait_for_notification(
-#             notifications.AddOnInstallConfirmation).install()
-#         firefox.browser.wait_for_notification(
-#             notifications.AddOnInstallComplete
-#         ).close()
+@pytest.mark.nondestructive
+@pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
+                    reason='Skip install on Release and Beta Firefox.')
+def test_install_of_test_pilot_addon(
```
b4handjr: Looking at the build log I have some thoughts:

It looks like there is some sort of popup that is shown on the html report within circleci/artifacts. The test is kind of just looking for the standard webextension install notification.

frontend/test/ui/test_install.py
```diff
@@ -45,53 +45,53 @@ def test_bottom_install_button(base_url, selenium, firefox, notifications):
...
-#     firefox.browser.wait_for_notification(
-#         notifications.AddOnInstallComplete).close()
-#     exp_detail = Detail(selenium, base_url)
-#     assert exp_detail.email_popup.is_popup_displayed()
-#     exp_detail.email_popup.close()
-#     assert exp_detail.enabled_popup.is_popup_displayed()
+@pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
+                    reason='Skip install on Release and Beta Firefox.')
+@pytest.mark.nondestructive
+def test_install_and_enable(base_url, selenium, firefox, notifications):
```
b4handjr, line 25: Looking the log it seems as if the Dev Example gets enabled but ```experiment.install_and_enable()``` is looking for something else. Specifically its within the ```self.wait.until()``` found in that class. I'd check there and make sure whatever it is waiting for matches the new page/styling.

frontend/test/ui/test_install.py
```diff
@@ -45,53 +45,53 @@ def test_bottom_install_button(base_url, selenium, firefox, notifications):
...
-#     assert exp_detail.email_popup.is_popup_displayed()
-#     exp_detail.email_popup.close()
-#     assert exp_detail.enabled_popup.is_popup_displayed()
-#     exp_detail.enabled_popup.close()
-#     experiment.disable()
-#     assert experiment.enable_button.is_displayed()
+@pytest.mark.nondestructive
+@pytest.mark.skipif(os.environ.get('SKIP_INSTALL_TEST') is not None,
+                    reason='Skip install on Release and Beta Firefox.')
+def test_enable_and_disable_experiment(
```
b4handjr, line 47: This looks like the same error as above, please see my comments there.

b4handjr (changes requested):

frontend/test/ui/test_install.py
b4handjr: Maybe this got changed? https://github.com/mozilla/testpilot/blob/master/frontend/test/ui/pages/desktop/detail.py#L45

meandavejustice:
@jrbenny35 Just pushed a change to the dev example experiment making it of the `addon` type again, which is what the integration tests expect. I probably would make sense to add tests and dev examples for each type. We now have:
* addon
* web experiment
* ios app
* android app

We also should think about having tests for experiments that have many types of these supported.

It should be noted that much of this ui is covered in our unit tests. We can probably find a healthy balance between the two.

b4handjr:
Cool sounds good. I'll have to take a look and see what could/should be added to these tests. Even just an install test could be good enough.

meandavejustice:
@jrbenny35 yeah, if we do tests for each type I was thinking:
- Addon : install test (we already have these)
- Web Experiment: test that button/link domain exists
- Mobile: test that App Store/Google Play store buttons are present

frontend/test/ui/test_install.py
meandavejustice: Seems to be the same, was just moved into ExperimentButtons https://github.com/mozilla/testpilot/blob/master/frontend/src/app/containers/ExperimentPage/ExperimentButtons.js#L143

meandavejustice:
@jrbenny35 Made a couple changes, would you mind taking a look again? I'm not sure why it can't find the email dialog. I'm seeing it locally

meandavejustice:
@jrbenny35 Could use your eyes on this, the current screenshot doesn't seem to match the error. Not sure what's going on. And still can't run these locally

b4handjr:
Why cant you run them locally?

meandavejustice:
@jrbenny35 I've never been able to get them to run on windows. Neither has @lmorchard. @fzzzy has also reported issues running the tests locally. I think the whole team has been pushing and waiting for circleci results each time there is an issue. Which is why we dread changing these

b4handjr:
The logs have this
```
> assert exp_detail.email_popup.is_popup_displayed()

frontend/test/ui/test_install.py:39:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
frontend/test/ui/pages/desktop/detail.py:25: in email_popup
return self.EmailPopup(self)
.tox/ui-tests/local/lib/python2.7/site-packages/pypom/region.py:73: in __init__
self.wait_for_region_to_load()
frontend/test/ui/pages/desktop/detail.py:74: in wait_for_region_to_load
lambda _: self.root.is_displayed())
```
 pointing to [this](https://github.com/mozilla/testpilot/blob/master/frontend/test/ui/pages/desktop/detail.py#L74). Does that selector still exist? It seems the popup is showing.

meandavejustice:
@jrbenny35 Yeah, that classname is present, https://github.com/mozilla/testpilot/blob/master/frontend/src/app/components/EmailDialog/index.js#L59 :thinking:

meandavejustice:
Here is the install process in screenshots from my local manual install.

Step 1: after clicking install txp and enable dev example:
txp-integration-email-dialog
Step 2: after dismissing the email dialog
txp-integration-add-dev
Step 3: after clicking "add sloth"
txp-integration-add-confim
Step 4: after confirming the previous doorhanger, the tour dialog is shown
txp-integration-tour-final

meandavejustice:
Put in a request for a loaner mac, should be here next week. Gunna put this on the backburner till I get that and can run the tests locally.

meandavejustice:
been messing with this past couple days, got integration tests running on the loaner mac, but am seeing different results than circleci. @jrbenny35 Is seeing the same issue.

The `test_bottom_install_button` test fails without showing the install dialog. Also the integration test seems to be launching my firefox stable release instead of the nightly install

meandavejustice:
Got these passing 100% on the macbook loaner. This is now only failing on circle ci

PR #3835, opened 2018-08-29
Allow sms refactor

### LINKED ISSUE #3636: Change sms fetch method and add test for MobileDialog
We need to restructure the way `allowSMS` state is set in the MobileDialog modal to be passed in as a prop. This will allow us to add more extensive testing

https://redux.js.org/advanced/async-actions

meandavejustice:
fixes #3636

### DIFF
--- a/frontend/src/app/actions/browser.js
+++ b/frontend/src/app/actions/browser.js
@@ -4,11 +4,23 @@ import type {
   FetchCountryCodeAction
 } from "../reducers/browser";
 
-export function fetchCountryCode(): FetchCountryCodeAction {
-  return fetch("url here").then((result) {
-    return {
-      type: "FETCH_COUNTRY_CODE",
-      payload: "country code here"
-    };
-  });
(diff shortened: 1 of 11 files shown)


frontend/src/app/actions/browser.js
```diff
@@ -0,0 +1,26 @@
...
+import type {
+  FetchCountryCodeAction
+} from "../reducers/browser";
+
+// There may be a better place for this, but this seems
+// the best for now.
+export const acceptedSMSCountries = ["US", "DE", "FR"];
+export const COUNTRY_CODE_ENDPOINT = "https://location.services.mozilla.com/v1/country";
+
+export function fetchCountryCode(): Promise<FetchCountryCodeAction> {
```
lmorchard, line 12: :+1: This looks like a good place for this, solves the refactor issue nicely!

lmorchard (approved):
Looks like the right refactor. Tests pass, looks like the mobile dialog works for me. r+

PR #3901, opened 2018-10-13
Retires Advance (closes #3900).

### LINKED ISSUE #3900: Retire Advance.
This will be effective 2018/10/17

chuckharmston:


content-src/experiments/advance.yaml
```diff
@@ -1,5 +1,6 @@
 addon_id: support@laserlike.com
 bug_report_url: "https://github.com/laserlikeinc/firefox-addon/issues"
+completed: '2018-10-10T06:00:00.000000Z'
```
pdehaan: Should this be 10/17?

chuckharmston: Eventually, yes; I was hoping this would trigger the PR deployment for testing, but that appears not to be working.

PR #4021, opened 2019-01-16
Add eol strings

meandavejustice:


lmorchard:
Thinking I'll just merge this after circleci passes, because we already reviewed these strings over on the EOL repo

PR #4030, opened 2019-01-21
Fix title attribute quoting for en-US locale

### LINKED ISSUE #4029: title attributes are weird on eol branch for "eolMessageFour"
Notice how we only get the first word of each `<a>` content ("Firefox Color" turns into `title="Firefox"`, "Side View" turns into `title="Side"`, etc):

```html
<p data-l10n-id="eolMessageFour">Other experiments like
  <a data-l10n-name="colorLink" ... title="⁨Firefox">⁨Firefox Color⁩</a>,
  <a data-l10n-name="sideViewLink" ... title="⁨Side">⁨Side View⁩</a>,
  <a data-l10n-name="notesLink" ... title="⁨Firefox">⁨Firefox Notes⁩</a>,
  <a data-l10n-name="priceWiseLink" ... title="⁨Price">⁨Price Wise⁩</a>, and
  <a data-l10n-name="emailTabsLink" ... title="⁨Email">⁨Email Tabs⁩</a>
will remain available to the ⁨Firefox⁩ community. If you have installed these experiments, you can continue to use them.
</p>
```

I'm not seeing the same behavior on "eolMessageThree", so it must be something subtle:

```html
<p data-l10n-id="eolMessageThree">
  We've also built great new app experiences like
  <a data-l10n-name="lockboxLink" ... title="⁨Firefox Lockbox⁩">⁨Firefox Lockbox⁩</a> and
  <a data-l10n-name="sendLink" ... title="⁨Firefox Send⁩">⁨Firefox Send⁩</a>
  which continue to push ⁨Firefox⁩ beyond the browser. Look for exciting new developments from both of these projects in 2019.
</p>
```

I do not know if having the `title=""` attribute with the exact same text as the anchor text improves a11y at all, but probably worth fixing since the current behavior would be notably worse for screen readers.

pdehaan:
Fixes #4029

### DIFF
--- a/locales/en-US/app.ftl
+++ b/locales/en-US/app.ftl
@@ -298,11 +298,11 @@ eolMessageThree =
     which continue to push {-brand} beyond the browser. Look for exciting new developments from both of these projects in 2019.
 eolMessageFour =
     Other experiments like
-    <a data-l10n-name="colorLink" title={-exp-color}>{-exp-color}</a>,
-    <a data-l10n-name="sideViewLink" title={-exp-side-view}>{-exp-side-view}</a>,
-    <a data-l10n-name="notesLink" title={-exp-notes}>{ -exp-notes}</a>,
-    <a data-l10n-name="priceWiseLink" title={-exp-price-wise}>{-exp-price-wise}</a>, and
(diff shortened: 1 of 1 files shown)


meandavejustice:
Thank you @pdehaan

PR #4046, opened 2019-03-20
Rename Price Wise to Price Tracker

### LINKED ISSUE #4045: Rename Price Wise to Price Tracker
We need to rename Price Wise to Price Tracker: https://github.com/mozilla/price-tracker/pull/308

pdehaan:
Fixes #4045


lmorchard:
FWIW, I think these changes need to happen on the l10n branch, [otherwise they'll be clobbered whenever (if ever) we next accept changes from Pontoon](https://github.com/mozilla/testpilot#accepting-work-from-l10n-branch). (tl;dr: As an expedient hack, l10n branch is still based on the old master branch, so we just copy the locales directory on top of the one in eol branch.)

pdehaan:
Odd. Changed base branch to l10n and everything auto closed.
Resubmitted as #4047, but I'll still have to do the URL change separately since I didn't see those files in the l10n branch (which I assume is tracking **master** and not **eol** branch).

PR #4047, opened 2019-03-20
Rename Price Wise to Price Tracker

pdehaan:


flodolo (approved):

lmorchard:
I think this should be fine, even with a failing integration test. We're not using the original site anymore anyway.

